PR #12, opened 2017-05-15
Skip invalid clusters

ikitiki:
— Ignore clusters with wrong specification
— catch quantity parsing errors

### DIFF
--- a/pkg/controller/postgresql.go
+++ b/pkg/controller/postgresql.go
@@ -184,21 +184,22 @@ func (c *Controller) queueClusterEvent(old, new *spec.Postgresql, eventType spec
 		clusterError error
 	)
 
-	if old != nil {
+	if old != nil { //update, delete
 		uid = old.Metadata.GetUID()
 		clusterName = util.NameFromMeta(old.Metadata)
 		if eventType == spec.EventUpdate && new.Error == nil && old != nil {
 			eventType = spec.EventAdd
 			clusterError = new.Error
+		} else {
+			clusterError = old.Error
 		}
-		clusterError = old.Error
-	} else {
+	} else { //add, sync
 		uid = new.Metadata.GetUID()
 		clusterName = util.NameFromMeta(new.Metadata)
 		clusterError = new.Error
 	}
 
-	if clusterError != nil {
+	if clusterError != nil && eventType != spec.EventDelete {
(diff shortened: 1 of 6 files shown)


pkg/controller/postgresql.go
```diff
@@ -37,16 +37,22 @@ func (c *Controller) clusterListFunc(options api.ListOptions) (runtime.Object, e
...
+			continue
+		}
 		c.queueClusterEvent(nil, pg, spec.EventSync)
 
 		c.logger.Debugf("Sync of the '%s' cluster has been queued", util.NameFromMeta(pg.Metadata))
+		activeClustersCnt++
 	}
-	if len(objList) > 0 {
+	if activeClustersCnt > 0 {
 		c.logger.Infof("There are %d clusters currently running", len(objList))
```
alexeyklyukin: Perhaps should the number of clusters that has errors. Would be useful when we have a background cleaner.

pkg/controller/postgresql.go
```diff
@@ -168,17 +174,30 @@ func (c *Controller) processClusterEventsQueue(idx int) {
...
 	)
 
 	if old != nil {
 		uid = old.Metadata.GetUID()
 		clusterName = util.NameFromMeta(old.Metadata)
+		if eventType == spec.EventUpdate && new.Error == nil && old != nil {
+			eventType = spec.EventAdd
+			clusterError = new.Error
+		}
+		clusterError = old.Error
```
alexeyklyukin: should be in the else block.

pkg/controller/postgresql.go
```diff
@@ -168,17 +174,30 @@ func (c *Controller) processClusterEventsQueue(idx int) {
...
 	} else {
 		uid = new.Metadata.GetUID()
 		clusterName = util.NameFromMeta(new.Metadata)
+		clusterError = new.Error
 	}
+
+	if clusterError != nil {
+		c.logger.Debugf("Skipping %s event for invalid cluster %s (reason: %s)", eventType, clusterName, clusterError)
```
alexeyklyukin: Why would we do it for the spec.EventDelete?

alexeyklyukin:
👍

ikitiki:
👍

PR #18, opened 2017-05-16
Replace the statefulset if it cannot be updated.

alexeyklyukin:
Updates to statefulset spec for fields other than 'replicas' and
containers' are forbidden. However, it is possible to delete the old
statefulset without deleting its pods and create the new one, using the
changed specs. The new statefulset shall pick up the orphaned pods.

Change the statefulset's comparison to return the combined effect of
all checks, not just the first non-matching field.

The environment variables in the statefulset are now sorted based on the name.
As a result, a one-time upgrade of all clusters might be necessary.

### DIFF
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -159,10 +159,10 @@ func (c *Cluster) replaceStatefulSet(newStatefulSet *v1beta1.StatefulSet) error 
 	c.logger.Debugf("Replacing StatefulSet")
 
 	// Delete the current statefulset without deleting the pods
-	OrphanDepencies := true
+	orphanDepencies := true
 	oldStatefulset := c.Statefulset
 
-	options := v1.DeleteOptions{OrphanDependents: &OrphanDepencies}
+	options := v1.DeleteOptions{OrphanDependents: &orphanDepencies}
 	if err := c.KubeClient.StatefulSets(oldStatefulset.Namespace).Delete(oldStatefulset.Name, &options); err != nil {
 		return fmt.Errorf("Can't delete statefulset '%s': %s", statefulSetName, err)
 	}
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -2,12 +2,12 @@ package cluster
 
 import (
 	"fmt"
-	"time"
 
 	"k8s.io/client-go/pkg/api"
 	"k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/pkg/apis/apps/v1beta1"
 
+	"github.bus.zalan.do/acid/postgres-operator/pkg/util/retryutil"
 	"github.com/zalando-incubator/postgres-operator/pkg/spec"
 	"github.com/zalando-incubator/postgres-operator/pkg/util"
 	"github.com/zalando-incubator/postgres-operator/pkg/util/constants"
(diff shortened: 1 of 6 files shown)


pkg/util/constants/constants.go
```diff
@@ -1,31 +1,34 @@
...
+	ZalandoDnsNameAnnotation   = "external-dns.alpha.kubernetes.io/hostname"
+	ElbTimeoutAnnotationName   = "service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout"
+	ElbTimeoutAnnotationValue  = "3600"
+	KubeIAmAnnotation          = "iam.amazonaws.com/role"
+	ResourceName               = TPRName + "s"
+	PodRoleMaster              = "master"
+	PodRoleReplica             = "replica"
+	SuperuserKeyName           = "superuser"
+	ReplicationUserKeyName     = "replication"
+	RoleFlagSuperuser          = "SUPERUSER"
```
ikitiki: I'd add empty line before ``RoleFlagSuperuser`` and move ``StatefulsetDeletionTimeout`` just after ``ReplicationUserKeyName``

pkg/cluster/resources.go
```diff
@@ -143,6 +146,56 @@ func (c *Cluster) updateStatefulSet(newStatefulSet *v1beta1.StatefulSet) error {
...
+	options := v1.DeleteOptions{OrphanDependents: &OrphanDepencies}
+	if err := c.KubeClient.StatefulSets(oldStatefulset.Namespace).Delete(oldStatefulset.Name, &options); err != nil {
+		return fmt.Errorf("Can't delete statefulset '%s': %s", statefulSetName, err)
+	}
+	// make sure we clear the stored statefulset status if the subsequent create fails.
+	c.Statefulset = nil
+	// wait until the statefulset is truly deleted
+	c.logger.Debugf("Waiting for the statefulset to be deleted")
+
+	before := time.Now()
```
ikitiki: can we use ``retryutil.Retry()`` here?

alexeyklyukin: Good idea, thanks.

pkg/cluster/resources.go
```diff
@@ -2,12 +2,12 @@ package cluster
 
 import (
 	"fmt"
-	"time"
 
 	"k8s.io/client-go/pkg/api"
 	"k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/pkg/apis/apps/v1beta1"
 
+	"github.bus.zalan.do/acid/postgres-operator/pkg/util/retryutil"
```
ikitiki: 😱
``github.com/zalando-incubator/postgres-operator/pkg/util/retryutil``

Jan-M:
Can we double check and document our EBS / Persistent Volume reclaim settings? I would like to make sure that we really are doing this right when doing deletes outside of "deletes" triggered by deletes on the PostgreSQL object.

alexeyklyukin:
@Jan-M I can't parse your comment. What does PV reclaim have to do with this pull request?

pkg/cluster/resources.go
alexeyklyukin: Ups, or IDE came out to be too smart. Fixed.

pkg/cluster/resources.go
```diff
@@ -146,6 +149,53 @@ func (c *Cluster) updateStatefulSet(newStatefulSet *v1beta1.StatefulSet) error {
...
+func (c *Cluster) replaceStatefulSet(newStatefulSet *v1beta1.StatefulSet) error {
+	if c.Statefulset == nil {
+		return fmt.Errorf("There is no StatefulSet in the cluster")
+	}
+
+	statefulSetName := util.NameFromMeta(c.Statefulset.ObjectMeta)
+	c.logger.Debugf("Replacing StatefulSet")
+
+	// Delete the current statefulset without deleting the pods
+	OrphanDepencies := true
```
ikitiki: i'd make it start with a lowercase letter

pkg/cluster/resources.go
```diff
@@ -146,6 +149,53 @@ func (c *Cluster) updateStatefulSet(newStatefulSet *v1beta1.StatefulSet) error {
...
+	if err != nil {
+		fmt.Errorf("could not delete statefulset: %s", err)
+	}
+
+	// create the new statefulset with the desired spec. It would take over the remaining pods.
+	createdStatefulset, err := c.KubeClient.StatefulSets(newStatefulSet.Namespace).Create(newStatefulSet)
+	if err != nil {
+		return fmt.Errorf("Can't create statefulset '%s': %s", statefulSetName, err)
+	} else {
+		// check that all the previous replicas were picked up.
```
ikitiki: the idea was to get the StatefulSet object from the Kubernetes and check if number of desired replicas match current number

ikitiki:
👍

alexeyklyukin:
👍

PR #22, opened 2017-05-17
Return error instead of logging it

ikitiki:
In order to make code more homogenous it would make sense to return error in methods like SyncCluster and Delete instead of just logging it


alexeyklyukin:
👍

ikitiki:
👍

PR #23, opened 2017-05-18
Use named arguments in the DNS name format

ikitiki:
Instead of using sprintfplaceholders use named arguments


pkg/util/config/util.go
```diff
@@ -162,3 +165,23 @@ func processField(value string, field reflect.Value) error {
...
 	return nil
 }
+
+func (f *stringTemplate) Decode(value string) error {
+	*f = stringTemplate(value)
+
+	return nil
+}
+
+func (f *stringTemplate) Parse(a ...string) string {
```
a1exsh: The function name looks surprising to me, as it does the opposite of "parsing".

ikitiki: yeah, you're right.
what do you think about naming it ``Execute`` ?

a1exsh: `Format`?

ikitiki: ``Format`` sounds ok for me

pkg/util/config/config.go
```diff
@@ -43,19 +43,19 @@ type Config struct {
...
-	Workers            uint32 `name:"workers" default:"4"`
+	Namespace          string         `name:"namespace"`
+	EtcdHost           string         `name:"etcd_host" default:"etcd-client.default.svc.cluster.local:2379"`
+	DockerImage        string         `name:"docker_image" default:"registry.opensource.zalan.do/acid/spiloprivate-9.6:1.2-p4"`
+	ServiceAccountName string         `name:"service_account_name" default:"operator"`
+	DbHostedZone       string         `name:"db_hosted_zone" default:"db.example.com"`
+	EtcdScope          string         `name:"etcd_scope" default:"service"`
+	WALES3Bucket       string         `name:"wal_s3_bucket"`
+	KubeIAMRole        string         `name:"kube_iam_role"`
+	DebugLogging       bool           `name:"debug_logging" default:"true"`
```
alexeyklyukin: Did you intentionally change the debug_logging to true?

ikitiki: No, it was unintentionally, but I don't mind keeping it true by default

alexeyklyukin:
👍

ikitiki:
👍

PR #24, opened 2017-05-18
Infrastructure to extend volumes.

alexeyklyukin:
Implements EBS volume resizing #35. 

 In order to support volumes different from EBS and filesystems other than EXT2/3/4 the respective code parts were implemented as interfaces. Adding the new resize for the volume or the filesystem will require implementing the interface, but no other changes in the cluster code itself.

Volume resizing first changes the EBS and the filesystem, and only afterwards is reflected in the Kubernetes "PersistentVolume" object. This is done deliberately to be able to check if the volume needs resizing by peeking at the Size of the PersistentVolume structure.  We recheck, nevertheless, in the EBSVolumeResizer, whether the actual EBS volume size doesn't match the spec, since call to the AWS ModifyVolume is counted against the resize limit of once every 6 hours, even for those calls that shouldn't result in an actual resize (i.e. when the size matches the one for the running volume).

As a collateral, split the constants into multiple files, move the volume code into a separate file and fix minor issues related to the error reporting.


pkg/util/constants/constants.go
```diff
@@ -3,28 +3,41 @@ package constants
 import "time"
 
 const (
```
ikitiki: I think it's high time to split that constants file into separate smaller ones

alexeyklyukin:
👍

alexeyklyukin:
👍

Jan-M:
:+1:

pkg/cluster/filesystems.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/zalando-incubator/postgres-operator/pkg/spec"
+	"github.com/zalando-incubator/postgres-operator/pkg/util/constants"
+	"github.com/zalando-incubator/postgres-operator/pkg/util/filesystems"
+)
+
+func (c *Cluster) getPostgresFilesystemInfo(podName *spec.NamespacedName) (err error, device string, fstype string) {
```
ikitiki: please, put ``error`` in the end and remove output argument names

pkg/cluster/filesystems.go
```diff
@@ -0,0 +1,45 @@
...
+	out, err := c.ExecCommand(podName, "bash", "-c", fmt.Sprintf("df -T %s|tail -1", constants.PostgresDataMount))
+	if err != nil {
+		return err, "", ""
+	}
+	fields := strings.Fields(out)
+	if len(fields) < 2 {
+		return fmt.Errorf("too few fields in the df output"), "", ""
+	}
+	return nil, fields[0], fields[1]
```
ikitiki: empty line before return will make it look better

pkg/cluster/volumes.go
```diff
@@ -0,0 +1,178 @@
...
+func (c *Cluster) deletePersistenVolumeClaims() error {
+	c.logger.Debugln("Deleting PVCs")
+	ns := c.Metadata.Namespace
+	pvcs, err := c.listPersistentVolumeClaims()
+	if err != nil {
+		return err
+	}
+	for _, pvc := range pvcs {
+		c.logger.Debugf("Deleting PVC '%s'", util.NameFromMeta(pvc.ObjectMeta))
+		if err := c.KubeClient.PersistentVolumeClaims(ns).Delete(pvc.Name, c.deleteOptions); err != nil {
```
ikitiki: I think it would be wise to get the namespace from ``pvc``, not from the cluster:
``c.KubeClient.PersistentVolumeClaims(pvc.Namespace).``

pkg/cluster/volumes.go
```diff
@@ -0,0 +1,178 @@
...
+				if err != nil {
+					return fmt.Errorf("could not connect to the volume provider: %v", err)
+				}
+				defer resizer.DisconnectFromProvider()
+			}
+			volumeSize := quantityToGigabyte(pv.Spec.Capacity[v1.ResourceStorage])
+			if volumeSize > newSize {
+				return fmt.Errorf("cannot shrink persistent volume")
+			}
+			if volumeSize != newSize {
```
ikitiki: maybe:
```
if volumeSize == newSize {
   continue
}
```
?

pkg/cluster/volumes.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	manifestSize := quantityToGigabyte(newSize)
+	volumes, err := c.listPersistentVolumes()
+	if err != nil {
+		return nil, 0, fmt.Errorf("could not list persistent volumes: %v", err)
+	}
+	return volumes, manifestSize, nil
+}
+
+func getPodNameFromPersistentVolume(pv *v1.PersistentVolume) *spec.NamespacedName {
```
ikitiki: I think it should be named as ``getPersistentVolumeClaimRef``

alexeyklyukin: Wouldn't it be confusing that we return not a claim, but a pod name, as the code does `pv.Spec.ClaimRef.Name[len(constants.DataVolumeName)+1:]`?

ikitiki: yeah, you're right

pkg/cluster/filesystems.go
alexeyklyukin: The first is obviously a blunder, but the second I would do nothing about. I find it less error-prone to name multiple consecutive arguments of the same type in a declaration, even though the names are not used to return values.

ikitiki:
I think you need to run ``go fmt`` for all the modified files

pkg/cluster/volumes.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, pv := range volumes {
+		currentSize := quantityToGigabyte(pv.Spec.Capacity[v1.ResourceStorage])
+		if currentSize != manifestSize {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (c *Cluster) listVolumesWitManifestSize(newVolume spec.Volume) ([]*v1.PersistentVolume, int64, error) {
```
ikitiki: ``Wit``?

alexeyklyukin:
👍

ikitiki:
👍

PR #25, opened 2017-05-19
Remove etcd requests

ikitiki:
As for now we use etcd only to check if there's a key for the cluster we're about to create.
in order to minimize number of dependencies, I propose deleting described functionality.


ikitiki:
:+1:

alexeyklyukin:
👍

PR #30, opened 2017-05-22
Use queues for the pod events

ikitiki:
Make use of the FIFO queue for the pod events of the cluster

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/client-go/pkg/apis/apps/v1beta1"
 	"k8s.io/client-go/pkg/types"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 
 	"github.com/zalando-incubator/postgres-operator/pkg/spec"
 	"github.com/zalando-incubator/postgres-operator/pkg/util"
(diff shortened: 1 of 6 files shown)


pkg/cluster/cluster.go
```diff
@@ -460,7 +463,38 @@ func (c *Cluster) Delete() error {
 }
 
 func (c *Cluster) ReceivePodEvent(event spec.PodEvent) {
-	c.podEvents <- event
+	c.podEventsQueue.Add(event)
+}
+
+func (c *Cluster) podEventProcess(obj interface{}) error {
```
alexeyklyukin: I'd rename it to processPodEvent (action -> object, i.e. setStatus), or name it podEventProcessor.

pkg/cluster/cluster.go
```diff
@@ -460,7 +463,38 @@ func (c *Cluster) Delete() error {
...
+	}
+
+	return nil
+}
+
+func (c *Cluster) Run(stopCh <-chan struct{}) {
+	go c.processPodEventQueue(stopCh)
+}
+
+func (c *Cluster) processPodEventQueue(stopCh <-chan struct{}) {
```
alexeyklyukin: Can we drop it and embed its body into the Run?

pkg/cluster/cluster.go
```diff
@@ -460,7 +463,38 @@ func (c *Cluster) Delete() error {
...
+	c.podSubscribersMu.RUnlock()
+	if ok {
+		subscriber <- event
+	}
+
+	return nil
+}
+
+func (c *Cluster) Run(stopCh <-chan struct{}) {
+	go c.processPodEventQueue(stopCh)
```
alexeyklyukin: I'd make it, perhaps, return the *Cluster, so that one can write cl := cluster.New(....).Run()

alexeyklyukin:
Minor stylistic comments apart LGTM.

alexeyklyukin:
👍

pkg/cluster/cluster.go
ikitiki: I want to have a place where I can run all the future go routines.
besides, lambda func will not look cool in the Run() function

alexeyklyukin: it there will be more - let's leave it as it is now.

alexeyklyukin:
👍

ikitiki:
+1

ikitiki:
👍

pkg/cluster/cluster.go
```diff
@@ -460,7 +463,38 @@ func (c *Cluster) Delete() error {
...
+
+func (c *Cluster) podEventProcess(obj interface{}) error {
+	event, ok := obj.(spec.PodEvent)
+	if !ok {
+		return fmt.Errorf("could not cast to PodEvent")
+	}
+
+	c.podSubscribersMu.RLock()
+	subscriber, ok := c.podSubscribers[event.PodName]
+	c.podSubscribersMu.RUnlock()
```
ikitiki: yeah. good point

alexeyklyukin: @Jan-M I'd not use defer here. One should hold the lock for as short as possible, and avoid holding it over the statements and does the waiting on their own, like reading from the channel.  Besides, there is nothing to fail in the statement between the lock and unlock.

ikitiki: JFYI, https://github.com/beyondns/gotips/blob/master/tips32.md#23---defer-is-slow

PR #32, opened 2017-05-26
Change maintenance window format

ikitiki:
Change maintenance window format to more readable form:
``{weekday}:{start time %H:%M}-{end time %H:%M}``
- minor improvements

### DIFF
--- a/pkg/spec/postgresql.go
+++ b/pkg/spec/postgresql.go
@@ -3,7 +3,6 @@ package spec
 import (
 	"encoding/json"
 	"fmt"
-	"regexp"
 	"strings"
 	"time"
 
@@ -13,11 +12,9 @@ import (
 )
 
 type MaintenanceWindow struct {
-	StartTime    time.Time    // Start time
-	StartWeekday time.Weekday // Start weekday
-
-	EndTime    time.Time    // End time
-	EndWeekday time.Weekday // End weekday
+	Weekday   time.Weekday
+	StartTime time.Time // Start time
+	EndTime   time.Time // End time
 }
 
 type Volume struct {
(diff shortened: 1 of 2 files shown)


ikitiki:
👍

ikitiki:
👍

pkg/spec/postgresql.go
```diff
@@ -107,6 +108,12 @@ func parseTime(s string) (t time.Time, wd time.Weekday, wdProvided bool, err err
 		}
 		timeLayout = "Mon:15:04"
 		wdProvided = true
+		weekday, ok := weekdays[parts[0]]
+		if !ok {
+			err = fmt.Errorf("incorrect weekday")
+			return
+		}
+		wd = time.Weekday(weekday)
```
alexeyklyukin: Why not let time.Parse parse it instead as a part of the timeLayout?

pkg/spec/postgresql.go
```diff
@@ -165,9 +170,13 @@ func (m *MaintenanceWindow) UnmarshalJSON(data []byte) error {
 		return fmt.Errorf("'From' time must be prior to the 'To' time")
 	}
 
+	if (int(got.StartWeekday)+6)%7 > (int(got.EndWeekday)+6)%7 {
+		return fmt.Errorf("'From' weekday must be prior to the 'To' weekday")
```
alexeyklyukin: Why not compare directly?

pkg/spec/postgresql.go
ikitiki: because time.Parse does not actually parses it, it just checks for the syntax.
Extract from the [doc](https://golang.org/pkg/time/#Parse):

> The day of the week is checked for syntax but it is otherwise ignored.

pkg/spec/postgresql.go
ikitiki: because in ``time`` package week starts with Sunday

pkg/spec/postgresql.go
alexeyklyukin: gotcha!

pkg/spec/postgresql.go
alexeyklyukin: Got it; don't we need to check the weekdayProvidedTo (otherwise, got.StartWeekday == got.EndWeekday == 0)

pkg/spec/postgresql.go
```diff
@@ -165,9 +170,13 @@ func (m *MaintenanceWindow) UnmarshalJSON(data []byte) error {
 		return fmt.Errorf("'From' time must be prior to the 'To' time")
```
alexeyklyukin: Does it mean that we don't allow intervals like Mon 15:35 - Tue 13:25?

ikitiki: no, we do not allow such period. One should split it into two:
- 15:35-23:59
- 00:00-13:25

alexeyklyukin:
I think we should still allow the form of start_time - end_time (w/o the day), so that it will run every day (instead of writing Sun:, Mon:  and so on).

alexeyklyukin:
👍

ikitiki:
👍

PR #33, opened 2017-05-26
Fix password check in pguserpassword function

ikitiki:
Use correct md5 length while checking password of the user

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -41,7 +41,7 @@ func NameFromMeta(meta v1.ObjectMeta) spec.NamespacedName {
 }
 
 func PGUserPassword(user spec.PgUser) string {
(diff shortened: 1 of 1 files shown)


pkg/util/util.go
```diff
@@ -41,7 +41,7 @@ func NameFromMeta(meta v1.ObjectMeta) spec.NamespacedName {
 }
 
 func PGUserPassword(user spec.PgUser) string {
-	if (len(user.Password) == md5.Size && user.Password[:3] == md5prefix) || user.Password == "" {
+	if (len(user.Password) == md5.Size*2+len(md5prefix) && user.Password[:3] == md5prefix) || user.Password == "" {
```
ikitiki: ``md5.Size`` is in Bytes, we use Hexadecimal representation of it

ikitiki:
👍

alexeyklyukin:
👍

PR #41, opened 2017-06-07
Fix resync of clusters

ikitiki:
Run sync not oftener than resync_period interval.
Periodically run resync of the clusters

### DIFF
--- a/pkg/controller/postgresql.go
+++ b/pkg/controller/postgresql.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"time"
 	"reflect"
 
 	"k8s.io/client-go/pkg/api"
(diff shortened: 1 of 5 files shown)


pkg/controller/postgresql.go
```diff
@@ -37,6 +51,11 @@ func (c *Controller) clusterListFunc(options api.ListOptions) (runtime.Object, e
 		return nil, fmt.Errorf("could not extract list of postgresql objects: %v", err)
 	}
 
+	if time.Since(c.lastClusterSyncTime) <= c.opConfig.ResyncPeriod {
```
alexeyklyukin: This function can be called concurrently, therefore, I'd protect access to the c.lastClusterSyncTime with a lock.

pkg/util/constants/kubernetes.go
```diff
@@ -10,4 +10,7 @@ const (
 	K8sAPIPath                  = "/api"
 	StatefulsetDeletionInterval = 1 * time.Second
 	StatefulsetDeletionTimeout  = 30 * time.Second
+
+	QueueResyncPeriodPod = 10 * time.Minute
```
alexeyklyukin: Why changing them from the former default of 5 minutes?

pkg/controller/postgresql.go
```diff
@@ -37,6 +51,11 @@ func (c *Controller) clusterListFunc(options api.ListOptions) (runtime.Object, e
...
+		c.logger.Debugln("skipping resync of clusters")
+		return object, err
```
alexeyklyukin: I think we should return the list of objects to the shared informer in any case, but only add the `Sync` message not earlier (and, preferably, not later) than after the `ResyncPeriod` from the last sync.

pkg/controller/postgresql.go
```diff
@@ -37,6 +51,11 @@ func (c *Controller) clusterListFunc(options api.ListOptions) (runtime.Object, e
 		return nil, fmt.Errorf("could not extract list of postgresql objects: %v", err)
 	}
 
+	if time.Since(c.lastClusterSyncTime) <= c.opConfig.ResyncPeriod {
+		c.logger.Debugln("skipping resync of clusters")
```
alexeyklyukin: The problem with this approach is that in an extreme case it would take close to 2 * lastClusterSyncTime before the sync runs (if the informer launched that function in minute 1, the go routine called it on minute 5, bailed out and then called it for the second time at minute 10, given that the sync interval is 5 minutes).

alexeyklyukin:
👍

ikitiki:
👍

PR #47, opened 2017-06-09
Replace service annotations when updating services.

### LINKED ISSUE #45: Service updates do not change the metadata
Right now we have a bunch of clusters that transition from the old `zalando.org/dnsname` annotation to the new `external-dns.alpha.kubernetes.io/hostname` one. The sync code correctly detects that and calls patch on those services, but the patch does nothing, because it only changes the spec, and the annotation is in the metadata part of the service.

alexeyklyukin:
In case the whole annotation changes (like the external DNS) we
don't want to keep the old one hanging around. Unline specs, we
don't expect anyone except the operator to change the annotations.

Use StrategicMergePatchType in order to replace the annotations
map completely.

This fixes #45

### DIFF
--- a/pkg/cluster/util.go
+++ b/pkg/cluster/util.go
@@ -70,7 +70,8 @@ func metadataAnnotationsPatch(annotations map[string]string) string {
 		annotationsList = append(annotationsList, fmt.Sprintf(`"%s":"%s"`, name, value))
 	}
 	annotationsString := strings.Join(annotationsList, ",")
-	return fmt.Sprintf(`{"metadata":{"annotations": {"$patch":"replace", %s}}}`, annotationsString)
+	// TODO: perhaps use patchStrategy:"replace" json annotation instead of constructing the patch literally.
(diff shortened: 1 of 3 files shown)


pkg/cluster/util.go
```diff
@@ -63,6 +63,16 @@ func specPatch(spec interface{}) ([]byte, error) {
...
 }
 
+func metadataAnnotationsPatch(annotations map[string]string) string {
+	annotationsList := make([]string, 0, len(annotations))
+
+	for name, value := range annotations {
+		annotationsList = append(annotationsList, fmt.Sprintf(`"%s":"%s"`, name, value))
+	}
+	annotationsString := strings.Join(annotationsList, ",")
+	return fmt.Sprintf(`{"metadata":{"annotations": {"$patch":"replace", %s}}}`, annotationsString)
```
ikitiki: I'd put ``TODO`` here

alexeyklyukin:
👍

Jan-M:
:+1:

PR #55, opened 2017-06-26
WIP: Restrict LB by default

alexeyklyukin:
if allowedSourceRanges is not specified - lock down the load balancer to accept only local connections for security reasons.
Allow enabling/disabling the load balancer on the operator level.


ikitiki:
👍

PR #58, opened 2017-07-12
Avoid re-creating master pod if it is empty

alexeyklyukin:
Avoid error messages like this one:

time="2017-07-12T07:35:37Z" level=info msg="Recreating master pod '/'" cluster-name=team-example pkg=cluster worker=2
time="2017-07-12T07:35:37Z" level=error msg="could not sync cluster 'default/team-example': could not sync statefulsets: could not recreate pods: could not recreate master pod '/': could not delete pod: resource name may not be empty" pkg=controller worker=2


ikitiki:
👍

alexeyklyukin:
👍

PR #61, opened 2017-07-13
Added Meetup Slides about PG as a Service and Operator.

Jan-M:
Add slides, so I can link to here on twitter.


ikitiki:
why not https://www.slideshare.net ?

PR #63, opened 2017-07-14
Gracefully shutdown goroutines on operator exit

ikitiki:
Wait until goroutine gracefully shutdowns when it receives stop event

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -63,17 +63,17 @@ func New(controllerConfig *Config, operatorConfig *config.Config) *Controller {
 }
 
 func (c *Controller) Run(stopCh <-chan struct{}, wg *sync.WaitGroup) {
-	defer wg.Done()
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -140,11 +143,20 @@ func (c *Controller) initController() {
 	}
 }
 
-func (c *Controller) runInformers(stopCh <-chan struct{}) {
-	go c.postgresqlInformer.Run(stopCh)
+func (c *Controller) runPodInformer(stopCh <-chan struct{}, wg *sync.WaitGroup) {
+	defer wg.Done()
+	wg.Add(1)
```
alexeyklyukin: I'd move all wg.Add-s to the Run. BTW, we don't need wg.Done and wg.Add for the Run itself, there is no way for the execution to reach wg.Wait without exiting Run first.

PR #64, opened 2017-07-25
Make use of kubernetes client-go v4

ikitiki:
so we can get rid of k8s.io/kubernetes (and some others) as a dependency

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -58,7 +58,7 @@ func main() {
 		log.Fatalf("couldn't get REST config: %v", err)
 	}
 
-	c := controller.New(&config)
+	c := controller.NewController(&config)
 
 	c.Run(stop, wg)
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -47,17 +47,17 @@ func init() {
 func ControllerConfig() *controller.Config {
 	restConfig, err := k8sutil.RestConfig(KubeConfigFile, OutOfCluster)
 	if err != nil {
-		log.Fatalf("Can't get REST config: %s", err)
+		log.Fatalf("Can't get REST config: %v", err)
 	}
 
 	client, err := k8sutil.KubernetesClient(restConfig)
 	if err != nil {
-		log.Fatalf("Can't create client: %s", err)
+		log.Fatalf("Can't create client: %v", err)
 	}
 
 	restClient, err := k8sutil.KubernetesRestClient(restConfig)
 	if err != nil {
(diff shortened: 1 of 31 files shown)


cmd/main.go
```diff
@@ -76,32 +53,13 @@ func main() {
...
-	}
-	if configMapData["namespace"] == "" { // Namespace in ConfigMap has priority over env var
-		configMapData["namespace"] = podNamespace
-	}
-	if noDatabaseAccess {
-		configMapData["enable_database_access"] = "false"
-	}
-	if noTeamsAPI {
-		configMapData["enable_teams_api"] = "false"
+	config.RestConfig, err = k8sutil.RestConfig(KubeConfigFile, OutOfCluster)
```
alexeyklyukin: it looks like the RestConfig is the only reason why k8util is imported from main. We can move it completely to the controller. Moreover, I think we don't need to store it anywhere, since the sole reason of its existence so far is to run exec.

ikitiki: besides RestConfig there are some other vars passed to the Controller
see https://github.com/zalando-incubator/postgres-operator/pull/64/files/1659a9b06a2e6c22ce752508767cd145430b9647#diff-3002d695552ec7e7ebbaaf04a21d3439R26

alexeyklyukin: ok, actually I was to quick: RestConfig is still there in ExecCommand. But may it is possible to get it from the RestClient somehow?

The rest of the variables are good to me, (and they doesn't depend on k8util)

pkg/controller/controller.go
```diff
@@ -39,74 +43,99 @@ type Controller struct {
...
-	wg.Add(1)
+func (c *Controller) initClients() {
+	client, err := k8sutil.ClientSet(c.config.RestConfig)
+	if err != nil {
+		c.logger.Fatalf("couldn't create client: %v", err)
+	}
+	c.KubeClient = k8sutil.NewFromKubernetesInterface(client)
 
-	c.initController()
+	c.RestClient, err = k8sutil.KubernetesRestClient(*c.config.RestConfig)
```
alexeyklyukin: Why do we need to maintain a separate variable, given that we can extract RestClient from the KubeClient (i.e. exec.go:29)?

ikitiki: because restClient extracted from KubeClient is not the restClient for the kubernetes API group

alexeyklyukin:
👍

ikitiki:
👍

PR #70, opened 2017-07-31
Skip running empty set of queries

ikitiki:
Check if there are `alter user` sql statements to run before wrapping them with `DO` block


alexeyklyukin:
👍

ikitiki:
👍

PR #71, opened 2017-08-01
Fix setting debug logger level

ikitiki:
Instead of setting Level for the `logrus.Entry` we must set it for the `Logger` instead


ikitiki:
👍

alexeyklyukin:
👍

PR #73, opened 2017-08-01
Fix some golint warnings

ikitiki:
Fix the warnings generated by [gometalinter](https://github.com/alecthomas/gometalinter)


pkg/cluster/cluster.go
```diff
@@ -204,7 +211,7 @@ func (c *Cluster) Create() error {
 	}
 	c.logger.Infof("secrets have been successfully created")
 
-	ss, err := c.createStatefulSet()
+	ss, err = c.createStatefulSet()
```
alexeyklyukin: What is the issue with ss, err := c.createStatefulSet() syntax? AFAIK it would only define ss, w/o touching err (unless err is defined on an upper level)

alexeyklyukin:
👍

pkg/cluster/cluster.go
ikitiki: warning: declaration of "err" shadows declaration at ...

ikitiki:
👍

PR #81, opened 2017-08-15
Pass cluster name as a logger field

ikitiki:
Instead of mentioning cluster name in the logger message, use logger fields for that;
start all the log messages with a lowercase letters


ikitiki:
👍

alexeyklyukin:
👍

PR #83, opened 2017-08-15
Move controller config to the spec package 

ikitiki:
Moving config to the spec package will help us avoiding cyclic dependencies


ikitiki:
👍

alexeyklyukin:
👍

PR #85, opened 2017-08-15
Fix cluster event queue consumption

ikitiki:
Make cluster event queue non-blocking, so that we can dump the queue via API;
Rename Service field to the Services as it contains multiple objects;
Remove var shading


ikitiki:
👍

alexeyklyukin:
👍

PR #86, opened 2017-08-17
Discard cluster events from the queue on cluster delete

ikitiki:
If we delete a cluster we can discard all the events preceding to the deletion of the cluster


alexeyklyukin:
👍

ikitiki:
👍

PR #91, opened 2017-08-30
Use warningf instead of warnf

ikitiki:
make it more homogeneous

### DIFF
--- a/pkg/cluster/pg.go
+++ b/pkg/cluster/pg.go
@@ -69,7 +69,7 @@ func (c *Cluster) closeDbConn() (err error) {
 		}
 		c.pgDb = nil
 	}
-	c.logger.Warning("attempted to close an empty db connection object")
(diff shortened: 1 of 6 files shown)


ikitiki:
👍

pkg/cluster/pg.go
```diff
@@ -69,7 +69,7 @@ func (c *Cluster) closeDbConn() (err error) {
 		}
 		c.pgDb = nil
 	}
-	c.logger.Warning("attempted to close an empty db connection object")
+	c.logger.Warn("attempted to close an empty db connection object")
```
alexeyklyukin: Shouldn't we stick with Warning in addition to Warningf?

ikitiki: yeah, missed that

alexeyklyukin:
👍

ikitiki:
👍

PR #92, opened 2017-08-30
Add patroni api client

ikitiki:
currently patroni api client has only one method — failover

### DIFF
--- a/pkg/util/patroni/patroni.go
+++ b/pkg/util/patroni/patroni.go
@@ -63,12 +63,12 @@ func (p *Patroni) Failover(master *v1.Pod, candidate string) error {
 	}
 	defer resp.Body.Close()
 
-	bodyBytes, err := ioutil.ReadAll(resp.Body)
-	if err != nil {
-		return fmt.Errorf("could not read response: %v", err)
-	}
-
 	if resp.StatusCode != http.StatusOK {
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("could not read response: %v", err)
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/util/patroni/patroni.go
```diff
@@ -0,0 +1,76 @@
...
+func (p *Patroni) apiURL(masterPod *v1.Pod) string {
+	return fmt.Sprintf("http://%s:%d", masterPod.Status.PodIP, apiPort)
+}
+
+// Failover does manual failover via patroni api
+func (p *Patroni) Failover(master *v1.Pod, candidate string) error {
+	buf := &bytes.Buffer{}
+	err := json.NewEncoder(buf).Encode(map[string]string{"leader": master.Name, "member": candidate})
+
+	request, err := http.NewRequest(http.MethodPost, p.apiURL(master)+failoverPath, buf)
```
alexeyklyukin: a nitpick: I'd add failoverpath to the apiURL method.

ikitiki: what if we want to use other api endpoints?

alexeyklyukin:
LGTM, although I can see 2 corner cases (perhaps they deserve their own PR):

- it could happen that the failover succeeded, but we did not failover to the desired replica (if that replicas was not available), Patroni will return 200 and 'Failed over to "memberA" instead of "memberB"'. 

- it may happen that Patroni returns 503 and "Failover status unknown".

In both cases the caller should recheck the failover results and, possibly, repeat the attempt with another replica.

ikitiki:
👍

alexeyklyukin:
👍

PR #94, opened 2017-09-01
Kube cluster upgrade

ikitiki:
Once nodes get cordoned, perform graceful Pod move (with minimum number of failovers) from cordoned nodes to the new ones.


pkg/cluster/cluster.go
```diff
@@ -664,3 +664,32 @@ func (c *Cluster) GetStatus() *spec.ClusterStatus {
 		Error: c.Error,
 	}
 }
+
+// ManualFailover does manual failover to a candidate pod
+func (c *Cluster) ManualFailover(curMaster *v1.Pod, candidate spec.NamespacedName) error {
+	podLabelErr := make(chan error)
+	defer close(podLabelErr)
+
+	go func() {
```
alexeyklyukin: I see no purpose in these goroutine. Patroni will do the waiting for you and only return success after new master appears in the cluster. Besides, it won't terminate if the error is returned to the main function.

ikitiki: we need to be sure master has the label

Makefile
```diff
@@ -2,9 +2,6 @@
 
 BINARY ?= postgres-operator
 BUILD_FLAGS ?= -v
-ifeq ("$(shell uname)", "Darwin")
```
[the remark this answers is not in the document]
ikitiki: it got impossible to build a project using go 1.9 with a ``-i`` flag

pkg/controller/node.go
```diff
@@ -0,0 +1,165 @@
...
+		}
+	}
+
+	clusters := make(map[*cluster.Cluster]struct{})
+	masterPods := make(map[*v1.Pod]*cluster.Cluster)
+	replicaPods := make(map[*v1.Pod]*cluster.Cluster)
+	for _, p := range nodePods {
+		podName := util.NameFromMeta(p.ObjectMeta)
+
+		pod, err := c.KubeClient.Pods(p.Namespace).Get(p.Name, metav1.GetOptions{})
```
alexeyklyukin: can we get rid of it and just access nodePods?

pkg/controller/node.go
```diff
@@ -0,0 +1,165 @@
...
+		} else {
+			replicaPods[pod] = cl
+		}
+
+		if !ok {
+			c.logger.Warningf("%q pod is orphaned", podName)
+			continue
+		}
+
+		if util.MapContains(pod.Labels, c.opConfig.CordonedNodeLabels) {
```
alexeyklyukin: Discard, we already know the pod is running on a cordoned node.

pkg/controller/node.go
```diff
@@ -0,0 +1,165 @@
...
+			clusters[cl] = struct{}{}
+		}
+
+		if cluster.PostgresRole(role) == cluster.Master {
+			masterPods[pod] = cl
+		} else {
+			replicaPods[pod] = cl
+		}
+
+		if !ok {
```
alexeyklyukin: Looks like a misplaced piece from line 105.

pkg/cluster/k8sres.go
```diff
@@ -226,6 +226,26 @@ PATRONI_INITDB_PARAMS:
...
+		matchExpressions = append(matchExpressions, v1.NodeSelectorRequirement{
+			Key:      k,
+			Operator: v1.NodeSelectorOpNotIn,
+			Values:   []string{v},
+		})
+	}
+
+	return &v1.Affinity{
+		NodeAffinity: &v1.NodeAffinity{
+			PreferredDuringSchedulingIgnoredDuringExecution: []v1.PreferredSchedulingTerm{{
```
alexeyklyukin: Let's try Required instead of Preferred, since Preferred may lead to multiple failovers.

pkg/cluster/pod.go
```diff
@@ -91,20 +113,124 @@ func (c *Cluster) registerPodSubscriber(podName spec.NamespacedName) chan spec.P
...
+
+	node, err := c.KubeClient.Nodes().Get(newPod.Spec.NodeName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not get node %q: %v", pod.Spec.NodeName, err)
+	}
+
+	if util.MapContains(node.Labels, c.OpConfig.CordonedNodeLabels) {
+		return nil, fmt.Errorf("pod moved to %q node, which is not a non-cordoned node", node.Name)
+	}
+	c.logger.Infof("poo %q moved from %q to %q node", podName)
```
alexeyklyukin: missing old and new nodes.

pkg/cluster/pod.go
```diff
@@ -91,20 +113,124 @@ func (c *Cluster) registerPodSubscriber(podName spec.NamespacedName) chan spec.P
...
+	if newPod.Spec.NodeName == pod.Spec.NodeName {
+		return nil, fmt.Errorf("pod %q remained on the same node", podName)
+	}
+
+	node, err := c.KubeClient.Nodes().Get(newPod.Spec.NodeName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not get node %q: %v", pod.Spec.NodeName, err)
+	}
+
+	if util.MapContains(node.Labels, c.OpConfig.CordonedNodeLabels) {
```
alexeyklyukin: If we set PreferredDuringSchedulingIgnoredDuringExecution then it's ok to land on the CordonedNode, otherwise (if we choose the Required... ) then this check is fine.

alexeyklyukin: it's better to log this message and continue in any case.

pkg/cluster/pod.go
```diff
@@ -91,20 +113,124 @@ func (c *Cluster) registerPodSubscriber(podName spec.NamespacedName) chan spec.P
 	return ch
 }
 
-func (c *Cluster) recreatePod(pod v1.Pod) error {
+func (c *Cluster) movePod(pod *v1.Pod) (*v1.Pod, error) {
 	podName := util.NameFromMeta(pod.ObjectMeta)
+	if err := c.recreatePod(podName); err != nil {
+		return nil, fmt.Errorf("could not move old master pod: %v", err)
```
alexeyklyukin: not necessary master

pkg/cluster/pod.go
```diff
@@ -91,20 +113,124 @@ func (c *Cluster) registerPodSubscriber(podName spec.NamespacedName) chan spec.P
 	return ch
 }
 
-func (c *Cluster) recreatePod(pod v1.Pod) error {
+func (c *Cluster) movePod(pod *v1.Pod) (*v1.Pod, error) {
```
alexeyklyukin: Better name? (movePodOffCordonedNode)?

pkg/cluster/pod.go
```diff
@@ -91,20 +113,124 @@ func (c *Cluster) registerPodSubscriber(podName spec.NamespacedName) chan spec.P
...
 	podName := util.NameFromMeta(pod.ObjectMeta)
+	if err := c.recreatePod(podName); err != nil {
```
alexeyklyukin: Need to check first if we are on the cordoned node.

pkg/cluster/pod.go
```diff
@@ -91,20 +113,124 @@ func (c *Cluster) registerPodSubscriber(podName spec.NamespacedName) chan spec.P
...
+		return nil, fmt.Errorf("pod %q remained on the same node", podName)
+	}
+
+	node, err := c.KubeClient.Nodes().Get(newPod.Spec.NodeName, metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("could not get node %q: %v", pod.Spec.NodeName, err)
+	}
+
+	if util.MapContains(node.Labels, c.OpConfig.CordonedNodeLabels) {
+		return nil, fmt.Errorf("pod moved to %q node, which is not a non-cordoned node", node.Name)
```
alexeyklyukin: "not a non-cordoned" -> a cordoned

pkg/cluster/pod.go
```diff
@@ -91,20 +113,124 @@ func (c *Cluster) registerPodSubscriber(podName spec.NamespacedName) chan spec.P
...
+	pod, err = c.movePod(oldMaster)
+	if err != nil {
+		return fmt.Errorf("could not move pod: %v", err)
+	}
+	c.logger.Infof("pod %q has been moved to %q node", util.NameFromMeta(pod.ObjectMeta), pod.Spec.NodeName)
+
+	return nil
+}
+
+// MigrateReplicaPod recreates pod on a new
```
alexeyklyukin: on a new node

pkg/controller/util.go
```diff
@@ -122,3 +123,49 @@ func (c *Controller) podClusterName(pod *v1.Pod) spec.NamespacedName {
 
 	return spec.NamespacedName{}
 }
+
+func (c *Controller) getNodePods(nodeName string) ([]*v1.Pod, error) {
```
alexeyklyukin: not used

pkg/controller/util.go
```diff
@@ -122,3 +123,49 @@ func (c *Controller) podClusterName(pod *v1.Pod) spec.NamespacedName {
...
+			continue
+		}
+
+		pods = append(pods, &podList.Items[i])
+	}
+
+	return pods, nil
+}
+
+func (c *Controller) getPodsClusters(pods []*v1.Pod) ([]*cluster.Cluster, error) {
```
alexeyklyukin: not used

pkg/cluster/k8sres.go
```diff
@@ -226,6 +226,25 @@ PATRONI_INITDB_PARAMS:
 	return string(result)
 }
 
+func (c *Cluster) nodeAffinity() *v1.Affinity {
+	matchExpressions := make([]v1.NodeSelectorRequirement, 0)
+	for k, v := range c.OpConfig.CordonedNodeLabels {
```
Jan-M: Is the naming wrong? We want to schedule on nodes with "ready" thus the not cordoned nodes.

Jan-M:
Please also create a single pod disruption budget per cluster.

manifests/configmap.yaml
```diff
@@ -12,7 +12,7 @@ data:
   dns_name_format: '{cluster}.{team}.staging.{hostedzone}'
   docker_image: registry.opensource.zalan.do/acid/spiloprivate-9.6:1.2-p4
   etcd_host: etcd-client.default.svc.cluster.local:2379
-  secret_name_template: '{username}.{clustername}.credentials.{tprkind}.{tprgroup}'
+  secret_name_template: '{username}.{cluster}.credentials.{tprkind}.{tprgroup}'
```
Jan-M: how does this work with our currently deployed configuration?

pkg/cluster/util.go
```diff
@@ -337,15 +350,11 @@ func (c *Cluster) credentialSecretNameForCluster(username string, clusterName st
 
 	return c.OpConfig.SecretNameTemplate.Format(
 		"username", strings.Replace(username, "_", "-", -1),
-		"clustername", clusterName,
```
Jan-M: Can you add this line again? Then it works with old and new config map. We can then clean this up at a later stage.

PR #95, opened 2017-09-04
Fetch cluster resources by name, not by label selectors

ikitiki:
Currently when we load spilo cluster resources, we rely on label selectors, instead of the name. That approach might lead to a situation when we have a resource with a proper name, but without proper labels to match label selector. To avoid that we need to fetch resources directly by name, not by label selector


alexeyklyukin:
We still list pods, PVCs and PVs using labels, so it doesn't cover all cases where the labels doesn't match, but at leas it covers those that will otherwise prevent sync from running.

alexeyklyukin:
👍

ikitiki:
👍

PR #96, opened 2017-09-04
Fix cluster delete

ikitiki:
Delete cluster key from the clusters list only when delete procedure is finished.
Clusters map is used to determine cluster to send a Pod event to. If there's no cluster in the clusters map, pod event will be dismissed, which will result in pod deletion stuck.


alexeyklyukin:
👍

ikitiki:
👍

PR #100, opened 2017-09-14
Set DCS_ENABLE_KUBERNETES_API on empty etcd host var

ikitiki:
Use DCS_ENABLE_KUBERNETES_API=true environment to enable kubernetes native deployment

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -242,10 +242,6 @@ func (c *Cluster) generatePodTemplate(resourceRequirements *v1.ResourceRequireme
 			Value: constants.PostgresDataPath,
 		},
 		{
-			Name:  "ETCD_HOST",
-			Value: c.OpConfig.EtcdHost,
-		},
-		{
 			Name: "POD_IP",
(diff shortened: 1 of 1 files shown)


pkg/cluster/k8sres.go
```diff
@@ -304,11 +300,15 @@ func (c *Cluster) generatePodTemplate(resourceRequirements *v1.ResourceRequireme
 	if c.OpConfig.WALES3Bucket != "" {
 		envVars = append(envVars, v1.EnvVar{Name: "WAL_S3_BUCKET", Value: c.OpConfig.WALES3Bucket})
 	}
+
+	if c.OpConfig.EtcdHost == "" {
```
Jan-M: How is this handling not set? or null?

ikitiki: it can't be null, only empty string
https://github.com/zalando-incubator/postgres-operator/blob/master/pkg/util/config/config.go#L50

Jan-M:
:+1:

ikitiki:
👍

PR #101, opened 2017-09-15
Feature/validate role options

alexeyklyukin:
Only accept CREATE ROLE flags that doesn't have any params (i.e.
not ADMIN or CONNECTION LIMIT). Check that both flag and NOflag
are not used at the same time.

### DIFF
--- a/pkg/cluster/util.go
+++ b/pkg/cluster/util.go
@@ -34,36 +34,33 @@ func isValidFlag(flag string) bool {
 func invertFlag(flag string) string {
 	if flag[:2] == "NO" {
 		return flag[2:]
-	} else {
-		return "NO" + flag
 	}
+	return "NO" + flag
 }
 
-func normalizeUserFlags(userFlags []string) (flags []string, err error) {
+func normalizeUserFlags(userFlags []string) ([]string, error) {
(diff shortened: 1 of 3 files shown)


pkg/cluster/util.go
```diff
@@ -21,23 +21,47 @@ func isValidUsername(username string) bool {
...
+			return true
+		}
+	}
+	return false
+}
+
+func invertFlag(flag string) string {
+	if flag[:2] == "NO" {
+		return flag[2:]
+	} else {
```
ikitiki: no need for ``else``

pkg/cluster/util.go
```diff
@@ -21,23 +21,47 @@ func isValidUsername(username string) bool {
...
+		return "NO" + flag
+	}
+}
+
 func normalizeUserFlags(userFlags []string) (flags []string, err error) {
```
ikitiki: I think in this case there is no strong need for the named return values.
Besides, code like ``return nil, fmt.Errorf("bla")`` as for me looks more natural than 
```
err = fmt.Errorf("bla")
return
```

ikitiki:
👍

alexeyklyukin:
👍

PR #114, opened 2017-10-04
Perform manual failover on cluster rolling upgrade

ikitiki:
While performing rolling upgrade, before terminating master pod, do manual failover to a random replica


Jan-M:
:+1:

ikitiki:
👍

PR #117, opened 2017-10-06
Replace TPR with CRD

ikitiki:
ref https://github.com/zalando-incubator/postgres-operator/issues/98


Jan-M:
Do the example manifests need an update?

ikitiki:
actually example manifest has been changed:
https://github.com/zalando-incubator/postgres-operator/pull/117/files#diff-c2c42f271df26541b4bea761a08f6c3dL2

Jan-M:
:+1:

Jan-M:
:+1:

pkg/controller/util.go
```diff
@@ -28,36 +29,59 @@ func (c *Controller) makeClusterConfig() cluster.Config {
...
 func (c *Controller) clusterWorkerID(clusterName spec.NamespacedName) uint32 {
 	return crc32.ChecksumIEEE([]byte(clusterName.String())) % c.opConfig.Workers
 }
 
-func (c *Controller) createTPR() error {
-	tpr := thirdPartyResource(constants.TPRName)
+func (c *Controller) createCRD() error {
+	crd := &apiextv1beta1.CustomResourceDefinition{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: constants.CRDResource + "." + constants.CRDGroup,
```
alexeyklyukin: Do we need to specify the namespace here as well (since we are using the NamespacedScope)?

ikitiki: makes sense

ikitiki: actually no, CRD is cluster wide

alexeyklyukin:
👍

PR #124, opened 2017-10-10
Extend diagnostic api with worker status info

ikitiki:
ref https://github.com/zalando-incubator/postgres-operator/issues/119

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -260,7 +260,6 @@ func (c *Cluster) Create() error {
 	if err != nil {
 		c.logger.Errorf("could not list resources: %v", err)
 	}
-	c.setProcessName("")
 
 	return nil
 }
@@ -524,7 +523,6 @@ func (c *Cluster) Update(newSpec *spec.Postgresql) error {
 	}
 
 	c.setStatus(spec.ClusterStatusRunning)
-	c.setProcessName("")
 
 	return nil
 }
@@ -556,7 +554,6 @@ func (c *Cluster) Delete() error {
 			return fmt.Errorf("could not delete secret: %v", err)
 		}
 	}
-	c.setProcessName("")
 
 	return nil
 }
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -9,6 +9,7 @@ import (
 	"reflect"
 	"regexp"
 	"sync"
+	"time"
 
 	"github.com/Sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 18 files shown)


pkg/cluster/cluster.go
```diff
@@ -513,6 +524,7 @@ func (c *Cluster) Update(newSpec *spec.Postgresql) error {
 	}
 
 	c.setStatus(spec.ClusterStatusRunning)
+	c.setProcessName("")
```
alexeyklyukin: I'd really use defer here and in other places where you clear up the status (in case the relevant function exits with an error).

ikitiki: makes sense. thank you

pkg/controller/status.go
```diff
@@ -167,6 +169,24 @@ func (c *Controller) GetWorkersCnt() uint32 {
...
+		return nil, fmt.Errorf("worker has no status")
+	}
+
+	cl, ok := obj.(*cluster.Cluster)
+	if !ok {
+		return nil, fmt.Errorf("could not cast to Cluster struct")
+	}
+
+	return &spec.WorkerStatus{
+		CurrentCluster: util.NameFromMeta(cl.ObjectMeta),
```
alexeyklyukin: I think it's not concurrency safe: the properties and methods of the cluster are accessed here without any locks, neither locks are taken when they are modified. That may lead to the cluster.currentProcess being accessed while it is modified. I'd make a  special mutex and lock it on the setProcessName and GetCurrentProcess (and also the place where you change the start time).

alexeyklyukin:
The whole business of setProcessName scattered all over the code seems unnecessary ugly. Is it possible to abuse the logrus hook once again? We could do logger.WithFields('scope', 'ProcessName') in order to filter this out in a Fire callback and send to the special clusterStatus map. That might also allow us to avoid exposing cluster to the controller?

pkg/controller/status.go
Jan-M: Isn't it safe but maybe outdated?

ikitiki:
Current process name of the cluster belongs to the cluster object and must be stored there.

I don't like the idea of using ``logrus`` as a means of setting process name as ProcessName does not belong to logging anyhow.

alexeyklyukin:
Well, in fact we  don't care about what does the current cluster do, we care about what does our worker of choice do, and the cluster is just a place to report that activity, since most of the tasks the workers are doing involve the cluster. So to me that activity belongs to the controller that manages workers.

ikitiki:
If we are interested in what worker does we will end up with four processes: ADD, UPDATE, DELETE, SYNC of the cluster.
The whole Worker idea is to parallelize cluster operations. So worker status per se is the status of the cluster the worker is currently dealing with

alexeyklyukin:
👍

ikitiki:
👍

PR #126, opened 2017-10-11
Fix pod wait timeouts.

alexeyklyukin:
Previously, a timer had been reset on every message received through
the pod channel.


ikitiki:
👍

alexeyklyukin:
👍

PR #129, opened 2017-10-12
Use clear name for env var denoting namespace to watch

ikitiki:
Use clear name for env var denoting namespace to watch


Jan-M:
:+1:

ikitiki:
👍

PR #130, opened 2017-10-12
Set status of the cluster on sync fail/success

ikitiki:
In case sync of the cluster failed we need to set proper status to the cluster.
The same if sync of the cluster succeeded

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -11,9 +11,11 @@ import (
 
 // Sync syncs the cluster, making sure the actual Kubernetes objects correspond to what is defined in the manifest.
 // Unlike the update, sync does not error out if some objects do not exist and takes care of creating them.
-func (c *Cluster) Sync() (err error) {
+func (c *Cluster) Sync(newSpec *spec.Postgresql) (err error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
+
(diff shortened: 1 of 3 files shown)


pkg/cluster/sync.go
```diff
@@ -11,32 +11,43 @@ import (
 
 // Sync syncs the cluster, making sure the actual Kubernetes objects correspond to what is defined in the manifest.
 // Unlike the update, sync does not error out if some objects do not exist and takes care of creating them.
-func (c *Cluster) Sync() error {
+func (c *Cluster) Sync() (err error) {
```
alexeyklyukin: ISTM that adding named return values changes return fmt.Errorf to 
err = fmt.Errorf
return
which seems to be just extra lines. What am I missing here?

ikitiki: named return value has been introduced to catch the return value in the defer func
here: https://github.com/zalando-incubator/postgres-operator/pull/130/files/7d55f9e4bc7c11f0429a1d3753b942fd51dd3ae7#diff-bc849ac792cdc03f340b5083c721cae1R23

I'd be happy to see better solution

alexeyklyukin:
👍

alexeyklyukin:
👍

ikitiki:
👍

PR #136, opened 2017-10-16
Perform manual failover during the scale down

ikitiki:
While performing scale down, if the pod which is going to be removed happens to be a master pod, perform manual failover to the -0 pod

### DIFF
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -158,10 +158,6 @@ func (c *Cluster) preScaleDown(newStatefulSet *v1beta1.StatefulSet) error {
 		return fmt.Errorf("could not get master candidate pod: %v", err)
 	}
 
-	if c.podSpiloRole(masterCandidatePod) == Master {
-		return nil
-	}
-
 	// some sanity check
 	if !util.MapContains(masterCandidatePod.Labels, c.OpConfig.ClusterLabels) ||
 		!util.MapContains(masterCandidatePod.Labels, map[string]string{c.OpConfig.ClusterNameLabel: c.Name}) {
@@ -185,7 +181,7 @@ func (c *Cluster) updateStatefulSet(newStatefulSet *v1beta1.StatefulSet) error {
 	//scale down
 	if *c.Statefulset.Spec.Replicas > *newStatefulSet.Spec.Replicas {
 		if err := c.preScaleDown(newStatefulSet); err != nil {
-			return fmt.Errorf("could not scale down: %v", err)
+			c.logger.Warningf("could not scale down: %v", err)
 		}
 	}
 	c.logger.Debugf("updating statefulset")
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -130,7 +130,7 @@ func getPodIndex(podName string) (int32, error) {
 	postfix := parts[len(parts)-1]
 	res, err := strconv.ParseInt(postfix, 10, 32)
 	if err != nil {
-		return 0, fmt.Errorf("couldn not parse pod index: %v", err)
+		return 0, fmt.Errorf("could not parse pod index: %v", err)
 	}
 
 	return int32(res), nil
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -159,7 +159,7 @@ func (c *Cluster) preScaleDown(newStatefulSet *v1beta1.StatefulSet) error {
 		return fmt.Errorf("could not get master candidate pod: %v", err)
 	}
 
-	if c.podSpiloRole(masterCandidatePod) == string(Master) {
+	if c.podSpiloRole(masterCandidatePod) == Master {
 		return nil
 	}
 
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -128,7 +128,7 @@ func getPodNumb(podName string) (int32, error) {
 	}
 
 	postfix := parts[len(parts)-1]
-	res, err := strconv.ParseInt(postfix, 10, 64)
+	res, err := strconv.ParseInt(postfix, 10, 32)
 	if err != nil {
 		return 0, fmt.Errorf("couldn not parse postfix: %v", err)
 	}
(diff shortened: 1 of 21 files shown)


pkg/cluster/resources.go
```diff
@@ -119,13 +121,74 @@ func (c *Cluster) createStatefulSet() (*v1beta1.StatefulSet, error) {
...
 }
 
+func getPodNumb(podName string) (int32, error) {
+	parts := strings.Split(podName, "-")
+	if len(parts) == 0 {
+		return 0, fmt.Errorf("no postfix")
+	}
+
+	postfix := parts[len(parts)-1]
+	res, err := strconv.ParseInt(postfix, 10, 64)
```
alexeyklyukin: why not 32? You are converting the result to int32 anyway.

pkg/cluster/resources.go
```diff
@@ -119,13 +121,74 @@ func (c *Cluster) createStatefulSet() (*v1beta1.StatefulSet, error) {
 	return statefulSet, nil
 }
 
+func getPodNumb(podName string) (int32, error) {
```
alexeyklyukin: I'd also name it getPodIndex, as getPodNumb reads like you want to make the Pod numb.

pkg/cluster/resources.go
```diff
@@ -119,13 +121,74 @@ func (c *Cluster) createStatefulSet() (*v1beta1.StatefulSet, error) {
...
+		return fmt.Errorf("could not get master pod: %v", err)
+	}
+
+	podNum, err := getPodIndex(masterPod[0].Name)
+	if err != nil {
+		return fmt.Errorf("could not get pod number: %v", err)
+	}
+
+	//Check if scale down affects current master pod
+	if (*c.Statefulset.Spec.Replicas < podNum+1) ||
```
alexeyklyukin: Don't really get this condition: to make it should be newStatefulSet.Spec.Replicas >= podNum + 1 (i.e. if the master pod index is 2, and  the total number of replicas in the new statefulset is at least 3 - don't do anything).

pkg/cluster/resources.go
```diff
@@ -119,13 +121,74 @@ func (c *Cluster) createStatefulSet() (*v1beta1.StatefulSet, error) {
...
+func getPodIndex(podName string) (int32, error) {
+	parts := strings.Split(podName, "-")
+	if len(parts) == 0 {
+		return 0, fmt.Errorf("no pod index")
+	}
+
+	postfix := parts[len(parts)-1]
+	res, err := strconv.ParseInt(postfix, 10, 32)
+	if err != nil {
+		return 0, fmt.Errorf("couldn not parse pod index: %v", err)
```
alexeyklyukin: could not parse pod index.

pkg/cluster/resources.go
```diff
@@ -119,13 +121,74 @@ func (c *Cluster) createStatefulSet() (*v1beta1.StatefulSet, error) {
 	return statefulSet, nil
 }
 
+func getPodIndex(podName string) (int32, error) {
+	parts := strings.Split(podName, "-")
+	if len(parts) == 0 {
+		return 0, fmt.Errorf("no pod index")
```
alexeyklyukin: perhaps 'pod has no index part'?

pkg/cluster/resources.go
```diff
@@ -119,13 +121,73 @@ func (c *Cluster) createStatefulSet() (*v1beta1.StatefulSet, error) {
...
+		return nil
+	}
+
+	podName := fmt.Sprintf("%s-0", c.Statefulset.Name)
+	masterCandidatePod, err := c.KubeClient.Pods(c.OpConfig.Namespace).Get(podName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("could not get master candidate pod: %v", err)
+	}
+
+	if c.podSpiloRole(masterCandidatePod) == Master {
```
alexeyklyukin: I'd remove that. It won't happen because of line 151, since we cannot scale down to less than one pod. It might be that the masterCandidatePod will switch to master after line 151 but before line 161, but we cannot guarantee the same won't happen after line 161, and Failover will catch this anyway.

pkg/cluster/resources.go
```diff
@@ -119,13 +121,73 @@ func (c *Cluster) createStatefulSet() (*v1beta1.StatefulSet, error) {
...
 	c.setProcessName("updating statefulset")
 	if c.Statefulset == nil {
 		return fmt.Errorf("there is no statefulset in the cluster")
 	}
 	statefulSetName := util.NameFromMeta(c.Statefulset.ObjectMeta)
 
+	//scale down
+	if *c.Statefulset.Spec.Replicas > *newStatefulSet.Spec.Replicas {
+		if err := c.preScaleDown(newStatefulSet); err != nil {
+			return fmt.Errorf("could not scale down: %v", err)
```
alexeyklyukin: I think we should continue the scale down even if we couldn't move the master pod for some reason, i.e. show the warning here and go further.

pkg/util/patroni/patroni.go
```diff
@@ -78,3 +112,25 @@ func (p *Patroni) Failover(master *v1.Pod, candidate string) error {
 
 	return nil
 }
+
+// Status returns patroni status
+func (p *Patroni) Status(pod *v1.Pod) (*Status, error) {
```
alexeyklyukin: I don't see where this function is used?

pkg/util/patroni/patroni.go
```diff
@@ -78,3 +112,25 @@ func (p *Patroni) Failover(master *v1.Pod, candidate string) error {
...
+	}
+
+	resp, err := http.DefaultClient.Do(request)
+	if err != nil {
+		return nil, fmt.Errorf("could not make request: %v", err)
+	}
+
+	err = json.NewDecoder(resp.Body).Decode(status)
+	if err != nil {
+		return nil, fmt.Errorf("could not parse team API response: %v", err)
```
alexeyklyukin: looks like a copy-paste from team API.

pkg/util/patroni/patroni.go
ikitiki: no

alexeyklyukin: why does it need to be part of this PR then?

alexeyklyukin:
👍

ikitiki:
👍

PR #137, opened 2017-10-17
Retry connecting to pg

ikitiki:
- add connect_timeout parameter to the pg connection string 
- if for some net-related reasons we failed to connect to the postgresql db, try again


pkg/cluster/pg.go
```diff
@@ -45,25 +46,42 @@ func (c *Cluster) databaseAccessDisabled() bool {
...
 			if err2 := conn.Close(); err2 != nil {
 				c.logger.Errorf("error when closing PostgreSQL connection after another error: %v", err2)
+				return false, err2
 			}
-			return err
-		}
 
-		c.pgDb = conn
+			if _, ok := err.(*net.OpError); ok {
+				return false, nil
```
alexeyklyukin: Why would we hide this error?

ikitiki: you mean we should log it?

alexeyklyukin: yes, logging it would help us in finding the root cause.

alexeyklyukin:
👍

ikitiki:
👍

PR #139, opened 2017-10-18
 Fix division by zero when connecting to the DB.

alexeyklyukin:
Apparently the retry function's first parameter is the duration of
a single attempt and it cannot be zero.


ikitiki:
👍

alexeyklyukin:
👍

PR #141, opened 2017-10-19
Change spilo docker image to demospilo

CyberDem0n:
Image size is slightly more than 24MB, it doesn't contain wal-e and not suitable for production, but it is very good for demo purposes.


ikitiki:
👍

Jan-M:
:+1:

PR #147, opened 2017-10-23
Minor fixes around PDB (pod-distruption-budget) syncing:

alexeyklyukin:
- Call comparison function in the case of the sync as well as for update
- Include full cluster name in PDB name
- Assign cluster labels to the PDB object


alexeyklyukin:
👍

PR #151, opened 2017-10-30
Add node toleration config to PodSpec

geku:
Our goal is to provide "dedicated" Kubernetes nodes for Postgres instances and this PR introduces a config option to set a node toleration on the Postgres pods.

The configmap is extended with the option `toleration` which allows configuration of a Kubernetes toleration to match node taints. These nodes are then reserved for Postgres pods and don't get any other pods scheduled except the toleration is added.

If the `toleration` option is not configured, no toleration specs are added to the StatefulSet. If a toleration is defined like:

```
toleration: "key:postgres,operator:Exists,effect:NoSchedule"
```

following specification is added to the StatefulSet:

```
tolerations:
- effect: NoSchedule
  key: postgres
  operator: Exists
```

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -7,6 +7,7 @@ data:
   cluster_labels: application:spilo
   cluster_name_label: version
   pod_role_label: spilo-role
+  toleration: "key:postgres,operator:Exists,effect:NoSchedule"
   db_hosted_zone: db.example.com
   debug_logging: "true"
   dns_name_format: '{cluster}.{team}.staging.{hostedzone}'
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -7,7 +7,6 @@ data:
   cluster_labels: application:spilo
   cluster_name_label: version
   pod_role_label: spilo-role
-  toleration: "key:postgres,operator:Exists,effect:NoSchedule"
   db_hosted_zone: db.example.com
   debug_logging: "true"
   dns_name_format: '{cluster}.{team}.staging.{hostedzone}'
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -250,7 +250,12 @@ func (c *Cluster) nodeAffinity() *v1.Affinity {
 	}
 }
 
-func (c *Cluster) tolerations() []v1.Toleration {
+func (c *Cluster) tolerations(tolerationsSpec *[]v1.Toleration) []v1.Toleration {
+	// allow to override tolerations by postgresql manifest
+	if len(*tolerationsSpec) > 0 {
+		return *tolerationsSpec
+	}
+
 	podToleration := c.Config.OpConfig.PodToleration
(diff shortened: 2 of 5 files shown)


manifests/configmap.yaml
```diff
@@ -7,6 +7,7 @@ data:
   cluster_labels: application:spilo
   cluster_name_label: version
   pod_role_label: spilo-role
+  toleration: "key:postgres,operator:Exists,effect:NoSchedule"
```
Jan-M: This should  not be in the default configmap as this one is also used for minikube example and should work in most clusters which we for now assume do not have the postgres tain/toleration in place.

geku: Thanks for your reply. I thought about that and I included it for documentation purposes but understand your concern.
It does work with the toleration though as it does not require the taint on the node to get scheduled. But it's not really clean.
Any proposal where I can document it?

Jan-M: I thought using "PreferNoSchedule" is the soft border, but anyways, feel free maybe to put this into the readme as a small sub section after testing, as I believe this a valid use case/question others may have too.

pkg/util/config/config.go
```diff
@@ -25,6 +25,7 @@ type Resources struct {
 	ClusterLabels           map[string]string `name:"cluster_labels" default:"application:spilo"`
 	ClusterNameLabel        string            `name:"cluster_name_label" default:"cluster-name"`
 	PodRoleLabel            string            `name:"pod_role_label" default:"spilo-role"`
+	PodToleration          	map[string]string `name:"toleration" default:""`
```
alexeyklyukin: I think this should correspond to the default value in configmap.yaml (i.e. "key:postgres,operator:Exists,effect:NoSchedule").

geku:
I added the documentation and removed the setting from the general ConfigMap.

pkg/util/config/config.go
geku: After removing it from the general ConfigMap I think the default should be empty too.

pkg/cluster/k8sres.go
```diff
@@ -250,6 +250,22 @@ func (c *Cluster) nodeAffinity() *v1.Affinity {
 	}
 }
 
+func (c *Cluster) tolerations() []v1.Toleration {
+	podToleration := c.Config.OpConfig.PodToleration
```
Jan-M: I wonder, may be worth allowing "tolerations" field in postgresql manifest to overwrite default config, similar to other fields.

Adding tolerations to the top level of the manifest, would be aligned with e.g. resources also being there.

This would also allow specing more tolerations if needed, e.g. SSD or other custom settings specific to nodes.

Jan-M:
Georg, I added one more comment on allowing overwrite from the individual postgresql manifest, that should make it pretty generic and fine to integrate from my point of view.

Objections?

alexeyklyukin:
👍

Jan-M:
:+1:

Jan-M:
:+1:

Jan-M:
Could be merged like this, but we need to look at whether Sync/Modifications push changes down to pod level.

geku:
Thanks a lot!

PR #152, opened 2017-10-30
Custom Environment Variables for Postgres Pods

geku:
This PR adds the ability to configure a config map which is used by the Postgres pods as additional provider for environment variables. This allows to customize the Spilo image and configure it with environment variables. The config map with the additional settings is configured in the operator's main config map.

**postgres-operator ConfigMap**

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: postgres-operator
data:
  # referencing config map with custom settings
  pod_env_configmap: postgres-pod-config
  ...
```

**ConfigMap for custom settings**

```
apiVersion: v1
kind: ConfigMap
metadata:
  name: postgres-pod-config
  namespace: default
data:
  MY_CUSTOM_VAR: value
  FOR_POSTGRES: demo
```

This ConfigMap is then added to the StatefulSet/pods with following spec:

```
envFrom:
- configMapRef:
    name: postgres-pod-config
```

## Considerations

In our case we changed some scripts in the Spilo image to have a custom backup solution which needs to be configured with additional environment variables. I don't think we can contribute the Spilo changes back as they are very specific for our use cause. Therefore it doesn't make sense to add additional environment settings to the operator ConfigMap and have an explicit mapping in the operator code. And I think having the additional ConfigMap results in a very simple and generic way to allow custom variables.

What do you think? Is it useful for other use cases or against the operator paradigm in general?

Please let me know.

If the setting is omitted in the operator config, nothing is added to the pod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,38 @@ spec:
 
 Please be ware that the taint and toleration only ensures that no other pod gets scheduled to the "postgres" node but not that Postgres pods are placed on such a node. This can be achieved by setting a node affinity rule in the ConfigMap.
 
+#### Custom Pod Environment Variables
+
+It is possible to configure a config map which is used by the Postgres pods as an additional provider for environment variables.
+
+One use case is to customize the Spilo image and configure it with environment variables. The config map with the additional settings is configured in the operator's main config map:
+
+**postgres-operator ConfigMap**
+
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-operator
+data:
+  # referencing config map with custom settings
+  pod_environment_configmap: postgres-pod-config
+  ...
+```
+
+**referenced ConfigMap `postgres-pod-config`**
+
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-pod-config
+  namespace: default
+data:
+  MY_CUSTOM_VAR: value
+```
+
+This ConfigMap is then added as a source of environment variables to the Postgres StatefulSet/pods.
 
 # Setup development environment
 
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -368,6 +368,10 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *v1beta1.StatefulSet) *comp
 		needsRollUpdate = true
 		reasons = append(reasons, "new statefulset's container environment doesn't match the current one")
 	}
+	if !reflect.DeepEqual(container1.EnvFrom, container2.EnvFrom) {
+		needsRollUpdate = true
+		reasons = append(reasons, "new statefulset's container environment sources don't match the current one")
+	}
 
 	if needsRollUpdate || needsReplace {
 		match = false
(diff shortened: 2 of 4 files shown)


Jan-M:
I like the idea. Let's see what others comment here.

Though I wonder, why are you not building your own Docker Image and setting ENV vars there and just base it on ours? That is our current approach right now, slim last layer with our custom config.

Maybe sometime in the future there is a template for the whole stateful set somewhere that one could customize freely, did this for another operator before. Helps to understand what is created.

alexeyklyukin:
The approach looks good, but I'd introduce a more descriptive name instead of pod_env_configmap, perhaps just `set_pod_environment_from_configmap`?

Jan-M:
*pod_environment_configmap* would be fine with me

Jan-M:
@geku Just change the name of the field in the config map to "pod_environment_configmap" and then this looks good to us. Seems this is useful over fiddling with docker images.

geku:
Sorry was off a couple of days. I'll change the name and document it as well.

geku:
@Jan-M I rebased, changed the name and added some documentation to the readme.

Btw having a template for the pods sounds great but requires quite some more work.

Jan-M:
looks good to me. I did not know this env from feature.

Jan-M:
:+1:

Jan-M:
@alexeyklyukin Is there more work needed for the diff between statefulset pod template and the actual set? assuming the env config map is changed and the operator is restarted, does it apply the changes as we are used to?

alexeyklyukin:
@Jan-M yes, if you make changes inside the referenced configmap added in this PR the operator without changing the reference itself (i.e. the name of the configmap) the operator won't see the changes, since the statefulset will look absolutely the same, although in reality a pod restart is required to apply the changes within the configmap.

alexeyklyukin:
@geku a small fix to make statefulset comparison code aware of the envFrom: https://github.com/geku/postgres-operator/pull/1 (without it existing clusters would not pick up the environment from pod_environment_configmap).

alexeyklyukin:
@geku another idea, perhaps we should inject variables from the pod configmap into the resulting statefulset by ourself? That would allow us to detect when those variables have changed and do pods rolling update when necessary. Since that configmap scope is the operator, we can read it only once at start.

geku:
@alexeyklyukin I see how adding the env vars directly improves updating existing pods. But am I not sure what happens when the ConfigMap is updated? I guess nothing as the ConfigMap is not watched by the operator.

Jan-M:
This is true for both the conifgmap of the operator itself as for your configmap for the env variables.

For now we actually work with redeploying/restarting the operator on such changes before working on "runtime" config change support.

alexeyklyukin:
@geku the operator doesn't watch the configmap, but it does sync each cluster every sync period (5 minutes by default). During the sync the operator will generate the new statefulset definition based on. the updated contents of the configmap and trigger a rolling upgrade if old and new definitions don't match each other.

Here's a PR against your branch implementing this idea: https://github.com/geku/postgres-operator/pull/4

I aimed to make its  behaviour less surprising by sticking to the same rules as envSource: the configmap is only searched in the local cluster namespace, and pod variables have priority over those fetched from the PodEnvironmentConfigMap.

Jan-M:
@geku @alexeyklyukin what else is needed here?

geku:
@Jan-I just merged the change by @alexeyklyukin into my branch, looks good from my side.

alexeyklyukin:
@geku thank you!
Rebased it to the recent master with https://github.com/zalando-incubator/postgres-operator/pull/175 to eliminate conflicts. Going to run a few manual tests before merging.

geku:
Closing this PR in favor of the rebased branch on #175

alexeyklyukin:
@geku merged your changes, thanks a lot!

PR #153, opened 2017-11-01
Create new databases and change owners of existing ones during sync.

alexeyklyukin:
And Minor code refactoring.

PR #154, opened 2017-11-01
use sync methods while updating the cluster

ikitiki:
In order to avoid logic duplication we should use the same methods while updating the cluster as we use in sync

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -606,41 +606,36 @@ func (c *Cluster) initSystemUsers() {
 	}
 }
 
+// syncRobotUsers syncs adds users from the operator Spec
+// to the lsit of users known to the operator.
 func (c *Cluster) syncRobotUsers() error {
-	curUsers := make(map[string]struct{})
-	for k := range c.pgUsers {
-		curUsers[k] = struct{}{}
-	}
-	for username, userFlags := range c.Spec.Users {
-		if !isValidUsername(username) {
-			return fmt.Errorf("invalid username: '%v'", username)
+	for newUsername, newUserFlags := range c.Spec.Users {
+		if !isValidUsername(newUsername) {
+			return fmt.Errorf("invalid username: %q", newUsername)
 		}
 
-		flags, err := normalizeUserFlags(userFlags)
+		newUserFlagsNormalized, err := normalizeUserFlags(newUserFlags)
 		if err != nil {
-			return fmt.Errorf("invalid flags for user '%v': %v", username, err)
+			return fmt.Errorf("invalid flags for user %q: %v", newUsername, err)
 		}
 
-		if _, ok := c.pgUsers[username]; !ok {
-			delete(c.pgUsers, username)
-		}
-
-		if u, ok := c.pgUsers[username]; ok {
-			u.Flags = flags
-			continue
-		}
-		delete(curUsers, username)
-		c.pgUsers[username] = spec.PgUser{
-			Name:     username,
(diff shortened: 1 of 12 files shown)


pkg/cluster/sync.go
```diff
@@ -107,73 +78,159 @@ func (c *Cluster) Sync(newSpec *spec.Postgresql) (err error) {
...
 	}
-	c.logServiceChanges(role, c.Services[role], desiredSvc, false, reason)
 
-	if err := c.updateService(role, desiredSvc); err != nil {
-		return fmt.Errorf("could not update %s service to match desired state: %v", role, err)
+	c.logger.Infof("could not find the cluster's %s service", role)
+
+	if svc, err := c.createService(role); err != nil {
+		if k8sutil.ResourceAlreadyExists(err) {
+			c.logger.Infof("%s service %q already exists", role, util.NameFromMeta(svc.ObjectMeta))
```
alexeyklyukin: in this case we won't reach the outer else block, that sets c.Services[role] = sec, so c.Services[role] will be nil despite the fact that the actual service exists.

pkg/cluster/sync.go
```diff
@@ -107,73 +78,159 @@ func (c *Cluster) Sync(newSpec *spec.Postgresql) (err error) {
...
+	// Endpoint does not exist
+	if role == Replica && !c.Spec.ReplicaLoadBalancer {
+		return nil
+	}
+
+	c.logger.Infof("could not find the cluster's %s endpoint", role)
+
+	if ep, err := c.createEndpoint(role); err != nil {
+		if k8sutil.ResourceAlreadyExists(err) {
+			c.logger.Infof("%s endpoint %q already exists", role, util.NameFromMeta(ep.ObjectMeta))
```
alexeyklyukin: same issue as with the services, c.Endpoints will have no knowledge of the existing endpoint.

pkg/cluster/volumes.go
```diff
@@ -146,11 +146,11 @@ func (c *Cluster) resizeVolumes(newVolume spec.Volume, resizers []volumes.Volume
 }
 
 func (c *Cluster) volumesNeedResizing(newVolume spec.Volume) (bool, error) {
-	volumes, manifestSize, err := c.listVolumesWithManifestSize(newVolume)
+	vols, manifestSize, err := c.listVolumesWithManifestSize(newVolume)
```
alexeyklyukin: why rename volumes to vols? Doesn't seem to make the code clearer.

pkg/cluster/cluster.go
```diff
@@ -256,8 +257,8 @@ func (c *Cluster) Create() error {
 			return fmt.Errorf("could not create users: %v", err)
 		}
 
-		if err = c.createDatabases(); err != nil {
-			return fmt.Errorf("could not create databases: %v", err)
+		if err = c.syncDbs(); err != nil {
```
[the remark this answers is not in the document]
ikitiki: I think you should merge PR #153 and then I'll merge master into this branch

pkg/cluster/volumes.go
ikitiki: because we have package named ``volumes``, and having the variable with the same name is not a good idea

pkg/cluster/cluster.go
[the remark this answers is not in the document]
alexeyklyukin: will do

alexeyklyukin:
👍

ikitiki:
👍

PR #161, opened 2017-11-08
Allow to specify global user options in the operator configuration

alexeyklyukin:
Introduce new api_roles_configuration parameter, which is a map of name, value configuration parameters that are applied to the role obtained from the API with ALTER ROLE SET.

As a byproduct, add support for parameters to the PgUser structure, making it easier to implement support for custom options in other places, i.e. in the `users` field  of the cluster manifest.

There is a special case for the search_path in order to strip quotes from the value that would be normally quoted in order to be put into the configMap value when multiple search paths are specified.

Added the parameter configuration description in the README, we may find a better place to put those in the future.

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -39,3 +39,4 @@ data:
   pod_terminate_grace_period: 5m
   pdb_name_format: "postgres-{cluster}-pdb"
   eol_node_label: "eol:true"
+  api_roles_configuration: "log_statement:all"
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -39,4 +39,4 @@ data:
   pod_terminate_grace_period: 5m
   pdb_name_format: "postgres-{cluster}-pdb"
   eol_node_label: "eol:true"
-  api_roles_configuration: "log_statement:all"
+  team_api_role_configuration: "log_statement:all"
--- a/pkg/util/config/util.go
+++ b/pkg/util/config/util.go
@@ -175,6 +175,7 @@ const (
 	Plain parserState = iota
 	DoubleQuoted
 	SingleQuoted
+	Escape
 )
 
 // Split the pair candidates by commas not located inside open quotes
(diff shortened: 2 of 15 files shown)


manifests/configmap.yaml
```diff
@@ -39,3 +39,4 @@ data:
   pod_terminate_grace_period: 5m
   pdb_name_format: "postgres-{cluster}-pdb"
   eol_node_label: "eol:true"
+  api_roles_configuration: "log_statement:all"
```
Jan-M: is there a reason you are not calling it teams_api_roles_configuration?

alexeyklyukin: no particular one, good idea.

alexeyklyukin:
👍

pkg/util/config/util.go
```diff
@@ -166,6 +169,55 @@ func processField(value string, field reflect.Value) error {
...
+type parserState int
+
+const (
+	Plain parserState = iota
+	DoubleQuoted
+	SingleQuoted
+)
+
+// Split the pair candidates by commas not located inside open quotes
+// Escape characters are not supported for simplicity, as we don't
```
zerg-junior: Please add an explicit check for escape characters.

manifests/configmap.yaml
```diff
@@ -39,3 +39,4 @@ data:
   pod_terminate_grace_period: 5m
   pdb_name_format: "postgres-{cluster}-pdb"
   eol_node_label: "eol:true"
+  team_api_role_configuration: "log_statement:all"
```
zerg-junior: Please add  an example of specifying multiple config params. I feel the YAML comment will suffice.

alexeyklyukin: I'm reluctant to add anything to this configmap just for the demonstration purposes, as it has been serving a good place to show the operator defaults before and, obviously, I don't want to modify the operator defaults for this particular purpose. Instead, I've added the string with multiple keys to the parameter description in the README.

alexeyklyukin:
👍

manifests/configmap.yaml
zerg-junior: i think that's OK

zerg-junior:
:+1:

PR #164, opened 2017-11-13
Allow per-cluster setting of a docker image.

alexeyklyukin:
Add dockerImage cluster configuration parameter that overrides global
operator defaults when set to a non-empty value.


Jan-M:
:+1:

alexeyklyukin:
👍

Issue #166, opened 2017-11-17
Operator should not re-write password already set by the infrastructure roles


zerg-junior:
The role with the same name may be defined simultaneously at  `infrastructure-roles.yaml ` and `postgres-manifest.yaml`. In this case the operator currently overwrites the password defined at the  `infrastructure-roles.yaml ` with the random password.

alexeyklyukin:
Fixed by #168

PR #168, opened 2017-11-25
Avoid overwriting infrastructure roles.

alexeyklyukin:
For a role defined in both the infrastructure roles secret and the cluster manifest use the infrastructure role definition and add flags defined in the manifest. #166

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -629,7 +629,7 @@ func (c *Cluster) initRobotUsers() error {
 			// avoid overwriting the password if the user is already there. The flags should be
 			// merged here, but since there is no mechanism to define them for non-robot roles
 			// they are assigned from the robot user.
-			c.logger.Debugf("merging user %q data", username)
+			c.logger.Debugf("merging manifest and infrastructure user %q data", username)
 			user := c.pgUsers[username]
 			user.Flags = flags
 			c.pgUsers[username] = user
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -654,7 +654,7 @@ func (c *Cluster) initHumanUsers() error {
 		}
 
 		if _, present := c.pgUsers[username]; present {
-			c.logger.Warnf("overwriting existing user %q with the data from the teams API")
+			c.logger.Warnf("overwriting existing user %q with the data from the teams API", username)
 		}
 
 		c.pgUsers[username] = spec.PgUser{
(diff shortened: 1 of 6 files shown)


pkg/cluster/cluster.go
```diff
@@ -645,6 +653,10 @@ func (c *Cluster) initHumanUsers() error {
 			}
 		}
 
+		if _, present := c.pgUsers[username]; present {
+			c.logger.Warnf("overwriting existing user %q with the data from the teams API")
```
zerg-junior: 1. Should the username be passed to the Warnf call as a value ?
2. Is the data from the teams API the only thing that can trigger the overwrite ?

zerg-junior:
@alexeyklyukin Could you please add a small unit test for this case to avoid regressions in the future ?

pkg/cluster/cluster.go
alexeyklyukin: good catch for the username, teams API roles will overwrite roles with the same name obtained from the infrastructure roles or the cluster manifests, but cluster manifest roles will merge with the infrastructure one (same PR), so teams API roles are the only ones that can overwrite other roles.

alexeyklyukin:
Added tests for the initHumanUser and initRobotUsers

pkg/cluster/cluster.go
```diff
@@ -610,18 +612,27 @@ func (c *Cluster) initSystemUsers() {
 func (c *Cluster) initRobotUsers() error {
 	for username, userFlags := range c.Spec.Users {
 		if !isValidUsername(username) {
-			return fmt.Errorf("invalid username: '%v'", username)
+			return fmt.Errorf("invalid username: %q", username)
```
zerg-junior: Shall we briefly mention here how valid usernames look like ?

pkg/cluster/cluster.go
```diff
@@ -610,18 +612,27 @@ func (c *Cluster) initSystemUsers() {
...
+			c.pgUsers[username] = spec.PgUser{
+				Name:     username,
+				Password: util.RandomPassword(constants.PasswordLength),
+				Flags:    flags,
+			}
+		} else {
+			// avoid overwriting the password if the user is already there. The flags should be
+			// merged here, but since there is no mechanism to define them for non-robot roles
+			// they are assigned from the robot user.
+			c.logger.Debugf("merging user %q data", username)
```
zerg-junior: Shall this message also say in which places exactly user is defined twice ?

zerg-junior:
:+1:

pkg/cluster/cluster.go
alexeyklyukin: good point.

pkg/cluster/cluster.go
alexeyklyukin: It would be too long to fit into the error message, I think it's a task of the documentation.

alexeyklyukin:
👍

PR #171, opened 2017-11-30
Tweak our reaction to the cluster upgrade process.

alexeyklyukin:
Previously, the operator started to move the pods off the nodes to be
decomissioned by watching the eol_node_label value. Every new postgres
pod has been created with the anti-affinity to that label, making sure
that the pods being moved won't land on another to be decomissioned
node.

The changes introduce another label that indicates the ready node.  The
new pod affinity will ensure that the pod is only scheduled to the node
marked as ready, discarding the previous anti-affinity.  That way the
nodes can transition from the pending-decomission to the other statuses
(drained, terminating) without having pods suddently scaled to them.

In addition, rename the label that triggers the start of the upgrade
process to node_eol_label (for consistency with node_readiness_label)
and set its default vvalue to lifecycle-status:pending-decomission.

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -38,5 +38,6 @@ data:
   cluster_history_entries: "1000"
   pod_terminate_grace_period: 5m
   pdb_name_format: "postgres-{cluster}-pdb"
-  eol_node_label: "eol:true"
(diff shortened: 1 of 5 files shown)


alexeyklyukin:
👍

Jan-M:
:+1:

PR #172, opened 2017-12-04
Avoid overwriting critical users.

alexeyklyukin:
Disallow defining new users either in the cluster manifest, teams
API or infrastructure roles with the names mentioned in the new
protected_role_names parameter (it is a list).

Additionally, forbid defining a user with the name matching either
super_username or replication_username, so that we don't overwrite
system roles required for correct working of the operator itself.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -615,6 +615,9 @@ func (c *Cluster) initRobotUsers() error {
 			return fmt.Errorf("invalid username: %q", username)
 		}
 
+		if c.avoidProtectedOrSystemRole(username, "manifest robot role") {
+			continue
+		}
 		flags, err := normalizeUserFlags(userFlags)
 		if err != nil {
 			return fmt.Errorf("invalid flags for user %q: %v", username, err)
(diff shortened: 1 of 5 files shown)


pkg/cluster/cluster.go
```diff
@@ -687,6 +701,18 @@ func (c *Cluster) initInfrastructureRoles() error {
 	return nil
 }
 
+func (c *Cluster) avoidProtectedOrSystemRole(username, purpose string) bool {
```
zerg-junior: Probably `shouldAvoidProtectedOrSystemRole` or just `isProtectedOrSystemRole` ? I find adjectives easier to read in `if-then-else`

zerg-junior:
Is it worth it to add a quick test for this capability ?

pkg/cluster/cluster.go
alexeyklyukin: shouldAvoid is good. I thought about isProtectedOrSystemRole, but the function actually emits a warning inside, so it not just gives a value to make a decision, but also suggests interprets it.

alexeyklyukin:
I've added some additional checks to the existing tests.

alexeyklyukin:
👍

zerg-junior:
:+1:

PR #175, opened 2017-12-13
custom-env-rebased

alexeyklyukin:
Use a ConfigMap set by the operator configuration option `pod_environment_configmap` in order to supply arbitrary environment variables to Spilo.
On conflict between the options supplied by the operator and those listed in the pod environment ConfigMap the operator options win.

This is a rebase of https://github.com/zalando-incubator/postgres-operator/pull/152/ to the recent master to eliminate conflicts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,38 @@ spec:
 Please be aware that the taint and toleration only ensures that no other pod gets scheduled to a PostgreSQL node 
 but not that PostgreSQL pods are placed on such a node. This can be achieved by setting a node affinity rule in the ConfigMap.
 
+#### Custom Pod Environment Variables
+
+It is possible to configure a config map which is used by the Postgres pods as an additional provider for environment variables.
+
+One use case is to customize the Spilo image and configure it with environment variables. The config map with the additional settings is configured in the operator's main config map:
+
+**postgres-operator ConfigMap**
+
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-operator
+data:
+  # referencing config map with custom settings
+  pod_environment_configmap: postgres-pod-config
+  ...
+```
+
+**referenced ConfigMap `postgres-pod-config`**
+
+```
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-pod-config
+  namespace: default
+data:
+  MY_CUSTOM_VAR: value
+```
+
+This ConfigMap is then added as a source of environment variables to the Postgres StatefulSet/pods.
 
 # Setup development environment
 
--- a/README.md
+++ b/README.md
@@ -178,6 +178,8 @@ data:
 
 This ConfigMap is then added as a source of environment variables to the Postgres StatefulSet/pods.
 
+:exclamation: Note that there are environment variables defined by the operator itself in order to pass parameters to the Spilo image. The values from the operator for those variables will take precedence over those defined in the `pod_environment_configmap`.
+
 # Setup development environment
 
 The following steps guide you through the setup to work on the operator itself.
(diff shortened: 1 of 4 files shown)


pkg/cluster/k8sres.go
```diff
@@ -362,6 +363,26 @@ func (c *Cluster) generatePodTemplate(resourceRequirements *v1.ResourceRequireme
 	if cloneDescription.ClusterName != "" {
 		envVars = append(envVars, c.generateCloneEnvironment(cloneDescription)...)
 	}
+
+	var names []string
+	// handle environment variables from the PodEnvironmentConfigMap. We don't use envSource here as it is impossible
```
zerg-junior: What is  `envSource` ? I cannot find code that references it.

pkg/cluster/k8sres.go
```diff
@@ -362,6 +363,26 @@ func (c *Cluster) generatePodTemplate(resourceRequirements *v1.ResourceRequireme
...
+	// to track any changes to the object envSource points to. In order to emulate the envSource behavior, however, we
+	// need to make sure that PodConfigMap variables doesn't override those we set explicitely from the configuration
+	// parameters
+	envVarsMap := make(map[string]string)
+	for _, envVar := range envVars {
+		envVarsMap[envVar.Name] = envVar.Value
+	}
+	for name := range extraEnv {
+		if _, ok := envVarsMap[name]; !ok {
```
zerg-junior: I feel this case is well worth a warning  cause in the case of accidental name clash this may be superhard to debug.

README.md
```diff
@@ -145,6 +145,38 @@ spec:
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgres-pod-config
+  namespace: default
+data:
+  MY_CUSTOM_VAR: value
+```
+
+This ConfigMap is then added as a source of environment variables to the Postgres StatefulSet/pods.
```
zerg-junior: We probably should add a short paragraph here saying that at least k8s and  CDP set quite a lot of their own `env` variables so there is tiny but non-zero chance of name clash.

pkg/cluster/k8sres.go
```diff
@@ -426,8 +447,15 @@ func (c *Cluster) generateStatefulSet(spec *spec.PostgresSpec) (*v1beta1.Statefu
 	if err != nil {
 		return nil, fmt.Errorf("could not generate resource requirements: %v", err)
 	}
-
-	podTemplate := c.generatePodTemplate(resourceRequirements, &spec.Tolerations, &spec.PostgresqlParam, &spec.Patroni, &spec.Clone, &spec.DockerImage)
+	var extraEnv map[string]string
```
zerg-junior: sth like `customPodEnvVars` may be a better name ?

pkg/cluster/k8sres.go
alexeyklyukin: https://kubernetes.io/docs/api-reference/v1.8/#configmapenvsource-v1-core

README.md
alexeyklyukin: I wouldn't state the obvious, but it is still missing a note on the priority of the variables from the pod configmap compared to those set by the operator itself.

pkg/cluster/k8sres.go
alexeyklyukin: Good point

README.md
zerg-junior: Agreed, that's worth adding

alexeyklyukin:
👍

zerg-junior:
:+1:

geku:
Thanks a lot for your work!

PR #177, opened 2017-12-14
Warn on infrastructure-roles.yaml format violations

### LINKED ISSUE #167: Operator should warn if secrets are empty
The operator should log a warning if the secrets it reads appear to be empty.  Rationale: a user may inadvertently write a `.yaml` file with broken syntax so that the file looks meaningful to the user but empty to the operator.

zerg-junior:
Closes #167

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -137,11 +137,17 @@ Users:
 					c.logger.Warningf("unknown key %q", p)
 				}
 			}
+
+			delete(data, key)
 		}
 
 		if t.Name != "" {
 			result[t.Name] = t
 		}
+	}
+
+	if len(data) != 0 {
+		c.logger.Warningf("Unprocessed entries in the infrastructure roles' secret. That may mean infrastructure-roles.yaml has incorrect format; see the file in the Postgres operator repo for the expected format.")
 	}
 
 	return result, nil
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -147,7 +147,9 @@ Users:
 	}
 
 	if len(data) != 0 {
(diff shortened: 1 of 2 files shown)


zerg-junior:
I traced down the original problem from #167 to the way the operator processes `infrastructure-roles.yaml`. The operator properly creates the infra roles' secret, but passes to downstream sync routines only the data adhering [to a special format](https://github.com/zalando-incubator/postgres-operator/blob/0e255f82c6be8b57ba2da11bbbc5c6b9da0636ec/pkg/controller/util.go#L115).

For now, I documented the expected format of `infrastructure-roles.yaml` and added a warning for the case of suspected failure.

pkg/controller/util.go
```diff
@@ -137,13 +137,19 @@ Users:
...
+			delete(data, key)
 		}
 
 		if t.Name != "" {
 			result[t.Name] = t
 		}
 	}
 
+	if len(data) != 0 {
+		c.logger.Warningf("Unprocessed entries in the infrastructure roles' secret. That may mean infrastructure-roles.yaml has incorrect format; see the file in the Postgres operator repo for the expected format.")
```
alexeyklyukin: I'd show the number of entries and not mention the manifest file in the comments, since it may have a different name, also we may move the example file and forget to change this message. Something like (you can also avoid hard-coding the list of allowed keys):

```
c.logger.Warningf("%d unprocessed entries for the infrastructure roles", len(data))
c.logger.Info(`infrastructure role entries should be in the {key}{id} format, where {key} can be either of "user", "password", "inrole" and the {id} a monotonically increasing integer starting with 1`)
c.logger.Debugf("unprocessed entries: %#v", data)
```

zerg-junior:
:+1:

alexeyklyukin:
👍

PR #178, opened 2017-12-15
Introduce higher and lower bounds for the number of instances

alexeyklyukin:
Reduce the number of instances to the min_instances if it is lower and
to the max_instances if it is higher. -1 for either of those means there
is no lower or upper bound.

In addition, terminate the operator when there is a nonsense in the
configuration (i.e. max_instances < min_instances).

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,12 @@ This ConfigMap is then added as a source of environment variables to the Postgre
 
 :exclamation: Note that there are environment variables defined by the operator itself in order to pass parameters to the Spilo image. The values from the operator for those variables will take precedence over those defined in the `pod_environment_configmap`.
 
+### Limiting the number of instances in clusters with `min_instances` and `max_instances`
+
+As a preventive measure, one can restrict the minimum and the maximum number of instances permitted by each Postgres cluster managed by the operator.
+If either `min_instances` or `max_instances` is set to a non-zero value, the operator may adjust the number of instances specified in the cluster manifest to match either the min or the max boundary.
+For instance, of a cluster manifest has 1 instance and the min_instances is set to 3, the cluster will be created with 3 instances. By default, both parameters are set to -1.
+
 # Setup development environment
 
 The following steps guide you through the setup to work on the operator itself.
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -464,6 +464,8 @@ func (c *Cluster) generateStatefulSet(spec *spec.PostgresSpec) (*v1beta1.Statefu
 		return nil, fmt.Errorf("could not generate volume claim template: %v", err)
 	}
 
+	numberOfInstaces := c.getNumberOfInstances(spec)
+
 	statefulSet := &v1beta1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      c.statefulSetName(),
@@ -471,7 +473,7 @@ func (c *Cluster) generateStatefulSet(spec *spec.PostgresSpec) (*v1beta1.Statefu
 			Labels:    c.labelsSet(),
 		},
 		Spec: v1beta1.StatefulSetSpec{
-			Replicas:             &spec.NumberOfInstances,
+			Replicas:             &numberOfInstaces,
 			ServiceName:          c.serviceName(Master),
 			Template:             *podTemplate,
 			VolumeClaimTemplates: []v1.PersistentVolumeClaim{*volumeClaimTemplate},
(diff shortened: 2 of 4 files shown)


Jan-M:
:+1:

pkg/cluster/k8sres.go
```diff
@@ -464,14 +464,16 @@ func (c *Cluster) generateStatefulSet(spec *spec.PostgresSpec) (*v1beta1.Statefu
 		return nil, fmt.Errorf("could not generate volume claim template: %v", err)
 	}
 
+	numberOfInstaces := c.getNumberOfInstances(spec)
```
zerg-junior: instaNces ?

pkg/util/config/config.go
```diff
@@ -33,6 +34,8 @@ type Resources struct {
 	PodEnvironmentConfigMap string            `name:"pod_environment_configmap" default:""`
 	NodeEOLLabel            map[string]string `name:"node_eol_label" default:"lifecycle-status:pending-decommission"`
 	NodeReadinessLabel      map[string]string `name:"node_readiness_label" default:"lifecycle-status:ready"`
+	MaxInstances            int32             `name:"max_instances" default:"-1"`
```
zerg-junior: shall we comment here on the special meaning of -1 ?

pkg/util/config/config.go
```diff
@@ -123,3 +129,13 @@ func Copy(c *Config) Config {
 
 	return cfg
 }
+
+func validate(cfg *Config) (err error) {
+	if cfg.MinInstances > 0 && cfg.MaxInstances > 0 && cfg.MinInstances > cfg.MaxInstances {
+		err = fmt.Errorf("minimum number of instances %d is set higher than the maximum number %d")
+	}
+	if cfg.Workers == 0 {
+		err = fmt.Errorf("number of works should be higher than 0")
```
zerg-junior: workERs ?

pkg/util/config/config.go
```diff
@@ -123,3 +129,13 @@ func Copy(c *Config) Config {
...
+	}
+	return
+}
```
mgomezch: Nothing really wrong with this, but I'm curious: why check all errors but only panic on the last one? (not really important as there are just two, but could this grow in the future?  it would probably be most helpful to log all of them and then panic)

pkg/util/config/config.go
```diff
@@ -123,3 +129,13 @@ func Copy(c *Config) Config {
 
 	return cfg
 }
+
+func validate(cfg *Config) (err error) {
+	if cfg.MinInstances > 0 && cfg.MaxInstances > 0 && cfg.MinInstances > cfg.MaxInstances {
+		err = fmt.Errorf("minimum number of instances %d is set higher than the maximum number %d")
+	}
+	if cfg.Workers == 0 {
```
zerg-junior: cfg.Workers > 0 ? It may be tempting to supply -1 meaning no limit :)

pkg/cluster/k8sres.go
```diff
@@ -481,6 +483,25 @@ func (c *Cluster) generateStatefulSet(spec *spec.PostgresSpec) (*v1beta1.Statefu
...
+	newcur = cur
+
+	if max > 0 && newcur > max {
+		newcur = max
+	}
+	if min > 0 && newcur < min {
+		newcur = min
+	}
+	if newcur != cur {
+		c.logger.Infof("adjusted number of instances to %d (min: %d, max: %d)", newcur, min, max)
```
zerg-junior: should the log mention the old value as well ?

mgomezch:
:+1:

pkg/util/config/config.go
alexeyklyukin: Yes, didn't care much about that for now (also, there are many other parameters that could be validated), just picked up a low-hanging fruit in the config to validate, since we had no validation before altogether.

pkg/util/config/config.go
alexeyklyukin: it's unit32.

README.md
```diff
@@ -180,6 +180,12 @@ This ConfigMap is then added as a source of environment variables to the Postgre
 
 :exclamation: Note that there are environment variables defined by the operator itself in order to pass parameters to the Spilo image. The values from the operator for those variables will take precedence over those defined in the `pod_environment_configmap`.
 
+### Limiting the number of instances in clusters with `min_instances` and `max_instances`
+
+As a preventive measure, one can restrict the minimum and the maximum number of instances permitted by each Postgres cluster managed by the operator.
+If either `min_instances` or `max_instances` is set to a non-zero value, the operator may adjust the number of instances specified in the cluster manifest to match either the min or the max boundary.
```
zerg-junior: and this goes to the operator configmap, right ?

alexeyklyukin: Yes, but there is no need to mention that, since that is part of the description of the operator parameters.

pkg/util/config/config.go
alexeyklyukin: That went to the docs.

zerg-junior:
:+1:

alexeyklyukin:
👍

PR #184, opened 2017-12-18
Improve behavior on node decomissionining

alexeyklyukin:
- use the lack of lifecycle-status: ready label to trigger the migration
- make sure the nodes are examined on start of the operator in order to avoid missing a migration of the set of nodes when the operator is killed in the process.

### DIFF
--- a/pkg/controller/node.go
+++ b/pkg/controller/node.go
@@ -55,13 +55,21 @@ func (c *Controller) nodeUpdate(prev, cur interface{}) {
 		return
 	}
 
-	if nodePrev.Spec.Unschedulable && util.MapContains(nodePrev.Labels, c.opConfig.NodeEOLLabel) ||
-		!nodeCur.Spec.Unschedulable || !util.MapContains(nodeCur.Labels, c.opConfig.NodeEOLLabel) {
+	// do nothing if the node should have already triggered an update or
+	// if only one of the label and the unschedulability criteria are met.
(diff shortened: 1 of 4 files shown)


Jan-M:
Just for clarity, the migration is only triggered IF the ready label is missing and the node is marked as unscheduleable.

alexeyklyukin:
👍

Jan-M:
:+1:

pkg/controller/node.go
```diff
@@ -55,13 +59,23 @@ func (c *Controller) nodeUpdate(prev, cur interface{}) {
...
+	// do nothing if the node should have already triggered an update or
+	// if only one of the label and the unschedulability criteria are met.
+	if !c.nodeIsReady(nodePrev) || c.nodeIsReady(nodeCur) {
 		return
 	}
+	c.movePodsOffNode(nodeCur)
+}
+
+func (c *Controller) nodeIsReady(node *v1.Node) bool {
+	return (!node.Spec.Unschedulable || util.MapContains(node.Labels, c.opConfig.NodeReadinessLabel) ||
```
linki: shouldn't the `!` outside of the parenthesis?

alexeyklyukin: Hm, the node is considered ready if  is not labeled as unschedulable or if it has a ready label or if it is a master. Otherwise (if it is unschedulable and not master and doesn't have a ready label) the node is not considered ready and we move pods off it.

linki: Alright, your statement makes sense and I also see your tests now. Thanks for clarification 👍

PR #189, opened 2017-12-19
Add an example for cloning a backup from existing cluster

### LINKED ISSUE #188: Add an example clone option to the manifest

zerg-junior:
Closes #188

### DIFF
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -43,8 +43,9 @@ spec:
     loop_wait: &loop_wait 10
     retry_timeout: 10
     maximum_lag_on_failover: 33554432
-  # uncomment to restore a Postgres DB with point-in-time-recovery using the latest backup
-  # optional timestamp: use the latest backup before the timestamp
+  # restore a Postgres DB with point-in-time-recovery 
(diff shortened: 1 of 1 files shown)


manifests/complete-postgres-manifest.yaml
```diff
@@ -43,6 +43,11 @@ spec:
     loop_wait: &loop_wait 10
     retry_timeout: 10
     maximum_lag_on_failover: 33554432
+  # uncomment to restore a Postgres DB with point-in-time-recovery using the latest backup
+  # optional timestamp: use the latest backup before the timestamp
+  # clone:
+  #  cluster: "acid-batman"
+  #  endTimestamp: "2017-12-19T12:40:33+01:00" # timezone required (offset relative to UTC, see RFC 3339 section 5.6)
```
alexeyklyukin: The operator would actually involve different logic if endTimestamp is empty, cloning from the host that is named the same as the cluster name and port 5432 (connecting to the service that way, i.e. acid-batman would normally point to acid-batman.default.svc.cluster.local assuming the default namespace), reading the replication password from the secret of the respective cluster. That implies that the startup will fail if the cluster to clone from is not running and endTimestamp is not specified or empty.

alexeyklyukin:
👍

zerg-junior:
:+1:

PR #190, opened 2017-12-20
Scalyr agent sidecar for log shipping

mgomezch:
When the `scalyr_api_key` key is defined in the configmap, this adds a sidecar container named `scalyr-agent` to PostgreSQL pods that ships all logs to Scalyr.

The sidecar image is configurable with the `scalyr_image` configmap key and defaults to the currently newest CDP build from [the currently-private Spilo Scalyr sidecar repository](https://github.bus.zalan.do/acid/spilo-sidecar-scalyr) in the private image registry.

That image is very simple: it just installs the Scalyr agent and creates a configuration file that consumes the environment variables set by the Operator.  It should be easy to make that repository public if we build the image on top of [the official Scalyr agent image on Docker Hub](https://hub.docker.com/r/scalyr/scalyr-agent/).  This was not done for the private image as I understand compliance requires approved base images for anything deployed internally.

We could also use the public Scalyr agent image as the default directly, although the agent config file would have to be generated e.g. with an init container, which would add inconvenient complexity to the Operator code.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -318,9 +318,6 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *v1beta1.StatefulSet) *comp
 				needsRollUpdate = true
 				reasons = append(reasons, fmt.Sprintf("new statefulset's container %d environment sources don't match the current one", index))
 			}
-			if needsRollUpdate || needsReplace {
-				match = false
-			}
 		}
 	}
 	if len(c.Statefulset.Spec.Template.Spec.Containers) == 0 {
@@ -380,6 +377,10 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *v1beta1.StatefulSet) *comp
 			needsReplace = true
 			reasons = append(reasons, fmt.Sprintf("new statefulset's volumeClaimTemplates specification for volume %q doesn't match the current one", name))
 		}
+	}
+
+	if needsRollUpdate || needsReplace {
+		match = false
 	}
 
 	return &compareStatefulsetResult{match: match, reasons: reasons, rollingUpdate: needsRollUpdate, replace: needsReplace}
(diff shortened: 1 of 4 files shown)


pkg/cluster/k8sres.go
```diff
@@ -431,6 +435,33 @@ func (c *Cluster) generatePodTemplate(resourceRequirements *v1.ResourceRequireme
...
 
+	if c.OpConfig.ScalyrAPIKey != "" && c.OpConfig.ScalyrImage != "" {
+		podSpec.Containers = append(
+			podSpec.Containers,
+			v1.Container{
+				Name:            "scalyr-sidecar",
+				Image:           c.OpConfig.ScalyrImage,
+				ImagePullPolicy: v1.PullIfNotPresent,
+				Resources:       *resourceRequirementsScalyrSidecar,
+				VolumeMounts:    volumeMounts,
```
alexeyklyukin: Why is there a need to mount the postgres data volume to the scalyr agent container?

mgomezch: Because it needs to watch the log files for changes, and the log files are in the Postgres data volume.  We could also put the log files in a separate volume, but I don't know if we want to do this.

pkg/cluster/cluster.go
```diff
@@ -295,6 +295,33 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *v1beta1.StatefulSet) *comp
...
+			}
+			if !reflect.DeepEqual(container1.Env, container2.Env) {
+				needsRollUpdate = true
+				reasons = append(reasons, fmt.Sprintf("new statefulset's container %d environment doesn't match the current one", index))
+			}
+			if !reflect.DeepEqual(container1.EnvFrom, container2.EnvFrom) {
+				needsRollUpdate = true
+				reasons = append(reasons, fmt.Sprintf("new statefulset's container %d environment sources don't match the current one", index))
+			}
+			if needsRollUpdate || needsReplace {
```
alexeyklyukin: should be outside of the else condition.

mgomezch:
Tested on playground. :+1:

mgomezch:
:+1:

alexeyklyukin:
👍

PR #196, opened 2017-12-21
Add PostgreSQL pod namespace Scalyr sidecar environment

mgomezch:
Another tiny bit of information that could be useful for log filters once we start deploying clusters into separate namespaces.


mgomezch:
:+1:

alexeyklyukin:
👍

PR #199, opened 2018-01-04
Migrate only master pods. Migrate single masters.

alexeyklyukin:
Avoid migrating replica pods, since they will be handled by the
node draining anyway (the PDB specifies that only masters are to
be kept).

Allow migration of the single-pod clusters.

### DIFF
--- a/pkg/cluster/pod.go
+++ b/pkg/cluster/pod.go
@@ -24,7 +24,7 @@ func (c *Cluster) listPods() ([]v1.Pod, error) {
 	return pods.Items, nil
 }
 
-func (c *Cluster) getRolePods(role PostgresRole) ([]v1.Pod, error) {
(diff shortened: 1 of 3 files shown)


pkg/cluster/pod.go
```diff
@@ -24,7 +24,7 @@ func (c *Cluster) listPods() ([]v1.Pod, error) {
 	return pods.Items, nil
 }
 
-func (c *Cluster) getRolePods(role PostgresRole) ([]v1.Pod, error) {
+func (c *Cluster) getRolePods(role PostgresRole, emptyOk bool) ([]v1.Pod, error) {
```
Jan-M: Feels like this could be done and checked where invoked, but ok.

Jan-M:
:+1:

alexeyklyukin:
👍

PR #200, opened 2018-01-09
WIP: support multiple namespaces.

alexeyklyukin:
This particular commit makes the operator listen on all namespaces
and introduces an optional namespace to the clusters RESET API
endpoints.

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -30,9 +30,9 @@ type controllerInformer interface {
 	GetOperatorConfig() *config.Config
 	GetStatus() *spec.ControllerStatus
 	TeamClusterList() map[string][]spec.NamespacedName
-	ClusterStatus(team, cluster string) (*spec.ClusterStatus, error)
-	ClusterLogs(team, cluster string) ([]*spec.LogEntry, error)
-	ClusterHistory(team, cluster string) ([]*spec.Diff, error)
+	ClusterStatus(team, namespace, cluster string) (*spec.ClusterStatus, error)
+	ClusterLogs(team, namespace, cluster string) ([]*spec.LogEntry, error)
+	ClusterHistory(team, namespace, cluster string) ([]*spec.Diff, error)
 	ClusterDatabasesMap() map[string][]string
 	WorkerLogs(workerID uint32) ([]*spec.LogEntry, error)
 	ListQueue(workerID uint32) (*spec.QueueDump, error)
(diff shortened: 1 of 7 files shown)


pkg/apiserver/apiserver.go
```diff
@@ -30,9 +30,9 @@ type controllerInformer interface {
 	GetOperatorConfig() *config.Config
 	GetStatus() *spec.ControllerStatus
 	TeamClusterList() map[string][]spec.NamespacedName
-	ClusterStatus(team, cluster string) (*spec.ClusterStatus, error)
-	ClusterLogs(team, cluster string) ([]*spec.LogEntry, error)
-	ClusterHistory(team, cluster string) ([]*spec.Diff, error)
+	ClusterStatus(team, namespace, cluster string) (*spec.ClusterStatus, error)
```
zerg-junior: 1. Since we now move to multiple namespaces it might make sense to add extra API endpoints for debugging purposes such as:
a. List namespaces operator actually listens to.
b. List databases per namespace
c. List namespaces per team

2. Please add/change 1-2 examples of the new API in [the debugging documentation](https://github.com/zalando-incubator/postgres-operator#debugging-the-operator-itself)

pkg/apiserver/apiserver.go
```diff
@@ -149,7 +149,8 @@ func (s *Server) clusters(w http.ResponseWriter, req *http.Request) {
 	)
 
 	if matches := util.FindNamedStringSubmatch(clusterStatusURL, req.URL.Path); matches != nil {
-		resp, err = s.controller.ClusterStatus(matches["team"], matches["cluster"])
+		namespace, _ := matches["namespace"]
```
zerg-junior: Does a separate var for `namespace` exist because the namespace is undefined  in some requests  and we need to ignore the error in such cases ?

pkg/controller/node.go
```diff
@@ -80,7 +80,7 @@ func (c *Controller) moveMasterPodsOffNode(node *v1.Node) {
 	opts := metav1.ListOptions{
 		LabelSelector: labels.Set(c.opConfig.ClusterLabels).String(),
 	}
-	podList, err := c.KubeClient.Pods(c.opConfig.Namespace).List(opts)
+	podList, err := c.KubeClient.Pods("").List(opts)
```
zerg-junior: Please comment on the effect of using the empty string aka `""` as a namespace name. Does it mean "listen to all namespaces" ? I traced it down to the [`Namespace` method](https://github.com/kubernetes/client-go/blob/4b76cf9824ec474ca9b122449fee23807a51e786/rest/request.go#L243) of the k8s go-client, and still wasn't quite sure.

zerg-junior:
1.  How does this PR handle the [`WATCH_NAMESPACE` parameter](https://github.com/zalando-incubator/postgres-operator/blob/702d901bd9a92208b86e7504bb0ee1a133e98d54/cmd/main.go#L30) ?

2.   @alexeyklyukin suggests it makes sense to enable restricting operator to a few namespaces instead of always listening to all of them. That would help testing because there we would like to have an operator listening only to the testing namespace. Also there may be cases where the operator has access rights only to certain namespaces.
 Any ideas on that ? @Jan-M ?

zerg-junior:
Closed in favor of PR #225

Issue #203, opened 2018-01-11
Can't schedule pod on any node with HEAD


ksala:
Hi,

I've tried to compile the latest HEAD (56359d2) and use it on a local minikube cluster.
I can build the docker image successfully, but when deployed the pods in the stateful set can't be scheduled with this error:
```
No nodes are available that match all of the predicates: MatchNodeSelector (1).
```

I've tried with commit c17aabb (the one used in the same postgres-operator.yaml) and it works fine.
I've also tried on a cluster on GKE and get the same results.

Not sure if I'm doing something wrong here :/ Google is not really helping.
I can try to bifork and see which commit doesn't work, but I'd prefer to have someone check first if I'm doing something terribly wrong.

Thanks a lot!

---

Steps to reproduce:
```
git clone https://github.com/zalando-incubator/postgres-operator.git
cd postgres-operator
make deps
make docker
# edit manifest/postgres-operator.yaml to use the image just built
minikube start --vm-driver=kvm2
kubectl --context minikube create -f manifests/configmap.yaml
kubectl --context minikube create -f manifests/serviceaccount.yaml
kubectl --context minikube create -f manifests/postgres-operator.yaml
kubectl --context minikube create -f manifests/minimal-postgres-manifest.yaml
```
Logs:
Operator logs:
```
kubectl --context minikube logs -f postgres-operator-7b4c7cc6d7-94762
2018/01/11 16:29:41 Spilo operator 56359d2
[...]

PR #204, opened 2018-01-11
Set node_readiness_label default to an empty value.

alexeyklyukin:
Previously, it was set to the lifecycle-status:ready, breaking a
lot of minikube deployments. Also it was not possible before to run
with this label set to an empty value.

Document the effect of the label in the new section of the
documentation.

Addresses #203

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ This project is currently in active development. It is however already [used int
 
 There is a talk about this project delivered by Josh Berkus on KubeCon 2017: [Kube-native Postgres](https://www.youtube.com/watch?v=Zn1vd7sQ_bc)
 
-Please, report any issues discovered to htts://github.com/zalando-incubator/postgres-operator/issues.
+Please, report any issues discovered to https://github.com/zalando-incubator/postgres-operator/issues.
 
 ## Running and testing the operator
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ This project is currently in active development. It is however already [used int
 
 There is a talk about this project delivered by Josh Berkus on KubeCon 2017: [Kube-native Postgres](https://www.youtube.com/watch?v=Zn1vd7sQ_bc)
 
-Please, report any issues discovered to https://github.com/zalando-incubator/postgres-operator/issues.
+Please, report any issues discovered to httnodeps://github.com/zalando-incubator/postgres-operator/issues.
 
 ## Running and testing the operator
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -148,6 +148,30 @@ spec:
...
+give us the following guarantees:
+
+* no postgres pod runs on the node being not ready
+* no postgres master pod is killed during the scale event before the operator has a chance to move it off the node
+* when a postgres pod is killed by the operator because the node is decomissioned, it will not respawn on another node to-be-decomissioned.
+
+By default the `node_readiness_label` is set to an empty string, disabling this feature altogether. It can be set to a
+string containing one or more key:value parameters, i.e:
+```
+node_readiness_label: "lifecycle-status:ready,disagnostic-checks:ok"
```
zerg-junior: So this is effectively a *set* of labels used to determine if a node is ready to run postgres pods ?

README.md
```diff
@@ -148,6 +148,30 @@ spec:
 Please be aware that the taint and toleration only ensures that no other pod gets scheduled to a PostgreSQL node 
 but not that PostgreSQL pods are placed on such a node. This can be achieved by setting a node affinity rule in the ConfigMap.
 
+### Using the operator to minimize the amount of failovers during the cluster upgrade
+
+To make sure there is only one failover per each PostgreSQL deployment when nodes of the cluster are decomissioned
```
zerg-junior: I feel the explanation from Jan is worth adding here:

> We use this readiness label during migrations of the underlying nodes to avoid scheduling pods/masters on nodes that will also be recycled

README.md
```diff
@@ -148,6 +148,30 @@ spec:
...
+label the operator starts moving master pods out of it. In addition, the operator sets the `PodDisruptionBudget` for the
+postgres pods to make sure pods with the role `master` are not killed by the cluster scale out operations. Each postgres
+pod has a `nodeAffinity` set to avoid being scheduled on the nodes without the readiness label. Together, those measures
+give us the following guarantees:
+
+* no postgres pod runs on the node being not ready
+* no postgres master pod is killed during the scale event before the operator has a chance to move it off the node
+* when a postgres pod is killed by the operator because the node is decomissioned, it will not respawn on another node to-be-decomissioned.
+
+By default the `node_readiness_label` is set to an empty string, disabling this feature altogether. It can be set to a
```
zerg-junior: 1. Do we also need to set this default in the [`configmap.yaml`](https://github.com/zalando-incubator/postgres-operator/blob/dd0affc390b0779e9218d8521033434a339aa929/manifests/configmap.yaml#L42) ? 

2. It may not be immediately obvious that an empty string disables the feature. So we probably should at least emphasize that particular statement in the docs, e.g., by typing "by default " in bold.

Jan-M:
:+1:

zerg-junior:
:+1:

README.md
alexeyklyukin: Well, this is right below
`when a postgres pod is killed by the operator because the node is decomissioned, it will not respawn on another node to-be-decomissione`

README.md
alexeyklyukin: Setting it in a configmap is good idea, although I think the docs state the meaning of an empty value clear enough to avoid emphasizing it further.

README.md
alexeyklyukin: Yes, it is a set of labels, although it is used not only to determine if a node is ready ro run postgres pods, but also whether the pods should be migrated from a given node.

alexeyklyukin:
👍

Jan-M:
:+1:

PR #205, opened 2018-01-12
Overwrite scalyr api key if the relevant env var is present in the operator pod

### LINKED ISSUE #202: Allow overwrite of scalyr api key via operator env var
To allow referencing an existing secret, which is not possible in the config map, the scalyr api key should be overwritable by a pod environment variable.

Should be simply set and checked in init config method.

zerg-junior:
Closes #202

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"fmt"
+	"os"
 	"sync"
 
 	"github.com/Sirupsen/logrus"
@@ -107,6 +108,11 @@ func (c *Controller) initOperatorConfig() {
 	}
 
 	c.opConfig = config.NewFromMap(configMapData)
+
+	scalyrAPIKey := os.Getenv("SCALYR_API_KEY")
+	if scalyrAPIKey != "" {
+		c.opConfig.ScalyrAPIKey = scalyrAPIKey
+	}
 }
 
 func (c *Controller) initController() {


Jan-M:
:+1:

zerg-junior:
:+1:

mgomezch:
Just to confirm: we don't want to mount the secret on the Scalyr sidecar container for each database pod, and instead, we want to mount the secret as an environment variable on the operator itself, and have it generate StatefulSets with the API key written directly into its pod template?

Note that if we expect the Scalyr API key to ever change, the former approach (as opposed to the one currently implemented here) would allow us to update the Scalyr API key for all database pods in a cluster without recreating them, as the Scalyr sidecar script could reload the API key from a mounted file each time it tries to start the agent.  I don't know if this is worth doing, though.

mgomezch:
:+1:

alexeyklyukin:
I agree with Manuel's approach: why don't we put the API KEY to the secret and simply mount it for the Scalyr pod, passing only the secret's name to the operator config. This way, we won't expose the Scalyr key via the REST API and will avoid contaminating the operator code with Scalyr minutia.

alexeyklyukin:
The other complaint about this PR is that the env variables should be handed together in main, there is no reason for them to go to the controller.go.

alexeyklyukin:
One of the disadvantages of the current approach is that we will return our secret Scalyr key via the operator's REST API. We won't have this problem when passing only the secret name via the configuration.

Issue #149, opened 2017-10-24
Sync/Recreate Infrastructure Roles on operator startup


Jan-M:
It would be very helpful for the most basic rotation or in case of mistakes in infrastructure role secret if we would create all roles again on startup. This should include altering existing infrastructure roles and changing their password.

- [ ] on startup change password of existing infrastructure roles to what is currently in the secret.

Jan-M:
As we now found out the problem is actually that the once created secrets for the infrastructure roles still contain a password. So theses need to be updated along the way. @alexeyklyukin

CyberDem0n:
:+1:

PR #206, opened 2018-01-12
update individual role secrets from infrastructure roles

alexeyklyukin:
When the password in the infrastructure role is updated, re-generate the
secret for that role.

Previously, the password for an infrastructure role was always fetched from
the secret, making any updates to such role a no-op after the corresponding
secret had been generated.

Addresses #149

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -320,6 +320,10 @@ func (c *Cluster) syncSecrets() error {
 			if err2 != nil {
 				return fmt.Errorf("could not get current secret: %v", err2)
 			}
+			if secretUsername != string(curSecret.Data["username"]) {
+				c.logger.Warningf("secret %q does not contain the role %q", secretSpec.Name, secretUsername)
+				continue
+			}
 			c.logger.Debugf("secret %q already exists, fetching its password", util.NameFromMeta(curSecret.ObjectMeta))
 			if secretUsername == c.systemUsers[constants.SuperuserKeyName].Name {
 				secretUsername = constants.SuperuserKeyName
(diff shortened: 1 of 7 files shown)


pkg/cluster/sync.go
```diff
@@ -331,8 +335,17 @@ func (c *Cluster) syncSecrets() error {
 				userMap = c.pgUsers
 			}
 			pwdUser := userMap[secretUsername]
-			pwdUser.Password = string(curSecret.Data["password"])
-			userMap[secretUsername] = pwdUser
+			// if this secret belongs to the infrastructure role and the password has changed - replace it in the secret
+			if pwdUser.Password != string(curSecret.Data["password"]) && pwdUser.Origin == spec.RoleOriginInfrastructure {
+				c.logger.Debugf("updating the secret %q from the infrastructure roles", secretSpec.Name)
+				if _, err := c.KubeClient.Secrets(secretSpec.Namespace).Update(secretSpec); err != nil {
+					return fmt.Errorf("could not update infrastructure role secret for role %q: %v", secretUsername, err)
```
zerg-junior: Is it worth adding to this error message the infra secret name such as `postgresql-infrastructure-roles` ?

pkg/cluster/sync.go
```diff
@@ -331,8 +335,17 @@ func (c *Cluster) syncSecrets() error {
...
+				}
+			} else {
+				// for non-infrastructure role - update the role with the password from the secret
```
zerg-junior: 1. So the operator is supposed to forcibly overwrite the passwords using secrets if the passwords change for whatever reason ?
2. Are non-infra-role passwords supposed to change w/o changing the relevant secrets ? If not, does this code branch deserve a warning log statement ?

zerg-junior:
As a general comment, I feel this change is worth (1) a test and (b) a note in the `readme`

pkg/cluster/sync.go
Jan-M: It is basically a feature for the infrastructure roles.

alexeyklyukin: The only reason for changing the role's password could be changes in the infrastructure roles secret. The idea behind these PR is to enable anyone who has corresponding permissions to change the password for an infrastructure role throughout the cluster by simply changing it in the infrastructure roles secret, providing a single place for setting passwords (and other properties as well). Before that PR, your changes of the password in the infrastructure role had no effect unless you remove the corresponding secret generated for that role.

pkg/cluster/sync.go
alexeyklyukin: yes, good idea.

pkg/cluster/sync.go
alexeyklyukin: Answering question (2), the only user-definable non-infrastructure roles that have corresponding secrets are manifest roles, but we deliberately not allow to set passwords in the manifest. The only authoritative place to store those roles passwords are the corresponding secrets generated for them, so, contrary to the infrastructure roles, we don't have multiple places where we can set secrets and there is no need to sync them.

Jan-M:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
👍

erthalion:
:+1:

PR #208, opened 2018-01-15
Improve infrastructure role definitions

alexeyklyukin:
Allow defining infrasturcture roles via configmaps.
See manifests/configmap_infrastructure_roles.yaml for an example.
Passwords should still be specified in the infrastructure roles
secret.

In addition, merge role definitions for the roles with the same
name using roles origins instead of implicit reliance on the order
of their initialization in the operator.

### DIFF
--- a/manifests/configmap_infrastructure_roles.yaml
+++ b/manifests/configmap_infrastructure_roles.yaml
@@ -1,0 +1,17 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: postgresql-infrastructure-roles
+data:
+  batman: |
+    name: batman
+    inrole: [role1, role2]
+    options:
+      - superuser
+      - login
+    parameters:
+      log_statements: all
+  robin: |
+    name: robin
+    inrole:
+      - postgres
--- a/manifests/infrastructure-roles.yaml
+++ b/manifests/infrastructure-roles.yaml
@@ -11,7 +11,8 @@ data:
   # testuser
   user2: dGVzdHVzZXI=
   # foobar
-  password2: Zm9vYmFy
+  password2: c2VjcmV0
+  batman: anVzdGljZQ==
 kind: Secret
 metadata:
   name: postgresql-infrastructure-roles
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -629,10 +629,12 @@ func (c *Cluster) initSystemUsers() {
 	// secrets, therefore, setting flags like SUPERUSER or REPLICATION
 	// is not necessary here
 	c.systemUsers[constants.SuperuserKeyName] = spec.PgUser{
+		Origin:   spec.RoleOriginSystem,
 		Name:     c.OpConfig.SuperUsername,
 		Password: util.RandomPassword(constants.PasswordLength),
 	}
 	c.systemUsers[constants.ReplicationUserKeyName] = spec.PgUser{
+		Origin:   spec.RoleOriginSystem,
 		Name:     c.OpConfig.ReplicationUsername,
 		Password: util.RandomPassword(constants.PasswordLength),
 	}
@@ -653,6 +655,7 @@ func (c *Cluster) initRobotUsers() error {
 		}
 		if _, present := c.pgUsers[username]; !present {
 			c.pgUsers[username] = spec.PgUser{
+				Origin:   spec.RoleOriginManifest,
 				Name:     username,
 				Password: util.RandomPassword(constants.PasswordLength),
 				Flags:    flags,
@@ -696,6 +699,7 @@ func (c *Cluster) initHumanUsers() error {
 		}
 
 		c.pgUsers[username] = spec.PgUser{
+			Origin:		spec.RoleOriginTeamsAPI,
 			Name:       username,
 			Flags:      flags,
 			MemberOf:   memberOf,
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -653,24 +653,23 @@ func (c *Cluster) initRobotUsers() error {
 		if err != nil {
 			return fmt.Errorf("invalid flags for user %q: %v", username, err)
 		}
+		newRole := spec.PgUser{
+			Origin:   spec.RoleOriginManifest,
+			Name:     username,
+			Password: util.RandomPassword(constants.PasswordLength),
+			Flags:    flags,
+		}
 		if _, present := c.pgUsers[username]; !present {
-			c.pgUsers[username] = spec.PgUser{
-				Origin:   spec.RoleOriginManifest,
-				Name:     username,
-				Password: util.RandomPassword(constants.PasswordLength),
-				Flags:    flags,
-			}
+			c.pgUsers[username] = newRole
 		} else {
-			// avoid overwriting the password if the user is already there. The flags should be
-			// merged here, but since there is no mechanism to define them for non-robot roles
-			// they are assigned from the robot user.
-			c.logger.Debugf("merging manifest and infrastructure user %q data", username)
 			user := c.pgUsers[username]
-			user.Flags = flags
-			c.pgUsers[username] = user
+			if mergeRole, err := c.mergeRoleDefinitions(&user, &newRole); err != nil {
+				return err
+			} else {
+				c.pgUsers[username] = *mergeRole
+			}
 		}
 	}
-
 	return nil
 }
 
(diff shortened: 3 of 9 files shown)


pkg/spec/postgresql.go
```diff
@@ -223,6 +223,7 @@ type postgresqlCopy Postgresql
 
 // UnmarshalJSON converts a JSON into the PostgreSQL object.
 func (p *Postgresql) UnmarshalJSON(data []byte) error {
+	fmt.Printf("unmarshaling %s\n", string(data))
```
zerg-junior: Leftovers from debugging ?

pkg/controller/util.go
```diff
@@ -134,22 +143,43 @@ Users:
 				case "inrole":
 					t.MemberOf = append(t.MemberOf, s)
 				default:
-					c.logger.Warningf("unknown key %q", p)
+					c.logger.Warningf("unkonwn key %q", p)
```
zerg-junior: "unknown" :)

pkg/spec/types.go
```diff
@@ -28,6 +28,16 @@ const (
...
+type RoleOrigin int
+
+const (
+	RoleOriginSystem = iota
+	RoleOriginInfrastructure
+	RoleOriginManifest
+	RoleOriginTeamsAPI
+	RoleOriginUnknown
+)
+
```
zerg-junior: we have this defined in  two separate PRs. Shall we merge [that one](https://github.com/zalando-incubator/postgres-operator/pull/206) first ?

manifests/configmap_infrastructure_roles.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: postgresql-infrastructure-roles
+data:
+  batman: |
+    name: batman
+    inrole: [role1, role2]
+    options:
+      - superuser
+      - login
+    parameters:
```
zerg-junior: The difference between `options` and `parameters` is not immediately clear

pkg/cluster/cluster.go
```diff
@@ -725,6 +735,31 @@ func (c *Cluster) initInfrastructureRoles() error {
 	return nil
 }
 
+func (c *Cluster) mergeRoleDefinitions(roleA, roleB *spec.PgUser) (*spec.PgUser, error) {
+	// human roles always win
+	// infrastructure roles take precendence over the manifest roles
+	if roleA.Name != roleB.Name {
+		return nil, fmt.Errorf("could not merge role %#v with role %#v: role names don't match")
```
zerg-junior: Are references to `roleA` and `roleB` missing  in `fmt.Errorf`?

pkg/cluster/cluster.go
```diff
@@ -725,6 +735,31 @@ func (c *Cluster) initInfrastructureRoles() error {
 	return nil
 }
 
+func (c *Cluster) mergeRoleDefinitions(roleA, roleB *spec.PgUser) (*spec.PgUser, error) {
+	// human roles always win
+	// infrastructure roles take precendence over the manifest roles
```
zerg-junior: manifest role == role defined in the postgres manifest?

pkg/cluster/cluster.go
```diff
@@ -725,6 +735,31 @@ func (c *Cluster) initInfrastructureRoles() error {
...
+	if roleA.Name != roleB.Name {
+		return nil, fmt.Errorf("could not merge role %#v with role %#v: role names don't match")
+	}
+	for _, origin := range []spec.RoleOrigin{spec.RoleOriginSystem, spec.RoleOriginTeamsAPI} {
+		if roleA.Origin == origin {
```
zerg-junior: do we warn users somewhere if they define roles that coincidentally have the same name as one the system roles ?

pkg/cluster/cluster.go
```diff
@@ -725,6 +735,31 @@ func (c *Cluster) initInfrastructureRoles() error {
...
+			return roleA, nil
+		}
+		if roleB.Origin == origin {
+			return roleB, nil
+		}
+	}
+	// one of the roles originates from the manifest, another one from the infrastructure roles
+	if roleA.Origin == spec.RoleOriginInfrastructure {
+		roleA.Flags = roleB.Flags
```
zerg-junior: Does this code overwrite the flags of the infra role with the flags of the manifest role ?  In other words, do manifest roles' flags have higher priority here ?

pkg/controller/util.go
```diff
@@ -134,22 +143,43 @@ Users:
...
+						delete(secretData, role)
+					}
+					roleDescr.Name = role
+					roleDescr.Origin = spec.RoleOriginInfrastructure
+					result[role] = *roleDescr
+				}
+			}
+		}
+		if len(secretData) > 0 {
+			c.logger.Warningf("%d unprocessed entries in the infrastructure roles' secret", len(secretData))
```
zerg-junior: Worth a secret name ?

pkg/controller/util.go
```diff
@@ -134,22 +143,43 @@ Users:
...
+				if roleDescr, err := readDecodedRole(s); err != nil {
+					// check if we have a a password in a configmap
+					return nil, fmt.Errorf("could not decode role description: %v", err)
+				} else {
+					c.logger.Debugf("found role description for role %q: %+v", role, roleDescr)
+					if passwd, ok := secretData[role]; ok {
+						roleDescr.Password = string(passwd)
+						delete(secretData, role)
+					}
+					roleDescr.Name = role
```
mgomezch: This means repeating the name within the entry value's YAML is redundant and ultimately ignored, right?  I think this is better, but then it should be removed from the examples. (Unmarshalling should still work and just leave the name as an empty string, no?)

pkg/controller/util.go
```diff
@@ -134,22 +143,43 @@ Users:
...
+		// perhaps we have some map entries with usernames, passwords, let's check if we have those users in the configmap
+		if infraRolesMap, err := c.KubeClient.ConfigMaps(rolesSecret.Namespace).Get(rolesSecret.Name, metav1.GetOptions{}); err == nil {
+			// we have a configmap with username - json description, let's read and decode it
+			for role, s := range infraRolesMap.Data {
+				if roleDescr, err := readDecodedRole(s); err != nil {
+					// check if we have a a password in a configmap
+					return nil, fmt.Errorf("could not decode role description: %v", err)
+				} else {
+					c.logger.Debugf("found role description for role %q: %+v", role, roleDescr)
+					if passwd, ok := secretData[role]; ok {
```
mgomezch: We sometimes want to get the password (or even the username) from some known key in a different secret that is known to be reliably present in the cluster.  For example, we have to get the ZMON username and password out of this secret: https://github.com/zalando-incubator/kubernetes-on-aws/pull/797

It would be useful to extend the scope of this change a bit so that it allows us to specify infrastructure roles in the ConfigMap but asking for the role name and password to be read from given keys in other secrets.

This would of course require using a more complex struct than `PgUser` for the infrastructure roles ConfigMap values.  I think it's worth doing, though, as we do actually need this right now.

pkg/controller/util.go
erthalion: Looking at the code I'm not sure about how "ultimately" it's being ignored - so far it's ignored only in the result of this `getInfrastructureRoles` function.

pkg/controller/util.go
erthalion: Let's do this in acid/database-as-a-service#150

manifests/configmap_infrastructure_roles.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: postgresql-infrastructure-roles
+data:
+  batman: |
+    name: batman
+    inrole: [role1, role2]
+    user_flags:
+      - superuser
+      - login
+    db_parameters:
```
zerg-junior: Since it's our example infra role manifest, could you please shortly comment inline on what `inrole, user_flags and db_parameters` mean ?

manifests/infrastructure-roles.yaml
```diff
@@ -11,7 +11,8 @@ data:
   # testuser
   user2: dGVzdHVzZXI=
   # foobar
-  password2: Zm9vYmFy
+  password2: c2VjcmV0
+  batman: anVzdGljZQ==
```
zerg-junior: Please update the docs to mention (a) what infra roles are for and (b) that their passwords are still distributed as k8s secrets.

pkg/cluster/cluster.go
```diff
@@ -756,6 +762,38 @@ func (c *Cluster) initInfrastructureRoles() error {
...
+// return it (possibly with some modifications), taking into account following
+// rules:
+// * Human roles always win
+// * Infrastructure roles take precendence over the manifest roles
+// * If both two roles have RoleOriginSystem or RoleOriginTeamsAPI,
+//	 the last one (roleB) wins. It's necessary to be able to e.g. extend roles
+//	 flags as shown in cluster_test.go:TestInitHumanUsers
+func (c *Cluster) mergeRoleDefinitions(roleA, roleB *spec.PgUser) (*spec.PgUser, error) {
+	if roleA.Name != roleB.Name {
+		return nil, fmt.Errorf("could not merge role %#v with role %#v: role names don't match")
```
zerg-junior: missing references to values ?

manifests/configmap_infrastructure_roles.yaml
erthalion: I've added few commentaries for `inrole` and `db_parameters`, `user_flags` is self-explanatory I think.

erthalion:
:+1:

alexeyklyukin:
👍

alexeyklyukin:
👍

erthalion:
:-1:

erthalion:
:+1:

PR #209, opened 2018-01-15
Allow configuring a NodeSelector in the Postgres kind

ksala:
This add support for adding the NodeSelector label to the Postgresql kind.  
We use this in a GKE cluster which has two node pools, default and preemptible, to have the Postgres pods scheduled only on the normal nodes.

### DIFF
--- a/pkg/spec/postgresql.go
+++ b/pkg/spec/postgresql.go
@@ -105,6 +105,7 @@ type PostgresSpec struct {
 	ClusterName         string               `json:"-"`
 	Databases           map[string]string    `json:"databases,omitempty"`
 	Tolerations         []v1.Toleration      `json:"tolerations,omitempty"`
+	NodeSelector        map[string]string    `json:"NodeSelector,omitempty"`
 }
 
 // PostgresqlList defines a list of PostgreSQL clusters.
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -282,6 +282,7 @@ func (c *Cluster) generatePodTemplate(
 	tolerationsSpec *[]v1.Toleration,
 	pgParameters *spec.PostgresqlParam,
 	patroniParameters *spec.Patroni,
+	nodeSelector *map[string]string,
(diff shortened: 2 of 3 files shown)


pkg/spec/postgresql.go
```diff
@@ -105,6 +105,7 @@ type PostgresSpec struct {
 	ClusterName         string               `json:"-"`
 	Databases           map[string]string    `json:"databases,omitempty"`
 	Tolerations         []v1.Toleration      `json:"tolerations,omitempty"`
+	NodeSelector        map[string]string    `json:"NodeSelector,omitempty"`
```
alexeyklyukin: Please, use the camelCase for the parameter name

cehoffman:
I think having per cluster node affinity is useful. Did you know this can be done for all postgresql cluster operator config key `node_readiness_label` with a value of the form `<label>:<value>,<2nd label>:<2nd value>`. For example, in our configuration to only run on nodes with portworx.

```yaml
  node_readiness_label: "px/enabled:true"
```

alexeyklyukin:
@ksala I think what you are doing here can be indeed achieved with a ``node_readiness_label `` parameter as described at [at the procedure for doing cluster upgrades](https://github.com/zalando-incubator/postgres-operator#using-the-operator-to-minimize-the-amount-of-failovers-during-the-cluster-upgrade). In addition to making sure your pod is scheduled on the nodes with a certain label, this feature also migrates master pods for you when the node label changes to indicate the node is no longer part of the 'ready' node.

Do you see any use case outside of choosing the ready/not-ready nodes that can be covered by setting node labels per-cluster?

zerg-junior:
> what you are doing here can be indeed achieved with a node_readiness_label parameter
 
Given that, and also the fact that the `nodeSelector` will be deprecated in favor of node affinity (see [here "Affinity and anti-affinity"](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#step-two-add-a-nodeselector-field-to-your-pod-configuration#affinity-and-anti-affinity)), I suggest we close this PR.

PR #210, opened 2018-01-15
Consider container names in Statefulset diffs

mgomezch:
This includes a comparison on container names being equal in the decision of whether a Statefulset has been updated.


mgomezch:
:+1:

alexeyklyukin:
👍

PR #211, opened 2018-01-17
Use cluster UID as a suffix to the WAL bucket.

alexeyklyukin:
Populate `WAL_BUCKET_SCOPE_SUFFIX` Spilo environment variable with the cluster manifest UID, which will be used as a suffix for the S3 bucket. The end goal is to allow deleting a cluster and then deploying another one with the same name without the latest one reusing the S3 bucket of the previous one.

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -8,6 +8,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/pkg/apis/apps/v1beta1"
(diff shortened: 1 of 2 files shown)


alexeyklyukin:
👍

pkg/cluster/k8sres.go
```diff
@@ -500,6 +503,13 @@ func (c *Cluster) generatePodTemplate(
 	return &template
 }
 
+func getWALBucketScopeSuffix(uid string) string {
+	if uid != "" {
+		return fmt.Sprintf("-%s", uid)
+	}
+	return ""
```
erthalion: Does it makes sense to return an empty string from this function? What if there would be an error, that would lead to the same empty suffix for different clusters?

alexeyklyukin: It could be perfectly fine to call this function to finding out a prefix to clone with an empty uid, allowing us to clone the cluster that has been started before we introduced the prefixes (i.e. cloning the cluster A from the cluster B using S3 bucket s3://... /spilo/B).

erthalion:
:+1:

alexeyklyukin:
👍

Issue #214, opened 2018-02-01
Describe how to step through the operator with debugger


zerg-junior:
That's well worth adding to the debugging section of the  `READ.me`

zerg-junior:
Closed by #215

PR #215, opened 2018-02-01
Add a section about debugger

erthalion:
For #214 
So far only delve setup is described in details, but I think it can be convenient.

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,3 +305,46 @@ The operator also supports pprof endpoints listed at the [pprof package](https:/
 * /debug/pprof/profile
 * /debug/pprof/symbol
 * /debug/pprof/trace
+
+It's possible to attach a debugger to troubleshoot postgres-operator inside a
+docker container. It's possible with gdb and
+[delve](https://github.com/derekparker/delve). Since the latter one is a
+specialized debugger for golang, we will use it as an example. To use it you
+need:
+
+* Install delve locally
+
+```
+go get -u github.com/derekparker/delve/cmd/dlv
+```
+
+* Add following dependencies to the `Dockerfile`
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -305,3 +305,46 @@ The operator also supports pprof endpoints listed at the [pprof package](https:/
...
+* Update the `Makefile` to build the project with debugging symbols
+
+```
+-gcflags "-N -l"
+```
+
+* Run `postgres-operator` under the delve
+
+```
+CMD ["/root/go/bin/dlv", "--listen=:DLV_PORT", "--headless=true", "--api-version=2", "exec", "/postgres-operator"]
```
zerg-junior: Instead of `ENTRYPOINT` ?

README.md
```diff
@@ -305,3 +305,46 @@ The operator also supports pprof endpoints listed at the [pprof package](https:/
...
+
+```
+RUN apk --no-cache add go git musl-dev
+RUN go get github.com/derekparker/delve/cmd/dlv
+```
+
+* Update the `Makefile` to build the project with debugging symbols
+
+```
+-gcflags "-N -l"
```
zerg-junior: To the`linux` target ?

zerg-junior:
:+1:

erthalion:
:+1:

mgomezch:
Wouldn't it be better to add a `debug` target to the `Makefile` or make it build with debugging if `DEBUG` has some value in the environment, rather than simply documenting how to hack the `Makefile` manually?  Perhaps this could also be included as an option in the script to run under Minikube.

zerg-junior:
@mgomezch That's very true. I will play a little bit with the debugger and then script all that stuff in a separate PR.

PR #217, opened 2018-02-02
Add debug mode

erthalion:
You can specify an environment variable DEBUG 1/0 to enable/disable
debug mode. Debug mode assumes delve dependency for remote debugging,
and debugging symbols for go build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,19 @@ ifeq ($(RACE),1)
     CGO_ENABLED=1
 endif
 
+ifeq ($(DEBUG),1)
+	BUILD_FLAGS += -gcflags "-N -l"
+endif
+
 LOCAL_BUILD_FLAGS ?= $(BUILD_FLAGS)
 LDFLAGS ?= -X=main.version=$(VERSION)
-DOCKERFILE = docker/Dockerfile
+
+ifeq ($(DEBUG),1)
+	DOCKERFILE = docker/DebugDockerfile
+else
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -43,7 +53,7 @@ docker-context: scm-source.json linux
 	cp build/linux/${BINARY} scm-source.json docker/build/
 
 docker: ${DOCKERFILE} docker-context
-	cd docker && docker build --rm -t "$(IMAGE):$(TAG)" .
+	docker build --rm -t "$(IMAGE):$(TAG)" -f "${DOCKERFILE}" .
```
mgomezch: Maybe add a `-debug` suffix to the tag just so it's more visible?

mgomezch: Also: note this changes behavior a bit.  As you no longer `cd` into the `docker` directory, the Docker build context will be the top level directory of the repository, so more things will be copied to the Docker daemon for the build.  Also, the `COPY build/* /` instruction from the `Dockerfile` will actually be using `$repo_root/build` instead of the `$repo_root/docker/build` directory prepared by the recipe in the `docker-context` target: `mkdir -p docker/build; cp build/linux/${BINARY} scm-source.json docker/build/`.  I'm not sure if the binary would even be present in the root directory of the image at `/postgres-operator`.  Did you intend this?

erthalion: Yeah, it makes sense.

zerg-junior:
:+1:

mgomezch:
:+1:

PR #224, opened 2018-02-07
Tests for retry_utils

erthalion:
One can argue about how necessary they are,
but at least I remembered how to do golang.

### DIFF
--- a/pkg/util/retryutil/retry_util.go
+++ b/pkg/util/retryutil/retry_util.go
@@ -5,14 +5,39 @@ import (
 	"time"
 )
 
-// Retry calls ConditionFunc until it returns boolean true, a timeout expires or an error occurs.
+type RetryTicker interface {
+	Stop()
+	Tick()
+}
+
+type Ticker struct {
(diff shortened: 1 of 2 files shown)


pkg/util/retryutil/retry_util.go
```diff
@@ -5,14 +5,39 @@ import (
...
 	}
+	tick := &Ticker{time.NewTicker(interval)}
+	return RetryWorker(interval, timeout, f, tick)
+}
+
+func RetryWorker(
+	interval time.Duration,
+	timeout time.Duration,
+	f func() (bool, error),
+	tick RetryTicker) error {
```
mgomezch: I don't think this is too important, but maybe the function parameter should be at the end to make invocations with long hanging lambdas more readable.

erthalion: Yeah, it actually bothered me too a bit.

erthalion:
:+1:

mgomezch:
:+1:

PR #225, opened 2018-02-07
Support many namespaces

### LINKED ISSUE #128: Support for multiple namespaces
My expectation was that the Postgres operator will watch and create Postgres instances in multiple namespaces. So far I only got it working for the namespace where the operator is running. Is this by design or did I miss some settings?

Thanks for your help.

zerg-junior:
Closes #128 

Builds on PR #200  by @alexeyklyukin

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -4,8 +4,8 @@ metadata:
   name: postgres-operator
 data:
   # the env var with the same name in the operator pod may overwrite this value
-  # if neither is set, listen to the 'default' namespace
-  # if set to the empty string "", listen to all namespaces
+  # if neither is set or evaluates to the empty string, listen to the operator's own namespace
+  # if set to the "*", listen to all namespaces
   # watched_namespace: development
   service_account_name: operator
   cluster_labels: application:spilo
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -4,7 +4,8 @@ metadata:
   name: postgres-operator
 data:
   # the env var with the same name in the operator pod may overwrite this value
-  # if neither is set, listen to all namespaces
+  # if neither is set, listen to the 'default' namespace
+  # if set to the empty string "", listen to all namespaces
   # watched_namespace: development
   service_account_name: operator
   cluster_labels: application:spilo
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -3,8 +3,8 @@ kind: ConfigMap
 metadata:
   name: postgres-operator
 data:
-  # assumes the ns exists before the operator starts
-  # the env var with the same name may overwrite it in the operator pod
+  # the env var with the same name in the operator pod may overwrite this value
+  # if neither is set, listen to all namespaces
   # watched_namespace: development
   service_account_name: operator
   cluster_labels: application:spilo
--- a/manifests/postgres-operator.yaml
+++ b/manifests/postgres-operator.yaml
@@ -16,6 +16,7 @@ spec:
         imagePullPolicy: IfNotPresent
         env:
         # uncomment to overwrite a similar setting from operator configmap
+        # if set to the empty string, watch the 'default' namespace
         # - name: WATCHED_NAMESPACE
         #  valueFrom:
         #    fieldRef:
--- a/manifests/postgres-operator.yaml
+++ b/manifests/postgres-operator.yaml
@@ -16,7 +16,8 @@ spec:
         imagePullPolicy: IfNotPresent
         env:
         # uncomment to overwrite a similar setting from operator configmap
-        # if set to the empty string, watch the 'default' namespace
+        # if set to the empty string, watch the operator's own namespace
+        # if set to the "*", listen to all namespaces
         # - name: WATCHED_NAMESPACE
         #  valueFrom:
         #    fieldRef:
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -48,9 +48,9 @@ type Server struct {
 }
 
 var (
-	clusterStatusURL     = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)(/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*))?/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/?$`)
-	clusterLogsURL       = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)(/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*))?/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/logs/?$`)
-	clusterHistoryURL    = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)(/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*))?/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/history/?$`)
+	clusterStatusURL     = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/?$`)
+	clusterLogsURL       = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/logs/?$`)
+	clusterHistoryURL    = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/history/?$`)
 	teamURL              = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/?$`)
 	workerLogsURL        = regexp.MustCompile(`^/workers/(?P<id>\d+)/logs/?$`)
 	workerEventsQueueURL = regexp.MustCompile(`^/workers/(?P<id>\d+)/queue/?$`)
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -30,9 +30,9 @@ type controllerInformer interface {
 	GetOperatorConfig() *config.Config
 	GetStatus() *spec.ControllerStatus
 	TeamClusterList() map[string][]spec.NamespacedName
-	ClusterStatus(team, cluster string) (*spec.ClusterStatus, error)
-	ClusterLogs(team, cluster string) ([]*spec.LogEntry, error)
-	ClusterHistory(team, cluster string) ([]*spec.Diff, error)
+	ClusterStatus(team, namespace, cluster string) (*spec.ClusterStatus, error)
+	ClusterLogs(team, namespace, cluster string) ([]*spec.LogEntry, error)
+	ClusterHistory(team, namespace, cluster string) ([]*spec.Diff, error)
 	ClusterDatabasesMap() map[string][]string
 	WorkerLogs(workerID uint32) ([]*spec.LogEntry, error)
 	ListQueue(workerID uint32) (*spec.QueueDump, error)
@@ -48,9 +48,9 @@ type Server struct {
 }
 
 var (
-	clusterStatusURL     = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/?$`)
-	clusterLogsURL       = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/logs/?$`)
-	clusterHistoryURL    = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/history/?$`)
+	clusterStatusURL     = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)(/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*))?/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/?$`)
+	clusterLogsURL       = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)(/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*))?/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/logs/?$`)
+	clusterHistoryURL    = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)(/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*))?/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/history/?$`)
 	teamURL              = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/?$`)
 	workerLogsURL        = regexp.MustCompile(`^/workers/(?P<id>\d+)/logs/?$`)
 	workerEventsQueueURL = regexp.MustCompile(`^/workers/(?P<id>\d+)/queue/?$`)
@@ -149,7 +149,8 @@ func (s *Server) clusters(w http.ResponseWriter, req *http.Request) {
 	)
 
 	if matches := util.FindNamedStringSubmatch(clusterStatusURL, req.URL.Path); matches != nil {
-		resp, err = s.controller.ClusterStatus(matches["team"], matches["cluster"])
+		namespace, _ := matches["namespace"]
+		resp, err = s.controller.ClusterStatus(matches["team"], namespace, matches["cluster"])
 	} else if matches := util.FindNamedStringSubmatch(teamURL, req.URL.Path); matches != nil {
 		teamClusters := s.controller.TeamClusterList()
 		clusters, found := teamClusters[matches["team"]]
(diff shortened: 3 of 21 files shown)


pkg/controller/controller.go
```diff
@@ -97,18 +97,34 @@ func (c *Controller) initOperatorConfig() {
...
-	}
+	// by default, the operator listens to all namespaces
+	// by setting the env variable, one can restrict the operator to a single namespace
+	watchedNamespace, isPresentInEnv := os.LookupEnv("WATCHED_NAMESPACE")
+	if isPresentInEnv {
+		// special case: v1.NamespaceAll currently also evaluates to the empty string
+		// so when the env var is set to the empty string, use the default ns
+		// since the meaning of this env var is only one namespace
+		if watchedNamespace == "" {
+			c.logger.Infof("The WATCHED_NAMESPACE env var evaluates to the empty string, falling back to watching the 'default' namespace.\n", watchedNamespace)
```
erthalion: It feels little bit too implicit for me. Maybe it would be better if:
* when this variable is not present, listen to a default namespace (which would be close to current behavior)
* when this variable is there and not empty, listen to these namespaces
* when this variable is there and empty, throw a validation error
* add a special flag "watchAll" to watch all namespaces.

pkg/controller/controller.go
```diff
@@ -97,16 +97,25 @@ func (c *Controller) initOperatorConfig() {
...
 	}
 
-	// env var takes priority over the same param from the operator ConfigMap
-	watchedNamespace := os.Getenv("WATCHED_NAMESPACE")
-	if watchedNamespace != "" {
-		c.logger.Infof("Watch the %q namespace specified in the env variable WATCHED_NAMESPACE\n", watchedNamespace)
-		configMapData["watched_namespace"] = watchedNamespace
+	// by default, the operator listens to all namespaces
+	if configMapData["watched_namespace"] == "" {
+		c.logger.Infoln("The operator config map specifies no namespace to watch. Falling back to watching all namespaces.")
```
mgomezch: Note that if the key is missing or empty in the ConfigMap but the environment variable is specified, this will still be printed to the log.  While this arguably correct (as the log message does refer explicitly to the config map), it could be confusing to see this message say it will fall back to watching all namespaces, and right below that say it's actually going to do what the environment variable specifies.

pkg/controller/controller.go
```diff
@@ -97,16 +97,25 @@ func (c *Controller) initOperatorConfig() {
...
-		c.logger.Infoln("No namespace to watch specified. Fall back to watching the 'default' namespace.")
-		configMapData["watched_namespace"] = v1.NamespaceDefault
+	watchedNsEnvVar, isPresentInEnv := os.LookupEnv("WATCHED_NAMESPACE")
+	if isPresentInEnv {
+		c.logger.Infoln("The WATCHED_NAMESPACE env variable takes priority over the same param from the operator configMap\n")
+		// special case: v1.NamespaceAll currently also evaluates to the empty string
+		// so when the env var is set to the empty string, use the default ns
+		// since the meaning of this env var is only one namespace
+		if watchedNsEnvVar == "" {
+			c.logger.Infof("The WATCHED_NAMESPACE env var evaluates to the empty string, falling back to watching the 'default' namespace.\n", watchedNsEnvVar)
```
mgomezch: It's confusing and complex that an explicit empty string in the environment variable falls back to watching only `default`, but an explicit empty string in the ConfigMap falls back to watching everything.  Shouldn't explicit empty string settings have the same meaning for both?

pkg/controller/controller.go
```diff
@@ -97,16 +97,25 @@ func (c *Controller) initOperatorConfig() {
 		c.logger.Infoln("no ConfigMap specified. Loading default values")
 	}
 
-	// env var takes priority over the same param from the operator ConfigMap
-	watchedNamespace := os.Getenv("WATCHED_NAMESPACE")
-	if watchedNamespace != "" {
-		c.logger.Infof("Watch the %q namespace specified in the env variable WATCHED_NAMESPACE\n", watchedNamespace)
-		configMapData["watched_namespace"] = watchedNamespace
+	// by default, the operator listens to all namespaces
+	if configMapData["watched_namespace"] == "" {
```
mgomezch: Just to confirm: does this behave as expected, given that it will behave the same way when the key is missing and when it is explicitly set to the empty string?

zerg-junior:
@mgomezch  I updated the PR since you correctly raised very valid issues

pkg/controller/controller.go
```diff
@@ -97,16 +97,39 @@ func (c *Controller) initOperatorConfig() {
...
-	if configMapData["watched_namespace"] == "" {
-		c.logger.Infoln("No namespace to watch specified. Fall back to watching the 'default' namespace.")
-		configMapData["watched_namespace"] = v1.NamespaceDefault
+	if (isPresentInOperatorConfigMap) && (!isPresentInOperatorEnv) {
+
+		// special case: v1.NamespaceAll currently also evaluates to the empty string
+		// so when the param evaluates to the empty string, we use the default ns
+		// since the meaning of the param is a single namespace and *not* all namespaces
+		if watchedNsConfigMapVar == "" {
+			c.logger.Infof("The watched namespace field in the operator config map evaluates to the empty string, falling back to watching the 'default' namespace.\n")
```
mgomezch: I still don't think this is quite ideal :( Having the meaning of this variable tied to specifying one individual namespace has a couple of consequences that I would say are not too desirable:

1. No options beyond "one" or "all": If this configuration variable means specifically *one* namespace to watch, then it would be difficult to adapt this to use cases like smoke-testing the multiple namespaces feature if the tests run on a cluster that also has a non-testing operator running.  If this supported listing, say, 2 or 3 namespaces (comma-separated or somesuch) instead of just one (or "all" when absent), then you could have the smoke tests specify 2 or 3 namespaces to listen to, all whose names would include the random test namespace string.  That way, the test operator would not interfere with the non-test operator.  This would not completely fix the issue of such tests, because if the non-test operator listens on *all* namespaces anyway, then the non-test operator would interfere with the test operator; therefore it's possible this is a misplaced concern, and this needs a broader solution (maybe a pattern for what namespaces should be watched, or a pattern for excluded namespaces, or somesuch, although that would bring additional complexity, ugh).

2. No complete environment override: The current semantics mean that if the ConfigMap specifies a `watched_namespace`, even if blank (which would default to the `default` namespace), then it is impossible to override this behavior from an environment variable to make the operator listen to all namespaces.  This is not a big problem as the real benefit of the environment variable override is to get a specific namespace name from `valueFrom` in Kubernetes deployments, and if you want the Operator to listen on all namespaces and you control its environment, then you can (and must) also easily control the contents of the ConfigMap.  Still, to me it feels like it should be possible to override this anyway.

3. Too implicit: I agree with @erthalion that toggling nontrivial behavior like "please listen on all namespaces" through a very implicit mechanism like the *absence* of configuration keys is not ideal and may be surprising for users.  It may be better to have something explicit — it's fine for the default behavior to be "all namespaces", but it seems to me that it should at least deserve a way to explicitly ask for it by setting a key to some value, not just leaving it unset.  Having a separate key or an explicit `*` name/pattern or somesuch would also avoid the inability to override noted in (2).

All of these issues are possibly a matter of preference and/or small importance, so maybe this PR is fine as it is right now.  I still think it's worth mentioning so we explicitly acknowledge this is the design we want despite these issues.

zerg-junior: > No options beyond "one" or "all": 

I feel watching a few namespaces specified beforehand is beyond the scope of this issue. Let's finish the simpler use case first and extend it later if needed.

>  an absent key meaning to listen only on the default namespace, and a present key set to the empty string (in the ConfigMap or the environment, with the environment taking precedence) meaning listening on all namespaces. 

Makes perfect sense, will have a look into that.

mgomezch: Totally agree on limiting scope; I think that's a separate feature (but you will need it as soon as this lands as your smoke tests will break ;) )

alexeyklyukin: @mgomezch watching on multiple namespaces has its technical issues: you can easily create a watcher not tied to any namespaces or tied to a single one. For multiple ones, you need multiple watchers. That would add some complexity to the code, and the benefits are not clear to me, as our testing cases are completely covered with running the operator in a single namespace (in the test cluster, we should avoid running any operators that listen to all namespaces, as we don't own that cluster and other people may start experimenting with it).

alexeyklyukin:
I think watching all namespaces should be an explicit choice, indicated by a value that cannot correspond to a valid namespace name; '*' seems to be a good candidate. If we make it implicit it may bite us in the future because of people deploying multiple versions of the operator with the default parameters (i.e. on the playground) and interfering with each other with issues that a hard to diagnose.

The namespace unset or set to '' should lead to choosing a default namespace. I think the code that chooses between the namespace defined in the configuration and the one that is set via the environment variable is unnecessary complex: it should read the values of both the configuration parameter and the environment variable (in that specific order) and loop over the two, setting the value inside the operator and complaining if the value is already set to a non-default namespace before overriding it.

As a nitpick, I'd name the configuration parameter ``watch_namespace`` instead of ``watched_namespace`` to match the environment variable name.

The functions in the status.go (ClusterStatus, ClusterLogs, ClusterHistory) accept empty string for the namespace, defaulting to the operator namespace. I think this is too simplistic, what we should do is to find all clusters with the name equal to the specific name in any namespaces and, when there are multiple clusters, complain, listing all the namespaces where the cluster with the particular name resides, enabling the caller to retry the request with the specific namespace.

mgomezch:
As noted by @Jan-M, one problem with `*` is that it suggests that maybe glob patterns work there.

pkg/apiserver/apiserver.go
```diff
@@ -48,9 +48,9 @@ type Server struct {
 }
 
 var (
-	clusterStatusURL     = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/?$`)
-	clusterLogsURL       = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/logs/?$`)
-	clusterHistoryURL    = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/history/?$`)
+	clusterStatusURL     = regexp.MustCompile(`^/clusters/(?P<team>[a-zA-Z][a-zA-Z0-9]*)/(?P<namespace>[a-zA-Z][a-zA-Z0-9-]*)/(?P<cluster>[a-zA-Z][a-zA-Z0-9-]*)/?$`)
```
alexeyklyukin: The namespace regex is wrong. The correct one is `[a-z0-9]([-a-z0-9]*[a-z0-9])?` (obtained from `kubectl create namespace INVALID-`)

manifests/configmap.yaml
```diff
@@ -3,8 +3,9 @@ kind: ConfigMap
 metadata:
   name: postgres-operator
 data:
-  # assumes the ns exists before the operator starts
-  # the env var with the same name may overwrite it in the operator pod
+  # the env var with the same name in the operator pod may overwrite this value
+  # if neither is set, listen to the 'default' namespace
+  # if set to the empty string "", listen to all namespaces
```
alexeyklyukin: We should listen to the operator namespace either when the parameters are empty or not present, and only listen to all namespaces when we have a '*'.

manifests/postgres-operator.yaml
```diff
@@ -16,6 +16,7 @@ spec:
         imagePullPolicy: IfNotPresent
         env:
         # uncomment to overwrite a similar setting from operator configmap
+        # if set to the empty string, watch the 'default' namespace
```
alexeyklyukin: we should watch for the operator namespace.

pkg/controller/controller.go
```diff
@@ -97,16 +97,40 @@ func (c *Controller) initOperatorConfig() {
...
 	}
 
-	// env var takes priority over the same param from the operator ConfigMap
-	watchedNamespace := os.Getenv("WATCHED_NAMESPACE")
-	if watchedNamespace != "" {
-		c.logger.Infof("Watch the %q namespace specified in the env variable WATCHED_NAMESPACE\n", watchedNamespace)
-		configMapData["watched_namespace"] = watchedNamespace
-	}
+	watchedNsConfigMapVar, isPresentInOperatorConfigMap := configMapData["watched_namespace"]
+	watchedNsEnvVar, isPresentInOperatorEnv := os.LookupEnv("WATCHED_NAMESPACE")
```
alexeyklyukin: we don't need LookupEnv here, since empty or missing WATCHED_NAMESPACE should produce the same default behavior of watching the operator namespace.

pkg/controller/controller.go
```diff
@@ -97,16 +97,40 @@ func (c *Controller) initOperatorConfig() {
...
+
+	if (!isPresentInOperatorConfigMap) && (!isPresentInOperatorEnv) {
 
-	if configMapData["watched_namespace"] == "" {
 		c.logger.Infof("No namespace to watch specified. By convention, the operator falls back to watching the namespace it is deployed to: '%v' \n", spec.GetOperatorNamespace())
 		configMapData["watched_namespace"] = spec.GetOperatorNamespace()
+
+	}
+
+	if (isPresentInOperatorConfigMap) && (!isPresentInOperatorEnv) {
```
alexeyklyukin: Looks like there is no reason for this block of code here. It should be something like:


```
configmapData['watched_namespace'] = util.Coalesce(namespaceFromEnvironment, util.Coalesce(namespaceFromConfigMap, spec.GetOperatorNamespace()))
if configmapData['watched_namespace']  == "*" {
    configmapData['watched_namespace'] = v1.NamespaceAll
    c.logger.Infof("Listening to all namespaces")
} else {
    c.logger.Infof("Listenting to the specific namespace %s",  configmapData['watched_namespace'] )
}

```

Perhaps it makes sense to wrap the code above into a function like `getEffectiveNamespace(environmentNamespace, configNamespace string)` that returns the result assigned to the configMapData and also emits info messages.

We can also add the Coalesce function to the util package,
```
func Coalesce(val, default string) {
     if val == "" {
         return default
    }
    return val
}
```

pkg/controller/controller.go
```diff
@@ -127,6 +151,16 @@ func (c *Controller) initOperatorConfig() {
 func (c *Controller) initController() {
 	c.initClients()
 	c.initOperatorConfig()
+
+	// earliest point where we can check if the namespace to watch actually exists
+	if c.opConfig.WatchedNamespace != v1.NamespaceAll {
+		_, err := c.KubeClient.Namespaces().Get(c.opConfig.WatchedNamespace, metav1.GetOptions{})
```
alexeyklyukin: I'd write the KubeClient check as 
```
if _, err := ; err != nil {
   c.logger.Fatalf(....)
}
```
The error text can be shortened to "could not find the watched namespace %q" (see https://www.postgresql.org/docs/10/static/error-style-guide.html for couldn't vs cannot).

Does it make sense to move this into a function that I mentioned above that figures the effective namespace and also checks whether it exists?

pkg/controller/status.go
```diff
@@ -14,9 +14,14 @@ import (
 )
 
 // ClusterStatus provides status of the cluster
-func (c *Controller) ClusterStatus(team, cluster string) (*spec.ClusterStatus, error) {
+func (c *Controller) ClusterStatus(team, namespace, cluster string) (*spec.ClusterStatus, error) {
+
+	if namespace == "" {
```
alexeyklyukin: Drop that (here and in other similar places), as we now require the namespace to be present by the caller

pkg/util/config/config.go
```diff
@@ -67,7 +67,7 @@ type Config struct {
 	Resources
 	Auth
 	Scalyr
-	WatchedNamespace         string            `name:"watched_namespace"`
+	WatchedNamespace         string            `name:"watched_namespace"` // may be v1.NamespaceAll, meaning watch all namespaces
```
alexeyklyukin: It should mention '*' instead of v1.NamespaceAll in the comment.

pkg/controller/controller.go
```diff
@@ -256,3 +248,26 @@ func (c *Controller) kubeNodesInformer(stopCh <-chan struct{}, wg *sync.WaitGrou
...
+	if namespace == "*" {
+
+		namespace = v1.NamespaceAll
+		c.logger.Infof("Listening to all namespaces")
+
+	} else {
+
+		if _, err := c.KubeClient.Namespaces().Get(namespace, metav1.GetOptions{}); err != nil {
+			// the namespace may be created manually at runtime
+			c.logger.Warnf("Could not find the watched namespace %q", namespace)
```
alexeyklyukin: Does it make sense to continue knowing that the namespace we are asked to listen to doesn't exist?

zerg-junior: I am unsure about that. For now it makes sense only for manual testing/experimentation

alexeyklyukin: then I think we can replace it back with Fatalf, which would give you the visible outcome for manual testing and will avoid running no-op operators for the production use case in case someone made a mistake in the configuration.

pkg/util/config/config.go
```diff
@@ -67,7 +67,7 @@ type Config struct {
 	Resources
 	Auth
 	Scalyr
-	WatchedNamespace         string            `name:"watched_namespace"`
+	WatchedNamespace         string            `name:"watched_namespace"` // may be "*", meaning watch all namespaces
```
erthalion: Shouldn't it be more complete like this?
```
may be "*", meaning watch all namespaces, or "", meaning watch a namespace where was deployed to
```

pkg/controller/controller.go
```diff
@@ -256,3 +248,25 @@ func (c *Controller) kubeNodesInformer(stopCh <-chan struct{}, wg *sync.WaitGrou
 
 	c.nodesInformer.Run(stopCh)
 }
+
+func (c *Controller) getEffectiveNamespace(namespaceFromEnvironment, namespaceFromConfigMap string) string {
+
+	namespace := util.Coalesce(namespaceFromEnvironment, util.Coalesce(namespaceFromConfigMap, spec.GetOperatorNamespace()))
```
erthalion: If you use a chain of `Coalesce`, maybe let's define it to accept variable number of strings?

```
func Coalesce(values ...string) string {
}
```

pkg/cluster/sync.go
```diff
@@ -24,6 +24,7 @@ func (c *Cluster) Sync(newSpec *spec.Postgresql) (err error) {
 
 	defer func() {
 		if err != nil {
+			c.logger.Warningf("error while syncing cluster state: %v", err)
```
erthalion: Is this somehow related to the subject of PR, or just a random improvement?

mgomezch: General +1 for random improvements regardless of the answer to the previous question.

erthalion:
:+1:

pkg/cluster/sync.go
zerg-junior: This was a random improvement ensuring we do not silently swallow an error.

erthalion:
:+1:

zerg-junior:
:+1:

PR #231, opened 2018-02-13
Operator should default to the watched namespace

### LINKED ISSUE #155: Operator should default to the watched namespace
Right now the operator seems to rely on the access to the default namespace. In particular, it tries to fetch security tokens from the default namespace instead of the watched one.

zerg-junior:
Closes #155

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Watching a namespace for an operator means tracking requests to change Postgresq
 
 By default, the operator watches the namespace it is deployed to. You can change this by altering the `WATCHED_NAMESPACE` env var in the operator deployment manifest or the `watched_namespace` field in the operator configmap. In the case both are set, the env var takes the precedence.
 
-Note that for an operator to manage pods in the watched namespace, the operator's service account (as specified in the operator deployment manifest) has to have appropriate privileges to access the watched namespace. The watched namespace also needs to have a (possibly different) service account that allows database pods to talk to the Kubernetes API.
+Note that for an operator to manage pods in the watched namespace, the operator's service account (as specified in the operator deployment manifest) has to have appropriate privileges to access the watched namespace. The watched namespace also needs to have a (possibly different) service account in the case database pods need to talk to the Kubernetes API (e.g. when using Kubernetes-native configuration of Patroni).
 
 ### Create ConfigMap
 
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Watching a namespace for an operator means tracking requests to change Postgresq
 
 By default, the operator watches the namespace it is deployed to. You can change this by altering the `WATCHED_NAMESPACE` env var in the operator deployment manifest or the `watched_namespace` field in the operator configmap. In the case both are set, the env var takes the precedence.
 
-Note that for an operator to create pods in the watched namespace, one needs to create the `operator` service account in the namespace.
+Note that for an operator to manage pods in the watched namespace, the operator's service account (as specified in the operator deployment manifest) has to have appropriate privileges to access the watched namespace. The watched namespace also needs to have a (possibly different) service account that allows database pods to talk to the Kubernetes API.
 
 ### Create ConfigMap
 
--- a/README.md
+++ b/README.md
@@ -61,6 +61,23 @@ to test your that your setup is working.
 
 Note: if you use multiple Kubernetes clusters, you can switch to Minikube with `kubectl config use-context minikube`
 
+### Select the namespace to deploy to
+
+The operator can run in a namespace other than `default`. For example, to deploy it to the `test` namespace, run the following:
+
+    kubectl create namespace test
+    kubectl config set-context minikube --namespace=test
+
+All subsequent `kubectl` commands will work with the `test` namespace. The operator  will run in this namespace and look up needed resources - such as its config map - there.
+
+### Specify the namespace to watch
+
+Watching a namespace for an operator means tracking requests to change Postgresql clusters in the namespace such as "increase the number of Postgresql replicas to 5" and reacting to the requests, in this example by actually scaling up. 
+
+By default, the operator watches the namespace it is deployed to. You can change this by altering the `WATCHED_NAMESPACE` env var in the operator deployment manifest or the `watched_namespace` field in the operator configmap. In the case both are set, the env var takes the preference.
+
+Note that for an operator to create pods in the watched namespace, one needs to create the `operator` service account in the namespace.
+
 ### Create ConfigMap
 
 ConfigMap is used to store the configuration of the operator
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Note: if you use multiple Kubernetes clusters, you can switch to Minikube with `
 
 ### Select the namespace to deploy to
 
-The operator can run in a namespace other than `default`. For example, to deploy it to the `test` namespace, run the following:
+The operator can run in a namespace other than `default`. For example, to use the `test` namespace, run the following before deploying the operator's manifests:
 
     kubectl create namespace test
     kubectl config set-context minikube --namespace=test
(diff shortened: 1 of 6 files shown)


cmd/main.go
```diff
@@ -27,13 +28,26 @@ func init() {
 	flag.BoolVar(&config.NoTeamsAPI, "noteamsapi", false, "Disable all access to the teams API")
 	flag.Parse()
 
+	cmd := exec.Command("cat", "/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
mgomezch: Don't you want [ioutil.ReadFile](https://golang.org/pkg/io/ioutil/#ReadFile) here?

alexeyklyukin: I'd actually assume there is a MY_POD_NAMESPACE environment variable set to metadata.namespace and read that variable.

zerg-junior: >  I'd actually assume there is a MY_POD_NAMESPACE

I observe this env var neither in `minikube` nor in actual k8s deployments.

cmd/main.go
```diff
@@ -27,13 +28,26 @@ func init() {
...
+	if err != nil {
+		log.Fatalf("Unable to detect operator namespace from within its pod  due to %v", err)
+	}
+
 	configMapRawName := os.Getenv("CONFIG_MAP_NAME")
 	if configMapRawName != "" {
-		err := config.ConfigMapName.Decode(configMapRawName)
+
+		operatorNamespace := string(operatorNamespaceBytes)
+		namespacedConfigMapName := operatorNamespace + "/" + configMapRawName
```
alexeyklyukin: Seems to be some unnecessary work. I'd rather introduce a DecodeWithDefaultNamespace function in spec/types.go, that receives 2 arguments, a value, and a namespace to fill-in as a default one. The logic from Decode will be moved there, and the Decode itself will call DecodeWithNamespace(value, v1.NamespaceDefault). The code in main will call DecodeWithDefaultNamespace directly, supplying the operator namespace as a second option.

cmd/main.go
alexeyklyukin: Sure, it needs to be added to our Kubernetes deployment manifest. That's how Kubernetes Downward API work https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/

Jan-M: Isnt there an env var set by default for this?

zerg-junior: the answer is no, as far as i can see from manual examination of the env

Jan-M:
```
POD_NAMESPACE=default
```

Use this if it is supplied by Kubernetes 1.8+

cmd/main.go
zerg-junior: this is the only place where `Decode` is actually used; right now i am unsure if it is sensible to split it into 2 functions.

alexeyklyukin: Nope, AFAIK Decode implements an interface and it used, for instance, when Decoding values from JSON in the operator configuration (i.e. OAuthTokenSecretName or InfrastructureRolesSecretName).

alexeyklyukin: Actually, given that we likely want those objects to default to the operator namespace as well, maybe it makes sense to just change the Decode to use the operator namespace instead of v1.DefaultNamespace.

zerg-junior: good point, will address in the commit that makes operator fetch manifests from the namespace it is deployed into

mgomezch:
:+1:

cmd/main.go
```diff
@@ -27,13 +28,28 @@ func init() {
 	flag.BoolVar(&config.NoTeamsAPI, "noteamsapi", false, "Disable all access to the teams API")
 	flag.Parse()
 
+	operatorNamespaceBytes, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
```
zerg-junior: put into const

erthalion:
:+1:

zerg-junior:
@mgomezch  @erthalion  updated. Could you please check the code and documentation match each other and the original issue ?

README.md
```diff
@@ -61,6 +61,23 @@ to test your that your setup is working.
 
 Note: if you use multiple Kubernetes clusters, you can switch to Minikube with `kubectl config use-context minikube`
 
+### Select the namespace to deploy to
+
+The operator can run in a namespace other than `default`. For example, to deploy it to the `test` namespace, run the following:
```
mgomezch: Minor nitpick, but I'd say "run the following before deploying the operator's manifests" or somesuch, as just running those commands on their own does not deploy the operator to the `test` namespace as stated.

mgomezch: Minor nitpick, but I'd say "run the following before deploying the operator's manifests" or somesuch, as just running those commands does not deploy the operator as stated.

README.md
```diff
@@ -61,6 +61,23 @@ to test your that your setup is working.
...
+    kubectl create namespace test
+    kubectl config set-context minikube --namespace=test
+
+All subsequent `kubectl` commands will work with the `test` namespace. The operator  will run in this namespace and look up needed resources - such as its config map - there.
+
+### Specify the namespace to watch
+
+Watching a namespace for an operator means tracking requests to change Postgresql clusters in the namespace such as "increase the number of Postgresql replicas to 5" and reacting to the requests, in this example by actually scaling up. 
+
+By default, the operator watches the namespace it is deployed to. You can change this by altering the `WATCHED_NAMESPACE` env var in the operator deployment manifest or the `watched_namespace` field in the operator configmap. In the case both are set, the env var takes the preference.
```
mgomezch: `s/the preference/precedence/`

cmd/main.go
```diff
@@ -29,11 +29,21 @@ func init() {
...
+		config.Namespace = operatorNamespace
+
+		namespacedConfigMapName := operatorNamespace + "/" + configMapRawName
+
+		log.Printf("Looking for the operator configmap at the same namespace the operator resides. Fully qualified configmap name: %v", namespacedConfigMapName)
+
+		err := config.ConfigMapName.Decode(namespacedConfigMapName)
 		if err != nil {
-			log.Fatalf("incorrect config map name")
+			log.Fatalf("incorrect config map name: %v", namespacedConfigMapName)
```
erthalion: Am I right that `Fatalf` would log a message and then do `exit(1)` ? I would prefer to be this more explicit.

mgomezch: +1, I guess this is fine if it's unsurprising/idiomatic Go, but it looks really surprising to me that `log.$SOMETHING` calls `exit` 🤔

zerg-junior: > Am I right that Fatalf would log a message and then do exit(1) 

Yes, you're right :)

> it's really surprising to me that log.Whatever() calls exit

i haven't found much on the Web except [this discussion](https://stackoverflow.com/a/33890104) 

> this is fine if it's unsurprising/idiomatic Go

i would assume it is idiomatic Golang given that  libs published by Google have exactly that function

pkg/controller/controller.go
```diff
@@ -256,3 +261,10 @@ func (c *Controller) kubeNodesInformer(stopCh <-chan struct{}, wg *sync.WaitGrou
 
 	c.nodesInformer.Run(stopCh)
 }
+
+func (c *Controller) mustHaveOperatorServiceAccountInNamespace(namespace string) {
```
erthalion: Just a small complaint - as for me it's too [long](https://projects.haykranen.nl/java/). Maybe give this function some short name, and then explain all the details in commentary? E.g:

```go
// Check that a namespace has a service account in it
func (c *Controller) checkServiceAccount(namespace string) {
}
```

zerg-junior: your suggestion makes sense, will do

cmd/main.go
mgomezch: ¯\_(ツ)_/¯

pkg/controller/controller.go
```diff
@@ -131,6 +131,11 @@ func (c *Controller) initController() {
 
 	c.logger.Infof("config: %s", c.opConfig.MustMarshal())
 
+	c.checkOperatorServiceAccount(c.config.Namespace)
```
mgomezch: As discussed, this shouldn't be here: the service account whose name is specified in `c.opConfig.ServiceAccountName` needs to be present on the watched namespace (so that database pods can use the Kubernetes API), not on the namespace where the operator is deployed.

If the namespace where the operator is deployed does not have a service account named as the one specified in the operator's deployment manifest, then the operator would simply fail to deploy before this code is even running anywhere.

erthalion:
:+1:

mgomezch:
:+1:

erthalion:
:+1:

zerg-junior:
:+1:

PR #232, opened 2018-02-14
Add a new step to run unit tests

erthalion:
This necessary step was missing. Also I update a delivery.yaml version and split image build & push steps.


mgomezch:
Push first, test later?  Valid if this is what we want, of course, but is it?

alexeyklyukin:
👍

erthalion:
:+1:

PR #234, opened 2018-02-16
Define the operator RBAC

alexeyklyukin:
Note that the account here is named zalando-postgres-operator and not
the 'operator' default that is created in the serviceaccount.yaml and
also used by the operator configmap to create new postgres clusters.

This is done intentionally, as to avoid breaking those setups that
already work. Ideally, the operator should be run under the
zalando-postgres-operator service account. However, the service account
used to run Postgres clusters does not require all those privileges and
is described at
https://github.com/zalando/patroni/blob/master/kubernetes/patroni_k8s.yaml

The service account defined here acquires some privileges not really
used by the operator (i.e. we only need list and watch on configmaps),
this is also done intentionally to avoid breaking things if someone
decides to configure the same service account in the operator's
configmap to run postgres clusters.

### DIFF
--- a/manifests/operator-rbac.yaml
+++ b/manifests/operator-rbac.yaml
@@ -1,0 +1,62 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: zalando-postgres-operator
+
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: Role
+metadata:
+  name: zalando-postgres-operator
+rules:
+- apiGroups:
+  - acid.zalan.do
+  resources:
+  - postgresqls
+  verbs:
+  - "*"
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - endpoints
+  - pods
+  - services
+  - statefulsets
+  - persistentvolumeclaims
+  - secrets
+  verbs:
+  - "*"
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - policy
+  resources:
+  - poddisruptionbudgets
+  verbs:
+  - "*"
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: zalando-postgres-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: operator
+subjects:
+- kind: ServiceAccount
+  name: zalando-postgres-operator
--- a/manifests/operator-rbac.yaml
+++ b/manifests/operator-rbac.yaml
@@ -89,7 +89,7 @@ rules:
   - get
   - patch
 - apiGroups:
-  - ""
+  - apps
   resources:
   - statefulsets
   verbs:
--- a/manifests/operator-rbac.yaml
+++ b/manifests/operator-rbac.yaml
@@ -96,6 +96,7 @@ rules:
   - create
   - delete
   - get
+  - list
   - patch
 - apiGroups:
   - ""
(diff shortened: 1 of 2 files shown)


manifests/operator-rbac.yaml
```diff
@@ -0,0 +1,62 @@
...
+kind: Role
+metadata:
+  name: zalando-postgres-operator
+rules:
+- apiGroups:
+  - acid.zalan.do
+  resources:
+  - postgresqls
+  verbs:
+  - "*"
```
mikkeloscar: Would be nice to be explicit about what actions it needs.

manifests/operator-rbac.yaml
```diff
@@ -0,0 +1,62 @@
...
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
```
adarshaj: This did not work as we are defining a `Role` whereas the operation being performed is at cluster level, so there should be another separate `ClusterRole` and corresponding `ClusterRoleBinding` and actions being performed at cluster level must be separated out.

mikkeloscar: I guess everything will be at the cluster level assuming the operator is running with multi-namespace support.

alexeyklyukin: we don't have support for multiple namespaces at the moment, but this patch https://github.com/zalando-incubator/postgres-operator/pull/225 aims to add support for listening to manifests and creating clusters either in a single namespace (as it is today), or in all namespaces, therefore, defining a `ClusterRole` makes sense to me. It will also solve the issue of creating customresourcedefinitions.

alexeyklyukin: apparently, another argument for the cluster role is that the operator needs to list nodes in order to migrate from the nodes approaching EOL (as indicated by the lack of the `node_readiness_label`).

manifests/operator-rbac.yaml
```diff
@@ -0,0 +1,129 @@
...
+  - service
+  verbs:
+  - create
+  - delete
+  - get
+  - patch
+- apiGroups:
+  - ""
+  resources:
+  - statefulsets
```
adarshaj: This also needed `list`. also I had to set apiGroups to `*`, still debugging, the operator isn't fully functional with this rbac

adarshaj:
> Note that the account here is named zalando-postgres-operator and not
the 'operator' default that is created in the serviceaccount.yaml and
also used by the operator configmap to create new postgres clusters.

I re-used zalando-postgres-operator as service account for postgres pod too, but it failed to come up because this rbac lacks `list` on configmaps

alexeyklyukin:
@adarshaj Right, the zalando-postgres-operator ClusterRole is intentionally set to have only those permissions required for the operator itself. The list on ConfigMaps is required by Patroni (btw, it's better to use EndPoints), and one can use the RBAC defined [in the patroni repository](https://github.com/zalando/patroni/blob/master/kubernetes/patroni_k8s.yaml ) for the ServiceAccount actual clusters run with (configured as a service_account_name in the operator ConfigMap)

alexeyklyukin:
@adarshaj also, as you noticed, the statefulset API group was set incorrectly to an empty value, the correct one is apps (fixed in https://github.com/zalando-incubator/postgres-operator/pull/234/commits/d15682b7bddcee3060d2a7e5561ac6b5ac3c021f)

manifests/operator-rbac.yaml
zerg-junior: > patch #225 aims to add support for listening to manifests and creating clusters either in a single namespace (as it is today), or in all namespaces,

it is now merged, so we do have support for multiple namespaces and thus should opt for `ClusterRole`

zerg-junior:
I feel the difference between `zalando-postgres-operator` and the `operator` accounts is worth putting into the `read.me`

zerg-junior:
:+1:

alexeyklyukin:
👍

zerg-junior:
:+1:

PR #238, opened 2018-02-16
Make operator watch its own namespace instead of controller's one

### LINKED ISSUE #237: Operator should watch its own namespace instead of the controller namespace
The controller namespace is currently unused and evaluates to the empty string at runtime.

zerg-junior:
Closes #237


mgomezch:
:+1:

zerg-junior:
:+1:

PR #240, opened 2018-02-19
Some improvements for golint, ineffassign and misspell

erthalion:
Improvements are mostly style related, like:

* drop else when there is a return statement in the previous if

* comments for retry_utils

* names consistency

### DIFF
--- a/pkg/apiserver/apiserver.go
+++ b/pkg/apiserver/apiserver.go
@@ -163,8 +163,7 @@ func (s *Server) clusters(w http.ResponseWriter, req *http.Request) {
 			clusterNames = append(clusterNames, cluster.Name[len(matches["team"])+1:])
 		}
 
-		s.respond(clusterNames, nil, w)
-		return
+		resp, err = clusterNames, nil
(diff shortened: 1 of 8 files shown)


pkg/apiserver/apiserver.go
```diff
@@ -176,11 +175,9 @@ func (s *Server) clusters(w http.ResponseWriter, req *http.Request) {
 				res[team] = append(res[team], cluster.Name[len(team)+1:])
 			}
 		}
-
-		s.respond(res, nil, w)
-		return
+		err = nil
```
alexeyklyukin: nil is the default.

pkg/util/retryutil/retry_util_test.go
```diff
@@ -36,13 +36,13 @@ func TestRetryWorkerOneFalse(t *testing.T) {
 
 	tick := &mockTicker{t, 0}
 	result := RetryWorker(1, 3, tick, func() (bool, error) {
-		counter += 1
+		counter++
 
 		if counter <= 1 {
 			return false, nil
-		} else {
```
alexeyklyukin: should be return counter > 1, nil.

erthalion:
:+1:

PR #242, opened 2018-02-20
Change the suffix delimiter to slash.

alexeyklyukin:
This allows using S3 API in order to simplify finding all
folders that are different only by a suffix, since the
suffix delimiter will not occur in the suffix itself.


Jan-M:
:+1:

mgomezch:
:+1:

alexeyklyukin:
👍

Issue #241, opened 2018-02-20
Cleanup after the deleted cluster


alexeyklyukin:
Right now it is not possible to delete an existing cluster and then create a new one with the same name. When stumbling upon the "/initialize" key from the previous cluster, a Patroni node decide to  become a replica of an existing cluster instead of running initidb. Patroni does not delete all objects it creates, namely, non-expiring keys in the DCS, for instance, `/initialize`. There is simply no good point in the Patroni lifecycle to do this. 

Therefore, the operator should take care of deleting those keys: it knows the intention of the user to delete the cluster and it can do this after all Patroni nodes are shut down.

The difficulty is the variety of DCS implementations supported by Patroni (Etcd, Consul, Zookeeper, Kubernetes) and the operator would have to know of all of them. So, as a first step, we can make sure we delete old endpoints created when running in Kubernetes. For the rest of the DCSs, it would be good to come up with a more generic solution that doesn't hard-code the knowledge about how to delete the keys stored there in the operator.

avaczi:
@alexeyklyukin Wouldn't it make sense to implement the DCS cleanup in Patroni itself?  It would then be completely transparent from the operator POV.

alexeyklyukin:
There is no good point in the Patroni lifecycle to do it. Patroni manages individual PostgreSQL instances. Naturally, you want to delete keys when the instance is shut down; however, you don't know whether the shutdown is performed because the cluster is being deleted, or simply this particular instance should be stopped - this knowledge is completely outside of Patroni.

Ideally, we want to do it when all Patroni instances are stopped by the party that knows about the intention of the user to delete the cluster, but when all Patroni instances are shut down - how would you call Patroni to cleanup?

alexeyklyukin:
Theoretically, it should be possible to have a shutdown procedure by writing a special /cluster/shutdown key to the DCS: when the nodes see that key, they delete everything but the shutdown and the /member key from the DCS and then stop gracefully, deleting their member keys, and the last member (according to the /member key) also deletes the shutdown key (which also can be set to expire after a certain large interval in case the last member dies).

alexeyklyukin:
This is, however, error-prone, as determining the last /member will lead to race conditions; perhaps a better way would be for a single node to acquire the shutdown lock and waiting for the member keys other then the one belonging to the lock bearer to expire before removing the shutdown key and shutting down itself at last.

alexeyklyukin:
[...]

PR #244, opened 2018-02-20
Remove Patroni leftover objects on cluster deletion.

alexeyklyukin:
To address #241

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -583,6 +583,10 @@ func (c *Cluster) Delete() error {
 		}
 	}
 
+	if err := c.deletePatroniClusterObjects(); err != nil {
+		return fmt.Errorf("could not remove leftover patroni objects; %v", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


erthalion:
:+1:

Jan-M:
:+1:

alexeyklyukin:
👍

pkg/cluster/cluster.go
```diff
@@ -817,3 +822,57 @@ func (c *Cluster) shouldDeleteSecret(secret *v1.Secret) (delete bool, userName s
...
+	}
+	return nil
+}
+
+// TODO: figure out the interface shenanigans to avoid code duplication below.
+func (c *Cluster) deletePatroniClusterEndpoints() error {
+	for _, suffix := range patroniObjectSuffixes {
+		configEndpointName := fmt.Sprintf("%s-%s", c.Name, suffix)
+
+		if ep, err := c.KubeClient.Endpoints(c.Namespace).Get(configEndpointName, metav1.GetOptions{}); err == nil {
```
erthalion: Why not pass `c.KubeClient.Endpoints(c.Namespace)` as an argument here? In this way we can avoid this code duplication.

alexeyklyukin: I don't think it is that easy. Endpoint and ConfigMap implement 2 different interfaces and Get functions actually return different types (*v1.Endpoint and *v1.ConfigMap and respectively).

alexeyklyukin:
👍

erthalion:
:+1:

alexeyklyukin:
👍

PR #250, opened 2018-02-22
Make errors in the cluster list function visible.

alexeyklyukin:
Sometimes the operator does not pick up clusters right away when
they are created. The change attempts to shed light on the
reason behind that.


mgomezch:
:+1:

alexeyklyukin:
👍

PR #251, opened 2018-02-23
Move container comparison logic to a separate function

erthalion:
As part of random improvements idea: move out some part of `compareStatefulSetWith` with cyclomatic complexity 23 to a separate function to simplify it.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -309,33 +309,7 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *v1beta1.StatefulSet) *comp
 		needsRollUpdate = true
 		reasons = append(reasons, "new statefulset's container specification doesn't match the current one")
 	} else {
-		for index, container1 := range c.Statefulset.Spec.Template.Spec.Containers {
-			container2 := statefulSet.Spec.Template.Spec.Containers[index]
-			if container1.Name != container2.Name {
-				needsRollUpdate = true
-				reasons = append(reasons, fmt.Sprintf("new statefulset's container %d name doesn't match the current one", index))
-			}
-			if container1.Image != container2.Image {
-				needsRollUpdate = true
-				reasons = append(reasons, fmt.Sprintf("new statefulset's container %d image doesn't match the current one", index))
-			}
-			if !reflect.DeepEqual(container1.Ports, container2.Ports) {
-				needsRollUpdate = true
-				reasons = append(reasons, fmt.Sprintf("new statefulset's container %d ports don't match the current one", index))
-			}
-			if !compareResources(&container1.Resources, &container2.Resources) {
-				needsRollUpdate = true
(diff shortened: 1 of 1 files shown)


pkg/cluster/cluster.go
```diff
@@ -403,6 +377,66 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *v1beta1.StatefulSet) *comp
...
+type ContainerCheck struct {
+	condition func(a, b v1.Container) bool
+	reason    string
+}
+
+// compareContainers: compare containers from two stateful sets
+// and return:
+// * whether or not roll update is needed
+// * a list of reasons in a human readable format
+func (c *Cluster) compareContainers(setA, setB *v1beta1.StatefulSet) (bool, []string) {
```
Jan-M: This function could just have been copy & paste from the function above and it would have helped or?

Transforming 20 lines into 50 lines does not immediately convince me that the simple "if" is not good enough.

Jan-M: Could have maybe looked shorter if it would have been just:

checks.append("container name does not match"lambda x,y: x.name != y.name)
checks.append(...)

erthalion: Number of lines grew mostly because of formatting stuff, I don't think it's a valid metric here.
I also don't think that it's possible to make it shorter, but I can easily return the implementation
with if conditions. Even this would be better than it's now.

Jan-M:
:+1:

erthalion:
:-1:

erthalion:
:+1:

mgomezch:
Given that this bit of code here already uses `reflect` anyway, we could consider using it even more aggressively: https://stackoverflow.com/questions/18930910/golang-access-struct-property-by-name

e.g.

```go
	comparedFields := []string {"Name", "Image", "Ports", "Env", "EnvFrom"}
	comparators := map[string]interface{}{
		"Resources": compareResources,
	}

	func getField(v interface{}, field string) interface{} {
		return reflect.Indirect(reflect.ValueOf(v)).FieldByName(field)
	}

	for index, containerA := range setA.Spec.Template.Spec.Containers {
		containerB := setB.Spec.Template.Spec.Containers[index]
		for _, field := range comparedFields {
			fieldA := getField(a, field)
			fieldB := getField(b, field)
			comparator := comparators[field]
			if comparator != nil {
				comparator = reflect.DeepEqual
			}
			if (fieldA == nil) != (fieldB == nil) || !comparator(fieldA, fieldB) {
				needsRollUpdate = true
				reasons = append(reasons, fmt.Sprintf("new statefulset's container %d field %s doesn't match the current one", index, field))
			}
		}
	}
```

(NB: I didn't test this and I've never used this library so I have no idea if it works, but as far as I can tell it should probably work fine with only minor adjustments (maybe changing the parameter types of `compareResources` and possibly adding some casts inside or somesuch).  And it's easy to extend, clean, short, fits reasonably in the main function, and we're *already* relying on dynamic type information from the `reflection` module anyway, so introducing this stringy field access is no big deal IMHO)

PR #252, opened 2018-02-23
Add 'team' label to pods, stateful sets, secrets and pod disruption budgets

### LINKED ISSUE #245: Add team label to pods
It would be nice, if the postgres-operator would apply a team label to pods, identical to the team label on the postgresql manifest.

In a cluster shared by multiple teams this makes it easier to select all pods owned by a team. In particular I want to take advantage of this to monitor all pods of my team using zmon.

zerg-junior:
closes #245

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -491,7 +491,7 @@ func (c *Cluster) generatePodTemplate(
 
 	template := v1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels:    c.labelsSet(),
+			Labels:    c.labelsSet(true),
 			Namespace: c.Namespace,
 		},
 		Spec: podSpec,
@@ -559,7 +559,7 @@ func (c *Cluster) generateStatefulSet(spec *spec.PostgresSpec) (*v1beta1.Statefu
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      c.statefulSetName(),
 			Namespace: c.Namespace,
-			Labels:    c.labelsSet(),
+			Labels:    c.labelsSet(true),
 		},
 		Spec: v1beta1.StatefulSetSpec{
 			Replicas:             &numberOfInstances,
@@ -653,7 +653,7 @@ func (c *Cluster) generateSingleUserSecret(namespace string, pgUser spec.PgUser)
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      c.credentialSecretName(username),
 			Namespace: namespace,
-			Labels:    c.labelsSet(),
+			Labels:    c.labelsSet(true),
 		},
 		Type: v1.SecretTypeOpaque,
 		Data: map[string][]byte{
@@ -780,7 +780,7 @@ func (c *Cluster) generatePodDisruptionBudget() *policybeta1.PodDisruptionBudget
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      c.podDisruptionBudgetName(),
 			Namespace: c.Namespace,
-			Labels:    c.labelsSet(),
+			Labels:    c.labelsSet(true),
 		},
 		Spec: policybeta1.PodDisruptionBudgetSpec{
 			MinAvailable: &minAvailable,
--- a/pkg/cluster/pod.go
+++ b/pkg/cluster/pod.go
@@ -13,7 +13,7 @@ import (
 
 func (c *Cluster) listPods() ([]v1.Pod, error) {
 	listOptions := metav1.ListOptions{
-		LabelSelector: c.labelsSet().String(),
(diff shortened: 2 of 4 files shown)


zerg-junior:
For example, `kubectl get pod acid-test-cluster-0 --show-labels` on minikube returns labels:
`    
application=spilo,controller-revision-hash=acid-test-cluster-567f6866db,spilo-role=master,team=ACID,version=acid-test-cluster
`

mgomezch:
:+1:

zerg-junior:
:+1:

mgomezch:
I'm not sure I'm right about this, but I fear this may not be so simple :(

See e.g. how this set of labels is used to get [StatefulSets](https://github.com/zalando-incubator/postgres-operator/blob/c4aab502b35122ccfdbb55f0302bb3f7ee033e0c/pkg/cluster/util.go#L264), [pods](https://github.com/zalando-incubator/postgres-operator/blob/c4aab502b35122ccfdbb55f0302bb3f7ee033e0c/pkg/cluster/util.go#L280), [pods again](https://github.com/zalando-incubator/postgres-operator/blob/cca73e30b7fd10a27dde1014a691bf21c6518410/pkg/cluster/pod.go#L16), [pods again again](https://github.com/zalando-incubator/postgres-operator/blob/cca73e30b7fd10a27dde1014a691bf21c6518410/pkg/cluster/pod.go#L271), [PVCs](https://github.com/zalando-incubator/postgres-operator/blob/86803406db39c5ced5dcf5063fb6604d132911f8/pkg/cluster/volumes.go#L22)… what I mean is there are many places in the operator where this set of labels is expected to be the one that identifies resources we are using, and maybe existing resources could fail to be tracked with continuity if we change this set of labels and deploy the new operator into a cluster with existing resources managed by the operator.

@alexeyklyukin, can you confirm if it's okay to just add extra informational labels here, and whether we need to do anything special to migrate existing resources without leaving anything orphaned or otherwise broken?

alexeyklyukin:
I'm missing the idea behind this PR. For sure you can find the team by just looking at the initial part of the cliuster name.

mgomezch:
See the details of the requested feature in the linked issue: https://github.com/zalando-incubator/postgres-operator/issues/245

This would permit filtering any Operator-managed resource by team uniformly with a label selector, including those whose names are not simply prefixed with the team name (e.g. secrets, etc), as opposed to only those that have the team name as a prefix.  Generally, filtering resources by a pattern on the name is not straightforward in the Kubernetes API or in `kubectl`, while label selectors are the standard/preferred tool to achieve this in Kubernetes.

@alexeyklyukin and I discussed this briefly and there are a number of details that we need to flesh out before solving this.  One is that we would prefer to implement this as a general feature that would read all labels included in the database cluster manifest and propagate them to all downstream resources created subsequently by the Operator in response to this manifest.  Doing it just for the `team` label seems too hardcoded.

One problem that needs to be solved before we implement this is to decide how to handle any possible clashes with the labels used for other purposes; e.g. those specified in `cluster_labels`, `cluster_name_label` and `pod_role_label`, or those managed by other systems like `controller-revision-hash`.  Of course, one simple option here is to just say "don't do that", but some safeguards would be nice.

Also, it could be expected that changing the labels on the manifest of an existing cluster would propagate the change to existing downstream resources, which would require additional effort by the Operator perhaps by extending the sync process.  At this point, a tangentially related issue comes up: to consider what happens to existing resources if the labels specified in the operator's `ConfigMap` change.

This requires some more thought, especially to delimit the scope of this change.

avaczi:
Not sure the PR is the right place to discuss these, but:

> Also, it could be expected that changing the labels on the manifest of an existing cluster would propagate the change to existing downstream resources

Which is the very case we are already facing, namely team structure changes.  AFAIK, currently it is a problem to change a cluster's manifest when the owning team changes.  Labels all the way down would possibly help with it.

Jan-M:
:+1:

alexeyklyukin:
@zerg-junior 

If we do it in a current form, it will use team label to list objects, breaking compatibility with objects created by the previous version of the operator.

Would it make sense to add a flag extraLabels=true/false to the `labelsSet` function until we come with a better generic solution. The generate functions should  call it with a flag set to true, adding, potentially, some labels that are not strictly required for fetching objects. The list/wait functions should set it to false, using the minimum-possible set of labels to fetch a given object.

zerg-junior:
@mgomezch @alexeyklyukin updated. The `team` flag is now added only when pods, statefulsets, secrets, and budgets are created. All functions that list kubernetes objects still work with the minimum set of labels.

mgomezch:
:+1:

I would've preferred [avoiding the boolean trap](https://ariya.io/2011/08/hall-of-api-shame-boolean-trap), though.

alexeyklyukin:
👍

zerg-junior:
:+1:

PR #258, opened 2018-02-27
[WIP] Always create replica service

### LINKED ISSUE #257: Enable read replica support

zerg-junior:
Closes #257

### DIFF
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -12,8 +12,8 @@ spec:
     zalando:
     - superuser
     - createdb
-  # enableMasterLoadBalancer: true
-  # enableReplicaLoadBalancer: true 
+  enableMasterLoadBalancer: true
+  enableReplicaLoadBalancer: true 
   allowedSourceRanges: # load balancers' source ranges for both master and replica services
   - 127.0.0.1/32
   databases:
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -12,7 +12,7 @@ spec:
     zalando:
     - superuser
     - createdb
-  enableMasterLoadBalancer: true
+  # enableMasterLoadBalancer: true
   # enableReplicaLoadBalancer: true 
   allowedSourceRanges: # load balancers' source ranges for both master and replica services
   - 127.0.0.1/32
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -13,7 +13,7 @@ spec:
     - superuser
     - createdb
   enableMasterLoadBalancer: true
-  # enableReplicaLoadBalancer: true # uncomment to enable a LB for a replica k8s service
+  # enableReplicaLoadBalancer: true 
   allowedSourceRanges: # load balancers' source ranges for both master and replica services
   - 127.0.0.1/32
   databases:
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -12,8 +12,9 @@ spec:
     zalando:
     - superuser
     - createdb
-  useLoadBalancer: true
-  allowedSourceRanges: #Load balancer source ranges
+  enableMasterLoadBalancer: true
+  # enableReplicaLoadBalancer: true # uncomment to enable a LB for a replica k8s service
+  allowedSourceRanges: # load balancers' source ranges for both master and replica services
   - 127.0.0.1/32
   databases:
     foo: zalando
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -36,7 +36,8 @@ data:
   team_admin_role: "admin"
   teams_api_url: http://fake-teams-api.default.svc.cluster.local
   workers: "4"
-  enable_load_balancer: "true"
+  enable_master_load_balancer: "true"
+  # enable_replica_load_balancer: "true"
   api_port: "8080"
   ring_log_lines: "100"
   cluster_history_entries: "1000"
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -38,8 +38,8 @@ data:
   workers: "4"
   # turn on/off load balancers for all Postgres clusters managed by the operator
   # LB settings in cluster manifests take priority over these settings
-  # enable_master_load_balancer: "true"
-  # enable_replica_load_balancer: "true"
+  enable_master_load_balancer: "true"
+  enable_replica_load_balancer: "false"
   api_port: "8080"
   ring_log_lines: "100"
   cluster_history_entries: "1000"
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -36,7 +36,9 @@ data:
   team_admin_role: "admin"
   teams_api_url: http://fake-teams-api.default.svc.cluster.local
   workers: "4"
-  enable_master_load_balancer: "true"
+  # turn on/off load balancers for all Postgres clusters managed by the operator
+  # LB settings in cluster manifests take priority over these settings
+  # enable_master_load_balancer: "true"
   # enable_replica_load_balancer: "true"
   api_port: "8080"
   ring_log_lines: "100"
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -478,8 +478,7 @@ func (c *Cluster) Update(oldSpec, newSpec *spec.Postgresql) error {
 
 	// Service
 	if !reflect.DeepEqual(c.generateService(Master, &oldSpec.Spec), c.generateService(Master, &newSpec.Spec)) ||
-		!reflect.DeepEqual(c.generateService(Replica, &oldSpec.Spec), c.generateService(Replica, &newSpec.Spec)) ||
-		oldSpec.Spec.ReplicaLoadBalancer != newSpec.Spec.ReplicaLoadBalancer {
+		!reflect.DeepEqual(c.generateService(Replica, &oldSpec.Spec), c.generateService(Replica, &newSpec.Spec)) {
 		c.logger.Debugf("syncing services")
 		if err := c.syncServices(); err != nil {
 			c.logger.Errorf("could not sync services: %v", err)
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -217,9 +217,7 @@ func (c *Cluster) Create() error {
 	c.setStatus(spec.ClusterStatusCreating)
 
 	for _, role := range []PostgresRole{Master, Replica} {
-		if role == Replica && !c.Spec.ReplicaLoadBalancer {
-			continue
-		}
+
 		if c.Endpoints[role] != nil {
 			return fmt.Errorf("%s endpoint already exists in the cluster", role)
 		}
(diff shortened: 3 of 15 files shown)


zerg-junior:
This PR makes the `ReplicaLoadBalancer` param from the Postgres manifest govern the creation of a load balancer for a replica service and *not* the creation of the replica service itself as the param did before. If the param is not explicitly set in the manifest, the value `useLoadBalancer` determines if load balancers are used for both master and replica services.

The replica service is now always created.

pkg/spec/postgresql.go
```diff
@@ -96,9 +96,11 @@ type PostgresSpec struct {
...
 	AllowedSourceRanges []string `json:"allowedSourceRanges"`
 	DockerImage         string   `json:"dockerImage,omitempty"`
-	// EnableLoadBalancer  is a pointer, since it is important to know if that parameters is omitted from the manifest
-	UseLoadBalancer     *bool                `json:"useLoadBalancer,omitempty"`
-	ReplicaLoadBalancer bool                 `json:"replicaLoadBalancer,omitempty"`
+	// EnableLoadBalancer  is a pointer, since it is important to know if that parameters is omitted from the Postgres manifest
+	// in that case UseLoadBalancer == nil and the value is taken from the operator config
+	UseLoadBalancer *bool `json:"useLoadBalancer,omitempty"`
+	// if ReplicaLoadBalancer == nil (is unset), value of UseLoadBalancer determines if a balancer for replicas is created
+	ReplicaLoadBalancer *bool                `json:"replicaLoadBalancer,omitempty"`
```
erthalion: Just a rhetorical question - why so unobvious, true/false and a nil pointer? Maybe it would be more straightforward to introduce some sort of enum True/False/Unset ?

Jan-M: Why do we need to know where the value is coming from?

pkg/cluster/cluster.go
```diff
@@ -480,8 +478,7 @@ func (c *Cluster) Update(oldSpec, newSpec *spec.Postgresql) error {
 
 	// Service
 	if !reflect.DeepEqual(c.generateService(Master, &oldSpec.Spec), c.generateService(Master, &newSpec.Spec)) ||
-		!reflect.DeepEqual(c.generateService(Replica, &oldSpec.Spec), c.generateService(Replica, &newSpec.Spec)) ||
-		oldSpec.Spec.ReplicaLoadBalancer != newSpec.Spec.ReplicaLoadBalancer {
+		!reflect.DeepEqual(c.generateService(Replica, &oldSpec.Spec), c.generateService(Replica, &newSpec.Spec)) {
```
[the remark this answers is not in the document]
zerg-junior: > Why did you remove ReplicaLoadBalancer here?

AFAIK this check is redundant. Change in `ReplicaLoadBalancer` should be caught by  `DeepEqual(c.generateService(Replica, &oldSpec.Spec), c.generateService(Replica, &newSpec.Spec)`

pkg/cluster/k8sres.go
```diff
@@ -669,6 +669,20 @@ func (c *Cluster) generateSingleUserSecret(namespace string, pgUser spec.PgUser)
...
+
+	// handle LB for the replica service separately if ReplicaLoadBalancer is set in the Postgres manifest
+	if role == Replica && spec.ReplicaLoadBalancer != nil {
+		return *spec.ReplicaLoadBalancer
+	}
+
+	// create a load balancer for the master service if either Postgres or operator manifests tell to do so
+	// if ReplicaLoadBalancer is unset and LB for master service is created, also create a balancer for replicas
+	return (spec.UseLoadBalancer != nil && *spec.UseLoadBalancer) ||
+		(spec.UseLoadBalancer == nil && c.OpConfig.EnableLoadBalancer)
```
mgomezch: It took me a while to understand this logic as it's written.

The cluster gets a replica load balancer according to:

1. the value of `replicaLoadBalancer` if specified;
2. otherwise, the value of `useLoadBalancer` if specified;
3. otherwise, the value of `enableLoadBalancer` if specified;
4. otherwise, yes.

For the primary service, it's the same but skipping step 1.

This looks kind of like

```
coalesce(
  role == replica ? spec.ReplicaLoadBalancer : nil,
  spec.UseLoadBalancer,
  c.OpConfig.EnableLoadBalancer,
  true,
)
```

…except that the `c.OpConfig.EnableLoadBalancer` field is not actually treated as nullable by representing it with a pointer to bool, but instead is represented as a non-pointer struct field defaulting to `true` on JSON deserialization, which is an odd but minor inconsistency.

This is unfortunately a bit unwieldy to express clearly in Go (oh wow), but maybe this is a bit more clear?

```
func Coalesce(values ...*bool) *bool {
	for _, value := range values {
		if value != nil {
			return value
		}
	}
	return nil
}

func Conditional(condition bool, value *bool) *bool {
	if condition {
		return value
	}
	return nil
}

…

default := true
Coalesce(
  Conditional(role == replica, spec.ReplicaLoadBalancer),
  spec.UseLoadBalancer,
  c.OpConfig.EnableLoadBalancer,
  &default,
)
```

This is kind of complicated and verbose and not necessarily better than how the logic is currently spelled out (and it's also unfortunate that `Coalesce` is of limited use since either `bool` has to be hardwired or casts would be needed to use it), so it may be best left as it's currently written.

I think a more explicit comment would help, though.

Jan-M: By the way, the default should not be "true". It should be exactly what comes from opConfig EnableLoadBalancer, which is either true or false.

mgomezch: Oops, yes, that final bit of code is wrong (even has the wrong type there); it would have to be like

```
Coalesce(
  Conditional(role == replica, spec.ReplicaLoadBalancer),
  spec.UseLoadBalancer,
  &c.OpConfig.EnableLoadBalancer,
)
```

(The default exists but is set when the struct is parsed from JSON)

erthalion: @mgomezch I already proposed one to have a function like your modification `Coalesce` and I like this idea, but in this case maybe just a few simple if conditions would be better? Like:

```go
if role == Replica && spec.ReplicaLoadBalancer != nil {
    return *spec.ReplicaLoadBalancer
}

if spec.UseLoadBalancer != nil {
    return *spec.UseLoadBalancer
}

return c.OpConfig.EnableLoadBalancer
```

pkg/spec/postgresql.go
```diff
@@ -96,16 +96,24 @@ type PostgresSpec struct {
...
-	Databases           map[string]string    `json:"databases,omitempty"`
-	Tolerations         []v1.Toleration      `json:"tolerations,omitempty"`
+
+	// vars that enable load balancers are pointers because it is important to know if any of them is omitted from the Postgres manifest
+	// in that case the var evaluates to nil and the value is taken from the operator config
+	EnableMasterLoadBalancer  *bool `json:"enableMasterLoadBalancer,omitempty"`
+	EnableReplicaLoadBalancer *bool `json:"enableReplicaLoadBalancer,omitempty"`
+
+	// deprecated load balancer settings mantained for backward compatibility
+	// see "Load balancers" operator docs (PR #258)
```
erthalion: Does it necessary to have such references e.g. "PR #258" ? You can get the same information from git history.

pkg/util/config/config.go
```diff
@@ -67,20 +67,23 @@ type Config struct {
...
-	KubeIAMRole              string            `name:"kube_iam_role"`
-	DebugLogging             bool              `name:"debug_logging" default:"true"`
-	EnableDBAccess           bool              `name:"enable_database_access" default:"true"`
-	EnableTeamsAPI           bool              `name:"enable_teams_api" default:"true"`
-	EnableTeamSuperuser      bool              `name:"enable_team_superuser" default:"false"`
-	TeamAdminRole            string            `name:"team_admin_role" default:"admin"`
-	EnableLoadBalancer       bool              `name:"enable_load_balancer" default:"true"`
+	WatchedNamespace          string `name:"watched_namespace"` // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
+	EtcdHost                  string `name:"etcd_host" default:"etcd-client.default.svc.cluster.local:2379"`
+	DockerImage               string `name:"docker_image" default:"registry.opensource.zalan.do/acid/spiloprivate-9.6:1.2-p4"`
```
erthalion: Is it ok to have this version hardcoded here?

pkg/cluster/k8sres.go
```diff
@@ -689,12 +731,12 @@ func (c *Cluster) generateService(role PostgresRole, spec *spec.PostgresSpec) *v
...
 
-	// Examine the per-cluster load balancer setting, if it is not defined - check the operator configuration.
-	if (spec.UseLoadBalancer != nil && *spec.UseLoadBalancer) ||
-		(spec.UseLoadBalancer == nil && c.OpConfig.EnableLoadBalancer) {
+	if c.shouldCreateLoadBalancerForService(role, spec) {
 
 		// safe default value: lock load balancer to only local address unless overridden explicitly.
 		sourceRanges := []string{localHost}
+
+		// source ranges are the same for balancers for master and replica services
```
erthalion: We already have this comment in `postgresql.go`, so let's get rid of it here.

pkg/cluster/k8sres.go
```diff
@@ -669,6 +669,48 @@ func (c *Cluster) generateSingleUserSecret(namespace string, pgUser spec.PgUser)
...
+
+	case Master:
+
+		if spec.UseLoadBalancer != nil {
+			c.logger.Debugf("The Postgres manifest for the cluster %v sets the deprecated `useLoadBalancer` param. Consider using the `enableMasterLoadBalancer` instead.", c.Name)
+			return *spec.UseLoadBalancer
+		}
+
+		// `enable_load_balancer`` governs LB for a master service
+		// there is no equivalent operator configmap option for the replica LB
```
erthalion: `there is no equivalent...` this is confusing, taking into account `EnableReplicaLoadBalancer`, did you mean there is nothing like that in the old version?

manifests/configmap.yaml
```diff
@@ -36,7 +36,10 @@ data:
   team_admin_role: "admin"
   teams_api_url: http://fake-teams-api.default.svc.cluster.local
   workers: "4"
-  enable_load_balancer: "true"
+  # turn on/off load balancers for all Postgres clusters managed by the operator
+  # LB settings in cluster manifests take priority over these settings
+  # enable_master_load_balancer: "true"
```
erthalion: I still prefer these options being uncommented here with values 'false', what do you think?

manifests/complete-postgres-manifest.yaml
```diff
@@ -12,8 +12,9 @@ spec:
     zalando:
     - superuser
     - createdb
-  useLoadBalancer: true
-  allowedSourceRanges: #Load balancer source ranges
+  # enableMasterLoadBalancer: true
```
erthalion: I still prefer to have these options here uncommented with value 'false', since it's a default - what do you think?

pkg/util/config/config.go
zerg-junior: Outside the scope of this PR, but afaik yes. IMO the reason is that  we want our defaults be applicable across wide variety of installations since users won't change most of them most of the time. From that viewpoint, it makes sense to hardcode a version of Spilo that is certainly stable.

pkg/cluster/k8sres.go
zerg-junior: > did you mean there is nothing like that in the old version?

Yes, exactly. What I mean here is that in the deprecated set of params for configuring LB at the *operator* level - that is, for all Postgres clusters  the operator watches - there is no separate param for replica LBs.  Good catch :+1:

manifests/complete-postgres-manifest.yaml
zerg-junior: > have these options here uncommented with value 'false', since it's a default

These are two different behaviors. Not specifying the LB option in the Postgres manifest means "use relevant operator settings" . Setting the LB option to `false`  explicitly says "no load balancer please"

erthalion:
:1st_place_medal:

erthalion:
:+1:

zerg-junior:
:+1:

PR #260, opened 2018-03-01
Set WAL_BUCKET_SCOPE_PREFIX to the empty string

### LINKED ISSUE #259: Set WAL_BUCKET_SCOPE_PREFIX to the empty string

zerg-junior:
closes #259


Jan-M:
:+1:

zerg-junior:
:+1:

mgomezch:
For the record, this works around a bug in the cloning code that fails when the prefix variable is not set, instead of assuming an empty prefix when the variable is absent.

PR #261, opened 2018-03-01
Fix clone for origins in non-default namespaces.

alexeyklyukin:
By default, spilo sets WAL_BUCKET_SCOPE_PREFIX depending on the cluster
namespace, possibly to a non-empty string. However, we won't be able to
clone those clusters, as the clone prefix is always set to an empty string.

We could go the other way around and set both WAL_BUCKET_SCOPE_PREFIX
and CLONE_WAL_BUCKET_SCOPE_PREFIX to a non-default value that depends
on the cluster's namespace, but it seems that we don't need this
feature for now (no conflict will occur even for clusters with the
same name and different namespaces because of the SCOPE_SUFFIX) and
it requires some additional testing first.

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -361,6 +361,7 @@ func (c *Cluster) generatePodTemplate(
 	if c.OpConfig.WALES3Bucket != "" {
 		envVars = append(envVars, v1.EnvVar{Name: "WAL_S3_BUCKET", Value: c.OpConfig.WALES3Bucket})
 		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getWALBucketScopeSuffix(string(uid))})
+		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
 	}
 
 	if c.patroniUsesKubernetes() {


alexeyklyukin:
👍

Jan-M:
Sounds like this needs a few minutes of thinking, we will have a look. This would limit cloning to the same namespace basically?

alexeyklyukin:
No, on the contrary. In order to clone something from S3, you need to specify the exact name of a folder in a bucket. Right now, new clusters in non-default namespaces prepend the namespace to the cluster name to construct such a folder (this is done by spilo and is overridden exactly by the WAL_BUCKET_SCOPE_PREFIX), however, the cloning code explicitly sets the prefix to an empty string, so it will never construct a correct folder to clone for clusters in non-default namespaces.

zerg-junior:
> new clusters in non-default namespaces prepend the namespace to the cluster name to construct such a folder (this is done by spilo and is overridden exactly by the WAL_BUCKET_SCOPE_PREFIX)

Does it mean that setting WAL_BUCKET_SCOPE_PREFIX to the empty string will require a cluster name to be unique across all namespaces ?

alexeyklyukin:
@zerg-junior 
>Does it mean that setting WAL_BUCKET_SCOPE_PREFIX to the empty string will require a cluster name to be unique across all namespaces for saving data to S3 to be succesful ?

No; it used to be like this before we introduced the UID-based suffix for S3 buckets. Different clusters will get different UIDs and will get different S3 buckets.

zerg-junior:
:+1:

PR #270, opened 2018-04-06
Remove the check for the clone cluster name.

alexeyklyukin:
Clone cluster name may not adhere to the same rules as the names of the
clusters managed by the operator, as it might be just a name inside the
S3 bucket.

### DIFF
--- a/pkg/spec/postgresql.go
+++ b/pkg/spec/postgresql.go
@@ -3,6 +3,7 @@ package spec
 import (
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -76,6 +77,11 @@ const (
 	ClusterStatusInvalid      PostgresStatus = "Invalid"
 )
 
+const (
+	serviceNameMaxLength = 63
+	clusterNameMaxLength = serviceNameMaxLength - len("-repl")
+)
+
 // Postgresql defines PostgreSQL Custom Resource Definition Object.
 type Postgresql struct {
 	metav1.TypeMeta   `json:",inline"`
@@ -126,7 +132,10 @@ type PostgresqlList struct {
 	Items []Postgresql `json:"items"`
 }
 
-var weekdays = map[string]int{"Sun": 0, "Mon": 1, "Tue": 2, "Wed": 3, "Thu": 4, "Fri": 5, "Sat": 6}
+var (
+	weekdays         = map[string]int{"Sun": 0, "Mon": 1, "Tue": 2, "Wed": 3, "Thu": 4, "Fri": 5, "Sat": 6}
+	serviceNameRegex = regexp.MustCompile(`^[a-z]([-a-z0-9]*[a-z0-9])?$`)
+)
 
 func parseTime(s string) (time.Time, error) {
 	parts := strings.Split(s, ":")
@@ -225,8 +234,27 @@ func extractClusterName(clusterName string, teamName string) (string, error) {
 	if strings.ToLower(clusterName[:teamNameLen+1]) != strings.ToLower(teamName)+"-" {
 		return "", fmt.Errorf("name must match {TEAM}-{NAME} format")
 	}
+	if len(clusterName) > clusterNameMaxLength {
+		return "", fmt.Errorf("name cannot be longer than %d characters", clusterNameMaxLength)
+	}
+	if !serviceNameRegex.MatchString(clusterName) {
+		return "", fmt.Errorf("name must confirm to a valid service name (DNS-1035)")
+	}
 
 	return clusterName[teamNameLen+1:], nil
+}
+
+func validateCloneClusterDescription(clone *CloneDescription) error {
+	// when cloning from the basebackup (no end timestamp) check that the cluster name is a valid service name
+	if clone.ClusterName != "" && clone.EndTimestamp == "" {
+		if !serviceNameRegex.MatchString(clone.ClusterName) {
+			return fmt.Errorf("clone cluster name must confirm to a valid service name (DNS-1035)")
+		}
+		if len(clone.ClusterName) > serviceNameMaxLength {
+			return fmt.Errorf("clone cluster name must be no longer than %d characters", serviceNameMaxLength)
+		}
+	}
+	return nil
 }
 
 type postgresqlListCopy PostgresqlList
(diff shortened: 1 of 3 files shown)


Jan-M:
:+1:

alexeyklyukin:
👎 tests are still failing

mgomezch:
This means anyone able to deploy a manifest can clone from any cluster's backups, and thus read all the data in other clusters.  Just checking: this does not break any currently existing privilege boundaries?

alexeyklyukin:
It pretty much depends on how different teams are separated on the infrastructure level; if all of them share the backup location there is no way this PR makes things worse (as one could, for instance, just restore the backup of other team to your own cluster manually w/o any assistance from the operator)

alexeyklyukin:
👍

alexeyklyukin:
👍

pkg/spec/postgresql.go
```diff
@@ -252,22 +280,16 @@ func (p *Postgresql) UnmarshalJSON(data []byte) error {
 	}
 	tmp2 := Postgresql(tmp)
 
-	clusterName, err := extractClusterName(tmp2.ObjectMeta.Name, tmp2.Spec.TeamID)
-	if err == nil {
-		tmp2.Spec.ClusterName = clusterName
-	} else {
+	if clusterName, err := extractClusterName(tmp2.ObjectMeta.Name, tmp2.Spec.TeamID); err != nil {
```
Jan-M: will this code now support restoring from S3 where the luster name is basically: "foobar" (spilo ec2 deployments)

pkg/spec/postgresql_test.go
```diff
@@ -43,14 +43,29 @@ var clusterNames = []struct {
 	{"acid-test", "acid", "test", nil},
```
Jan-M: seems like "foobar" is not covered but should work as this is team "foo" and cluster "bar" in ec2 spilo way

pkg/spec/postgresql.go
alexeyklyukin: That line doesn't validate the clone cluster name, it happens at https://github.com/zalando-incubator/postgres-operator/pull/270/files/04d6a7607508fef669faacfb57ef9ef443b587da#diff-44fe837275fc3b64cc6c85eb90ce74c9R247

There, the name for the cluster clone that comes from S3 is not validated.

pkg/spec/postgresql.go
```diff
@@ -225,10 +234,29 @@ func extractClusterName(clusterName string, teamName string) (string, error) {
 	if strings.ToLower(clusterName[:teamNameLen+1]) != strings.ToLower(teamName)+"-" {
 		return "", fmt.Errorf("name must match {TEAM}-{NAME} format")
 	}
+	if len(clusterName) > clusterNameMaxLength {
+		return "", fmt.Errorf("name cannot be longer than %d characters", clusterNameMaxLength)
+	}
+	if !serviceNameRegex.MatchString(clusterName) {
+		return "", fmt.Errorf("name must confirm to a valid service name (DNS-1035)")
```
zerg-junior: RFC 1035 is a  50+ pages document. Can you please specify the section that provides overview of the naming  conventions and thus the regexp ?

pkg/spec/postgresql.go
```diff
@@ -225,10 +234,29 @@ func extractClusterName(clusterName string, teamName string) (string, error) {
...
+	}
 
 	return clusterName[teamNameLen+1:], nil
 }
 
+func validateCloneClusterDescription(clone *CloneDescription) error {
+	// when cloning from the basebackup (no end timestamp) check that the cluster name is a valid service name
+	if clone.ClusterName != "" && clone.EndTimestamp == "" {
+		if !serviceNameRegex.MatchString(clone.ClusterName) {
```
zerg-junior: Do existing cluster clones always have names adhering to this format ? that should be enforced by k8s, right ?

pkg/spec/postgresql.go
```diff
@@ -252,22 +280,16 @@ func (p *Postgresql) UnmarshalJSON(data []byte) error {
...
-	if err == nil {
-		tmp2.Spec.ClusterName = clusterName
-	} else {
+	if clusterName, err := extractClusterName(tmp2.ObjectMeta.Name, tmp2.Spec.TeamID); err != nil {
 		tmp2.Error = err
 		tmp2.Status = ClusterStatusInvalid
+	} else if err := validateCloneClusterDescription(&tmp2.Spec.Clone); err != nil {
+		tmp2.Error = err
+		tmp2.Status = ClusterStatusInvalid
+	} else {
```
zerg-junior: i feel the summary of [this discussion](https://github.com/zalando-incubator/postgres-operator/pull/270#issuecomment-379200852) is worth a comment here:

> anyone able to deploy a manifest can clone from any cluster's backups, and thus read all the data in other clusters.
>> It pretty much depends on how different teams are separated on the infrastructure level;

i.e. it is outside of the operator scope to check this

pkg/spec/postgresql.go
```diff
@@ -76,6 +77,11 @@ const (
 	ClusterStatusInvalid      PostgresStatus = "Invalid"
 )
 
+const (
+	serviceNameMaxLength = 63
```
zerg-junior: is it required by k8s  ?

pkg/spec/postgresql_test.go
```diff
@@ -435,11 +468,12 @@ func TestPostgresUnmarshal(t *testing.T) {
...
 		err := cluster.UnmarshalJSON(tt.in)
 		if err != nil {
-			if err.Error() != tt.err.Error() {
+			if tt.err == nil || err.Error() != tt.err.Error() {
 				t.Errorf("Unmarshal expected error: %v, got: %v", tt.err, err)
 			}
-
 			continue
+		} else if tt.err != nil {
+			t.Errorf("Expected error: %v", tt.err)
```
zerg-junior: and go no error, right ?

alexeyklyukin: yes

pkg/spec/postgresql.go
alexeyklyukin: yes

pkg/spec/postgresql.go
alexeyklyukin: Yes

alexeyklyukin: Yes

alexeyklyukin:
👍

alexeyklyukin:
👍

zerg-junior:
:+1:

PR #271, opened 2018-04-06
Remove team from statefulset selector

alexeyklyukin:
I was never supposed to be there, but implicitely statefulset creates a selector out of meta.labels field. That is the problem with recent Kubernetes, since statefulset cannot pick up pods with non-matching label selectors, and we rely on statefulset picking up old pods after statefulset replacement.


Jan-M:
:+1:

alexeyklyukin:
👍

Jan-M:
:+1:

Issue #272, opened 2018-04-06
Resume an incomplete rolling upgrade


alexeyklyukin:
When the Statefulset update for a cluster requires a rolling upgrade that cannot be retried on an error, as we don't check whether individual pods match the new statefulset definition (we used to do it before, but it was to cumbersome.) We need some durable flag to set when there is pending rolling upgrade and clear it once it finishes.  One idea is to introduce a status like field to the cluster description.

alexeyklyukin:
So the plan is the following:

* each time we need a rolling upgrade because of the statefulset changes, record this fact in the cluster structure by setting a boolean flag to True. Once the flag is set, do the  rolling upgrade and reset the flag at the end. If we make the flag a boolean pointer we can use the nil value to indicate that the rolling upgrade state is unknown and we need to look at the pod labels to figure it out.

* when the rolling upgrade flag is set to nil, go to each pod for that cluster and compare its `controller-revision-hash` with the revisions in the statefulset status to figure out if the pod belongs to the current version of the statefulset. Depending on the outcome, if at least one pod is not up-to-date, set the flag to False.

As a result, we could detect the need for a rolling upgrade and at the same time limit ourselves to only one scan of the pods during the very first sync of the cluster since the start of the operator.

PR #275, opened 2018-04-09
Check rolling updates using controller revisions.

alexeyklyukin:
Compare pods controller revisions with the one for the statefulset to determine whether the pod is running the latest revision and, therefore, no rolling update is necessary. This is performed only during the operator start, afterwards the rolling update status that is stored locally in the cluster structure is used for all rolling update decisions.

Per #272


alexeyklyukin:
This one is deprecated by https://github.com/zalando-incubator/postgres-operator/pull/286, which provides a more reliable way to detect rolling updates (and, as a bonus, a mechanism to trigger rolling updates on-demand that is potentially useful when debugging issues around statefulsets).

PR #276, opened 2018-04-16
Fix clone timestamp key in example manifest

mgomezch:
It was set to `endTimestamp`, but it should be `timestamp`.


mgomezch:
:+1:

Jan-M:
:+1:

PR #277, opened 2018-04-18
Deploy service account for pod creation on demand

zerg-junior:
Stateful sets need a service account to create Spilo pods. This PR automates deployment of such an account, thereby reducing manual work required to deploy a database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ By default, the operator watches the namespace it is deployed to. You can change
 
 Note that for an operator to manage pods in the watched namespace, the operator's service account (as specified in the operator deployment manifest) has to have appropriate privileges to access the watched namespace. The operator may not be able to function in the case it watches all namespaces but lacks access rights to any of them (except Kubernetes system namespaces like `kube-system`). The reason is that for multiple namespaces operations such as 'list pods' execute at the cluster scope and fail at the first violation of access rights.
 
-The watched namespace also needs to have a (possibly different) service account in the case database pods need to talk to the Kubernetes API (e.g. when using Kubernetes-native configuration of Patroni).
+The watched namespace also needs to have a (possibly different) service account in the case database pods need to talk to the Kubernetes API (e.g. when using Kubernetes-native configuration of Patroni). The operator checks that the `pod_service_account_name` exists in the target namespace, and, if not, deploys there the `pod_service_account_definition`. In this definition, the operator overwrites the account's name to match `pod_service_account_name` and the namespace to match the target namespace. The operator  performs **no** further syncing of this account.
 
 ### Create ConfigMap
 
--- a/README.md
+++ b/README.md
@@ -87,7 +87,16 @@ By default, the operator watches the namespace it is deployed to. You can change
 
 Note that for an operator to manage pods in the watched namespace, the operator's service account (as specified in the operator deployment manifest) has to have appropriate privileges to access the watched namespace. The operator may not be able to function in the case it watches all namespaces but lacks access rights to any of them (except Kubernetes system namespaces like `kube-system`). The reason is that for multiple namespaces operations such as 'list pods' execute at the cluster scope and fail at the first violation of access rights.
 
-The watched namespace also needs to have a (possibly different) service account in the case database pods need to talk to the Kubernetes API (e.g. when using Kubernetes-native configuration of Patroni). The operator checks that the `pod_service_account_name` exists in the target namespace, and, if not, deploys there the `pod_service_account_definition`. In this definition, the operator overwrites the account's name to match `pod_service_account_name` and the namespace to match the target namespace. The operator  performs **no** further syncing of this account.
+The watched namespace also needs to have a (possibly different) service account in the case database pods need to talk to the Kubernetes API (e.g. when using Kubernetes-native configuration of Patroni). The operator checks that the `pod_service_account_name` exists in the target namespace, and, if not, deploys there the `pod_service_account_definition` from the operator [`Config`](pkg/util/config/config.go) with the default value of:
+
+```yaml
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+ name: operator
+```
+
+ In this definition, the operator overwrites the account's name to match `pod_service_account_name` and the `default` namespace to match the target namespace. The operator  performs **no** further syncing of this account.
 
 ### Create ConfigMap
 
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -256,6 +256,11 @@ func (c *Cluster) Create() error {
 	}
 	c.logger.Infof("pod disruption budget %q has been successfully created", util.NameFromMeta(pdb.ObjectMeta))
 
+	if err = c.syncPodServiceAccounts(); err != nil {
+		return fmt.Errorf("could not sync pod service accounts: %v", err)
+	}
+	c.logger.Infof("pod service accounts have been successfully synced")
+
 	if c.Statefulset != nil {
 		return fmt.Errorf("statefulset already exists in the cluster")
 	}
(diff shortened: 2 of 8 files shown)


pkg/util/config/config.go
```diff
@@ -67,21 +68,25 @@ type Config struct {
...
-	EnableMasterLoadBalancer  bool   `name:"enable_master_load_balancer" default:"true"`
-	EnableReplicaLoadBalancer bool   `name:"enable_replica_load_balancer" default:"false"`
+	PodServiceAccount v1.ServiceAccount // has to be struct value, not a pointer
+	WatchedNamespace  string            `name:"watched_namespace"` // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
+	EtcdHost          string            `name:"etcd_host" default:"etcd-client.default.svc.cluster.local:2379"`
+	DockerImage       string            `name:"docker_image" default:"registry.opensource.zalan.do/acid/spiloprivate-9.6:1.2-p4"`
+	// re-use one account for both Spilo pods and the operator; this grants extra privileges to pods
+	ServiceAccountName          string `name:"service_account_name" default:"operator"`
+	PodServiceAccountName       string `name:"pod_service_account_name" default:"operator"`
+	PodServiceAccountDefinition string `name:"pod_service_account_definition" default:"apiVersion: v1\nkind: ServiceAccount\nmetadata:\n  name: operator\n"`
```
alexeyklyukin: I'd make it a stringTemplate and add {serviceAccountName} inside the template; that way, it would be more obvious that the name is just a template

pkg/util/config/config.go
```diff
@@ -67,21 +68,25 @@ type Config struct {
...
-	WALES3Bucket              string `name:"wal_s3_bucket"`
-	KubeIAMRole               string `name:"kube_iam_role"`
-	DebugLogging              bool   `name:"debug_logging" default:"true"`
-	EnableDBAccess            bool   `name:"enable_database_access" default:"true"`
-	EnableTeamsAPI            bool   `name:"enable_teams_api" default:"true"`
-	EnableTeamSuperuser       bool   `name:"enable_team_superuser" default:"false"`
-	TeamAdminRole             string `name:"team_admin_role" default:"admin"`
-	EnableMasterLoadBalancer  bool   `name:"enable_master_load_balancer" default:"true"`
-	EnableReplicaLoadBalancer bool   `name:"enable_replica_load_balancer" default:"false"`
+	PodServiceAccount v1.ServiceAccount // has to be struct value, not a pointer
```
alexeyklyukin: Just as a comment, not an immediate call for action, I think everything would be more clear to users if we start calling this PatroniServiceAccount and  also rename the whole project to PatroniOperator. Won't happen anytime soon most likely.

alexeyklyukin: Why should it be part for the Config structure; I'd rather move it to the controller itself; also, why cannot it be a pointer?

pkg/util/config/config.go
```diff
@@ -67,21 +68,25 @@ type Config struct {
...
+	WatchedNamespace  string            `name:"watched_namespace"` // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
+	EtcdHost          string            `name:"etcd_host" default:"etcd-client.default.svc.cluster.local:2379"`
+	DockerImage       string            `name:"docker_image" default:"registry.opensource.zalan.do/acid/spiloprivate-9.6:1.2-p4"`
+	// re-use one account for both Spilo pods and the operator; this grants extra privileges to pods
+	ServiceAccountName          string `name:"service_account_name" default:"operator"`
```
alexeyklyukin: Hm... I can't find the reason why does the operator need to know about its own service account, given that now we have PodServiceAccountName; looks like k8res.go:438 needs to be changed to it. Afterwards, we can remove serviceAccountName altogether.

README.md
```diff
@@ -87,7 +87,7 @@ By default, the operator watches the namespace it is deployed to. You can change
...
 
-The watched namespace also needs to have a (possibly different) service account in the case database pods need to talk to the Kubernetes API (e.g. when using Kubernetes-native configuration of Patroni).
+The watched namespace also needs to have a (possibly different) service account in the case database pods need to talk to the Kubernetes API (e.g. when using Kubernetes-native configuration of Patroni). The operator checks that the `pod_service_account_name` exists in the target namespace, and, if not, deploys there the `pod_service_account_definition`. In this definition, the operator overwrites the account's name to match `pod_service_account_name` and the namespace to match the target namespace. The operator  performs **no** further syncing of this account.
```
alexeyklyukin: I think we should include the default definition into this explanation.

pkg/util/config/config.go
Jan-M: That does sound true to me too.

pkg/util/config/config.go
Jan-M: Right now probably ok, I am just wondering if/when we arrive at a kind of "sync" for this, if we may need the ability to spec a service account name different than the template, e.g. to roll out service accounts first. But let's stay with the simple solution here for now until we need more.

pkg/util/config/config.go
zerg-junior: This service account is accessed during cluster creation through the copy of the operator `Config` struct present in the every `Cluster` struct.

> Why should it be part for the Config structure; I'd rather move it to the controller itself

that is also an option, by what then would be a correct way to reference this `Controller` struct from within cluster `Create()` ?

> why cannot it be a pointer?

just to prevent potential race condition in the case the account needs to be deployed to two different namespaces simultaneously.

zerg-junior:
@alexeyklyukin updated

pkg/util/config/config.go
alexeyklyukin: > that is also an option, by what then would be a correct way to reference this Controller struct from within cluster Create() ?

We can also copy it when the cluster is created from the controller into the cluster. The point is that it doesn't belong to the operator config, I'd rather have a separate field in the Cluster structure for it.

> To prevent potential race condition in the case the account needs to be deployed to two different namespaces simultaneously. If it were a pointer, then two copies of the operator Config might inadvertently re-use the same "namespace" field in the same service account struct.

>To prevent potential race condition in the case the account needs to be deployed to two different namespaces simultaneously. If it were a pointer, then two copies of the operator Config might inadvertently re-use the same "namespace" field in the same service account struct.

Well, Create at the end of the day is just a REST API call that shouldn't modify the definition passed to it. I think you can remove the namespace setting from the service description altogether and rely on the Services(Namespace) to create the service account in the right namespace.

pkg/cluster/cluster.go
```diff
@@ -194,6 +195,38 @@ func (c *Cluster) initUsers() error {
...
 
+/*
+  Ensures the service account required by StatefulSets to create pods exists in a namespace before a PG cluster is created there so that a user does not have to deploy the account manually.
+
+  The operator does not sync these accounts after creation.
+*/
+func (c *Cluster) createPodServiceAccounts() error {
+
+	podServiceAccountName := c.Config.OpConfig.PodServiceAccountName
+	c.setProcessName(fmt.Sprintf("creating pod service account in the namespace %v", c.Namespace))
```
alexeyklyukin: This should go only after the check return there is no existing service account

pkg/cluster/cluster.go
```diff
@@ -194,6 +195,38 @@ func (c *Cluster) initUsers() error {
...
+
+	_, err := c.KubeClient.ServiceAccounts(c.Namespace).Get(podServiceAccountName, metav1.GetOptions{})
+
+	if err != nil {
+		c.logger.Infof("the pod service account %q cannot be retrieved in the namespace %q; stateful sets in the namespace may be unable to create pods. Trying to deploy the account.", podServiceAccountName, c.Namespace)
```
alexeyklyukin: Here we should just state our attempt to create the service account, a message that says that the statefulset might be unable to create pods is confusing.

alexeyklyukin:
👍

zerg-junior:
:+1:

alexeyklyukin:
👍

PR #278, opened 2018-04-23
Tolerate issues of the Teams API

zerg-junior:
The sync should continue in case  team service issue/unexpected response; return an empty list of team members as a fallback value. The change affects `Create()`, `Sync()` and `Update()`


zerg-junior:
:+1:

Issue #279, opened 2018-04-24
Yet another upgrade oversight


alexeyklyukin:
Looks like our upgrade process can still produce some unnecessary downtime. Consider the scenario:

- there are two nodes, one (A) hosting the primary and the other (B) the standby PostgreSQL cluster.
- first, node B is about to go down, the operator choses to do nothing since it doesn't have any primaries
- now it takes some time for the standby to be respawned on some node
- next, node A goes down, the operator looks for the standby instances of the cluster, finds none (the only standby is still being recreated) and decides to shut down master, as their no candidates to failover to.

In the original implementation of the upgrade we used to wait for both primaries and standbys by both having a PDB that covered both of them and running the code that migrated the standbys and waited until the pod was actually re-created.

The later modifications made moving of the replicas unnecessary, since they don't affect the downtime. Unfortunately, that lead to the aforementioned issues of shutting down all replicas and  not being able to find the proper candidate for moving the master off the node to be decommissioned.

One idea to solve that would be to examine the number of replicas that are supposed to run for  a given cluster and wait for at least one of them to come up in order to move the master.

alexeyklyukin:
After discussing this with @mgomezch we realized that waiting for the replica to come up might be a costly operation, especially if there are multiple clusters on the same node; essentially, if we have 3 nodes and each master takes 5 minutes for the replica to come up (which is not unheard of even in the case of a single replica and more likely the more clusters we have on the node), making the total amount of time the 3rd node will wait already 15 minutes. At some point, it might exceed the timeout for the node to be killed unconditionally, making the whole process pointless. So a followup to the initial proposal would be to parallelize the process, both doing multiple nodes at a time and also doing multiple clusters concurrently from the same name.

PR #281, opened 2018-04-27
Improve the pod moving behavior during the Kubernetes cluster upgrade.

alexeyklyukin:
Fix an issue of not waiting for at least one replica to become ready
(if the Statefulset indicates there are replicas) when moving the master
pod off the decomissioned node. Resolves the first part of #279.

Small fixes to error messages.

### DIFF
--- a/pkg/cluster/pod.go
+++ b/pkg/cluster/pod.go
@@ -149,13 +149,19 @@ func (c *Cluster) movePodFromEndOfLifeNode(pod *v1.Pod) (*v1.Pod, error) {
 }
 
 func (c *Cluster) masterCandidate(oldNodeName string) (*v1.Pod, error) {
+
+	// Wait until at least one replica pod will come up
+	if err := c.waitForAnyReplicaLabelReady(); err != nil {
+		c.logger.Warningf("could not find at least one ready replica: %v", err)
+	}
+
 	replicas, err := c.getRolePods(Replica)
 	if err != nil {
 		return nil, fmt.Errorf("could not get replica pods: %v", err)
 	}
 
 	if len(replicas) == 0 {
-		c.logger.Warningf("single master pod for cluster %q, migration will cause disruption of the service")
+		c.logger.Warningf("no available master candidates, migration will cause longer downtime of the master instance")
 		return nil, nil
 	}
 
@@ -168,12 +174,16 @@ func (c *Cluster) masterCandidate(oldNodeName string) (*v1.Pod, error) {
 			}
 		}
 	}
-	c.logger.Debug("no available master candidates on live nodes")
+	c.logger.Warningf("no available master candidates on live nodes")
 	return &replicas[rand.Intn(len(replicas))], nil
 }
 
 // MigrateMasterPod migrates master pod via failover to a replica
 func (c *Cluster) MigrateMasterPod(podName spec.NamespacedName) error {
+	var (
+		masterCandidatePod *v1.Pod
+	)
+
 	oldMaster, err := c.KubeClient.Pods(podName.Namespace).Get(podName.Name, metav1.GetOptions{})
 
 	if err != nil {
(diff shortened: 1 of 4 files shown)


alexeyklyukin:
👎 the race condition is apparently still there

alexeyklyukin:
👍

alexeyklyukin:
As a note, we always wait for the label without checking the status of the pod (it could be that the pod is in termination phase but the label is still there.). This, however, should be a subject of a separate PR to avoid having too many things crammed into this one.

pkg/cluster/pod.go
```diff
@@ -168,12 +174,16 @@ func (c *Cluster) masterCandidate(oldNodeName string) (*v1.Pod, error) {
 			}
 		}
 	}
-	c.logger.Debug("no available master candidates on live nodes")
+	c.logger.Warningf("no available master candidates on live nodes")
```
zerg-junior: Warning**f** or just `Warning` ?

pkg/cluster/util.go
```diff
@@ -295,35 +300,55 @@ func (c *Cluster) waitPodLabelsReady() error {
 			c.OpConfig.PodRoleLabel: string(Replica),
 		}).String(),
 	}
-	pods, err := c.KubeClient.Pods(namespace).List(listOptions)
-	if err != nil {
-		return err
+	podsNumber = 1
+	if !anyReplica {
```
zerg-junior: `anyReplica`  -> sth like `canUseAnyReplica` ?

pkg/cluster/util.go
```diff
@@ -295,35 +300,55 @@ func (c *Cluster) waitPodLabelsReady() error {
...
+		pods, err := c.KubeClient.Pods(namespace).List(listOptions)
+		if err != nil {
+			return err
+		}
+		podsNumber = len(pods.Items)
+		c.logger.Debugf("Waiting for %d pods to become ready", podsNumber)
+	} else {
+		c.logger.Debugf("Waiting for any replica pod to become ready")
```
zerg-junior: Debug or Debug**f** ?

pkg/cluster/util.go
```diff
@@ -295,35 +300,55 @@ func (c *Cluster) waitPodLabelsReady() error {
...
-	if err != nil {
-		return err
+	podsNumber = 1
+	if !anyReplica {
+		pods, err := c.KubeClient.Pods(namespace).List(listOptions)
+		if err != nil {
+			return err
+		}
+		podsNumber = len(pods.Items)
+		c.logger.Debugf("Waiting for %d pods to become ready", podsNumber)
```
zerg-junior: probably that line  should mention that the operator waits for all replicas of that PG cluster and then state their number ?

pkg/cluster/util.go
```diff
@@ -295,35 +300,55 @@ func (c *Cluster) waitPodLabelsReady() error {
...
-			if err2 != nil {
-				return false, err2
+			masterCount := 0
+			if !anyReplica {
+				masterPods, err2 := c.KubeClient.Pods(namespace).List(masterListOption)
+				if err2 != nil {
+					return false, err2
+				}
+				if len(masterPods.Items) > 1 {
+					return false, fmt.Errorf("too many masters")
```
zerg-junior: worth adding an exact number ?

pkg/cluster/util.go
```diff
@@ -295,35 +300,55 @@ func (c *Cluster) waitPodLabelsReady() error {
...
 			replicaPods, err2 := c.KubeClient.Pods(namespace).List(replicaListOption)
 			if err2 != nil {
 				return false, err2
 			}
-			if len(masterPods.Items) > 1 {
-				return false, fmt.Errorf("too many masters")
-			}
-			if len(replicaPods.Items) == podsNumber {
+			replicaCount := len(replicaPods.Items)
+			if anyReplica && replicaCount > 1 {
```
zerg-junior: why strictly greater than 1?

pkg/cluster/util.go
alexeyklyukin: Both should produce the same result.

pkg/cluster/util.go
alexeyklyukin: For the debugging purposes it's important to know how many replicas we are waiting for. The person examining the output would already know the context from the prior debug messages in the same session (i.e. when we are waiting for all pods they will see it is because of the statefulset.

pkg/cluster/util.go
alexeyklyukin: A typo, good point!

pkg/cluster/util.go
alexeyklyukin: makes sense.

alexeyklyukin:
👍

PR #282, opened 2018-04-27
Fix a bug with syncing services

alexeyklyukin:
Avoid showing "there is no service in the cluster" when syncing a
service for the cluster if the operator has been restarted after
the cluster had been created.


alexeyklyukin:
👍

Jan-M:
:+1:

Issue #280, opened 2018-04-26
Create the pod service account if missing during sync


mgomezch:
During processing of a sync event for a database cluster, the pod service account should be queried on the database cluster's target namespace, and if the pod service account is missing, then it should be created.

Note that this is a *separate* issue from deciding what to do if a service account is found in the cluster's target namespace, in which case syncing is unfortunately more complex.

(For now, an effective workaround for any existing database clusters that fail to start due to a missing service account is to deploy another database cluster to the same namespace, because then the service account creation logic will kick in and the stuck StatefulSet should heal.)

zerg-junior:
Won't fix for now since the motivation behind this issue is a specific use case, namely:
1. a PG cluster is started under the older operator version that does not check/deploy  a pod service account  during cluster creation
2.  The cluster won't able to start due to the lack of the service account 
3. At this moment, the operator is upgraded to the version that deploys service account during creation.

i see this situation is a quite unlikely known issue  and not worth a specific workaround in `Sync()` that would affect all clusters.

PR #283, opened 2018-04-27
Minor log improvements

zerg-junior:
result of investigating issue #280

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -293,7 +293,7 @@ func (c *Cluster) Create() error {
 	if err = c.createPodServiceAccounts(); err != nil {
 		return fmt.Errorf("could not create pod service account %v : %v", c.OpConfig.PodServiceAccountName, err)
 	}
-	c.logger.Infof("pod service accounts have been successfully synced")
+	c.logger.Infof("pod service accounts have been successfully found/created")
 
 	if c.Statefulset != nil {
 		return fmt.Errorf("statefulset already exists in the cluster")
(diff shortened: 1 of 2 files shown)


pkg/cluster/pod.go
```diff
@@ -155,7 +155,7 @@ func (c *Cluster) masterCandidate(oldNodeName string) (*v1.Pod, error) {
 	}
 
 	if len(replicas) == 0 {
-		c.logger.Warningf("single master pod for cluster %q, migration will cause disruption of the service")
+		c.logger.Warningf("single master pod for cluster %q, migration will cause disruption of the service", c.Name)
```
Jan-M: is this needed? dont we have the log context displaying this?

pkg/cluster/cluster.go
```diff
@@ -293,7 +293,7 @@ func (c *Cluster) Create() error {
 	if err = c.createPodServiceAccounts(); err != nil {
 		return fmt.Errorf("could not create pod service account %v : %v", c.OpConfig.PodServiceAccountName, err)
 	}
-	c.logger.Infof("pod service accounts have been successfully synced")
+	c.logger.Infof("pod service accounts have been successfully found/created")
```
alexeyklyukin: Drop it, we already show relevant  messages inside createPodServiceAccount.

pkg/cluster/pod.go
alexeyklyukin: I've already fixed that as a part of https://github.com/zalando-incubator/postgres-operator/pull/281/ (and the working here is confusing, since it could be interpreted as something related to the Kubernetes service).

zerg-junior:
closed as relevant changes  are introduced in #281

PR #287, opened 2018-05-09
Deprecate old LB options, fix endpoint sync.

alexeyklyukin:
- deprecate useLoadBalancer, replicaLoadBalancer from the manifest
  and enable_load_balancer from the operator configuration. Those
  options become no-op with this commit.
- Make sure the endpoint being created during the sync receives proper
  addresses subset. This is more critical for the replicas, as for the
  masters Patroni will normally re-create the endpoint before the
  operator.
- Update the README and unit tests.

### DIFF
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -343,11 +343,17 @@ func (c *Cluster) deleteService(role PostgresRole) error {
 	return nil
 }
 
-func (c *Cluster) createEndpoint(role PostgresRole) (*v1.Endpoints, error) {
+func (c *Cluster) createEndpoint(role PostgresRole, sync bool) (*v1.Endpoints, error) {
+	var (
+		subsets []v1.EndpointSubset
+	)
 	c.setProcessName("creating endpoint")
-	subsets := make([]v1.EndpointSubset, 0)
-	if role == Master {
-		//TODO: set subsets to the master
+	if sync {
+		subsets = c.generateEndpointSubsets(role)
+	} else {
+		// Patroni will populate the master endpoint for the new cluster
+		// The replica endpoint will be filled-in by the service selector.
+		subsets = make([]v1.EndpointSubset, 0)
 	}
 	endpointsSpec := c.generateEndpoint(role, subsets)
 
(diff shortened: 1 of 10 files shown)


alexeyklyukin:
👍

pkg/cluster/resources.go
```diff
@@ -343,11 +343,17 @@ func (c *Cluster) deleteService(role PostgresRole) error {
 	return nil
 }
 
-func (c *Cluster) createEndpoint(role PostgresRole) (*v1.Endpoints, error) {
+func (c *Cluster) createEndpoint(role PostgresRole, sync bool) (*v1.Endpoints, error) {
```
mgomezch: I always find unnamed boolean function parameters to be a bit unclear.  Instead of using a `bool` parameter here, could we use the cluster status from `c.GetStatus()` and compare it to `ClusterStatusCreating`?

pkg/controller/postgresql.go
```diff
@@ -287,6 +298,46 @@ func (c *Controller) processClusterEventsQueue(idx int, stopCh <-chan struct{},
 	}
 }
 
+func (c *Controller) warnOnDeprecatedPostgreSQLSpecParameters(spec *spec.PostgresSpec) {
+
+	deprecate := func(deprecated, replacement string) {
+		c.logger.Warningf("Parameter %q is deprecated. Consider setting %q instead")
```
zerg-junior: missing `deprecated`, `replacement` in the warning ?

pkg/cluster/resources.go
```diff
@@ -343,11 +343,17 @@ func (c *Cluster) deleteService(role PostgresRole) error {
...
+		subsets []v1.EndpointSubset
+	)
 	c.setProcessName("creating endpoint")
-	subsets := make([]v1.EndpointSubset, 0)
-	if role == Master {
-		//TODO: set subsets to the master
+	if sync {
+		subsets = c.generateEndpointSubsets(role)
+	} else {
+		// Patroni will populate the master endpoint for the new cluster
```
zerg-junior: "populate" here - assing IP addr and port ?

zerg-junior:
:+1:

mgomezch:
:+1:

alexeyklyukin:
👍

mgomezch:
:+1:

PR #293, opened 2018-05-18
Sync persistent volumes before statefulsets.

alexeyklyukin:
Avoid the condition of waiting for the pod that cannot start
PostgreSQL because it ran out of disk space.


alexeyklyukin:
👍

Jan-M:
:+1:

zerg-junior:
:+1:

PR #294, opened 2018-05-18
Fix crash during sync.

alexeyklyukin:
Do not use statefulset number of pods to figure out running ones
for volume resizing, since the statefulset pointer could be nil.
Instead, look at the actual running pods.


zerg-junior:
:+1:

alexeyklyukin:
👍

Issue #218, opened 2018-02-02
Incomplete deletion if resources upon delete postgresql


avaczi:
Under circumstances (for example, missing pod disruption budget) a `kubectl delete postgresql {name}` will return successfully, but leaving behind resources.  The case of PDB leaves behind the service and the endpoint, for example.

The reason is the early return upon errors in https://github.com/zalando-incubator/postgres-operator/blob/bf4406d2a49deb73830eeb72cb1b35cdccd846e0/pkg/cluster/cluster.go#L568-L583

The logic has to be revised so that we don't leave behind stuff, at least not when the error stems from the nonexistence of the resource.

mgomezch:
Log output for the incomplete deletion:

```
time="2018-02-02T09:39:05Z" level=info msg="\"DELETE\" event has been queued" cluster-name=default/XXX pkg=controller worker=3
time="2018-02-02T09:39:05Z" level=info msg="deletion of the cluster started" cluster-name=default/XXX pkg=controller worker=3
time="2018-02-02T09:39:05Z" level=debug msg="deleting statefulset" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=info msg="statefulset \"default/XXX\" has been deleted" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=debug msg="deleting pods" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=debug msg="no pods to delete" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=debug msg="deleting PVCs" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=debug msg="no PVCs to delete" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=info msg="not removing secret \"standby.XXX.credentials.postgresql.YYY\" for the system user \"standby\"" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=debug msg="deleting secret \"default/ZZZ.XXX.credentials.postgresql.YYY\"" cluster-name=default/XXX pkg=cluster worker=3
time="2018-02-02T09:39:05Z" level=info msg="secret \"default/ZZZ.XXX.credentials.postgresql.YYY\" has been deleted" cluster-name=default/XXX pkg=cluster worker=3
[...]

PR #295, opened 2018-05-18
Avoid terminating delete on errors.

alexeyklyukin:
When there is an error happening upon deletion of the Kubernetes object
belonging to the cluster being removed, it makes no sense to abort the
deletion: the manifest will be removed anyway, therefore all the objects
after the one we aborted at will stay forever.


alexeyklyukin:
This partially addresses #218 , as it will not stop deletion when the object to delete is not there; however, it will also continue on any other error, as delete is a one-shot operation: once the CRD is removed the operator leaves no traces of the cluster existence (we could avoid removing the in-memory structures, but that wouldn't suffice in case the operator is restarted somewhere after the unsuccessful deletion).

alexeyklyukin:
👍

zerg-junior:
:+1:

PR #299, opened 2018-05-23
Use Patroni API to set bootstrap-only options.

alexeyklyukin:
Call Patroni API /config in order to set special options that are
ignored when set in the configuration file, such as max_connections.
Per https://github.com/zalando-incubator/postgres-operator/issues/297

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -229,7 +229,7 @@ func (c *Cluster) syncStatefulSet() error {
 	var (
 		podsRollingUpdateRequired bool
 	)
-	// the following code should not return before the end of this function in a non-error case
+	// NB: Be careful to consider the codepath that acts on podsRollingUpdateRequired before returning early.
 	sset, err := c.KubeClient.StatefulSets(c.Namespace).Get(c.statefulSetName(), metav1.GetOptions{})
 	if err != nil {
 		if !k8sutil.ResourceNotFound(err) {
@@ -343,7 +343,8 @@ func (c *Cluster) checkAndSetGlobalPostgreSQLConfiguration() error {
 				c.logger.Warningf("could not patch postgres parameters with a pod %s: %v", podName, err)
 			}
 		}
-		return fmt.Errorf("could not call Patroni API to set Postgres options: failed on %d pods", len(pods))
+		return fmt.Errorf("could not reach Patroni API to set Postgres options: failed on every pod (%d total)",
+			len(pods))
 	}
 	return nil
 }
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -326,19 +326,24 @@ func (c *Cluster) checkAndSetGlobalPostgreSQLConfiguration() error {
 	}
 
 	if len(optionsToSet) > 0 {
-		// TODO: eliminate unnecessary calls to listPods during sync
 		pods, err := c.listPods()
 		if err != nil {
 			return err
 		}
 		if len(pods) == 0 {
-			return fmt.Errorf("could not call patroni API: cluster has no pods")
+			return fmt.Errorf("could not call Patroni API: cluster has no pods")
 		}
-		// TODO: try to call all pods until success
-		c.logger.Debugf("calling Patroni API to set the following Postgres options: %v", optionsToSet)
-		if err := c.patroni.SetPostgresParameters(&pods[0], optionsToSet); err != nil {
-			return err
+		for _, pod := range pods {
+			podName := util.NameFromMeta(pod.ObjectMeta)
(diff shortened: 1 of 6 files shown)


pkg/cluster/sync.go
```diff
@@ -229,6 +229,7 @@ func (c *Cluster) syncStatefulSet() error {
 	var (
 		podsRollingUpdateRequired bool
 	)
+	// the following code should not return before the end of this function in a non-error case
```
zerg-junior: 1. "The following" == code until the end of the function?
2. Can the code leave the cluster in inconsistent state in the case of early non-error return ?

pkg/cluster/sync.go
```diff
@@ -303,6 +312,42 @@ func (c *Cluster) syncStatefulSet() error {
...
+			podName := util.NameFromMeta(pod.ObjectMeta)
+			c.logger.Debugf("calling Patroni API on a pod %s to set the following Postgres options: %v",
+				podName, optionsToSet)
+			if err := c.patroni.SetPostgresParameters(&pod, optionsToSet); err == nil {
+				return nil
+			} else {
+				c.logger.Warningf("could not patch postgres parameters with a pod %s: %v", podName, err)
+			}
+		}
+		return fmt.Errorf("could not call Patroni API to set Postgres options: failed on %d pods", len(pods))
```
zerg-junior: What exactly does this error message intend to say ? That if Patroni fails to apply a param, it fails on all pods at once ?

pkg/cluster/volumes.go
```diff
@@ -103,10 +103,10 @@ func (c *Cluster) resizeVolumes(newVolume spec.Volume, resizers []volumes.Volume
 
 	for _, pv := range pvs {
 		volumeSize := quantityToGigabyte(pv.Spec.Capacity[v1.ResourceStorage])
-		if volumeSize > newSize {
-			return fmt.Errorf("cannot shrink persistent volume")
-		}
-		if volumeSize == newSize {
+		if volumeSize >= newSize {
+			if volumeSize > newSize {
+				c.logger.Warningf("cannot shrink persistent volume")
```
zerg-junior: Do we log somewhere in the neighborhood of this statement which exactly volumes/sizes are we talking about ?

pkg/util/patroni/patroni.go
```diff
@@ -74,6 +69,28 @@ func (p *Patroni) Failover(master *v1.Pod, candidate string) error {
 
 		return fmt.Errorf("patroni returned '%s'", string(bodyBytes))
 	}
+	return nil
+}
+
+// Failover does manual failover via Patroni POST API call
```
zerg-junior: Manual as in "make Patroni do that " ?

pkg/util/patroni/patroni.go
```diff
@@ -14,13 +14,15 @@ import (
...
 	failoverPath = "/failover"
+	configPath   = "/config"
 	apiPort      = 8008
 	timeout      = 30 * time.Second
 )
 
 // Interface describe patroni methods
 type Interface interface {
 	Failover(master *v1.Pod, candidate string) error
+	SetPostgresParameters(server *v1.Pod, options map[string]string) error
```
zerg-junior: Is this method dedicated only to params that require Postgres restart ?

pkg/cluster/sync.go
alexeyklyukin: The rolling update is applied at the end of the function, and a code that will return beforehand will miss the rolling update.

pkg/cluster/sync.go
alexeyklyukin: That calling Patroni API failed, and we tried %d pods.

pkg/cluster/volumes.go
alexeyklyukin: Hm.. this change seems to be unrelated to the actual PR. Not sure how did it get there.

pkg/util/patroni/patroni.go
alexeyklyukin: yes, 'manual' might be not the best name here.

pkg/util/patroni/patroni.go
alexeyklyukin: Not necessary.

pkg/cluster/sync.go
zerg-junior: > we tried %d pods.

i first understood the original err msg as "update failed on all `len(pods)` pods"

alexeyklyukin:
@zerg-junior I did fix some warning messages and comments per your review. Please, take another look.

zerg-junior:
:+1:

alexeyklyukin:
👍

PR #300, opened 2018-05-24
Fix a crash on node migration.

alexeyklyukin:
After an unsuccessful initial cluster sync it may happen that the
cluster statefulset is empty. This has been made more likely since
88d6a7be3, since it has introduced syncing volumes before statefulsets,
and the volume sync may fail for different reasons (i.e. the volume has
been shrinked, or too many calls to Amazon).


alexeyklyukin:
👍

szuecs:
:+1:

Issue #292, opened 2018-05-17
Quickstart: How to access acid-minimal-cluster (Minikube) from host?


hjacobs:
I just tried out the quickstart (introduced with #289), but I don't know how to access the new "acid-minimal-cluster" PostgreSQL cluster:

```
$ kubectl create -f manifests/configmap.yaml  # configuration
configmap "postgres-operator" created
$ kubectl create -f manifests/operator-service-account-rbac.yaml  # identity and permissions
serviceaccount "zalando-postgres-operator" created
clusterrole "zalando-postgres-operator" created
clusterrolebinding "zalando-postgres-operator" created
$ kubectl create -f manifests/postgres-operator.yaml  # deployment
deployment "postgres-operator" created
$ kubectl get pod
NAME                                 READY     STATUS    RESTARTS   AGE
postgres-operator-6474c766b4-tgp82   1/1       Running   0          8s
$ kubectl create -f manifests/minimal-postgres-manifest.yaml 
postgresql "acid-minimal-cluster" created
$ kubectl get pod
NAME                                 READY     STATUS    RESTARTS   AGE
acid-minimal-cluster-0               0/1       Pending   0          2s
postgres-operator-6474c766b4-tgp82   1/1       Running   0          17s
$ kubectl get pod
NAME                                 READY     STATUS    RESTARTS   AGE
acid-minimal-cluster-0               1/1       Running   0          20s
acid-minimal-cluster-1               1/1       Running   0          6s
postgres-operator-6474c766b4-tgp82   1/1       Running   0          35s
$ kubectl get svc
NAME                        TYPE           CLUSTER-IP       EXTERNAL-IP   PORT(S)          AGE
acid-minimal-cluster        LoadBalancer   10.102.211.199   <pending>     5432:32114/TCP   25s
acid-minimal-cluster-repl   ClusterIP      10.106.83.140    <none>        5432/TCP         25s
kubernetes                  ClusterIP      10.96.0.1        <none>        443/TCP          1m
$ minikube service acid-minimal-cluster
Waiting, endpoint for service is not ready yet...
```

The service will never become ready as it's of type `LoadBalancer`.

Jan-M:
FYI: We are looking into this right now.

zerg-junior:
This behavior seems to be minikube-specific and relates to the recent upgrades that made `minikube` use Kubernetes 1.10+ 

[...]

PR #301, opened 2018-05-24
Fix conf for minikube

zerg-junior:
Partially addresses #292

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -12,6 +12,7 @@ data:
   debug_logging: "true"
   workers: "4"
   docker_image: registry.opensource.zalan.do/acid/spilo-10:1.4-p6
+  pod_service_account_name: "zalando-postgres-operator"
   secret_name_template: '{username}.{cluster}.credentials'
   # etcd_host: ""
   super_username: postgres
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -11,7 +11,7 @@ data:
 
   debug_logging: "true"
   workers: "4"
-  docker_image: registry.opensource.zalan.do/acid/demospilo-10:1.3-p3
+  docker_image: registry.opensource.zalan.do/acid/spilo-10:1.4-p6
   secret_name_template: '{username}.{cluster}.credentials'
   # etcd_host: ""
   super_username: postgres
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -14,7 +14,7 @@ data:
   docker_image: registry.opensource.zalan.do/acid/spilo-10:1.4-p6
   pod_service_account_name: "zalando-postgres-operator"
   secret_name_template: '{username}.{cluster}.credentials'
-  # etcd_host: ""
+  etcd_host: ""
   super_username: postgres
   enable_teams_api: "false"
   # enable_team_superuser: "false"
--- a/manifests/operator-service-account-rbac.yaml
+++ b/manifests/operator-service-account-rbac.yaml
@@ -37,6 +37,9 @@ rules:
   - create
   - delete
   - get
+  - list
+  - patch
+  - watch # for Spilo replica
 - apiGroups:
   - ""
   resources:
@@ -79,6 +82,7 @@ rules:
   - get
   - list
   - watch
+  - patch
 - apiGroups:
   - ""
   resources:
--- a/manifests/postgres-operator.yaml
+++ b/manifests/postgres-operator.yaml
@@ -12,7 +12,7 @@ spec:
       serviceAccountName: zalando-postgres-operator
       containers:
       - name: postgres-operator
-        image: registry.opensource.zalan.do/acid/postgres-operator:4c8dfd7
+        image: registry.opensource.zalan.do/acid/postgres-operator:e09e62f
         imagePullPolicy: IfNotPresent
         env:
         # provided additional ENV vars can overwrite individual config map entries  


manifests/configmap.yaml
```diff
@@ -11,9 +11,10 @@ data:
 
   debug_logging: "true"
   workers: "4"
-  docker_image: registry.opensource.zalan.do/acid/demospilo-10:1.3-p3
+  docker_image: registry.opensource.zalan.do/acid/spilo-10:1.4-p6
+  pod_service_account_name: "zalando-postgres-operator"
   secret_name_template: '{username}.{cluster}.credentials'
-  # etcd_host: ""
+  etcd_host: ""
```
zerg-junior: This is done only to overwrite [the default hostname](https://github.com/zalando-incubator/postgres-operator/blob/987b43456b080518c23a1242caa5c199ae7c1407/pkg/util/config/config.go#L72) that does not resolve on `minikube`.

What is our current policy on using/configuring etcd ?

mgomezch: Correct me if I'm wrong; as far as I know:

* etcd is still supported as a configuration alternative for Operator-managed Spilo clusters, and currently there are no plans to stop supporting it.

* etcd is definitely not the recommended configuration — the recommended configuration is using the Kubernetes API as the DCS backing Operator-managed Spilo clusters, which requires setting `etcd_host` to the empty string in the operator `ConfigMap`.

* I would argue `etcd_host` should default to the empty string.

manifests/operator-service-account-rbac.yaml
```diff
@@ -37,6 +37,9 @@ rules:
   - create
   - delete
   - get
+  - list
+  - patch
+  - watch # for Spilo replica
```
alexeyklyukin: I don't recall cases of watching for endpoints within the operator

manifests/configmap.yaml
Jan-M: Yes, ETCD remains supported by Patroni but is not our suggested way to run on Kubernetes. This would just be another service to be maintained that we can avoid for the "native" experience.

manifests/configmap.yaml
```diff
@@ -11,9 +11,10 @@ data:
 
   debug_logging: "true"
   workers: "4"
-  docker_image: registry.opensource.zalan.do/acid/demospilo-10:1.3-p3
+  docker_image: registry.opensource.zalan.do/acid/spilo-cdp-10:1.4-p8
+  pod_service_account_name: "zalando-postgres-operator"
   secret_name_template: '{username}.{cluster}.credentials'
-  # etcd_host: ""
+  etcd_host: ""
```
Jan-M: is there a reason we need this line? could just remove it completely from the example?

mgomezch: Currently it's required because the default for this `ConfigMap` field is nonempty](https://github.com/zalando-incubator/postgres-operator/blob/987b43456b080518c23a1242caa5c199ae7c1407/pkg/util/config/config.go#L72) and [that](https://github.com/zalando-incubator/postgres-operator/blob/c99cdd791584335203f5329e0c2f1b35fcfb938d/pkg/cluster/util.go#L464-L466) means [Spilo is set up to use etcd as DCS](https://github.com/zalando-incubator/postgres-operator/blob/e09e62f259a7bf5ba10099cbf2fb52a5ab2c5d57/pkg/cluster/k8sres.go#L402-L406) instead of the Kubernetes API.

manifests/operator-service-account-rbac.yaml
zerg-junior: I doubled-checked, omitting this line raises the following exception in a _replica_:

> Traceback (most recent call last):                                                                  [5/1916]
  File "/usr/local/lib/python3.6/dist-packages/patroni/dcs/kubernetes.py", line 396, in watch
    _request_timeout=(1, timeout + 1)):
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/watch/watch.py", line 115, in stream
    resp = func(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/patroni/dcs/kubernetes.py", line 50, in wrapper
    return getattr(self._api, func)(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/apis/core_v1_api.py", line 12528, in list_n
amespaced_endpoints
    (data) = self.list_namespaced_endpoints_with_http_info(namespace, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/apis/core_v1_api.py", line 12630, in list_n
amespaced_endpoints_with_http_info
    collection_formats=collection_formats)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/api_client.py", line 335, in call_api
    _preload_content, _request_timeout)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/api_client.py", line 148, in __call_api
    _request_timeout=_request_timeout)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/api_client.py", line 371, in request
    headers=headers)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/rest.py", line 250, in GET
    query_params=query_params)
  File "/usr/local/lib/python3.6/dist-packages/kubernetes/client/rest.py", line 240, in request
    raise ApiException(http_resp=r)
kubernetes.client.rest.ApiException: (403)
Reason: Forbidden
HTTP response headers: HTTPHeaderDict({'Content-Type': 'application/json', 'X-Content-Type-Options': 'nosnif
f', 'Date': 'Thu, 24 May 2018 15:20:53 GMT', 'Content-Length': '285'})
HTTP response body: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"endpoin
ts is forbidden: User \\"system:serviceaccount:default:zalando-postgres-operator\\" cannot watch endpoints i
n the namespace \\"default\\"","reason":"Forbidden","details":{"kind":"endpoints"},"code":403}\n'

hjacobs:
:+1:

manifests/configmap.yaml
```diff
@@ -28,7 +28,7 @@ data:
   db_hosted_zone: db.example.com
   master_dns_name_format: '{cluster}.{team}.staging.{hostedzone}'
   replica_dns_name_format: '{cluster}-repl.{team}.staging.{hostedzone}'
-  enable_master_load_balancer: "false"
+  enable_master_load_balancer: "true"
```
alexeyklyukin: Why does it make sense to set its default to true when aiming for minikube users?

zerg-junior: we need the LB service of such type here for it to be [accessible outside of minikube](https://github.com/kubernetes/minikube/issues/384#issuecomment-234409957).

alexeyklyukin:
👍

zerg-junior:
:+1:

Issue #264, opened 2018-03-07
add monitoring


theRealWardo:
what does zalando do for postgres monitoring with any databases run via this operator?

I was thinking of building https://github.com/wrouesnel/postgres_exporter into the database container and having that be monitored via our prometheus operator.

is there any existing plans to add monitoring directly into this project in some way? if not, is there a need for a more detailed discussion/approach prior to contribution or shall I do as the contribution guidelines say and just hack away and send a PR?

Jan-M:
Quick answer is no, there is no intent to make the operator "monitor" anything. Ideally the operator focuses on "operation" and more specifically on the provisioning and modifying part. The "ops" part we largely leave to Patroni which is very well suited for taking care of the cluster itself.

The operator however does contain a very slim API to allow monitoring it from the outside.

At Zalando we use ZMON (zmon.io) for all monitoring. But there is other options here like Prometheus.

We are running Postgres with the bg_mon extension exposing a lot of Postgres data via a rest API on port 8080 so this helps a lot I think.

theRealWardo:
thanks for the quick reply! to be clear I'm not proposing monitoring the operator itself but rather the database it is operating on. if there is something in the operator that you monitor and feel others should monitor please do let me know! otherwise our system will probably just be monitoring that the pod is up and running.

what I'd like to add to this operator to facilitate that is flag would would add a simple named monitoring port on the [`ServiceSpec`](https://github.com/zalando-incubator/postgres-operator/blob/c2ca2307bc0fca9aa8f6d050689c4bc18ad550ba/pkg/cluster/k8sres.go#L682). that would enable me to have a `ServiceMonitor` (custom resource) which my Prometheus operator would then be able to turn into scrape targets for my Prometheus instance. does that sound reasonable?

Jan-M:
I forgot one tool here, just teasing it as we have not release it yet, but teams rely on our new pgview web interface to monitor their DBs too and it has proven very useful.

![image](https://user-images.githubusercontent.com/1443395/37088392-7be43462-21fd-11e8-8ae2-e7fa2d3b2853.png)

theRealWardo:
for that kind of web dashboard thing we've been running https://github.com/ankane/pghero which has definitely helped us a couple times but it doesn't hook into our alerting systems which is what I'm really trying to achieve here.

Jan-M:
[...]

PR #308, opened 2018-05-31
arbitrary sidecar support

theRealWardo:
see the discussion in #264

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -611,7 +611,7 @@ func (c *Cluster) getSidecarContainer(sidecar spec.Sidecar, index int, volumeMou
 		Resources:       *resources,
 		VolumeMounts:    volumeMounts,
 		Env:             env,
-		Posts:           sidecar.Ports,
+		Ports:           sidecar.Ports,
 	}, nil
 }
 
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -318,8 +318,9 @@ func (c *Cluster) generatePodTemplate(
 	patroniParameters *spec.Patroni,
 	cloneDescription *spec.CloneDescription,
 	dockerImage *string,
-	customPodEnvVars map[string]string,
-) *v1.PodTemplateSpec {
+	sidecars *[]spec.Sidecar,
+	customPodEnvVars map[string]string
+) (*v1.PodTemplateSpec, error) {
 	spiloConfiguration := c.generateSpiloJSONConfiguration(pgParameters, patroniParameters)
 
 	envVars := []v1.EnvVar{
@@ -525,6 +526,16 @@ func (c *Cluster) generatePodTemplate(
 		)
 	}
 
+	if sidecars != nil && len(*sidecars) > 0 {
+		for index, sidecar := range *sidecars {
+			sc, err := c.getSidecarContainer(sidecar, index, volumeMounts)
+			if err != nil {
+				return nil, err
+			}
+			podSpec.Containers = append(podSpec.Containers, *sc)
+		}
+	}
+
 	template := v1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:    c.labelsSet(true),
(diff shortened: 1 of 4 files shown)


pkg/cluster/k8sres.go
```diff
@@ -525,6 +527,52 @@ func (c *Cluster) generatePodTemplate(
...
+							FieldRef: &v1.ObjectFieldSelector{
+								APIVersion: "v1",
+								FieldPath:  "metadata.namespace",
+							},
+						},
+					},
+					// TODO: figure out auth.
+					/*
+						{
+							Name:  "POSTGRES_USER",
```
alexeyklyukin: You can expose them from the secrets, i.e.
```
ValueFrom: &v1.EnvVarSource{
				SecretKeyRef: &v1.SecretKeySelector{
					LocalObjectReference: v1.LocalObjectReference{
						Name: c.credentialSecretName(c.OpConfig.SuperUsername),
					},
					Key: "password",
				},
			},
```

pkg/cluster/k8sres.go
```diff
@@ -525,6 +527,52 @@ func (c *Cluster) generatePodTemplate(
...
+	monitoringContainerImage := c.OpConfig.MonitoringDockerImage
+	if monitoringDockerImage != nil && *monitoringDockerImage != "" {
+		monitoringContainerImage = *monitoringDockerImage
+	}
+	if monitoringContainerImage != "" {
+		podSpec.Containers = append(
+			podSpec.Containers,
+			v1.Container{
+				Name:            "monitoring-sidecar",
+				Image:           c.monitoringImage(),
```
alexeyklyukin: Looks like it is not defined anywhere.

pkg/cluster/k8sres.go
theRealWardo: done!

pkg/cluster/k8sres.go
theRealWardo: yea this was a sketch and I was trying to figure out how to get my environment set up... all fixed up and working now.

theRealWardo:
okay so I tried to do something like `Sidecars []v1.Container` rather than a custom `Sidecar` type, but there were some problems specifically with the handling of resources there.

that of course means that there's a bit more of a painful path to supporting things like adding custom volumes/mounts here, but I think this is probably a decent start. just let me know what you think!

alexeyklyukin:
@theRealWardo thanks for updating the PR. I see little use in modifying individual cluster manifests. What if you run 100 clusters managed by the operator, would you ask each one of them to change its definition to incorporate the same container image (be it monitoring or log support)? Instead, we can pass this information via the operator configuration to configure all clusters managed by the operator at once and uniformly. It could be a parameter of the operator itself, i.e:
```yaml
sidecar_containers: "name:image, name:image, name:image". 
```
We could pass the necessary minimum of environment variables like credentials to those containers via the `c.generatePodTemplate` and alike and use `pod_environment_configmap ` to propagate the rest.

theRealWardo:
@alexeyklyukin while I do agree there is a use case for cluster wide sidecars, I would propose supporting both. I think that there is a use case for configuring an individual cluster's sidecars differently from other clusters' sidecars.

I also believe this project has poor support for structured data in the operator config. how am I supposed to specify ports for example? the structured string in the operator config is extremely hard to read and therefore error prone. it further requires maintaining separate object definition from the standard Kubernetes objects, making the learning curve higher, and thus reducing potential traction of this operator. for these reasons, I avoided the operator config for now and have proposed a change to the cluster definition.

theRealWardo:
@alexeyklyukin friendly ping. would love to get this or something like it in. we've been running this code in our staging environment fine for the last week or so.  😄

alexeyklyukin:
> @alexeyklyukin while I do agree there is a use case for cluster wide sidecars, I would propose supporting both. I think that there is a use case for configuring an individual cluster's sidecars differently from other clusters' sidecars.

I see none. What is your use case for that?

> I also believe this project has poor support for structured data in the operator config. how am I supposed to specify ports for example? the structured string in the operator config is extremely hard to read and therefore error prone. it further requires maintaining separate object definition from the standard Kubernetes objects, making the learning curve higher, and thus reducing potential traction of this operator. for these reasons, I avoided the operator config for now and have proposed a change to the cluster definition.


Fair enough, I have some plans to address that.

alexeyklyukin:
@theRealWardo I have created https://github.com/zalando-incubator/postgres-operator/pull/331 based on your PR that adds support for global containers and does some refactoring. I have considered doing it against your own branch, but since the latest master is not merged in there the PR shows too many unrelated changes.

I'd appreciate if you take a look.

PR #309, opened 2018-05-31
Split already existing documentation into parts

erthalion:
To improve the documentation we need to split it into smaller parts:

* quickstart (in the readme)
* general concepts
* tutorials
* how to
* references

And then add the missing information. So far I just split the existing
documentation and left references almost empty. I assume that references
may duplicate the rest of the documentation in a way that the doc will
have references to this section, that contains all the formal details.


Jan-M:
Can we find a better distinction than "how to" and "tutorial"? We should make it somehow clear one is aiming for the developer/user of Postgres manifests while the other is aiming towards the "admin" deploying and mainting the operator/PostgreSQL infrastructure.

* User Guide?
* Configuration Guide?
* Administrator Guide vs User Guide?

alexeyklyukin:
I'd rename howtos to features. Also, it seems there are no links from the README to other, more specific parts of the documentation.

erthalion:
> I'd rename howtos to features.

Hm...isn't reference the more proper place for the comprehensive list of features?

hjacobs:
For reference: [What nobody tells you about documentation](https://www.divio.com/en/blog/documentation/) --- that's how we usually structure documentation now

erthalion:
> For reference: What nobody tells you about documentation --- that's how we usually structure documentation now.

thanks @hjacobs - I remember it, but couldn't find in my mail box.

docs/user.md
```diff
@@ -0,0 +1,183 @@
+# How to create a new db cluster
```
Jan-M: Can we write "Postgres" cluster? Especially with the wording database in Postgres being a logical unit within a Cluster this can be confusing.

alexeyklyukin:
👍

erthalion:
:+1:

PR #312, opened 2018-06-01
Add an example of clone feature

erthalion:
Since this feature is quite important it makes sense to mention it in the user part.

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -184,8 +184,31 @@ spec:
 
 ## How to clone an existing PostgreSQL cluster
 
-To spin up a new cluster as a clone of the existing one, you need to provide a
-clone section in the spec:
+You can spin up a new cluster as a clone of the existing one, using a clone
+section in the spec. There are two options here:
+
+* Clone directly from a source cluster using `pg_basebackup`
+
+* Clone from an S3 bucket
+
+### Clone directly
+
+```yaml
+apiVersion: "acid.zalan.do/v1"
+kind: postgresql
+
+metadata:
+  name: acid-test-cluster
+spec:
+  clone:
+    cluster: "acid-batman"
+```
+
+Here `cluster` is a name of a source cluster that is going to be cloned. The
+cluster to clone is assumed to be running and the clone procedure invokes
+`pg_basebackup` from it.
+
+### Clone from S3
 
 ```yaml
 apiVersion: "acid.zalan.do/v1"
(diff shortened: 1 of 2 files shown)


docs/user.md
```diff
@@ -181,3 +181,40 @@ spec:
...
+metadata:
+  name: acid-test-cluster
+spec:
+  clone:
+    uid: "efd12e58-5786-11e8-b5a7-06148230260c"
+    cluster: "acid-batman"
+    timestamp: "2017-12-19T12:40:33+01:00"
+```
+
+Here `cluster` is a name of a target cluster that is going to be cloned. If
```
alexeyklyukin: `cluster` is the name of the cluster to be cloned.

docs/user.md
```diff
@@ -181,3 +181,40 @@ spec:
...
+`timestamp` is not empty, then a new cluster will be cloned from an S3 bucket,
```
alexeyklyukin: the operator does not create S3 buckets. Perhaps "a new cluster will be cloned from S3 ... "

docs/user.md
```diff
@@ -181,3 +181,40 @@ spec:
...
+that was created by operator,
+using the latest backup before the `timestamp`. In this case `uid` field is
```
alexeyklyukin: In this case,

docs/user.md
```diff
@@ -181,3 +181,40 @@ spec:
...
+also mandatory - operator will use it to find an S3 bucket. You can find this
```
alexeyklyukin: the correct folder inside the S3 bucket.

docs/user.md
```diff
@@ -181,3 +181,40 @@ spec:
...
+field from metadata of a target cluster:
```
alexeyklyukin: target is the one to be cloned to. I'd say it is the source (or origin) cluster.

docs/user.md
```diff
@@ -181,3 +181,40 @@ spec:
...
+
+```yaml
+apiVersion: acid.zalan.do/v1
+kind: postgresql
+metadata:
+  name: acid-test-cluster
+  uid: efd12e58-5786-11e8-b5a7-06148230260c
+```
+
+If `timestamp` is empty or absent, a new cluster will be cloned from an
```
alexeyklyukin: the cluster to clone is assumed  to be running and the clone procedure invokes pg_basebackup from it. Note that the timestamp parameter mandates the timezone.

docs/user.md
zerg-junior: correct key maybe ? afaik there are no folders within S3 buckets, keys just mimic directory structure using names with slashes.

alexeyklyukin: although  directory (or folder) is universally understood in the context of S3 (I think WAL-E mentions it at least once on their page), using key would things less ambiguous, so +1

alexeyklyukin:
👍

erthalion:
:+1:

PR #314, opened 2018-06-04
Fix exec into pods to resize volumes for multi-container pods.

alexeyklyukin:
The original code assumed only one container per pod.


mgomezch:
:+1:

alexeyklyukin:
👍

PR #315, opened 2018-06-05
Add section about volume increase

erthalion:
Another important feature to document

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -242,3 +242,34 @@ metadata:
 
 Note that timezone required for `timestamp` (offset relative to UTC, see RFC
 3339 section 5.6)
+
+## Increase volume size
+
+PostgreSQL operator supports statefulset volume resize if you're using the
+operator on top of AWS. For that you need to apply manifest with a new size,
+and the operator will find out that a volume size needs to be changed:
+
+```
+apiVersion: "acid.zalan.do/v1"
+kind: postgresql
+
+metadata:
+  name: acid-test-cluster
+spec:
+  volume:
+    size: 5Gi # new volume size
+```
+
+You can only increase a volume size in this way, if a new requested size is
+smaller than the previous one nothing will be done. After this update all the
+new pods in a statefulset will be created with a new volume size. To increase a
+volume size on already existing pods in a statefulset, the operator will
+perform the following steps:
+
+* modify EBS volume size
+
+* resize an actuall filesystem use `resize2fs`
+
+Note that if before a volume size was increased a statefulset was scaled down
+and (after the change was applied) scaled back, those pods that were down will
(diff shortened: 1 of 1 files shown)


docs/user.md
```diff
@@ -242,3 +242,34 @@ metadata:
 
 Note that timezone required for `timestamp` (offset relative to UTC, see RFC
 3339 section 5.6)
+
+## Increase volume size
+
+PostgreSQL operator supports statefulset volume resize if you're using the
+operator on top of AWS. For that you need to apply manifest with a new size,
```
alexeyklyukin: I'd change it something like (not clear about which size you are talking about, could be more than one volume that needs to be changed):

'you need to change the size field of the volume description in the cluster manifest and apply it; the operator compares the new volume size with the previous on and acts on differences.'

docs/user.md
```diff
@@ -242,3 +242,34 @@ metadata:
...
+kind: postgresql
+
+metadata:
+  name: acid-test-cluster
+spec:
+  volume:
+    size: 5Gi # new volume size
+```
+
+You can only increase a volume size in this way, if a new requested size is
```
alexeyklyukin: You can only enlarge the volume with the process described above, shrinking is not supported and will emit a warning.

docs/user.md
```diff
@@ -242,3 +242,34 @@ metadata:
...
+smaller than the previous one nothing will be done. After this update all the
+new pods in a statefulset will be created with a new volume size. To increase a
+volume size on already existing pods in a statefulset, the operator will
```
alexeyklyukin: to enlarge persistent volumes attached to the running pods, the operator performs the following actions:

docs/user.md
```diff
@@ -242,3 +242,34 @@ metadata:
...
+perform the following steps:
+
+* modify EBS volume size
```
alexeyklyukin: call AWS API to change the volume size (AWS rate-limits this operation to no more than once every 6 hours)

docs/user.md
```diff
@@ -242,3 +242,34 @@ metadata:
...
+
+* resize an actuall filesystem use `resize2fs`
```
alexeyklyukin: resize the filesystem with `resize2fs` by doing the equivalent of `kubectl exec` on the pod.

docs/user.md
```diff
@@ -242,3 +242,34 @@ metadata:
...
+
+Note that if before a volume size was increased a statefulset was scaled down
```
alexeyklyukin: something like '
if the statefulset is scaled down before resizing the size changes are only applied to the volumes attached to the running pods; the size of the volumes that correspond to the previously running pods is not changed.'

alexeyklyukin:
👍

erthalion:
:+1:

PR #319, opened 2018-06-06
[WIP] Prototype automatic submission of role bindings

### LINKED ISSUE #310: Clarify RBAC configuration
The current RBAC setup needs the following improvements:

1) Documentation on the general setup of RBAC rules. In particular, we assume the [`zalando-postgres-operator`](https://github.com/zalando-incubator/postgres-operator/blob/master/manifests/operator-service-account-rbac.yaml#L8) **cluster** role exists. We also need to clarify that [the example rbac setup](https://github.com/zalando-incubator/postgres-operator/blob/master/manifests/operator-service-account-rbac.yaml) re-uses the (powerful) operator's service account for the pods, and that pods can be configured to use a separate account. 

2) Automatic submission  of a `RoleBinding` from  the `zalando-postgres-operator` role to a pod service account in non-default namespaces.

3) Removal of all mentions of `operator` service account in favor of `zalando-postgres-operator`, both in the docs and the source code.

zerg-junior:
closes #310

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 	"k8s.io/client-go/pkg/apis/apps/v1beta1"
 	policybeta1 "k8s.io/client-go/pkg/apis/policy/v1beta1"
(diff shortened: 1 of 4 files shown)


zerg-junior:
retired in favor of  #344

PR #326, opened 2018-06-14
Allow configuring the operator via the YAML manifest.

alexeyklyukin:
Up until now, the operator read its own configuration from the
configmap.  That has a number of limitations, i.e. when the
configuration value is not a scalar, but a map or a list. We use a
custom code based on github.com/kelseyhightower/envconfig to decode
non-scalar values out of plain text keys, but that breaks when the data
inside the keys contains both YAML-special elememtns (i.e. commas) and
complex quotes, one good example for that is search_path inside
`team_api_role_configuration`. In addition, reliance on the configmap
forced a flag structure on the configuration, making it hard to write
and to read (see
https://github.com/zalando-incubator/postgres-operator/pull/308#issuecomment-395131778).

This commit allows to supply the operator configuration in a proper YAML
file. That required registering a custom CRD to support the operator
configuration and provide an example at
manifests/postgresql-operator-default-configuration.yaml. At the moment,
both old configmap and the new CRD configuration is supported, so no
compatibility issues, however, in the future I'd like to deprecate the
configmap-based configuration altogether. Contrary to the
configmap-based configuration, the CRD one doesn't embed defaults into
the operator code, however, one can use the
manifests/postgresql-operator-default-configuration.yaml as a starting
point in order to build a custom configuration.

Since previously `ReadyWaitInterval` and `ReadyWaitTimeout` parameters
used to create the CRD were taken from the operator configuration, which
is not possible if the configuration itself is stored in the CRD object,
I've added the ability to specify them as environment variables
`CRD_READY_WAIT_INTERVAL` and `CRD_READY_WAIT_TIMEOUT` respectively.

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -294,7 +294,7 @@ either. In the CRD-based configuration those options are grouped under the
   [kube2iam](https://github.com/jtblin/kube2iam) project on AWS. The default is empty.
 
 * **aws_region**
-  AWS region used to store ESB volumes.
+  AWS region used to store ESB volumes. The default is `eu-central-1`.
 
 ## Debugging the operator
 
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -1,9 +1,54 @@
+There are two mutually-exclusive methods to set the Postgres Operator
+configuration.
 
-Postgres operator is configured via a ConfigMap defined by the
-`CONFIG_MAP_NAME` environment variable. Variable names are underscore-separated
-words.
+* ConfigMaps-based, the legacy one.  The configuration is supplied in a
+  key-value configmap, defined by the `CONFIG_MAP_NAME` environment variable.
+  Non-scalar values, i.e. lists or maps, are encoded in the value strings using
+  the comma-based syntax for lists and coma-separated `key:value` syntax for
+  maps. String values containing ':' should be enclosed in quotes. The
+  configuration is flat, parameter group names below are not reflected in the
+  configuration structure. There is an
+  [example](https://github.com/zalando-incubator/postgres-operator/blob/master/manifests/configmap.yaml)
+
+* CRD-based configuration.  The configuration is stored in the custom YAML
+  manifest, an instance of the custom resource definition (CRD) called
+  `postgresql-operator-configuration`.  This CRD is registered by the operator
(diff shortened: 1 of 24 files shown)


docs/reference/operator_parameters.md
```diff
@@ -1,9 +1,54 @@
...
+  maps. String values containing ':' should be enclosed in quotes. The
+  configuration is flat, parameter group names below are not reflected in the
+  configuration structure. There is an
+  [example](https://github.com/zalando-incubator/postgres-operator/blob/master/manifests/configmap.yaml)
+
+* CRD-based configuration.  The configuration is stored in the custom YAML
+  manifest, an instance of the custom resource definition (CRD) called
+  `postgresql-operator-configuration`.  This CRD is registered by the operator
+  during the start when `POSTGRES_OPERATOR_CONFIGURATION_OBJECT` variable is
+  set to a non-empty value. The CRD-based configuration is a regular YAML
```
zerg-junior: is this env var set in the operator deployment manifest?

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -0,0 +1,78 @@
...
+    replication_username: standby
+  kubernetes:
+    pod_service_account_name: operator
+    pod_terminate_grace_period: 5m
+    pdb_name_format: "postgres-{cluster}-pdb"
+    secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"
+    oauth_token_secret_name: postgresql-operator
+    pod_role_label: spilo-role
+    cluster_labels:
+        application: splio
```
zerg-junior: spILo :)

docs/reference/operator_parameters.md
alexeyklyukin: It could be, but right now it is not.

alexeyklyukin:
👍

Jan-M:
Only thing I am missing is the CRD def YAML for manual deployment. I would assume our example then changes to deploy config CRD, followed by default object and then the operator deployment over the earler config map?

alexeyklyukin:
This is the eventual plan, although this doesn't need to be part of this changeset. I think nothing stops us from adding the  CRD as a separate PR, to avoid blocking other changes on this one.

Jan-M:
:+1:

alexeyklyukin:
👍

PR #331, opened 2018-06-26
Support for per-cluster and operator global sidecars

alexeyklyukin:
- Based on the work of @theRealWardo to to implement per-cluster sidecards

- Add support for operator-global sidecards (currently crude, as we only allow the list of docker images, and the optional variables (besides mandatory `POD_NAME`, `POD_NAMESPACE`, `POSTGRES_USER` and `POSTGRES_PASSWORD` ) can only be propagated via either `pod_environment_configmap` or Pod Presets. Should be improved once https://github.com/zalando-incubator/postgres-operator/pull/326 is merged).

- Do refactoring around the code that generates the statefulset, particularly, decouple most of the helper routines from the cluster object in order to simplify their testing in the future, as well as to convert the scalar configuration internally to a sidecar structure to use the common code path.


pkg/util/config/config.go
```diff
@@ -68,9 +68,10 @@ type Config struct {
 	Auth
 	Scalyr
 
-	WatchedNamespace string `name:"watched_namespace"`    // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
-	EtcdHost         string `name:"etcd_host" default:""` // special values: the empty string "" means Patroni will use k8s as a DCS
-	DockerImage      string `name:"docker_image" default:"registry.opensource.zalan.do/acid/spilo-cdp-10:1.4-p8"`
+	WatchedNamespace string            `name:"watched_namespace"`    // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
+	EtcdHost         string            `name:"etcd_host" default:""` // special values: the empty string "" means Patroni will use k8s as a DCS
+	DockerImage      string            `name:"docker_image" default:"registry.opensource.zalan.do/acid/spilo-cdp-10:1.4-p8"`
+	Sidecars         map[string]string `name:"sidecars"`
```
Jan-M: If I see this right, this is a map of sidecar container name to docker images? Maybe we should make it more clear from the variable name? That is not sidecar name to yaml?

Jan-M:
How would this solution help us with our Scalyr Sidecar? I dont see how to inject the key on global level?

alexeyklyukin:
Well, outside of making the code less horrible by converting the scalyr hard-code to the same sidecar structure used to inject other sidecars so far it doesn't. However, I don't think it should be our goal to cram those things into the operator; what I propose instead is to rely on the Kubernetes feature called `PodPresets` to inject environment variables and remove all duplication of it (namely, https://postgres-operator.readthedocs.io/en/latest/administrator/#custom-pod-environment-variables) from the operator altogether.

pkg/util/config/config.go
alexeyklyukin: While this is [documented](https://postgres-operator.readthedocs.io/en/refactoring-sidecars/reference/operator_parameters/#general) I wouldn't mind a better name. Perhaps `sidecar_docker_images`?

alexeyklyukin:
👍

pkg/cluster/k8sres.go
```diff
@@ -610,6 +735,81 @@ func (c *Cluster) generateStatefulSet(spec *spec.PostgresSpec) (*v1beta1.Statefu
...
 
+func getEffectiveDockerImage(globalDockerImage, clusterDockerImage string) string {
+	if clusterDockerImage == "" {
+		return globalDockerImage
+	}
+	return clusterDockerImage
+}
+
+func generateScalarSidecarSpec(clusterName, APIKey, serverURL, dockerImage string, containerResources *spec.Resources) (sidecar *spec.Sidecar, present bool) {
+	if APIKey == "" || serverURL == "" || dockerImage == "" {
```
zerg-junior: does this mean that if this particular sidecar is misconfigured, we silently ignore this fact w/o logging it?

pkg/util/config/config.go
```diff
@@ -68,9 +68,10 @@ type Config struct {
 	Auth
 	Scalyr
 
-	WatchedNamespace string `name:"watched_namespace"`    // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
-	EtcdHost         string `name:"etcd_host" default:""` // special values: the empty string "" means Patroni will use k8s as a DCS
-	DockerImage      string `name:"docker_image" default:"registry.opensource.zalan.do/acid/spilo-cdp-10:1.4-p8"`
+	WatchedNamespace string            `name:"watched_namespace"`    // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
+	EtcdHost         string            `name:"etcd_host" default:""` // special values: the empty string "" means Patroni will use k8s as a DCS
+	DockerImage      string            `name:"docker_image" default:"registry.opensource.zalan.do/acid/spilo-cdp-10:1.4-p8"`
+	Sidecars         map[string]string `name:"sidecar_docker_images"`
```
zerg-junior: what about adding a short example to the operator manifest ?

docs/reference/operator_parameters.md
```diff
@@ -15,6 +15,11 @@ words.
   your own Spilo image from the [github
   repository](https://github.com/zalando/spilo).
 
+* **sidecar_docker_images**
+  a map of sidecar names to docker images for the containers to run alongside
+  Spilo. In case of the name conflict with the definition in the cluster
+  manifest the cluster-specific one is preferred.
```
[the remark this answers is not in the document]
alexeyklyukin: yes

pkg/util/config/config.go
alexeyklyukin: not sure it actually needs sidecars.

pkg/cluster/k8sres.go
```diff
@@ -409,134 +519,89 @@ func (c *Cluster) generatePodTemplate(
...
-		if _, ok := envVarsMap[name]; !ok {
-			names = append(names, name)
-		} else {
-			c.logger.Warningf("variable %q value from %q is ignored: conflict with the definition from the operator",
-				name, c.OpConfig.PodEnvironmentConfigMap)
+
+	return envVars
+}
+
+// deduplicateEnvVars makes sure there are no duplicate in the target envVar array. While Kubernetes already does, it
```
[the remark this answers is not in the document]
alexeyklyukin: yes, looks like I've lost a word or two :-(

pkg/cluster/k8sres.go
```diff
@@ -409,134 +519,89 @@ func (c *Cluster) generatePodTemplate(
...
+func deduplicateEnvVars(input []v1.EnvVar, containerName string, logger *logrus.Entry) []v1.EnvVar {
+	result := make([]v1.EnvVar, 0)
+	names := make(map[string]int)
+
+	for i, va := range input {
+		if names[va.Name] == 0 {
+			names[va.Name] += 1
+			result = append(result, input[i])
+		} else if names[va.Name] == 1 {
+			names[va.Name] += 1
```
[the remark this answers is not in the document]
alexeyklyukin: To avoid producing this log message on each occurrences of the duplicate parameter (as opposed to only once when the duplicate is detected.)

pkg/cluster/k8sres.go
```diff
@@ -409,134 +519,89 @@ func (c *Cluster) generatePodTemplate(
...
+			logger.Warningf("variable %q is defined in %q more than once, the subsequent definitions are ignored",
```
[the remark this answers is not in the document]
alexeyklyukin: I would say the purpose of this warning is to prevent such cases, not to assist in constructing them to get what you want.  I don't want to overload it with an extra debugging info.

zerg-junior:
:+1:

alexeyklyukin:
👍

zerg-junior:
:+1:

theRealWardo:
awesome! thanks for carrying this through!

PR #332, opened 2018-06-27
Update travis configuration.

alexeyklyukin:
- explicitely set sudo to false, since we don't need it and it
  slows-down builds.
- use the newest go toolchain.


alexeyklyukin:
👍

PR #338, opened 2018-07-06
Use the StorageClassName field of the volumeClaimTemplate.

alexeyklyukin:
The old way of specifying it with the annotation is deprecated and not
available in recent Kubernetes versions. We will keep it there anyway
until upgrading to the new go-client that is incompatible with those
versions.

Per https://github.com/zalando-incubator/postgres-operator/issues/337


alexeyklyukin:
👍

Jan-M:
:+1:

PR #343, opened 2018-07-13
Switchover must wait for the inner goroutine before it returns.

alexeyklyukin:
Otherwise, two corner cases may happen:

 - waitForPodLabel writes to the podLabelErr channel that has been
   already closed by the outer routine

 - the outer routine exists and the caller subscribes to the pod
   the inner goroutine has already subscribed to, resulting in panic.

 The previous commit https://github.com/zalando-incubator/postgres-operator/commit/fe47f9ebeadd54639913296735158b42d17ee012
 that touched that code added the cancellation channel, but didn't bother
 to actually wait for the goroutine to be cancelled.

 Per report from @valer-cara.
 Original issue: https://github.com/zalando-incubator/postgres-operator/issues/342

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -879,17 +879,17 @@ func (c *Cluster) Switchover(curMaster *v1.Pod, candidate spec.NamespacedName) (
 	wg.Add(1)
 
 	go func() {
-		ch := c.registerPodSubscriber(candidate)
 		defer wg.Done()
+		ch := c.registerPodSubscriber(candidate)
 		defer c.unregisterPodSubscriber(candidate)
 
 		role := Master
 
 		select {
 		case <-stopCh:
-		case podLabelErr <- func() error {
-			_, err2 := c.waitForPodLabel(ch, stopCh, &role)
-			return err2
+		case podLabelErr <- func() (err error) {
+			_, err = c.waitForPodLabel(ch, stopCh, &role)
+			return
 		}():
 		}
 	}()
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -867,15 +867,20 @@ func (c *Cluster) GetStatus() *spec.ClusterStatus {
 }
 
 // Switchover does a switchover (via Patroni) to a candidate pod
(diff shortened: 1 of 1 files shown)


pkg/cluster/cluster.go
```diff
@@ -867,41 +867,52 @@ func (c *Cluster) GetStatus() *spec.ClusterStatus {
...
+	// wait until the goroutine terminates, since unregisterPodSubscriber
+	// must be called before the outer return; otherwsise we risk subscribing to the same pod twice.
+	wg.Wait()
+	// close the label waiting channel no sooner than the waiting goroutine terminates.
+	close(podLabelErr)
 
-	if err := <-podLabelErr; err != nil {
-		return fmt.Errorf("could not get master pod label: %v", err)
-	}
+	return
```
valer-cara: @alexeyklyukin - _minor_ observation, I've noticed most code uses explicit returns. `err2` drew my attention, that's why I've mentioned this. Using an explicit return would get rid of `err2` and it'd all just be `err` all over the place

pkg/cluster/cluster.go
```diff
@@ -867,41 +867,52 @@ func (c *Cluster) GetStatus() *spec.ClusterStatus {
...
+
 	podLabelErr := make(chan error)
 	stopCh := make(chan struct{})
-	defer close(podLabelErr)
+
+	wg.Add(1)
 
 	go func() {
 		ch := c.registerPodSubscriber(candidate)
+		defer wg.Done()
```
valer-cara: Should this be below the next defer just to make sure that it unblocks _after_ it unregistered pod subscriber? Not sure if ordering is an issue here, just thought there might be a possibility

pkg/cluster/cluster.go
alexeyklyukin: Good point, thanks!

pkg/cluster/cluster.go
alexeyklyukin: Moving wg.Done() to the top of this function makes sense, although the defer order is correct, since they are pushed to the stack and then unrolled in the last-in-first-out order.

Jan-M:
:+1:

pkg/cluster/cluster.go
valer-cara: that's right, my mistake :D

PR #344, opened 2018-07-16
[WIP] Submit service account during initial Event processing

zerg-junior:
Move submitting service account to a controller; this should not block a controller itself since this code runs in a separate go-routine.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/pkg/api/v1"
+	rbacv1beta1 "k8s.io/client-go/pkg/apis/rbac/v1beta1"
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/zalando-incubator/postgres-operator/pkg/apiserver"
@@ -52,7 +53,8 @@ type Controller struct {
 
 	workerLogs map[uint32]ringlog.RingLogger
 
-	PodServiceAccount *v1.ServiceAccount
+	PodServiceAccount            *v1.ServiceAccount
+	PodServiceAccountRoleBinding *rbacv1beta1.RoleBinding
 }
 
 // NewController creates a new controller
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -163,10 +163,58 @@ func (c *Controller) initPodServiceAccount() {
 	// actual service accounts are deployed at the time of Postgres/Spilo cluster creation
 }
 
+func (c *Controller) initRoleBinding() {
+
+	// service account on its own lacks any rights starting with k8s v1.8
+	// operator binds it to the cluster role with sufficient priviliges
+	// we assume the role is created by the k8s administrator
+	if c.opConfig.PodServiceAccountRoleBindingDefinition == "" {
+		c.opConfig.PodServiceAccountRoleBindingDefinition = `
+		{ 
+			"apiVersion": "rbac.authorization.k8s.io/v1beta1",
+			"kind": "RoleBinding", 
+			"metadata": { 
+				   "name": "zalando-postgres-operator"
+			},
+			"roleRef": {
+				"apiGroup": "rbac.authorization.k8s.io",
+				"kind": "ClusterRole",
+				"name": "zalando-postgres-operator"
+			},
+			"subjects": [
+				{
+					"kind": "ServiceAccount",
+					"name": "operator"
+				}
+			]	
+		}`
+	}
+	c.logger.Info("Parse role bindings")
+	// re-uses k8s internal parsing. See k8s client-go issue #193 for explanation
+	decode := scheme.Codecs.UniversalDeserializer().Decode
+	obj, groupVersionKind, err := decode([]byte(c.opConfig.PodServiceAccountRoleBindingDefinition), nil, nil)
+
+	switch {
+	case err != nil:
+		panic(fmt.Errorf("Unable to parse the definiton of the role binding  for the pod service account definiton from the operator config map: %v", err))
+	case groupVersionKind.Kind != "RoleBinding":
+		panic(fmt.Errorf("role binding definiton in the operator config map defines another type of resource: %v", groupVersionKind.Kind))
+	default:
+		c.PodServiceAccountRoleBinding = obj.(*rbacv1beta1.RoleBinding)
+		c.PodServiceAccountRoleBinding.Namespace = ""
+		c.PodServiceAccountRoleBinding.Subjects[0].Name = c.PodServiceAccount.Name
+		c.logger.Info("successfully parsed")
+
+	}
+
+	// actual roles bindings are deployed at the time of Postgres/Spilo cluster creation
+}
+
 func (c *Controller) initController() {
 	c.initClients()
 	c.initOperatorConfig()
 	c.initPodServiceAccount()
+	c.initRoleBinding()
 
 	c.initSharedInformers()
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -55,6 +55,7 @@ type Controller struct {
 
 	PodServiceAccount            *v1.ServiceAccount
 	PodServiceAccountRoleBinding *rbacv1beta1.RoleBinding
+	namespacesWithDefinedRBAC    sync.Map
 }
 
 // NewController creates a new controller
--- a/pkg/controller/postgresql.go
+++ b/pkg/controller/postgresql.go
@@ -24,7 +24,7 @@ import (
 )
 
 var (
-	namespacesWithPodServiceAccount = new(sync.Map)
+	namespacesWithDefinedRBAC = new(sync.Map)
 )
 
 func (c *Controller) clusterResync(stopCh <-chan struct{}, wg *sync.WaitGroup) {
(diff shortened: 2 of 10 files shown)


alexeyklyukin:
Since we are operating in a controller, let's make a sync.Map to cache the results of the `c.KubeClient.ServiceAccounts(namespace).Get(podServiceAccountName, metav1.GetOptions{})` check (and subsequent service account creation status) to avoid running this query more than necessary.

pkg/controller/postgresql.go
```diff
@@ -457,3 +466,40 @@ func (c *Controller) postgresqlDelete(obj interface{}) {
...
+
+  The operator does not sync these accounts after creation.
+*/
+func (c *Controller) createPodServiceAccounts(event spec.ClusterEvent) error {
+
+	namespace := event.NewSpec.GetNamespace()
+	podServiceAccountName := c.opConfig.PodServiceAccountName
+
+	if _, ok := namespacesWithPodServiceAccount.Load(namespace); ok {
+		c.logger.Infof(fmt.Sprintf("The required pod service account %v already exists in the namespace %v", podServiceAccountName, namespace))
```
alexeyklyukin: Showing this for every add, update or sync operation is a sure way to annoy the user.  I'd opt for now displaying anything here.

pkg/controller/postgresql.go
```diff
@@ -457,3 +466,40 @@ func (c *Controller) postgresqlDelete(obj interface{}) {
...
+		return nil
+	}
+
+	_, err := c.KubeClient.ServiceAccounts(namespace).Get(podServiceAccountName, metav1.GetOptions{})
+
+	if err != nil {
```
alexeyklyukin: We need to make sure the error is corresponds to the object not found. Use the `k8util.ResourceNotFound(err)` function for this. Otherwise, return the error.

pkg/controller/postgresql.go
```diff
@@ -457,3 +466,40 @@ func (c *Controller) postgresqlDelete(obj interface{}) {
...
+
+		c.logger.Infof(fmt.Sprintf("creating pod service account in the namespace %v", namespace))
+		c.logger.Infof("the pod service account %q cannot be retrieved in the namespace %q. Trying to deploy the account.", podServiceAccountName, namespace)
```
alexeyklyukin: this seems to be a leftover from some prior changes.

pkg/controller/postgresql.go
```diff
@@ -457,3 +466,40 @@ func (c *Controller) postgresqlDelete(obj interface{}) {
...
+		// to prevent a race condition when setting a namespace for many clusters
+		sa := *c.PodServiceAccount
+		_, err = c.KubeClient.ServiceAccounts(namespace).Create(&sa)
+		if err != nil {
+			return fmt.Errorf("cannot deploy the pod service account %q defined in the config map to the %q namespace: %v", podServiceAccountName, namespace, err)
+		}
+
+		namespacesWithPodServiceAccount.Store(namespace, podServiceAccountName)
+		c.logger.Infof("successfully deployed the pod service account %q to the %q namespace", podServiceAccountName, namespace)
+	}
```
alexeyklyukin: if the Get call returned us the service account we also need to store it in the namespacesWithPodServiceAccount map.

pkg/controller/postgresql.go
```diff
@@ -23,6 +23,10 @@ import (
 	"github.com/zalando-incubator/postgres-operator/pkg/util/ringlog"
 )
 
+var (
+	namespacesWithPodServiceAccount = new(sync.Map)
```
alexeyklyukin: Please, make it a part of the controller structure (not exporting it is fine).

zerg-junior:
@alexeyklyukin updated

pkg/controller/postgresql.go
```diff
@@ -457,3 +463,78 @@ func (c *Controller) postgresqlDelete(obj interface{}) {
...
+		if _, err = c.KubeClient.ServiceAccounts(namespace).Create(&sa); err != nil {
+			return fmt.Errorf("cannot deploy the pod service account %v defined in the config map to the %v namespace: %v", podServiceAccountName, namespace, err)
+		}
+
+		c.logger.Infof("successfully deployed the pod service account %v to the %v namespace", podServiceAccountName, namespace)
+	} else {
+		return err
+	}
+
+	return nil
```
alexeyklyukin: we should just return err at the end of the function, getting rid of else block completely.

pkg/controller/postgresql.go
```diff
@@ -457,3 +463,78 @@ func (c *Controller) postgresqlDelete(obj interface{}) {
...
+		rb := *c.PodServiceAccountRoleBinding
+		_, err = c.KubeClient.RoleBindings(namespace).Create(&rb)
+		if err != nil {
+			return fmt.Errorf("cannot bind the pod service account %q defined in the config map to the cluster role in the %q namespace: %v", podServiceAccountName, namespace, err)
+		}
+
+		c.logger.Infof("successfully deployed the role binding for the pod service account %q to the %q namespace", podServiceAccountName, namespace)
+
+	} else {
+		return err
```
alexeyklyukin: we should just return err at the end of the function, getting rid of else block completely (same as with createPodServiceAccount)

pkg/util/config/config.go
```diff
@@ -75,19 +75,20 @@ type Config struct {
...
-	LogS3Bucket                 string `name:"log_s3_bucket"`
-	KubeIAMRole                 string `name:"kube_iam_role"`
-	DebugLogging                bool   `name:"debug_logging" default:"true"`
-	EnableDBAccess              bool   `name:"enable_database_access" default:"true"`
-	EnableTeamsAPI              bool   `name:"enable_teams_api" default:"true"`
-	EnableTeamSuperuser         bool   `name:"enable_team_superuser" default:"false"`
-	TeamAdminRole               string `name:"team_admin_role" default:"admin"`
-	EnableMasterLoadBalancer    bool   `name:"enable_master_load_balancer" default:"true"`
-	EnableReplicaLoadBalancer   bool   `name:"enable_replica_load_balancer" default:"false"`
+	PodServiceAccountDefinition            string `name:"pod_service_account_definition" default:""`
```
alexeyklyukin: Should be added to the CRD-cofnuguration (in the crd_config.go). Also new parameters are not documented.

zerg-junior:
@alexeyklyukin updated

pkg/controller/postgresql.go
```diff
@@ -457,3 +463,74 @@ func (c *Controller) postgresqlDelete(obj interface{}) {
...
+/*
+  Ensures the pod service account and role bindings exists in a namespace before a PG cluster is created there so that a user does not have to deploy these credentials manually.
+  StatefulSets require the service account to create pods; Patroni requires relevant RBAC bindings to access endpoints.
+
+  The operator does not sync accounts/role bindings after creation.
+*/
+func (c *Controller) submitRBACCredentials(event spec.ClusterEvent) error {
+
+	namespace := event.NewSpec.GetNamespace()
+	if _, ok := c.namespacesWithDefinedRBAC.Load(namespace); ok {
```
alexeyklyukin: Now we use a single flag to indicate the namespace has service accounts and role binding defined, but in fact it may define the service account and fail on the role binding, so inside each of those functions we also need to handle "object already exist" (k8utils.ResourceAlreadyExists(err))  case as a non-error one.

pkg/util/config/crd_config.go
```diff
@@ -31,17 +31,18 @@ type PostgresUsersConfiguration struct {
...
-	WatchedNamespace              string              `json:"watched_namespace,omitempty"`
-	PDBNameFormat                 stringTemplate      `json:"pdb_name_format,omitempty"`
-	SecretNameTemplate            stringTemplate      `json:"secret_name_template,omitempty"`
-	OAuthTokenSecretName          spec.NamespacedName `json:"oauth_token_secret_name,omitempty"`
-	InfrastructureRolesSecretName spec.NamespacedName `json:"infrastructure_roles_secret_name,omitempty"`
-	PodRoleLabel                  string              `json:"pod_role_label,omitempty"`
-	ClusterLabels                 map[string]string   `json:"cluster_labels,omitempty"`
-	ClusterNameLabel              string              `json:"cluster_name_label,omitempty"`
-	NodeReadinessLabel            map[string]string   `json:"node_readiness_label,omitempty"`
+	PodServiceAccountDefinition            string              `json:"pod_service_account_definition,omitempty"`
```
alexeyklyukin: We also need to copy new fields from the CRD definition into the configmap-based one, see operator_config.go:importConfigurationFromCRD

pkg/controller/controller.go
```diff
@@ -162,6 +165,53 @@ func (c *Controller) initPodServiceAccount() {
...
+		{ 
+			"apiVersion": "rbac.authorization.k8s.io/v1beta1",
+			"kind": "RoleBinding", 
+			"metadata": { 
+				   "name": "zalando-postgres-operator"
+			},
+			"roleRef": {
+				"apiGroup": "rbac.authorization.k8s.io",
+				"kind": "ClusterRole",
+				"name": "zalando-postgres-operator"
```
alexeyklyukin: I wonder if it makes sense to configure the role we bind to?

zerg-junior: We assume this cluster role is created by the cluster administrator; the role binding itself is configurable via operator configmap or crd.

alexeyklyukin:
👍

zerg-junior:
:+1:

PR #350, opened 2018-07-20
Avoid showing an extra error when resizing volume fails

alexeyklyukin:
Do not show 'persistent volumes are not compatible' error for the
volumes that failed to be resized because of the other reasons (i.e.
the new size is smaller than the existing one).


Jan-M:
:+1:

alexeyklyukin:
👍

PR #352, opened 2018-07-31
Client go upgrade v6

alexeyklyukin:
Upgrade go-client to v6, implement deepcopy for CRDs, cope with some minor API changes (and packages moved).

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -11,10 +11,13 @@ import:
 - package: github.com/lib/pq
 - package: github.com/motomux/pretty
 - package: k8s.io/apimachinery
-  version: kubernetes-1.9.0
+  version: kubernetes-1.9.9
 - package: k8s.io/apiextensions-apiserver
-  version: kubernetes-1.9.0
+  version: kubernetes-1.9.9
 - package: k8s.io/client-go
   version: ^6.0.0
+- package: k8s.io/code-generator
+  version: kubernetes-1.9.9
+- package: k8s.io/gengo
 - package: gopkg.in/yaml.v2
 - package: github.com/mohae/deepcopy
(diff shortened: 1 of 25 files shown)


alexeyklyukin:
👍

glide.yaml
```diff
@@ -10,38 +10,14 @@ import:
...
-  - kubernetes/typed/extensions/v1beta1
-  - pkg/api
-  - pkg/api/v1
-  - pkg/apis/apps/v1beta1
-  - pkg/apis/extensions/v1beta1
-  - rest
-  - tools/cache
-  - tools/clientcmd
-  - tools/remotecommand
+  version: ^6.0.0
```
zerg-junior: what's so specific about version  6 ? afaik current docs [recommend v8](https://github.com/kubernetes/client-go#client-go)

pkg/controller/postgresql.go
```diff
@@ -147,12 +147,12 @@ func (d *crdDecoder) Decode() (action watch.EventType, object runtime.Object, er
 
 func (c *Controller) clusterWatchFunc(options metav1.ListOptions) (watch.Interface, error) {
 	options.Watch = true
+	// MIGRATION: FieldsSelectorParam(nil)
```
zerg-junior: does it mean that the code line was removed during migration cause the method is  not there anymore ?

zerg-junior:
I successfully run the operator with the updated client locally, so +1

zerg-junior:
:+1:

glide.yaml
alexeyklyukin: We want to do one step at a time, upgrade to v6, achieve our goal of implementing priority, then go further with 8.

pkg/controller/postgresql.go
alexeyklyukin: Yes

PR #353, opened 2018-08-01
Allow configuring pod priority globally and per cluster.

alexeyklyukin:
Allow to specify pod priority class for all pods managed by the operator,
as well as for those belonging to individual clusters.

Controlled by the pod_priority_class_name operator configuration
parameter and the podPriorityClassName manifest option.

See https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass
for the explanation on how to define priority classes since Kubernetes 1.8.

Some import order changes are due to go fmt.
Removal of OrphanDependents deprecated field.


Jan-M:
Great to see this being resolved. @aermakov-zalando

pkg/cluster/cluster.go
```diff
@@ -601,12 +601,12 @@ func (c *Cluster) Delete() {
...
 
 	for _, obj := range c.Secrets {
-		if delete, user := c.shouldDeleteSecret(obj); !delete {
+		if doDelete, user := c.shouldDeleteSecret(obj); !doDelete {
 			c.logger.Warningf("not removing secret %q for the system user %q", obj.GetName(), user)
 			continue
 		}
 		if err := c.deleteSecret(obj); err != nil {
-			c.logger.Warningf("could not delete secret: %v", err)
+			c.logger.Warningf("could not doDelete secret: %v", err)
```
zerg-junior: 'delete' reads better here.

docs/reference/cluster_manifest.md
```diff
@@ -89,7 +89,14 @@ Those are parameters grouped directly under  the `spec` key in the manifest.
...
   for details on tolerations and possible values of those keys. When set, this
   value overrides the `pod_toleration` setting from the operator. Optional.
-
+  
+* **podPriorityClassName**
+   a name of the [priority
+   class](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass)
+   that should be assigned to the cluster pods. When not specified, the value
+   is taken from the `pod_priority_class_name` operator parameter, if not set
+   then the default priority class is taken.
```
zerg-junior: probably worth noting here that the desired priority class must exist beforehand in the target k8s cluster; otherwise pod creation will fail

zerg-junior:
a general question regarding this capability : how will preemption work with pods originating from stateful sets ? [current docs](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass) seem to be short on this topic

pkg/cluster/cluster.go
alexeyklyukin: ugh, batch replace went wrong.

docs/reference/cluster_manifest.md
alexeyklyukin: good point

alexeyklyukin:
@zerg-junior I assume it doesn't care whether the pod to be preempted belongs to a statefulset; the expectations are that the lower-priority pod can be rescheduled in the future.

This is less a problem for a typical use-case of running PostgreSQL together with a stateless application in a single cluster: normally we expect all postgres pods to run with the same priority and nothing  PostgreSQL is usually the only service that runs in a statefulset.

alexeyklyukin:
@zerg-junior addressed your code review?

zerg-junior:
:+1:

alexeyklyukin:
👍

PR #364, opened 2018-08-07
Populate list of clusters in the controller at startup.

alexeyklyukin:
Assign the list of clusters in the controller with the up-to-date list
of Postgres manifests on Kubernetes during the startup.

Node migration routines launched asynchronously to the cluster
processing rely on an up-to-date list of clusters in the controller to
detect clusters affected by the migration of the node and lock them
when doing migration of master pods. Without the initial list the
operator was subject to race conditions like the one described at
https://github.com/zalando-incubator/postgres-operator/issues/363

Restructure the code to decouple list cluster function required by the
postgresql informer from the one that emits cluster sync events. No
extra work is introduced, since cluster sync already runs in a separate
goroutine (clusterResync).

Introduce explicit initial cluster sync at the end of
acquireInitialListOfClusters instead of relying on an implicit one
coming from list function of the PostgreSQL informer.

Some minor refactoring.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -326,6 +326,18 @@ func (c *Controller) initSharedInformers() {
 func (c *Controller) Run(stopCh <-chan struct{}, wg *sync.WaitGroup) {
 	c.initController()
 
+	// start workers reading from the events queue to prevent the initial sync from blocking on it.
+	for i := range c.clusterEventQueues {
+		wg.Add(1)
+		c.workerLogs[uint32(i)] = ringlog.New(c.opConfig.RingLogLines)
+		go c.processClusterEventsQueue(i, stopCh, wg)
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -326,18 +326,25 @@ func (c *Controller) initSharedInformers() {
 func (c *Controller) Run(stopCh <-chan struct{}, wg *sync.WaitGroup) {
 	c.initController()
 
+	// start workers reading from the events queue to prevent the initial sync from blocking on it.
```
zerg-junior: that basically means that workers must always start before `acquireInitialListOfClusters`  executes since it queues the 1st Sync event, right ?

alexeyklyukin: Yes, and if there are no consumers of that queue `acquireInitialListOfClusters`  function will block forever...

PR #368, opened 2018-08-10
Clarify what a default values is

zerg-junior:


Jan-M:
Imho not adding much, configuration is alway overwritten by local configuration. Could be moved to the bottom of the page with two links to where to find the defaults in the code/or example manifest so users can verify.

PR #371, opened 2018-08-20
[WIP] Grant 'superuser' to the members of  Postgres admin teams

zerg-junior:
Initial version, do not merge.

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -213,8 +213,10 @@ report success as a result of the last operation applied to them.
 
 The operator is capable of maintaining roles of multiple kinds within a Postgres database cluster:
 
-1. **System roles** are roles necessary for the proper work of Postgres itself such as a replication role or the initial superuser role. Operator delegates creating such roles to Patroni and only establishes relevant secrets.
+1. **System roles** are roles necessary for the proper work of Postgres itself such as a replication role or the initial superuser role. The operator delegates creating such roles to Patroni and only establishes relevant secrets.
 
-2. **Infrastructure roles** are roles for processes originating from external systems, e.g. monitoring robots. The operator creates such roles assuming k8s secrets with their credentials exist beforehand.
+2. **Infrastructure roles** are roles for processes originating from external systems, e.g. monitoring robots. The operator creates such roles in all PG clusters it manages assuming k8s secrets with the relevant credentials exist beforehand.
 
-3. **Human users**. Human users can originate from either the Teams API or a Postgres manifest. Operator differentiates between product teams that may be granted admin rights to maintain there own Postgres cluster, and the Postgres administrator teams that get superuser access to all PG databases running in a k8s cluster for the purposes of maintaining and troubleshooting.
+3. **Per-cluster robot users** are also roles for processes originating from external systems but defined for an individual Postgres cluster in its manifest. A typical example is a role for connections from an application that uses the database.
+
+4. **Human users** originate from the Teams API that returns list of the team members given a team id. Operator differentiates between (a) product teams that own a particular Postgres cluster and are granted admin rights to maintain it, and (b) Postgres superuser teams that get the superuser access to all PG databases running in a k8s cluster for the purposes of maintaining and troubleshooting.
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -208,3 +208,13 @@ generated from the current cluster manifest. There are two types of scans: a
 `sync scan`, running every `resync_period` seconds for every cluster, and the
 `repair scan`, coming every `repair_period` only for those clusters that didn't
 report success as a result of the last operation applied to them.
+
+## Postgres roles supported by the operator
+
+The operator is capable of maintaining roles of multiple kinds within a Postgres database cluster:
+
+1. **System roles** are roles necessary for the proper work of Postgres itself such as a replication role or the initial superuser role. Operator delegates creating such roles to Patroni and only establishes relevant secrets.
+
+2. **Infrastructure roles** are roles for processes originating from external systems, e.g. monitoring robots. The operator creates such roles assuming k8s secrets with their credentials exist beforehand.
+
+3. **Human users**. Human users can originate from either the Teams API or a Postgres manifest. Operator differentiates between product teams that may be granted admin rights to maintain there own Postgres cluster, and the Postgres administrator teams that get superuser access to all PG databases running in a k8s cluster for the purposes of maintaining and troubleshooting.
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -377,6 +377,9 @@ key.
   List of roles that cannot be overwritten by an application, team or
   infrastructure role. The default is `admin`.
 
+* **postgres_admin_teams**
+  List of teams which members need superuser roles in each PG database cluster to administer Postgres and maintain infrastructure built around it. The default is `postgres_admin`.
+
 ## Logging and REST API
 
 Parameters affecting logging and REST API listener. In the CRD-based configuration they are grouped under the `logging_rest_api` key.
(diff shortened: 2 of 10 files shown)


pkg/cluster/cluster.go
```diff
@@ -760,6 +763,17 @@ func (c *Cluster) initHumanUsers() error {
 	return nil
 }
 
+func (c *Cluster) initHumanUsers() error {
+
+	for _, postgresAdminTeam := range c.OpConfig.PostgresAdminTeams {
+		c.initTeamMembers(postgresAdminTeam, true /*isPostgresAdminTeam*/)
```
golangci: Error return value of `c.initTeamMembers` is not checked

pkg/cluster/cluster.go
```diff
@@ -760,6 +763,17 @@ func (c *Cluster) initHumanUsers() error {
...
+	}
+
+	c.initTeamMembers(c.Spec.TeamID, false /*isPostgresAdminTeam*/)
```
golangci: Error return value of `c.initTeamMembers` is not checked

docs/administrator.md
```diff
@@ -208,3 +208,13 @@ generated from the current cluster manifest. There are two types of scans: a
...
+
+## Postgres roles supported by the operator
+
+The operator is capable of maintaining roles of multiple kinds within a Postgres database cluster:
+
+1. **System roles** are roles necessary for the proper work of Postgres itself such as a replication role or the initial superuser role. Operator delegates creating such roles to Patroni and only establishes relevant secrets.
+
+2. **Infrastructure roles** are roles for processes originating from external systems, e.g. monitoring robots. The operator creates such roles assuming k8s secrets with their credentials exist beforehand.
+
+3. **Human users**. Human users can originate from either the Teams API or a Postgres manifest. Operator differentiates between product teams that may be granted admin rights to maintain there own Postgres cluster, and the Postgres administrator teams that get superuser access to all PG databases running in a k8s cluster for the purposes of maintaining and troubleshooting.
```
alexeyklyukin: There are also per-cluster robot users.

pkg/cluster/cluster.go
alexeyklyukin: What if your team is already listed in both the admin and the owner team, would resolveTeamMembers handle this correctly?

pkg/cluster/util.go
```diff
@@ -225,9 +227,9 @@ func (c *Cluster) getTeamMembers() ([]string, error) {
 		return []string{}, nil
 	}
 
-	teamInfo, err := c.teamsAPIClient.TeamInfo(c.Spec.TeamID, token)
+	teamInfo, err := c.teamsAPIClient.TeamInfo(teamID, token)
 	if err != nil {
-		c.logger.Warnf("could not get team info, returning empty list of team members: %v", err)
+		c.logger.Warnf("could not get team info for team %v, returning empty list of team members: %v", teamID, err)
```
alexeyklyukin: please, use %q instead of %v for simple string variables.

docs/reference/operator_parameters.md
```diff
@@ -377,6 +377,9 @@ key.
   List of roles that cannot be overwritten by an application, team or
   infrastructure role. The default is `admin`.
 
+* **postgres_admin_teams**
```
Jan-M: This is confusing. We give "admin" to owning teams and call the superusers admins. Let's call it by what is is. postgres_superuser_teams?

docs/administrator.md
zerg-junior: By robot users here do we mean for instance roles created for applications connections ?

pkg/cluster/cluster.go
Jan-M: Imho no need for this comment in the function call.

pkg/util/config/config.go
```diff
@@ -103,6 +103,7 @@ type Config struct {
 	TeamAPIRoleConfiguration map[string]string `name:"team_api_role_configuration" default:"log_statement:all"`
 	PodTerminateGracePeriod  time.Duration     `name:"pod_terminate_grace_period" default:"5m"`
 	ProtectedRoles           []string          `name:"protected_role_names" default:"admin"`
+	PostgresAdminTeams       []string          `name:"postgres_admin_teams" default:"postgres_admin"`
```
Jan-M: Please make this an empty default, don't want people to create a team called postgres_admin and suddenly get super power.

pkg/util/config/config.go
```diff
@@ -103,7 +103,7 @@ type Config struct {
 	TeamAPIRoleConfiguration map[string]string `name:"team_api_role_configuration" default:"log_statement:all"`
 	PodTerminateGracePeriod  time.Duration     `name:"pod_terminate_grace_period" default:"5m"`
 	ProtectedRoles           []string          `name:"protected_role_names" default:"admin"`
-	PostgresAdminTeams       []string          `name:"postgres_admin_teams" default:"postgres_admin"`
+	PostgresSuperuserTeams   []string          `name:"postgres_superuser_teams" default:"postgres_superusers"`
```
Jan-M: Still with a default ;)

pkg/cluster/cluster.go
```diff
@@ -760,6 +763,30 @@ func (c *Cluster) initHumanUsers() error {
...
+func (c *Cluster) initHumanUsers() error {
+
+	for _, postgresSuperuserTeam := range c.OpConfig.PostgresSuperuserTeams {
+		err := c.initTeamMembers(postgresSuperuserTeam, true /*isPostgresSuperuserTeam*/)
+		if err != nil {
+			return fmt.Errorf("Cannot create a team %q of Postgres superusers: %v", postgresSuperuserTeam, err)
+		}
+	}
+
+	for _, postgresSuperuserTeam := range c.OpConfig.PostgresSuperuserTeams {
```
Jan-M: is there no "in" operator? for String array?

pkg/cluster/cluster.go
```diff
@@ -760,6 +763,30 @@ func (c *Cluster) initHumanUsers() error {
...
+		if postgresSuperuserTeam == c.Spec.TeamID {
+			c.logger.Infof("Team %q owning the cluster is also a team of superusers. Created superuser roles for its members instead of admin roles.", c.Spec.TeamID)
+			return nil
+		}
+	}
+
+	err := c.initTeamMembers(c.Spec.TeamID, false /*isPostgresSuperuserTeam*/)
```
Jan-M: Seems this can be skipped if superuser team contains team already? (hinted by Oleksii)

zerg-junior: it is skipped in the lines ahead

pkg/cluster/cluster.go
zerg-junior: [apparently no due to the lack of generics in the language](https://stackoverflow.com/questions/10485743/contains-method-for-a-slice)

erthalion: @zerg-junior but why this part can't be merged with the previous loop?

pkg/cluster/cluster.go
```diff
@@ -760,6 +763,30 @@ func (c *Cluster) initHumanUsers() error {
 	return nil
 }
 
+func (c *Cluster) initHumanUsers() error {
+
+	for _, postgresSuperuserTeam := range c.OpConfig.PostgresSuperuserTeams {
+		err := c.initTeamMembers(postgresSuperuserTeam, true /*isPostgresSuperuserTeam*/)
```
erthalion: I would say if we need this commentary, it's better to incorporate this information into the argument itself - e.g. replace boolean with enum.

zerg-junior: this is just  a convenience commentary, imo not worth an enum for just two cases

Jan-M:
:+1:

PR #378, opened 2018-08-31
Move CRD definitions into a formal API to allow access from other controllers

coderanger:
This helps when building other operators on top of this, so I can do stuff like:

```go
t := &envtest.Environment{
    // ...
    CRDs: []*apiextv1beta1.CustomResourceDefinition{postgresv1.PostgresCRD()},
}
```

when testing my own operator code.

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/crds.go
+++ b/pkg/apis/acid.zalan.do/v1/crds.go
@@ -1,0 +1,54 @@
+package v1
+
+import (
+	"github.com/zalando-incubator/postgres-operator/pkg/apis/acid.zalan.do"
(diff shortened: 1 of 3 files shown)


Jan-M:
Looks good to me. Would be interested to know what other ideas you have for your own "CRDs" or what you want to operate? Mostly curious as we do some assumptions around what we manage/operate.

coderanger:
The initial plan for my operator is to make a Postgres object, wait for it to initialize, run a Job for the database migrations if needed, and then set all the various deployments for the Python app itself. It will probably grow more from there, but I’m starting small :)

PR #380, opened 2018-09-05
Fix the casing of github.com/Sirupsen/logrus to match what the project itself uses

coderanger:
Dep enforces this. I have no idea why it just started bothering me about it but it seemed like a safe thing to fix. Reference https://github.com/sirupsen/logrus/blob/master/example_basic_test.go#L4 or the default name capitalization in github to confirm I'm not off the mark?

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -100,7 +100,7 @@ imports:
   version: b2aad2c9a95d14eb978f29baa6e3a5c3c20eef30
 - name: github.com/peterbourgon/diskv
   version: 5f041e8faa004a95c88a202771f4cc3e991971e6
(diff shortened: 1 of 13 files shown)


coderanger:
This was apparently a whole big thing back in 2017 https://github.com/sirupsen/logrus/issues/553#issuecomment-306591437 https://github.com/sirupsen/logrus/issues/570#issuecomment-313933276 and I just never noticed until I happened to pull in both postgres-operator and another lib that uses logrus. Apparently lowercase is the Official Solution™ even though it can make your glide cache cranky.

redbaron:
this made glide.lock hash incorrect :-/

PR #381, opened 2018-09-11
Minor improvements

zerg-junior:
1. Remove zappr conf. We switched to using GitHub's code owners instead.
2. Update example manifest, operator config map and `run_operator_locally.sh` to create a cluster in a non-default namespace. This is useful for quick local testing of multi-namespace support.
3. Add a `test` target to the makefile. I tend to forget the relevant `go` command.
4. Minor documentation updates
5. Add Oleksii's talk at Atmosphere


Makefile
```diff
@@ -76,3 +76,6 @@ vet:
 
 deps:
 	@glide install --strip-vendor
+
+test:
+	@go test ./...
```
erthalion: I would even say `go test ./... -v`

PR #383, opened 2018-09-11
Some tweaks to ensure compat with newer Go

coderanger:
I upgraded to 1.11 on my laptop and thought I would clean up some minor things. Newer gofmt, a tweak to a format string to please newer `vet`, and fixing one test where the error message appears to vary between Go versions.


coderanger:
The JSON error thing is apparently already fixed upstream: https://github.com/golang/go/issues/27464

I just have bad timing.

PR #389, opened 2018-09-17
Unify warnings about unmovable pods

zerg-junior:
Ease debugging the situation when a PG master pod cannot be moved from a node that became unschedulable:

1. Report all such pods in one place instead of scattering warnings through the operator log
2. Report all the master ports on a node even if they are leftovers from the PG clusters that no longer exist


pkg/controller/node.go
```diff
@@ -142,14 +149,17 @@ func (c *Controller) moveMasterPodsOffNode(node *v1.Node) {
 		cl.Unlock()
 	}
 
-	totalPods := len(masterPods)
+	totalPods := len(nodePods)
```
erthalion: I think it's wrong - we want to report not all the pods, but only master pods that we wanted to migrate, since:

> %d out of %d master pods have been moved out from the node %q"

Which means that `totalPods` should be equal to the number of all master pods, that we want to move, and probably we need another counter at line 133. Although it's already a bit too complicated for such task, so I would probably even create a filter util function to filter pods with different conditions.

zerg-junior: > we want to report not all the pods, but only master pods that we wanted to migrate

that was exactly the issue with the original code - it reported only the master pods that the operator wanted to migrated and omitted those that originated from no-longer-existing clusters. It took us some time to figure out that there are other leftover (master) pods on the node that prevent the node from being shut down. 

> we need another counter at line 133

what for ?

erthalion: But then the message `could not move %d master pods from the node %q` is wrong, since we count all the pods that we can't move, not only masters - right?

zerg-junior: I double checked, both the method's name `moveMasterPodsOffNode` and [the code](https://github.com/zalando-incubator/postgres-operator/blob/master/pkg/controller/node.go#L103) suggest we only move master pods. This makes sense to me since replicas can be transparently re-created by deployments, but re-creating a master requires a failover first that has to be explicitly triggered

pkg/controller/node.go
```diff
@@ -55,15 +58,16 @@ func (c *Controller) nodeUpdate(prev, cur interface{}) {
 		return
 	}
 
-	if util.MapContains(nodeCur.Labels, map[string]string{"master": "true"}) {
```
zerg-junior: this check is covered by `nodeIsReady(nodeCur)`

PR #391, opened 2018-09-19
Improve reporting about rolling updates

zerg-junior:
1. Fix the bug that prevented logging the reasons for a rolling updates in some of the cases and remove the `todo` related to that
2. Use the word `update` everywhere instead of `upgrade` to be consistent with `k8s` docs.
3. Refer to containers in diffs not only by index in the `[]containers`, but also by name.
4. Briefly document how find reasons for rolling updates.
5. Minor typo/logging improvements.

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+
 	"k8s.io/api/core/v1"
 	policybeta1 "k8s.io/api/policy/v1beta1"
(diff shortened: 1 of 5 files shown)


pkg/cluster/sync.go
```diff
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+
```
erthalion: Just for the records - I think it makes sense to avoid an unnecessary changes next time.

PR #392, opened 2018-09-21
Properly overwrite empty allowed source ranges for load balancers

zerg-junior:
1. Fix the bug with the default range value being always overwritten. Even when `allowedSourceRanges` was unset in a PG manifest, the relevant slice was empty and had zero length, thus causing the overwrite of the supposed default value with the empty slice. Switching to `>0`  instead of `>=0` should fix this.
2. Log the final source ranges that make it to a service spec for debugging purposes. Note that the `generateService()` responsible for the spec is called in many places, and there is a chance that the desired source range is already applied or will not be applied for some reason such as a `Sync` failure. 
3. Briefly document the intended behavior of the `allowedSourceRanges` field


pkg/cluster/k8sres.go
```diff
@@ -961,13 +961,15 @@ func (c *Cluster) generateService(role PostgresRole, spec *acidv1.PostgresSpec)
 		// safe default value: lock load balancer to only local address unless overridden explicitly.
 		sourceRanges := []string{localHost}
 
-		allowedSourceRanges := spec.AllowedSourceRanges
-		if len(allowedSourceRanges) >= 0 {
-			sourceRanges = allowedSourceRanges
+		// spec.AllowedSourceRanges evaluates to the empty slice of zero length
+		// when omitted or set to 'null'/empty sequence in the PG manifest
+		if len(spec.AllowedSourceRanges) > 0 {
+			sourceRanges = spec.AllowedSourceRanges
```
alexeyklyukin: Minor nitpick, but it seems we don't need the sourceRanges variable altogether, i.e
```
if len(spec.AllowedSourceRanges) > 0 {
   serviceSpec.LoadBalancerSourceRanges = spec.AllowedSourceRanges
} else {
   serviceSpec.LoadBalancerSourceRanges = []string{localHost}
}
```

docs/administrator.md
```diff
@@ -198,7 +198,9 @@ services to an outer network, one can attach load balancers to them by setting
 cluster manifest. In the case any of these variables are omitted from the
 manifest, the operator configmap's settings `enable_master_load_balancer` and
 `enable_replica_load_balancer` apply. Note that the operator settings affect
-all Postgresql services running in a namespace watched by the operator.
+all Postgresql services running in all namespaces watched by the operator.
+
+To limit the range of IP adresses that can reach a load balancer, speficy the desired ranges in the `allowedSourceRanges` field (applies to both master and replica LBs). To prevent exposing LBs to the entire Internet, this field is set by default to `127.0.0.1/32`. To return to this default, explicitly set the field to the empty sequence `[]`; setting it to `null` or omitting entirely may not work due to [k8s handling of null fields](https://kubernetes.io/docs/concepts/overview/object-management-kubectl/declarative-config/#how-apply-calculates-differences-and-merges-changes).
```
erthalion: That's formulated strangely:

> this field is set by default to `127.0.0.1/32`

but

> omitting entirely may not work

Don't we assume a default value if it was omitted?

zerg-junior: good point, I will expand the explanation

Issue #394, opened 2018-10-04
Invalid regexp for rest api cluster status endpoint


miv:
Current apiserver regexp for cluster status is very narrow and doesn't support dashes in teamId.
https://github.com/zalando-incubator/postgres-operator/blob/master/pkg/apiserver/apiserver.go#L52

For example cluster with `teamId: "db-cluster-1"` will not be found using this regexp

erthalion:
Thanks for reporting the issue, #400 should address the problem.

PR #400, opened 2018-10-31
API url regexps

erthalion:
Make regexps for API more flexible, to allow e.g. teamId contains dashes. Solves #394


Jan-M:
:+1:

miv:
Thanks! When new official image is planned for release?

PR #409, opened 2018-11-09
Set HUMAN_ROLE Spilo env var

### LINKED ISSUE #407: Wrong PAM role added to pg_hba
I'd like to use the PAM support and noticed that the correct role name is not used in all places.

**Steps to reproduce**
I have set a PAM role in the operator config map:
`pam_role_name: mypamusers`. I replaced the entry in this file: https://github.com/zalando-incubator/postgres-operator/blob/master/manifests/configmap.yaml)

Then I commented out the lines for `pg_hba:` in the cluster definition in order to use the defaults that add the PAM support. I used this file as my starting point: https://github.com/zalando-incubator/postgres-operator/blob/master/manifests/complete-postgres-manifest.yaml

When I launch a new cluster using the two files above, I observe two things:
- the `pg_hba.conf` file in the postgres pod looks like this:
```
# Do not edit this file manually!
# It will be overwritten by Patroni!
local   all             all                                   trust
hostssl all             +zalandos    127.0.0.1/32       pam
host    all             all                127.0.0.1/32       md5
hostssl all             +zalandos    ::1/128            pam
host    all             all                ::1/128            md5
hostssl replication     standby all                md5
hostnossl all           all                all                reject
hostssl all             +zalandos    all                pam
hostssl all             all                all                md5
```
- a role `zalandos` is created in the postgres database

It appears that in both cases Spilo uses the ENV variable `HUMAN_ROLE` and it defaults to "zalandos" (see https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L349)

**Expected behaviour**
- Spilo's `HUMAN_ROLE` variable should be set to value of `pam_role_name`
- `pg_hba.conf` should use the value of `pam_role_name`
- no role named `zalandos` should be added to the postgres database

sdudoladov:
Set  the env var `HUMAN_ROLE` in a Spilo container; fixes #407


zerg-junior:
> I would like to know that this cannot be put into the spilo config json, dont like additional env vars when we have a big config object already.

that won't work, I've run a  few experiments 😐  the current PR version with the env var works, though

delivery.yaml
```diff
@@ -22,7 +22,7 @@ pipeline:
             go version
         - desc: 'Install Docker'
           cmd: |
-            curl -sSL https://get.docker.com/ | sh
+            curl -fLOsS https://delivery.cloud.zalando.com/utils/ensure-docker && sh ensure-docker && rm ensure-docker
```
erthalion: As far as I remember, we don't need this anymore for CDP.

PR #422, opened 2018-11-24
add .golangci.yml

jirfag:
This pull request fixes warnings in an analysis by [golangci.com](https://golangci.com/r/github.com/zalando-incubator/postgres-operator)

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,5 @@
+# https://github.com/golangci/golangci/wiki/Configuration
+
+service:
+  prepare:
+    - make deps


erthalion:
Thanks, looks great! I just wonder, why on the wiki page "Service configuration" marked as Beta, does it mean this feature is active only during beta testing?

jirfag:
it means that the feature is active only for some customers, including you. It won't be disabled in the future :) I am planning to enable it to all customers in 1-2 weeks

jirfag:
thank you!

PR #425, opened 2018-11-30
[WIP] Add 'admin' option to create role

### LINKED ISSUE #424: Roles created from manifest are missing "with admin" option
The roles created via the operator are not grantable by the admins as it is missing the with admin flag.

It should be configurable in the config map weather this grant should happen but the default can be yes, assuming we would not be the only ones suffering from this issue.

The admin role name is configurable already, so this one should be uses with "WITH ADMIN admin" when roles are created or altered.

From the documentation:

*ADMIN role_name*

The ADMIN clause is like ROLE, but the named roles are added to the new role WITH ADMIN OPTION, giving them the right to grant membership in this role to others.

https://www.postgresql.org/docs/9.6/sql-createrole.html

zerg-junior:
closes  #424

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -365,7 +365,7 @@ key.
   role name to grant to team members created from the Teams API. The default is
   `admin`, that role is created by Spilo as a `NOLOGIN` role.
 
-* **add_admin_flag_to_robot_roles**
+* **enable_admin_role_for_users**
    if `true`, the `team_admin_role` will have the rights to grant roles coming from PG manifests. Such roles will be created as in "CREATE ROLE 'role_from_manifest' ... ADMIN 'team_admin_role'". The default is `true`.
 
 * **pam_role_name**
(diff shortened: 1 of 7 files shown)


Jan-M:
I am fine with this, but slight objection to the config variable name. Not sure we need the word "robot". This word is not clearly defined, it is simply "users" from the manifest.

docs/reference/operator_parameters.md
```diff
@@ -365,6 +365,9 @@ key.
   role name to grant to team members created from the Teams API. The default is
   `admin`, that role is created by Spilo as a `NOLOGIN` role.
 
+* **add_admin_flag_to_robot_roles**
```
Jan-M: enable_admin_role_for_users

Could be good enough, the others are "admin/superusers" and "team members" so maybe "users" is good enough.

Jan-M:
Can the last comment be adressed, so we can close this?

PR #427, opened 2018-12-06
Add shm_volume option

### LINKED ISSUE #426: Add tempfs mounts for increased shmem usage
With PG10+ parallel queries may need more than the Docker Limit of 64MB of shared mem to run.

We have seen this fail here and there.

https://docs.okd.io/latest/dev_guide/shared_memory.html

config map should be extended by:

enable_shared_memory_mount: True

Default True is ok, given the long existence of PG10 should avoid others to run into this problem.

statefulset pod template then needs to be extended similar to the example above.

erthalion:
Add possibility to mount an extra memory volume to increase available shm. Fixes #426

### DIFF
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -97,6 +97,12 @@ Those are parameters grouped directly under  the `spec` key in the manifest.
    is taken from the `pod_priority_class_name` operator parameter, if not set
    then the default priority class is taken. The priority class itself must be defined in advance.
    
+* **shmVolume**
+  Start a database pod without limitations on shm memory. By default docker
+  limit `/dev/shm` to `64M`, which could be not enough if PostgreSQL uses
+  parallel workers heavily. If this option is enabled, to the target database
+  pod will be mounted a new tmpfs volume to remove this limitation.
+
 ## Postgres parameters
 
 Those parameters are grouped under the `postgresql` top-level key.
@@ -111,6 +117,7 @@ Those parameters are grouped under the `postgresql` top-level key.
   a dictionary of postgres parameter names and values to apply to the resulting
   cluster. Optional (Spilo automatically sets reasonable defaults for
   parameters like work_mem or max_connections).
+
 
 ## Patroni parameters
 
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -96,12 +96,18 @@ Those are parameters grouped directly under  the `spec` key in the manifest.
    that should be assigned to the cluster pods. When not specified, the value
    is taken from the `pod_priority_class_name` operator parameter, if not set
    then the default priority class is taken. The priority class itself must be defined in advance.
-   
+
 * **shmVolume**
   Start a database pod without limitations on shm memory. By default docker
-  limit `/dev/shm` to `64M`, which could be not enough if PostgreSQL uses
-  parallel workers heavily. If this option is enabled, to the target database
-  pod will be mounted a new tmpfs volume to remove this limitation.
+  limit `/dev/shm` to `64M` (see e.g. the [docker
+  issue](https://github.com/docker-library/postgres/issues/416), which could be
+  not enough if PostgreSQL uses parallel workers heavily. If this option is
+  present and value is `true`, to the target database pod will be mounted a new
+  tmpfs volume to remove this limitation. If it's not present, the decision
+  about mounting a volume will be made based on operator configuration
+  (`shm_volume`, which is `true` by default). It it's present and value is
+  `false`, then no volume will be mounted no matter how operator was configured
+  (so you can override the operator configuration).
 
 ## Postgres parameters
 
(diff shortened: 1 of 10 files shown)


pkg/cluster/k8sres.go
```diff
@@ -420,6 +421,10 @@ func generatePodTemplate(
 		Tolerations:                   *tolerationsSpec,
 	}
 
+	if shmVolume == true {
```
golangcibot: should omit comparison to bool constant, can be simplified to shmVolume (from `megacheck`)

docs/reference/cluster_manifest.md
```diff
@@ -97,6 +97,12 @@ Those are parameters grouped directly under  the `spec` key in the manifest.
    is taken from the `pod_priority_class_name` operator parameter, if not set
    then the default priority class is taken. The priority class itself must be defined in advance.
    
+* **shmVolume**
```
zerg-junior: 1. @Jan-M  just to double check - the original issue description talked about modifying the configmap, not only cluster manifests. Can it be the case that we will need to enable this feature for a bunch of PG clusters, and then it will be easier to enable it in the operator and let it take care of the rest  ?  IMHO most clusters will no need this feature, and we can leave it on the level of the cluster manifest as it is currently implemented

2. is the intention of this feature to mount a `tmpfs` volume at `/dev/shm` without restricting the volume's size ?

3. ``shmVolume: true`` is probably not the most descriptive name for that option. Sth like `enableExtraSharedMemory` sounds imo better cause it reflects the purpose of the feature.

pkg/util/constants/postgresql.go
```diff
@@ -10,4 +10,8 @@ const (
 
 	PostgresConnectRetryTimeout = 2 * time.Minute
 	PostgresConnectTimeout      = 15 * time.Second
+
+	ShmVolumeName = "dshm"
+	ShmVolumePath = "/dev/shm"
+	MinShmSize    = 64 * 1024 * 1024
```
zerg-junior: where is `MinShmSize` used ? and the value is in bytes, isn't it ?

docs/reference/cluster_manifest.md
```diff
@@ -97,6 +97,12 @@ Those are parameters grouped directly under  the `spec` key in the manifest.
...
+  Start a database pod without limitations on shm memory. By default docker
+  limit `/dev/shm` to `64M`, which could be not enough if PostgreSQL uses
+  parallel workers heavily. If this option is enabled, to the target database
+  pod will be mounted a new tmpfs volume to remove this limitation.
```
zerg-junior: 1. default value ?
2. related [docker-library issue](https://github.com/docker-library/postgres/issues/416)

pkg/cluster/k8sres.go
```diff
@@ -882,6 +893,31 @@ func (c *Cluster) getNumberOfInstances(spec *acidv1.PostgresSpec) int32 {
...
+	volumes := append(podSpec.Volumes, v1.Volume{
+		Name: constants.ShmVolumeName,
+		VolumeSource: v1.VolumeSource{
+			EmptyDir: &v1.EmptyDirVolumeSource{
+				Medium: "Memory",
+			},
+		},
+	})
+
+	mounts := append(podSpec.Containers[0].VolumeMounts,
```
zerg-junior: Containers[0] == postgres container .  worth a comment/name since not everyone is familiar with this implicit assumption.

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -51,7 +51,7 @@ type PostgresSpec struct {
 	Tolerations          []v1.Toleration      `json:"tolerations,omitempty"`
 	Sidecars             []Sidecar            `json:"sidecars,omitempty"`
 	PodPriorityClassName string               `json:"pod_priority_class_name,omitempty"`
-	ShmVolume            bool                 `json:"shmVolume,omitempty"`
```
Jan-M: should we make this name nicer?

enableSharedMemoryVolume?

pkg/util/config/config.go
```diff
@@ -38,6 +38,7 @@ type Resources struct {
 	NodeReadinessLabel      map[string]string `name:"node_readiness_label" default:""`
 	MaxInstances            int32             `name:"max_instances" default:"-1"`
 	MinInstances            int32             `name:"min_instances" default:"-1"`
+	ShmVolume               bool              `name:"shm_volume" default:"true"`
```
Jan-M: I thing we use "enable" prefix in other config options?

enableShardedMemoryVolume ?

erthalion: Well, I actually disagree. The data type itself for this value tells that it's enable/disable option, so add it to the name would be redundant.

pkg/cluster/k8sres.go
```diff
@@ -396,6 +397,16 @@ func generateSidecarContainers(sidecars []acidv1.Sidecar,
 	return nil, nil
 }
 
+// Check whether or not we're requested to mount an shm volume,
+// taking into account that PostgreSQL manifest has precedence.
+func mountShmVolumeNeeded(opConfig config.Config, pgSpec *acidv1.PostgresSpec) bool {
```
zerg-junior: `isShmVolumeMountNeeded ` - adjectives are easier to read in the context of boolean expressions

PR #435, opened 2019-01-01
Fix reference to enable_database_access in operator_parameters.md

aknuds1:
AFAICT the parameter `enable_db_access` referred to in docs/reference/operator_parameters.md should really be `enable_database_access`.


erthalion:
Hm...looks like judging from [this](https://github.com/zalando-incubator/postgres-operator/blob/master/pkg/util/config/config.go#L89) you're right, thanks for noticing!

PR #417, opened 2018-11-14
fix golint fail #415

u5surf:
#415 

i fix golint fail
result:
```
$ golint -set_exit_status $(go list ./...)
$
```
please check this PR 
https://goreportcard.com/report/github.com/zalando-incubator/postgres-operator#golint
is obsoluted result?? is not HEAD result??

PR #438, opened 2019-01-03
Extend #417 (fix golint failures)

zerg-junior:
Based on earlier work #417  by @u5surf  . This PR can be retired, [if the permission is given](https://github.com/zalando-incubator/postgres-operator/pull/417#issuecomment-451108870) to modify the original branch.


pkg/util/retryutil/retry_util.go
```diff
@@ -17,8 +17,10 @@ type Ticker struct {
 	ticker *time.Ticker
 }
 
+// Stop returns ticker.Stop
 func (t *Ticker) Stop() { t.ticker.Stop() }
 
+// Tick returns ticker.C
```
erthalion: Let's give it more meaningful commentary, e.g. Wrapper around ticker to make it testable.

pkg/cluster/types.go
```diff
@@ -20,6 +21,7 @@ const (
 	Replica PostgresRole = "replica"
 )
 
+// PodEventType describes value of event type
```
erthalion: It sounds strange, how data type can describe a value? Maybe 'PodEventType is a data type that represents type of a pod-related event'?

pkg/apis/acid.zalan.do/v1/util.go
```diff
@@ -83,6 +84,7 @@ func validateCloneClusterDescription(clone *CloneDescription) error {
 	return nil
 }
 
+// Success of the last operation on a Postgres cluster
```
erthalion: Why the last operation? Isn't it just for the current status, that was passed as an argument?

pkg/apis/acid.zalan.do/v1/register.go
```diff
@@ -8,15 +8,20 @@ import (
...
 
+// APIVersion of the `postgresql` and `operator` CRDs
 const (
 	APIVersion = "v1"
 )
 
 var (
 	// localSchemeBuilder and AddToScheme will stay in k8s.io/kubernetes.
+
+	// SchemeBuilder is runtime.SchemeBuilder
```
erthalion: "An instance of runtime.SchemeBuilder, global for this package"?

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -103,12 +116,14 @@ type TeamsAPIConfiguration struct {
...
 }
 
+// LoggingRESTAPIConfiguration defines Logging API conf
 type LoggingRESTAPIConfiguration struct {
 	APIPort               int `json:"api_port,omitempty"`
 	RingLogLines          int `json:"ring_log_lines,omitempty"`
 	ClusterHistoryEntries int `json:"cluster_history_entries,omitempty"`
 }
 
+// ScalyrConfiguration defines the configration of ScalyrAPI
```
erthalion: `configuration *for* ScalyrAPI`?

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -79,18 +88,22 @@ type LoadBalancerConfiguration struct {
...
+// AWSGCPConfiguration defines the configuration for AWS
+// TODO complete Google Cloud Platform (GCP) configuration
 type AWSGCPConfiguration struct {
 	WALES3Bucket string `json:"wal_s3_bucket,omitempty"`
 	AWSRegion    string `json:"aws_region,omitempty"`
 	LogS3Bucket  string `json:"log_s3_bucket,omitempty"`
 	KubeIAMRole  string `json:"kube_iam_role,omitempty"`
 }
 
+// OperatorDebugConfiguration defines the configration of debug
```
erthalion: `defines options for debug mode`?

Jan-M: Do these types with useful names really need comments? :dancer:

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
Jan-M: See above, make it short and quick.

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
erthalion: Yes, it's questionable. The original purpose was to silence the golint, since it complains about missing comments. I thought we could do this in a more meaningful way,

PR #442, opened 2019-01-08
Add logical backup

sdudoladov:
Add logical backup by managing k8s cron jobs.

> Jan-M 
> The operator will just manage another resource, which is a k8s cron job and sync it with the configuration. Kubernetes takes over the handling/spawning/retry, and e.g. monitoring will have to be solved somewhere else.


docs/user.md
```diff
@@ -308,3 +308,12 @@ every 6 hours.
...
 volumes that correspond to the previously running pods is not changed.
+
+## Logical backups
+
+If you add
+```
+logical_backup:
+  enable_logical_backup: true
+```
+to the cluster manifest, the operator will start a k8s cron job that will periodically execute `pg_dumpall` on the target PG cluster and upload results to an S3 bucket. Note that due to the [limitation of Kubernetes cron jobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations) it is highly advisable to set up additional monitoring for this feature; such monitoring is outside of the scope of operator responsibilities. See configuration reference for details.
```
avaczi: A link to the reference would be welcome

docs/administrator.md
```diff
@@ -226,3 +226,13 @@ The operator is capable of maintaining roles of multiple kinds within a Postgres
...
+
+## Logical backups
+
+The operator can launch k8s cron jobs to do periodic logical backups of all PG clusters under its control. The cron job spawns a separate pod with a single container that connects to one of the Postgres replicas for a backup. The operator updates cron jobs during Sync if a schedule or a docker image of a job changes. Notes:
+
+1. The provided  `registry.opensource.zalan.do/acid/logical-backup` image implements the backup via `pg_dumpall` and upload of (compressed) results to an S3 bucket; `pg_dumpall` requires a `superuser` access to a DB.
+
+2. Due to the [limitation of Kubernetes cron jobs](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations) it is highly advisable to set up additional monitoring for this feature; such monitoring is outside of the scope of operator responsibilities. 
+
+3. The operator does not remove old backups: it is the responsibility of a k8s cluster administrator to define and implement a retention policy for backups.
```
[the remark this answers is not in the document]
zerg-junior: for now not. the first task is to simply make the cron job running.

Jan-M:
Can we rephrase this a bit in the description. The operator is def. not responsible for spawning anything.

The operator will just manage another resource, which is a k8s cron job and sync it with the configuration. Kubernetes takes over the handling/spawning/retry, and e.g. monitoring will have to be solved somewhere else.

docs/administrator.md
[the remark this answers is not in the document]
Jan-M: Cron job data is exposed via k8s api and for backup data itself there is s3 for now or equivalent storage APIs. (Similar to what is already used for physical backups)

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -149,3 +150,10 @@ type OperatorConfigurationUsers struct {
 }
 
 type Duration time.Duration
+
+type OperatorLogicalBackupConfiguration struct {
+	EnableLogicalBackup      bool   `json:"enable_logical_backup,omitempty"`
```
Jan-M: Hmm we can remove the "logical_backup" here or? Given it is already within the struct and the nested key?

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -127,3 +128,9 @@ type UserFlags []string
 
 // PostgresStatus contains status of the PostgreSQL cluster (running, creation failed etc.)
 type PostgresStatus string
+
+// LogicalBackup contains config of a k8s cron job responsible for running pg_dumpall
+type LogicalBackup struct {
+	EnableLogicalBackup   bool   `json:"enable_logical_backup,omitempty"`
```
Jan-M:
```
logical_backup:
    enabled: true
    schedule: "0 2 * * *"
```

Jan-M: I am not sure if this should be nested in the manifest? what is more convenient? Similar to load balancers?

zerg-junior: at present we have both, for example volume size is nested:
```
volume:
    size: 1Gi
```
if we plan to add more options, we are imo better leaving it  nested

zerg-junior:
> So let's stick to capitalization.

sure, just mixed it with the Patroni params in the complete manifest

docs/reference/cluster_manifest.md
```diff
@@ -109,6 +111,12 @@ Those are parameters grouped directly under  the `spec` key in the manifest.
   is `false`, then no volume will be mounted no matter how operator was
   configured (so you can override the operator configuration).
 
+* **enableLogicalBackup**
+  Determines if the logical backup of this cluster should be taken and uploaded to S3. Default: false.
+
+* **logicalBackupSchedule**
+  Backup schedule in the Cron format. Default: "30 00 \* \* \*"
```
erthalion: It worth attaching the reference to the format:

https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#schedule

because it supports non-standard stuff, https://www.freebsd.org/cgi/man.cgi?crontab%285%29

docs/reference/operator_parameters.md
```diff
@@ -443,4 +443,22 @@ scalyr sidecar. In the CRD-based configuration they are grouped under the
   Memory limit value for the Scalyr sidecar. The default is `1Gi`.
 
 
+## Logical backup
+
+  These parameters configure a k8s cron job managed by the operator to produce Postgres logical backups.
+  In the CRD-based configuration those parameters are grouped under the `logical_backup` key.
+
+  * **enable_logical_backup**
+    Determines if the operator should initiate the backup of all Postgres clusters it controls. Default: false.
```
erthalion: I think it's formulated incorrectly, operator doesn't initiate the backup itself, it creates a job for that.

pkg/cluster/cluster.go
```diff
@@ -82,6 +83,8 @@ type Cluster struct {
 	currentProcess   Process
 	processMu        sync.RWMutex // protects the current operation for reporting, no need to hold the master mutex
 	specMu           sync.RWMutex // protects the spec for reporting, no need to hold the master mutex
+
+	backupJob        *batchv1beta1.CronJob  // periodical logical backups independent from WAL archiving
```
erthalion: Maybe we can be more specific in the variable name, `logicalBackupJob`? Just in case if there would be other kind of backups in the future.

pkg/cluster/cluster.go
```diff
@@ -296,6 +299,12 @@ func (c *Cluster) Create() error {
 		c.logger.Infof("databases have been successfully created")
 	}
 
+	if c.Postgresql.Spec.EnableLogicalBackup {
+		if err := c.createBackupCronJob(); err != nil {
+			return fmt.Errorf("could not create a k8s cron job for logical backups: %v", err)
```
erthalion: Should it be a fatal error or is it something we can proceed with?

sdudoladov: good question ! here I've chosen to follow the creation pattern of all other k8s resources in this method: if a resource cannot be created, exit with error

sdudoladov: the advantage of the approach is that  a misconfigured backup has more chances to fail fast and hence be immediately visible

pkg/cluster/k8sres.go
```diff
@@ -1251,3 +1253,68 @@ func (c *Cluster) getClusterServiceConnectionParameters(clusterName string) (hos
...
+
+	var (
+		err         error
+		podTemplate *v1.PodTemplateSpec
+	)
+
+	c.logger.Debug("Generating logical backup pod template")
+
+	defaultResources := c.makeDefaultResources()
+	resourceRequirements, err := generateResourceRequirements(c.Spec.Resources, defaultResources)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/cluster/sync.go
```diff
@@ -519,3 +530,47 @@ func (c *Cluster) syncDatabases() error {
...
+	var (
+		job        *batchv1beta1.CronJob
+		desiredJob *batchv1beta1.CronJob
+		err        error
+	)
+	c.setProcessName("syncing the logical backup job")
+
+	if job, err = c.KubeClient.CronJobsGetter.CronJobs(c.Namespace).Get(c.logicalBackupJob.Name, metav1.GetOptions{}); err == nil {
+
+		desiredJob, err = c.generateLogicalBackupJob()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/cluster/sync.go
```diff
@@ -519,3 +530,47 @@ func (c *Cluster) syncDatabases() error {
...
+	c.logger.Info("could not find the cluster's logical backup job")
+
+	if err = c.createBackupCronJob(); err == nil {
+		c.logger.Infof("created missing logical backup job %q", util.NameFromMeta(job.ObjectMeta))
+	} else {
+		if !k8sutil.ResourceAlreadyExists(err) {
+			return fmt.Errorf("could not create missing logical backup job: %v", err)
+		}
+		c.logger.Infof("logical backup job %q already exists", util.NameFromMeta(job.ObjectMeta))
+		if job, err = c.KubeClient.CronJobsGetter.CronJobs(c.Namespace).Get(c.logicalBackupJob.Name, metav1.GetOptions{}); err != nil {
```
golangcibot: SA4006: this value of `job` is never used (from `staticcheck`)

Jan-M:
Whats the default job start time? Can it be default 02:00 or something straight out of config?

sdudoladov:
> Whats the default job start time? Can it be default 02:00 or something straight out of config?

It is currently "30 00 * * *"

pkg/cluster/sync.go
```diff
@@ -519,3 +530,52 @@ func (c *Cluster) syncDatabases() error {
...
+	var (
+		job        *batchv1beta1.CronJob
+		desiredJob *batchv1beta1.CronJob
+		err        error
+	)
+	c.setProcessName("syncing the logical backup job")
+
+	// operator pod at startup syncs all clusters, logicalBackupJob will be nil at this point
+	if (c.Postgresql.Spec.EnableLogicalBackup == true) && (c.logicalBackupJob == nil) {
+		c.logicalBackupJob, err = c.generateLogicalBackupJob()
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/cluster/sync.go
```diff
@@ -519,3 +530,52 @@ func (c *Cluster) syncDatabases() error {
...
+func (c *Cluster) syncLogicalBackupJob() error {
+	var (
+		job        *batchv1beta1.CronJob
+		desiredJob *batchv1beta1.CronJob
+		err        error
+	)
+	c.setProcessName("syncing the logical backup job")
+
+	// operator pod at startup syncs all clusters, logicalBackupJob will be nil at this point
+	if (c.Postgresql.Spec.EnableLogicalBackup == true) && (c.logicalBackupJob == nil) {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `c.Postgresql.Spec.EnableLogicalBackup` (from `gosimple`)

pkg/util/config/config.go
```diff
@@ -66,12 +66,21 @@ type Scalyr struct {
 	ScalyrMemoryLimit   string `name:"scalyr_memory_limit" default:"1Gi"`
 }
 
+// LogicalBackup
+type LogicalBackup struct {
+	EnableLogicalBackup      bool   `name:"enable_logical_backup" default:"false"`
```
erthalion: Do we even need a separate flag for that? We can enable this functionality when any of `logical_backup_schedule`, `logical_backup_docker_image`, `logical_backup_s3_bucket` is not empty.

pkg/cluster/util.go
```diff
@@ -483,3 +484,13 @@ func (c *Cluster) GetSpec() (*acidv1.Postgresql, error) {
 func (c *Cluster) patroniUsesKubernetes() bool {
 	return c.OpConfig.EtcdHost == ""
 }
+
+func (c *Cluster) logLogicalBackupJobChanges(old, new *batchv1beta1.CronJob, reason string) {
```
erthalion: This is confusing, since the function is used only once, and doesn't really have any significant functionality.

pkg/cluster/sync.go
```diff
@@ -519,3 +530,52 @@ func (c *Cluster) syncDatabases() error {
...
+func (c *Cluster) syncLogicalBackupJob() error {
+	var (
+		job        *batchv1beta1.CronJob
+		desiredJob *batchv1beta1.CronJob
+		err        error
+	)
+	c.setProcessName("syncing the logical backup job")
+
+	// sync the job if it exists
+	// NB operator pod at startup syncs all clusters, c.logicalBackupJob will be nil during such Sync
```
erthalion: Probably there is a lack of explanation why it would be nil?

pkg/cluster/sync.go
```diff
@@ -519,3 +530,52 @@ func (c *Cluster) syncDatabases() error {
...
+			c.logLogicalBackupJobChanges(job, desiredJob, reason)
+			if err = c.patchLogicalBackupJob(desiredJob); err != nil {
+				return fmt.Errorf("could not update logical backup job to match desired state: %v", err)
+			}
+			c.logger.Info("the logical backup job is in the desired state now")
+		}
+		return nil
+	}
+	if !k8sutil.ResourceNotFound(err) {
+		return fmt.Errorf("could not get logical backp job: %v", err)
```
erthalion: Why we don't create missing job here?

pkg/cluster/sync.go
```diff
@@ -519,3 +530,52 @@ func (c *Cluster) syncDatabases() error {
...
+		desiredJob, err = c.generateLogicalBackupJob()
+		if err != nil {
+			return fmt.Errorf("could not generate the desired logical backup job state: %v", err)
+		}
+		if match, reason := k8sutil.SameLogicalBackupJob(job, desiredJob); !match {
+			c.logLogicalBackupJobChanges(job, desiredJob, reason)
+			if err = c.patchLogicalBackupJob(desiredJob); err != nil {
+				return fmt.Errorf("could not update logical backup job to match desired state: %v", err)
+			}
+			c.logger.Info("the logical backup job is in the desired state now")
```
erthalion: From the logs it would not be clear what happened (whether it was updated, or was already in the required state), I suggest to change to "logical backup job is synced"

pkg/cluster/cluster.go
```diff
@@ -567,6 +576,43 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
...
+				c.logger.Errorf("could not delete a k8s cron job for logical backups: %v", err)
+				updateFailed = true
+				return
+			}
+
+		}
+
+		// apply schedule changes
+		// this is the only parameter of logical backups a user can overwrite in the cluster manifest
+		if (newSpec.Spec.EnableLogicalBackup) &&
```
erthalion: Does this condition mean that, if we've updated a spec to enable logical backup, it will be created and then synced right away, so the same job will be done twice?

pkg/cluster/resources.go
```diff
@@ -609,6 +610,44 @@ func (c *Cluster) createRoles() (err error) {
 	return c.syncRoles()
 }
 
+func (c *Cluster) createLogicalBackupJob() (err error) {
```
erthalion: I'm confused - why `createLogicalBackupJob` is defined in `cluster/resources`, `deleteLogicalBackupJob` in `cluster/cluster` and `syncLogicalBackupJob` in `cluster/sync`?

pkg/cluster/cluster.go
```diff
@@ -1006,3 +1059,10 @@ func (c *Cluster) deletePatroniClusterConfigMaps() error {
 
 	return c.deleteClusterObject(get, deleteConfigMapFn, "configmap")
 }
+
+func (c *Cluster) deleteLogicalBackupJob() error {
+
+	c.logger.Debug("removing the logical backup job")
```
erthalion: As you mentioned before, if we'll delete a cronjob, a corresponding jobs will be also stopped. It means in turn that the backup would be interrupted, which in fact is something that can be important (e.g. if it takes long time). By itself it's ok, but we need to make this information visible, and since if I recall correctly we do not show `Debug` log messages by default, I suggest to raise it to something visible.

pkg/cluster/cluster.go
```diff
@@ -480,7 +488,8 @@ func compareResoucesAssumeFirstNotNil(a *v1.ResourceRequirements, b *v1.Resource
 }
 
 // Update changes Kubernetes objects according to the new specification. Unlike the sync case, the missing object.
-// (i.e. service) is treated as an error.
+// (i.e. service) is treated as an error
+// logical backup cron jobs are an exception: they can be freely created/deleted by users
```
erthalion: > they can be freely created/deleted by users

Can you elaborate ?

pkg/cluster/cluster.go
sdudoladov: good catch !

pkg/cluster/cluster.go
sdudoladov: we do log Debugf by default though

pkg/cluster/resources.go
sdudoladov: this follows the implementation of  StatefulSets. I'll move `delete..` though as it does not seem to match

pkg/cluster/sync.go
sdudoladov: that looks like a dead comment ; I removed this pointer earlier

pkg/util/config/config.go
sdudoladov: makes sense, will update the docs as well

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -174,3 +175,9 @@ type OperatorConfigurationUsers struct {
 
 //Duration shortens this frequently used name
 type Duration time.Duration
+
+type OperatorLogicalBackupConfiguration struct {
+	Schedule    string `json:"logical_backup_schedule,omitempty"`
+	DockerImage string `json:"logical_backup_docker_image,omitempty"`
+	S3Bucket    string `json:"logical_backup_s3_bucket,omitempty"`
```
FxKu: why so AWS-specific? Can it not be something like `logical_backup_target` everywhere?

sdudoladov: supporting other targets is outside the scope of this PR; such a change also has to be synced with the logical backup image env

Jan-M: I would still claim, we could just remove s3 from here.

That is probably sufficient to leave any further work to the docker image.

Google, Digital Ocean, they probably all just take a "path" to a "bucket".

sdudoladov:
:+1:

docs/reference/operator_parameters.md
```diff
@@ -476,4 +478,16 @@ scalyr sidecar. In the CRD-based configuration they are grouped under the
...
+## Logical backup
+
+  These parameters configure a k8s cron job managed by the operator to produce Postgres logical backups.
+  In the CRD-based configuration those parameters are grouped under the `logical_backup` key.
+
+  * **logical_backup_schedule**
+    Backup schedule in the cron format. Please take [the reference schedule format](https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/#schedule) into account. Default: "30 00 \* \* \*"
+  
+  * **logical_backup_docker_image**
+    Docker image for the pods of the cron job. Must implement backup logic and correctly handle pod and job restarts. Default: "registry.opensource.zalan.do/acid/logical-backup:master-42" (`pgdumpall` on the replica + upload to S3).
```
avaczi: `pg_dumpall`

avaczi: and is it wise to hardcode a version in the docs here?  Isn't there a better way?

docs/reference/operator_parameters.md
```diff
@@ -476,4 +478,16 @@ scalyr sidecar. In the CRD-based configuration they are grouped under the
...
+
+  * **logical_backup_s3_bucket**
+    S3 bucket to store backup results. The bucket has to be present and accessible by Postgres pods. Default: empty.
```
avaczi: Maybe some description where exactly the backup will end up?  I remember having it read elsewhere...

pkg/util/config/config.go
```diff
@@ -66,12 +66,20 @@ type Scalyr struct {
 	ScalyrMemoryLimit   string `name:"scalyr_memory_limit" default:"1Gi"`
 }
 
+// LogicalBackup
+type LogicalBackup struct {
+	LogicalBackupSchedule    string `name:"logical_backup_schedule" default:"30 00 * * *"`
+	LogicalBackupDockerImage string `name:"logical_backup_docker_image" default:"registry.opensource.zalan.do/acid/logical-backup"`
```
avaczi: is this valid without a tag?

sdudoladov: yes, this will fetch the image tagged `:latest`. This tag does [have its own problems](https://vsupalov.com/docker-latest-tag/), however in that particular case it should always fetch the last build image. That is exactly the image we intend to use ourselves.

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -91,4 +91,7 @@ configuration:
     # scalyr_api_key: ""
     # scalyr_image: ""
     # scalyr_server_url: ""
-
+  logical_backup_key:
```
FxKu: why is it called "_key"? I'm missing the enable_logical_backup flag here.

pkg/cluster/k8sres.go
```diff
@@ -1281,3 +1283,167 @@ func (c *Cluster) getClusterServiceConnectionParameters(clusterName string) (hos
...
+		TopologyKey: "kubernetes.io/hostname",
+	}
+	podAffinity := v1.Affinity{
+		PodAffinity: &v1.PodAffinity{
+			PreferredDuringSchedulingIgnoredDuringExecution: []v1.WeightedPodAffinityTerm{{
+				Weight:          1,
+				PodAffinityTerm: podAffinityTerm,
+			},
+			},
+		}}
```
FxKu: As there exists the function `generatePodAffinity`, I wondered if it can be used here. But seems, that this code here is just too different of `generatePodAffinity` is too specific. Pitty.

manifests/operator-service-account-rbac.yaml
```diff
@@ -142,7 +142,16 @@ rules:
...
+- apiGroups:
+  - batch
+  resources:
+  - cronjobs # enables logical backups
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
```
FxKu: update, too?

FxKu: Remember to add this in the helm chart template as well. Same for other parameters.

sdudoladov: > Remember to add this in the helm chart template as well. 

i've already them to `values.yaml`

pkg/cluster/k8sres.go
sdudoladov: you are right: despite its name, `generatePodAffinity` generates anti-affinity settings.

manifests/postgresql-operator-default-configuration.yaml
sdudoladov: > why is it called "_key"?

by mistake, will fix

> I'm missing the enable_logical_backup flag here. Same for configMap

There is no such key, because that particular feature is supposed  to be triggered by users only on very few selected PG clusters  ; it hardly makes sense to enable it for all DBs within a k8s cluster.

sdudoladov:
:+1:

manifests/operator-service-account-rbac.yaml
FxKu: referring to [this](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/templates/clusterrole.yaml) template file

FxKu:
:+1:

sdudoladov:
:+1:

PR #444, opened 2019-01-09
Not being able to register CRD is not a fatal error

redbaron:
Operator proceeds to checking if CRD is present and ready,
and if not, only then it is a fatal error.

### DIFF
--- a/pkg/controller/util.go
+++ b/pkg/controller/util.go
@@ -51,17 +51,18 @@ func (c *Controller) clusterWorkerID(clusterName spec.NamespacedName) uint32 {
 
 func (c *Controller) createOperatorCRD(crd *apiextv1beta1.CustomResourceDefinition) error {
 	if _, err := c.KubeClient.CustomResourceDefinitions().Create(crd); err != nil {
-		if !k8sutil.ResourceAlreadyExists(err) {
-			return fmt.Errorf("could not create customResourceDefinition: %v", err)
-		}
-		c.logger.Infof("customResourceDefinition %q is already registered and will only be updated", crd.Name)
+		if k8sutil.ResourceAlreadyExists(err) {
+			c.logger.Infof("customResourceDefinition %q is already registered and will only be updated", crd.Name)
 
-		patch, err := json.Marshal(crd)
-		if err != nil {
-			return fmt.Errorf("could not marshal new customResourceDefintion: %v", err)
-		}
-		if _, err := c.KubeClient.CustomResourceDefinitions().Patch(crd.Name, types.MergePatchType, patch); err != nil {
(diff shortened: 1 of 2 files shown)


zerg-junior:
@redbaron at the absence of the operator CRD, at which point do you expect the operator pod to fail ?

redbaron:
It fails on CRD wait status then , which remains Fatal error

```
time="2019-01-09T14:53:00Z" level=warning msg="Can't register customResourceDefinition operatorconfigurations.acid.zalan.do, will continue if it already exists" pkg=controller
time="2019-01-09T14:53:04Z" level=fatal msg="could not register Operator Configuration CustomResourceDefinition: customresourcedefinitions.apiextensions.k8s.io \"operatorconfigurations.acid.zalan.do\" not found" pkg=controller
```

Jan-M:
But isn't the message confusing, why will it continue to fail if it already exists?

redbaron:
> But isn't the message confusing, why will it continue to fail if it already exists?

I don't understand. If CRD already exists, then  irrespectively of CRD creation it moves to checking CRD status and only if CRD status check fails, operator fails too.

redbaron:
In other words, it fails on what is important : CRD is not registered in the cluster

pkg/controller/util.go
```diff
@@ -49,10 +49,11 @@ func (c *Controller) clusterWorkerID(clusterName spec.NamespacedName) uint32 {
 
 func (c *Controller) createOperatorCRD(crd *apiextv1beta1.CustomResourceDefinition) error {
 	if _, err := c.KubeClient.CustomResourceDefinitions().Create(crd); err != nil {
-		if !k8sutil.ResourceAlreadyExists(err) {
-			return fmt.Errorf("could not create customResourceDefinition: %v", err)
+		if k8sutil.ResourceAlreadyExists(err) {
+			c.logger.Infof("customResourceDefinition %q is already registered", crd.Name)
+		} else {
+			c.logger.Warnf("Can't register customResourceDefinition %s, will continue if it already exists", crd.Name)
```
sdudoladov: 1. Please log the error for the ease of debugging as the original code does

sdudoladov:
hey @redbaron :) any updates on this PR ?

pkg/controller/util.go
redbaron: done

FxKu:
Thanks for the update. With the new update mechanism there are also no confusing message anymore.

FxKu:
:+1:

sdudoladov:
:+1:

sdudoladov:
@redbaron thanks for the contribution :+1:

PR #445, opened 2019-01-09
Allow specifying init_containers in Postgres CRD

redbaron:


Jan-M:
1) Can you elaborate a bit more on what is your use case? As a lot of the bootstrapping can be hidden in the Docker image and should probably honor the leader election?

2) There should probably be a operator config option to enable/allow this or maybe whitelist Docker images to use in init containers? I dont think everyhone should just be allowed to spec this, even if for example the K8S infrastructure already blocks docker hub.

redbaron:
Sure,

> Can you elaborate a bit more on what is your use case? As a lot of the bootstrapping can be hidden in the Docker image and should probably honor the leader election?

In our case it is how we enable encryption at rest, it comes in the form of init container which is added to all apps with PVCs by our CI pipeline as init container and prepares mount point for application.  I think there can be other uses which supposed to be generic and therefore does not fit well with custom image rebuild for every app.

> There should probably be a operator config option to enable/allow this or maybe whitelist Docker images to use in init containers? I dont think everyhone should just be allowed to spec this, even if for example the K8S infrastructure already blocks docker hub.

I just followed Sidecar example, but allowed more flexibibility. As for restrictions, I am not opposed to adding it, but IMHO any form of restriction to be effective ought to be enforced at Kubernetes level in the form of validating  admission controller.

sdudoladov:
@CyberDem0n what is your opinion on the restriction we need to impose on this feature ?

redbaron:
I can, but how it is conceptually different from side containers which are not gated by operator  config?

sdudoladov:
> I can, but how it is conceptually different from side containers which are not gated by operator config?

True, there is no much difference. I'll merge this PR and create an umbrella issue to implement proper turning on/off of custom containers.

Thank you for the contribution :)

PR #452, opened 2019-01-11
Support subPath in generated container

redbaron:
This is something which is useful together https://github.com/zalando-incubator/postgres-operator/pull/445 or sidecars when they get support of arbitrary volume mounts.

These containers might want to use volume for their own need, but it should not interfere with Spilo/Patroni container, so safest way is to use subPath in Patroni container.

This change is completely transparent for Patroni, paths in container remain the same and nothing changes from it's point of view.

### DIFF
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -269,6 +269,9 @@ properties of the persistent storage that stores postgres data.
   documentation](https://kubernetes.io/docs/concepts/storage/storage-classes/)
   for the details on storage classes. Optional.
 
+* **subPath**
+  Subpath to use when mounting volume into Spilo container
+
 ### Sidecar definitions
 
 Those parameters are defined under the `sidecars` key. They consist of a list
--- a/pkg/apis/acid.zalan.do/v1/postgresql_type.go
+++ b/pkg/apis/acid.zalan.do/v1/postgresql_type.go
@@ -82,6 +82,7 @@ type MaintenanceWindow struct {
 type Volume struct {
 	Size         string `json:"size"`
 	StorageClass string `json:"storageClass"`
+	SubPath      string `json:"subPath,omitempty"`
 }
 
 // PostgresqlParam describes PostgreSQL version and pairs of configuration parameter name - values.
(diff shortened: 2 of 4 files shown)


sdudoladov:
@erthalion can you please have a look into this PR ?

Jan-M:
https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath-with-expanded-environment-variables

Is this an issue for clusters without this feature gate in 1.11 and how is the omission treated?

redbaron:
@Jan-M,  subPath expansion is the feature of the cluster, if it is enabled then every subpath is subject to expansion logic similar to one done for args and environment variables. subPath from this patch is used directly as-is in container's volume mounts subPath. subPath itself is not feature-gated and available since forever , only variable expansion in subpath value is subject to that feature gate.

FxKu:
@redbaron could you reflect your change also in [util_test](https://github.com/zalando/postgres-operator/blob/67b2fc40c027b0674bd77d9b1a5039b918839f17/pkg/apis/acid.zalan.do/v1/util_test.go#L135)?

Instead of `"volume":{"size":"","storageClass":""}` it has to be `"volume":{"size":"","storageClass":"", "subPath":""}`

Or you make subPath omitempty in the type declaration (which I would prefer): 
`SubPath string `json:"subPath,omitempty"`

redbaron:
@FxKu, rebased  and added `omitempty` tag

FxKu:
Thanks for updating :)
Looks fine for me now, although one of the tests in `util_test` could use an example for `subPath`

erthalion:
> rebased and added omitempty tag

Indeed, thanks!

> Looks fine for me now, although one of the tests in util_test could use an example for subPath

Agree, the PR looks good, but I would also ask for a simple unit test.

redbaron:
@FxKu, @erthalion , updated one test with subpath value

FxKu:
:+1:

redbaron:
Happy Monday! would be great to get it into 1.2

Jan-M:
:+1:

PR #461, opened 2019-01-23
Added possibility to add custom annotations to LoadBalancer service.

ermajn:
LoadBalancer services in bare metal configuration usually requires some annotations to overcome problem assigning external IP. This should help with that.

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -294,6 +294,11 @@ In the CRD-based configuration they are grouped under the `load_balancer` key.
   cluster.  Can be overridden by individual cluster settings. The default is
   `false`.
 
+* **custom_service_annotations**
+  when load balancing is enabled, LoadBalancer service is created and
+  this parameter takes service annotations that are applied to service.
+  Optional.
+
 * **master_dns_name_format** defines the DNS name string template for the
   master load balancer cluster.  The default is
   `{cluster}.{team}.{hostedzone}`, where `{cluster}` is replaced by the cluster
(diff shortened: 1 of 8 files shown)


docs/reference/operator_parameters.md
```diff
@@ -294,6 +294,11 @@ In the CRD-based configuration they are grouped under the `load_balancer` key.
   cluster.  Can be overridden by individual cluster settings. The default is
   `false`.
 
+* **custom_service_annotations**
+  when load balancing is enabled, LoadBalancer service is created and
+  this parameter takes service annotations that are applied to service.
+  Optional.
```
sdudoladov: is this applied to both replica and master LBs if they are enabled ?

pkg/cluster/k8sres.go
```diff
@@ -1093,6 +1093,13 @@ func (c *Cluster) generateService(role PostgresRole, spec *acidv1.PostgresSpec)
 			constants.ZalandoDNSNameAnnotation: dnsName,
 			constants.ElbTimeoutAnnotationName: constants.ElbTimeoutAnnotationValue,
 		}
+
+		if len(c.OpConfig.CustomServiceAnnotations) != 0 {
+			c.logger.Debugf("There are custom annotations defined, creating them.")
```
sdudoladov: should not these annotations be applied only when the service of type LB is created ?

ermajn: yes, and considering line 1090 annotations are applied only in case of LB type.

docs/reference/operator_parameters.md
ermajn: yes, that is correct, to both replica and master, but only LB type.

PR #462, opened 2019-01-28
Fix run operator locally

sdudoladov:
1. Bump up Spilo and operator images
2. Run `acid-minimal-cluster` in the `default` namespace, add an option to run it in `test`
3. Fix the command for starting minikube (caused by changes in its API)
4. Add an option to replace the operator image without restarting the minikube; that greatly speeds up local testing as minikube restarts may take minutes.

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -10,7 +10,7 @@ data:
 
   debug_logging: "true"
   workers: "4"
-  docker_image: registry.opensource.zalan.do/acid/spilo-cdp-11:1.5-p42
+  docker_image: registry.opensource.zalan.do/acid/spilo-11:1.5-p4
   pod_service_account_name: "zalando-postgres-operator"
   secret_name_template: '{username}.{cluster}.credentials'
   super_username: postgres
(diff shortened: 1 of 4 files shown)


manifests/configmap.yaml
```diff
@@ -10,7 +10,7 @@ data:
 
   debug_logging: "true"
   workers: "4"
-  docker_image: registry.opensource.zalan.do/acid/spilo-cdp-11:1.5-p42
+  docker_image: registry.opensource.zalan.do/acid/spilo-11:1.5-p4
```
redbaron: @sdudoladov , just curious what is the difference between `spilo-cdp-11` and `spilo-11`?

CyberDem0n: `spilo-cdp` is build and pushed by ci/cd system on every commit to the master branch.
Later it is tested and if everything is fine we are tagging it as a `spilo`.

redbaron: is `spilll-11` aligned with Patroni releases or it is for every commit as well, just after test?

PR #463, opened 2019-01-28
Retry moving master pods

### LINKED ISSUE #429: Ensure operator moves pods from a decommisioned nodes
The operator supposedly gets stuck when moving master pods form a decommissioned node.
We need to reproduce and investigate this situation, which cannot be done in `minikube` cause `minikube` is a single node k8s cluster. 

one situation where it is reported to happen is the decommission of multiple nodes. The operator apparently ignores a few decommission nodes after removing master pods / finding no pods to remove from the 1st one

The current solution is simply to restart the operator pod.

sdudoladov:
introduces retries to moving master pods away from decommissioned nodes. Master-replica switchovers will be retried with some delay that gives time for a healthy replica to start up.

partially fixes #429


redbaron:
I am curious if this manual pod move still necessary? Kubernetes has taint based eviction nowdays, so that pods on not ready nodes are moved after 5 minutes by default which is configurable:

```
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
```

Jan-M:
I am not sure, I am answering the right question.

We do a "manual" move of pods in the postgres operator to reduce the number of possible failovers during a rolling upgrade to exactly 1. In our cloud infrastructure and possible others, where all nodes are completely replaced you need to make 1 failover from a master on an old node to a replica on a new node, otherwise if you failover to a replica on an old node you need another failover later.

Thus we make sure to move a replica first (delete pod on an old node, comes back on new node).

pkg/controller/node.go
```diff
@@ -38,9 +41,22 @@ func (c *Controller) nodeAdd(obj interface{}) {
...
+			func() (bool, error) {
+				err := c.moveMasterPodsOffNode(node)
+				if  err != nil {
+					return false, fmt.Errorf("unable to move master pods off the unschedulable node; will retry after delay")
+				}
+				return true, nil
+			} )
+
+		if err != nil {
+			c.logger.Warning("failed to move master pods from the node %q: timeout expired", node.Name)
```
Jan-M: Let's add time the time limit here in the logs

pkg/util/config/config.go
```diff
@@ -80,6 +80,7 @@ type Config struct {
 	// value of this string must be valid JSON or YAML; see initPodServiceAccount
 	PodServiceAccountDefinition            string `name:"pod_service_account_definition" default:""`
 	PodServiceAccountRoleBindingDefinition string `name:"pod_service_account_role_binding_definition" default:""`
+	MasterPodMoveTimeout                   time.Duration `name:"master_pod_move_timeout" default:"10m"`
```
Jan-M: Sergey, let's bump this to 20min.

Jan-M:
@erthalion please have a look, too.

pkg/controller/node.go
```diff
@@ -83,7 +90,7 @@ func (c *Controller) moveMasterPodsOffNode(node *v1.Node) {
 	podList, err := c.KubeClient.Pods(c.opConfig.WatchedNamespace).List(opts)
 	if err != nil {
 		c.logger.Errorf("could not fetch list of the pods: %v", err)
-		return
+		return err
```
erthalion: If I understand correctly, `attemptToMoveMasterPodsOffNode` will be retried only if returned an error, i.e. only if we're on this line. The rest of the errors, that could happen (including switchover itself), are sort of hidden in `range masterPods` loop. Is that intended behaviour?

sdudoladov: > `attemptToMoveMasterPodsOffNode` will be retried only if returned an error

yes.

> The rest of the errors, that could happen (including switchover itself), are sort of hidden in `range masterPods` loop

this PR also introduces the following change 
> 	if leftPods := totalPods - movedPods; leftPods > 0 {
		       return fmt.Errorf("could not move master %d/%d pods from the %q node",
			leftPods, totalPods, nodeName)
	}

so whatever goes wrong is supposed to be caught here

docs/reference/operator_parameters.md
```diff
@@ -212,6 +212,11 @@ configuration they are grouped under the `kubernetes` key.
   class](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/#priorityclass)
   that should be assigned to the Postgres pods. The priority class itself must be defined in advance.
   Default is empty (use the default priority class).
+  
+ * **master_pod_move_timeout**
+   The period of time to wait for the success of Patroni switchovers from master pods on an unschedulable node
```
erthalion: Please make it more explicit, that it includes not only switchover, but the whole pod migration procedure.

pkg/controller/node.go
```diff
@@ -161,3 +170,21 @@ func (c *Controller) nodeDelete(obj interface{}) {
...
+			err := c.attemptToMoveMasterPodsOffNode(node)
+			if err != nil {
+				return false, fmt.Errorf("unable to move master pods off the unschedulable node; will retry after delay of 1 minute")
+			}
+			return true, nil
+		},
+	)
+
+	if err != nil {
+		c.logger.Warning("failed to move master pods from the node %q: timeout of %v minutes expired", node.Name, c.opConfig.MasterPodMoveTimeout)
```
erthalion: Just noticed, that here should be `Warningf`

Issue #465, opened 2019-01-29
Investigate taint-based eviction for pod migration


sdudoladov:
The k8s v1.13 brings taint-based eviction to the beta stage and *enables it by default*. We need to investigate if
1.  at the bare minimum this mechanism does not affect master pod migration
2. if we can rely solely on this mechanism to migrate master pods away from decommissioned nodes.

related issues:
#429  #463

redbaron:
I think whole work can be deferred to the Kubernetes and Patroni  like like following:

1. Patroni  updates  `tolerationSeconds` of a master pod for desired key, such as `node.kubernetes.io/unreachable` to be lower than replica's 

2. Patroni penalizes switchover candidates which are on tainted nodes

Net effect is that there is  only 1 switchover if situation in cluster allows it and as a bonus all failover/switchover decision is situated within Patroni, with no interference from Postgres operator (operator currently selects replicas to promote without taking into account their lag or other factors such as patroni tags, which doesn't look safe to me)

redbaron:
on a second part: it is possible to add tolerations to any key of effect "NoExecute" 

```
  - effect: NoSchedule
    operator: Exists
     tolerationSeconds: 600
```

sdudoladov:
I briefly documented the taint-based eviction; on our infra we did not run into problems with it.

PR #468, opened 2019-01-29
Document taint-based eviction

sdudoladov:
based on #465


docs/administrator.md
```diff
@@ -146,6 +146,11 @@ data:
   ...
 ```
 
+Note that the Kubernetes version 1.13 brings [taint-based eviction](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/#taint-based-evictions) to the beta stage and enables it by default.
+Postgres pods by default receive tolerations for `unreachable` and `noExecute` taints with the timeout of `5m`.
+Depending on you setup, you may want to adjust these parameters to prevent master pods from being evicted by Kubernetes runtime.
```
FxKu: your setup

FxKu: by the Kubernetes runtime

docs/administrator.md
```diff
@@ -146,6 +146,11 @@ data:
...
+To prevent eviction completely, specify the toleration without specifying the `tolerationSeconds` value (similar to how Kubernetes own DaemonSets are configured)
```
FxKu: specify the toleration by leaving out the ...

Kubernetes' own ...

sdudoladov:
@FxKu I updated the PR. Thanks for the review !

PR #477, opened 2019-02-01
Pass PodEnvironmentConfigMap to results

kupson:


erthalion:
Yes, I can confirm the issue, and after a short testing I don't see any drawbacks of proposed change.

PR #478, opened 2019-02-04
Add global option to enable/disable init containers and sidecars

### LINKED ISSUE #464: Optionally enable init/sidecar containers
The operator currently support running sidecar and init containers side-by-side with Postgres containers. For maintenance reasons , we would like to be able to enable/disable these features per k8s cluster. The task is to:
1. Add relevant toggles to the operator configuration (CRD and configmap) similar to existing boolean options such as `enable_shm_volume`. 
2. Log a warning if a user attempts to use init/sidecars containers when this feature is disabled.

FxKu:
Closes #464.
Error out on `createStatefulSet`, otherwise only warn whenever `generateStatefulSet` is called


pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -145,6 +145,8 @@ type OperatorConfigurationData struct {
 	ResyncPeriod               Duration                     `json:"resync_period,omitempty"`
 	RepairPeriod               Duration                     `json:"repair_period,omitempty"`
 	Sidecars                   map[string]string            `json:"sidecar_docker_images,omitempty"`
+	EnableSidecars             bool                         `json:"enable_sidecars"`
```
redbaron: I think it should be a `*bool*` for smooth upgrade path, otherwise you won't be tell apart if field is not specifies vs field is set to false

pkg/cluster/k8sres.go
```diff
@@ -735,6 +736,10 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*v1beta1.State
 		return nil, fmt.Errorf("could not generate resource requirements: %v", err)
 	}
 
+	if c.OpConfig.EnableInitContainers {
```
sdudoladov: The situation where the spec for init containers is provided, but the option itself is disabled, is worth logging a warning.

pkg/cluster/k8sres.go
```diff
@@ -792,9 +797,11 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*v1beta1.State
 	}
 
 	// generate sidecar containers
-	if sidecarContainers, err = generateSidecarContainers(sideCars, volumeMounts, defaultResources,
-		c.OpConfig.SuperUsername, c.credentialSecretName(c.OpConfig.SuperUsername), c.logger); err != nil {
-		return nil, fmt.Errorf("could not generate sidecar containers: %v", err)
+	if c.OpConfig.EnableSidecars {
```
sdudoladov: same for standard sidecards

pkg/cluster/k8sres.go
```diff
@@ -763,6 +764,14 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*v1beta1.State
 		return nil, fmt.Errorf("could not generate resource requirements: %v", err)
 	}
 
+	if spec.InitContainers != nil {
+		if c.OpConfig.EnableInitContainers {
+			initContainers = spec.InitContainers
+		} else {
+			c.logger.Warningf("InitContainers specified but globally disabled!")
```
redbaron: IMHO it is an example of invalid manifest and it should be rejected, instead of accepted to processing but with fine print in logs that no init containers are to be created. when users create PG object and see `status: Running` in it they assume it is all good and ready, but it is not what they asked

Jan-M:
How do init containers work on scaling out for example or in? Is changing init containers reflected correctly in the manifest? (does it make sense to react on changes of init container on a running cluster)

But I agree with the above comment, if init is specced in manifest and globally disabled cluster should probably not be created. Modifying an existing one is something else.

redbaron:
> How do init containers work on scaling out for example or in? Is changing init containers reflected correctly in the manifest? (does it make sense to react on changes of init container on a running cluster)

same as any other change to Stateful set's pod spec, it triggers graceful roll of all pods to new version of spec

Jan-M:
Side car I get, but what is the value of having init containers added once the statefulset is already up and runing (I guess I need to read a bit more of K8S docs here). Maybe mixing a bit of "init of postgres" with "init of container environment"

docs/reference/operator_parameters.md
```diff
@@ -80,10 +80,18 @@ Those are top-level keys, containing both leaf keys and groups.
   your own Spilo image from the [github
   repository](https://github.com/zalando/spilo).
 
+* **enable_init_containers**
+  global option to allow for creating init containers to run actions before
+  Spilo is started. Disabled by default.
```
sdudoladov: i am not so sure about the default here. In all likelihood there are people there running the operator with init/sidecars. If we silently disable the features, we may break their deployments.

FxKu: Set the default to true now.

pkg/cluster/resources.go
```diff
@@ -65,6 +65,17 @@ func (c *Cluster) listResources() error {
 
 func (c *Cluster) createStatefulSet() (*appsv1.StatefulSet, error) {
 	c.setProcessName("creating statefulset")
+	// check if it's allowed that spec contains initContainers
+	if c.Spec.InitContainers != nil && len(c.Spec.InitContainers) > 0 &&
```
sdudoladov: Does  the validation reject the manifest with empty  `initContainers` list  ?  that looks like a misconfiguration at first glance. Same for sidecar containers.

pkg/cluster/k8sres.go
```diff
@@ -872,9 +880,14 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
 	}
 
 	// generate sidecar containers
-	if sidecarContainers, err = generateSidecarContainers(sideCars, volumeMounts, defaultResources,
-		c.OpConfig.SuperUsername, c.credentialSecretName(c.OpConfig.SuperUsername), c.logger); err != nil {
-		return nil, fmt.Errorf("could not generate sidecar containers: %v", err)
+	if sideCars != nil && len(sideCars) > 0 {
+		if c.OpConfig.EnableSidecars != nil && !(*c.OpConfig.EnableSidecars) {
+			c.logger.Warningf("sidecars in use but globally disabled - next stateful set creation would fail")
```
sdudoladov: "in use" as "specified in the manifest" ?

go.sum
```diff
@@ -160,6 +160,7 @@ github.com/google/uuid v1.0.0/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+
 github.com/google/uuid v1.1.1 h1:Gkbcsh/GbpXz7lPftLA3P6TYMwjCLYm83jiFQZF/3gY=
 github.com/google/uuid v1.1.1/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+yHo=
 github.com/googleapis/gax-go/v2 v2.0.4/go.mod h1:0Wqv26UfaUD9n4G6kQubkQ+KchISgw+vpHVxEJEs9eg=
+github.com/googleapis/gnostic v0.0.0-20170426233943-68f4ded48ba9/go.mod h1:sJBsCZ4ayReDTBIg8b9dl28c5xFWyhBTVRp3pOg5EKY=
```
sdudoladov: what is the reason behind these dependency updates ?

docs/user.md
```diff
@@ -374,6 +378,10 @@ spec:
 
 `initContainers` accepts full `v1.Container` definition.
 
+**Note**: The operator will not launch a cluster if `initContainers` are
```
sdudoladov: `create` is more specific than `launch`

pkg/cluster/resources.go
FxKu: validation does not reject empty arrays

pkg/cluster/k8sres.go
FxKu: this code won't be reached on cluster creation because `createStatefulSet` would throw an error with this wording ("specified but disabled"). Therefore, during snyc and update I can assume that sidecars and initContainers are in use or running.

go.sum
FxKu: not sure, why it got introduced but updating deps won't hurt. To have a fresh list, I removed `go.mod` and `go.sum` and ran another `make deps`.

pkg/cluster/k8sres.go
sdudoladov: sure it is just that phrases "in use" and "globally" make sense for people familiar with operator internals and this feature in particular but not necessarily to those reading the logs

FxKu: updated the messages. Should be clearer now.

sdudoladov:
:+1:

FxKu:
:+1:

Issue #485, opened 2019-02-11
Enable cloning from arbitrary S3 buckets


sdudoladov:
Currently cloning a PG cluster is [restricted to work only with the S3 bucket designated for backups](https://github.com/zalando-incubator/postgres-operator/blob/3544cc90faeca5b425bdb447e65d669165301cfe/pkg/cluster/k8sres.go#L1173) in the operator conf. However there are cases when a cluster need to be cloned from another S3 bucket. The task is to extend the cloning to handle arbitrary S3 buckets specified by their full path.

sdudoladov:
closed by #487

PR #487, opened 2019-02-13
Override clone s3 bucket path

erthalion:
Add possibility to use a custom s3 bucket path for cloning a cluster
from an arbitraty bucket (e.g. from another k8s cluster).


sdudoladov:
resolves #485

pkg/cluster/k8sres.go
```diff
@@ -1169,10 +1171,25 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
 			})
 	} else {
 		// cloning with S3, find out the bucket to clone
+		cloneScope := cluster
+		cloneS3Bucket := c.OpConfig.WALES3Bucket
+		cloneSuffix := getBucketScopeSuffix(description.UID)
+
+		if description.S3WalPath != "" {
```
sdudoladov: don't we need to check that it is the valid S3 bucket name or at least log from where we are trying to clone to ease debugging?

sdudoladov:
we also need to document it including the necessary permissions' setup for S3 i am afraid

Jan-M:
For now we try to leave as much checking as possible to later stage when it comes to AWS infrastructure. Lets try not to couple it too much there, which makes me think maybe we should not call it s3 path. Dont know how it sounds and looks like on GKE.

erthalion:
> Lets try not to couple it too much there, which makes me think maybe we should not call it s3 path.

Then we need to have a flag "S3" anyway, because under hood we're transforming this path to a few spilo S3 related env variables.

pkg/cluster/k8sres.go
```diff
@@ -1169,10 +1198,25 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
...
 		// cloning with S3, find out the bucket to clone
+		cloneScope := cluster
+		cloneS3Bucket := c.OpConfig.WALES3Bucket
+		cloneSuffix := getBucketScopeSuffix(description.UID)
+
+		customVars := customCloneVars(description.S3WalPath)
+		// if the provided path seems valid, then override
+		if len(customVars) > 2 {
+			msg := "Use custom S3WalPath %s from the manifest"
+			c.logger.Warningf(msg, description.S3WalPath)
```
Jan-M: I see an issue here, unfortunately. This would block migrations towards K8S from non matching S3 paths? Where e.g. we do not have the uuid?

TeutoNet:
@erthalion for the issue #494 I've created a new pull request #497 to cover the settings required for cloning from a S3-compatible service.

erthalion:
@TeutoNet FYI: we agreed with @CyberDem0n that he will adjust spilo to also accept a custom bucket path, if provided. This will simplify this particular PR, and probably help in your case too.

Jan-M:
@CyberDem0n @erthalion how to proceed here?

pkg/cluster/k8sres.go
```diff
@@ -1169,10 +1170,35 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
...
+				},
+			}
+
+			result = append(result, envs...)
+		} else {
+			msg := "Use custom parsed S3WalPath %s from the manifest"
+			c.logger.Warningf(msg, description.S3WalPath)
+
+			result = append(result, v1.EnvVar{
+				Name:  "CLONE_WAL_S3_PREFIX",
```
CyberDem0n: It must be called `CLONE_WALE_S3_PREFIX`

Jan-M:
:+1:

erthalion:
:+1:

Issue #479, opened 2019-02-04
allow full pod spec (or maybe even statefulset spec)


redbaron:
Operator config already has init containers, side containers, pod service account name, environment vars configmap, resources and I am pretty sure more to come (for instance tolerations, DNSConfig, extra volumes/volume mounts).
 
It wont stop until all possible pod features will be reexported as arbitrary DIY  params , so maybe it is worth to allow full podSpec to be specified in operator config?  It then be used as a base and overlayed with any changes required by the runtime (like adding  environment variables with created credentials secrets or spilo configuration environment) when creating a statefulset.

redbaron:
@Jan-M , I can work on implementation if idea is generally accepted and we discuss design at least ot the high level

erthalion:
I agree, it sounds very appealing, and I myself missed this for testing purposes many times. But before jumping on it I guess we need to think through the possible use cases. As I see there are three possible options to configure pods:

* use only operator provided options (I guess, it's a most used option)
* use a part of a full podSpec and merge it with the rest of the options
* use a full podSpec to fully customize pods

And even taking a simplest possible approach (i.e. without a second case) I think it will require significant restructure of every code, which deals with configuration, to not make it looks like a hack. Another point is that it will make validation more complicated. At the end I think it makes sense to split this and tackle every question separately.

redbaron:
@erthalion ,my proposal is to have full podSpec specified, then operator unconditionally overlays any necessary modifications on top of it.  If no podSpec is set, then operator works as it is now.

Then there is choice what to do with existing DIY customizations when podSpec is set, as you outlined there are really options:

- merge with podSpec
- reject CRD as one having conflicting requirements
- silently ignore with warning in logs

I prefer second option as it cleanest and least surprising: anyone who wants full control over pod have to give up any shortcuts and specify it in full.   Any attempt to reconcile and merge old confgs with podSpec would raise too many questions with too few good answers, such as:

- does side_containers append, replace or replace all but 1 containers in podSpec?
- which scalar values wins? `postgres.podSpec.containers[0].image` or `posgres.dockerImage` ?

hedgss:
It is a really useful feature that covers a lot of cases when someone needs a custom configuration.
In my case, I need to be sure that Kubernetes pod uses storage in the same datacenter.
At this moment it is not possible to implement.
How long will this topic be in discussion state?
What is required to move this feature on the next stages?

paalkr:
[...]

PR #489, opened 2019-02-15
Add Pod Anti Affinity

TeutoNet:
Add pod anti affinity to enforce that postgres pods does not spawn on the same k8s node.


redbaron:
related https://github.com/zalando-incubator/postgres-operator/issues/479

redbaron:
In some cases it is OK to have 2 PG pods on the same node, for instance when testing switchover/failover

TeutoNet:
@redbaron You're right with #479, but it is a big change and it will take some time to get it done. Are there any timelines regarding this issue?

For production DBs it would be a worst case scenario to have 2 PG pods on the same node, once the node dies.

pkg/cluster/k8sres.go
```diff
@@ -290,6 +290,26 @@ func nodeAffinity(nodeReadinessLabel map[string]string) *v1.Affinity {
 	}
 }
 
+func generatePodAffinity(team string, version string) *v1.Affinity {
+	// generate pod anti affinity to avoid multiple on instances on the same node
+	matchLabels := make(map[string]string)
+
+	matchLabels["application"] = "spilo"
```
redbaron: this should come from `cluster_labels`  parameter in operator configuration

pkg/cluster/k8sres.go
```diff
@@ -437,9 +457,7 @@ func generatePodTemplate(
 		addShmVolume(&podSpec)
 	}
 
-	if nodeAffinity != nil {
-		podSpec.Affinity = nodeAffinity
```
redbaron: this changes previous behaviour where node affinity was set when node readiness labels are configured in operator configuration

pkg/cluster/k8sres.go
TeutoNet: Ok, I'll try to change it to use the `cluster_labels` instead the hardcoded once.

pkg/cluster/k8sres.go
TeutoNet: I've tried to respect the `node_readiness_label` with the pod anti affinity, but the operator crashed with nil pointer.

Jan-M:
This def. needs a toggle, maybe incl. a toggle per cluster.

While initially this may sound bad for availability it all depends.

In our K8S infrastructure on AWS this is also largely not happening for a single Postgres cluster as pods are bound to different AZs via the different EBS volumes.

But if your storage is EBS it may sometimes in a single AZ not matter so much if pods are on the same host.

Regarding #479 I am looking forward to ideas. In our initial impl. in Python this was somehow neater having the statefulset spec just injected as a String template.

TeutoNet:
@Jan-M I think it's a good idea to have a switch in the cluster config to enable pod anti affinity and a addtional property to set the pod affinity `TopologyKey`, default could be `kubernetes.io/hostname`.

pkg/cluster/k8sres.go
TeutoNet: @redbaron I've added the `node_readiness_label` to the pod anti affinity.

pkg/cluster/k8sres.go
TeutoNet: I've changed the code to use the `cluster_labels`.

TeutoNet:
@Jan-M Switch for pod anti affinity is implemented as well as the `TopologyKey`.

pkg/cluster/k8sres.go
```diff
@@ -290,6 +290,26 @@ func nodeAffinity(nodeReadinessLabel map[string]string) *v1.Affinity {
 	}
 }
 
+func generatePodAffinity(labels labels.Set, topologyKey string, nodeAffinity *v1.Affinity) *v1.Affinity {
+	// generate pod anti affinity to avoid multiple on instances on the same node
```
sdudoladov: pls change to sth like

> generate pod anti-affinity to avoid multiple pods of the same Postgres cluster on the same node

to make it a little bit more clear

TeutoNet: I've changed the comment as requested and also added documentation to admin docs and operator params reference.

CRD-based operator configuration is missing the parameters as I wasn't able to incorporate the changes.

TeutoNet: Implemented propagation of the pod anti affinity parameters to/from CRD-based operator configuration.

sdudoladov:
lgtm. thank you, @ocaner-biz

Issue #494, opened 2019-02-21
Inconsistency in S3 settings between Operator Configuration and Cluster Configuration


TeutoNet:
In the Operator Configuration you can specify a global s3 bucket for WAL files `wal_s3_bucket` and log files `log_s3_bucket`.

In the Spilo Env ConfigMap used to create Postgres Cluster  you can define the following additional parameters like:
* `AWS_ACCESS_KEY_ID`
* `AWS_SECRET_ACCESS_KEY`
* `WALE_S3_ENDPOINT`
* `AWS_S3_FORCE_PATH_STYLE`
* `AWS_INSTANCE_PROFILE`

This inconsistency causes issues once you try to clone a cluster from a s3 bucket.
The WAL files of the source database is stored in a custom s3 storage server (RADOS Gateway), but the destination instance tries to fetch the WAL files from AWS S3 storage server.

Currently I can't find any setting to propagate the required env variables for cloning.

sdudoladov:
related #487 .
@erthalion pls have a look

Jan-M:
@CyberDem0n probably the best one to answer the different spilo and clone vars?

CyberDem0n:
I think [this](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L543-L561) block of code could answer most of your questions.

Basically most of the env variables mentioned there could have `CLONE_` prefix and spilo will use them.

TeutoNet:
@CyberDem0n yeah you're right, but the environment variables need to be pre-populated by the postgres operator otherwise clone from non AWS S3 Storage won't work.

TeutoNet:
@CyberDem0n once I've added the missing `CLONE_` env variables to postgres operator I hit another issue, because the spilo image `registry.opensource.zalan.do/acid/spilo-cdp-11:1.5-p42` I'm using is missing the fixes from zalando/spilo#292 . 

Changed spilo image to `registry.opensource.zalan.do/acid/spilo-cdp-11:latest` and recovery works, but now backup/wal file uploads are broken, as the env variables `WALG_S3_SSE` is set to `AES256` which enables encrypted uploads and expects that an encryption key is defined:

```
[...]

PR #497, opened 2019-02-22
Add additional S3 settings for cloning

TeutoNet:
In order to enable cloning from non AWS S3 service additional settings are required, see #494 .


TeutoNet:
Is it possible that the CI is broken, because the upstream project moved from `zalando-incubator` to `zalando`?

docs/user.md
```diff
@@ -242,6 +242,25 @@ metadata:
...
+spec:
+  clone:
+    uid: "efd12e58-5786-11e8-b5a7-06148230260c"
+    cluster: "acid-batman"
+    timestamp: "2017-12-19T12:40:33+01:00"
+    s3_endpoint: https://s3.acme.org
+    s3_access_key_id: 0123456789abcdef0123456789abcdef
+    s3_secret_access_key: 0123456789abcdef0123456789abcdef
+    s3_force_path_style: true
+    aws_instance_profile: "0"
```
CyberDem0n: You don't need to propagate aws_instance_profile, it is automatically calculated in spilo based on the presence of s3_access_key_id and s3_secret_access_key

TeutoNet: Ok, should I remove the setting at all?

erthalion:
> Is it possible that the CI is broken, because the upstream project moved from `zalando-incubator` to `zalando`?

Most likely, we'll fix it shortly.

docs/user.md
TeutoNet: @CyberDem0n  I've removed the mentioned `aws_instance_profile`.

Jan-M:
@sdudoladov @CyberDem0n how to progress here? Is there anything in Spilo will make it simpler?

CyberDem0n:
There are a [bunch](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L595-L607) of env variables is Spilo to support backup and PITR with wal-e/wal-g.
Most of them could also be specified with `CLONE_` or `STANDBY_` prefix.

The list of settings might look huge, but unfortunately we can't make it smaller and I am pretty sure there are some settings already missing due to some changes in wal-g.

abh:
We don't use AWS S3, but have a compatible blob store -- this change seems to do exactly what we need...

TeutoNet:
I've updated my PR branch to include latest changes from master.

Jan-M:
@FxKu @sdudoladov looks good to me, if this helps with non aws non s3 setups this needs to be exposed.

sdudoladov:
:+1:

FxKu:
:+1:

FxKu:
Thanks @TeutoNet for your contribution.

PR #500, opened 2019-02-26
Update README.md

sdudoladov:
Include info about GSoC 2019 participation

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@
 
 <img src="docs/diagrams/logo.png" width="200">
 
+# Google Summer of Code
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,6 +8,9 @@
 
 <img src="docs/diagrams/logo.png" width="200">
 
+# Google Summer of Code
```
FxKu: maybe ## ? It looks a bit too prominent there at the top. We could also think of adding the GSoC logo: https://developers.google.com/open-source/gsoc/resources/marketing

Maybe use the logo + Link in the top of the README and the text from this PR somewhere else (either below or a new file?). Do you think, one could get the impression that the Postgres Operator is "only" a GSoC ( / students) project after all?

PR #502, opened 2019-02-28
test: add NewFromMap and Validate func

### LINKED ISSUE #416: Add unit tests to config util functions
The task is to add unit tests to the `validate()` and `newFromMap()` functions in [`/pkg/util/config/config.go`](https://github.com/zalando-incubator/postgres-operator/blob/master/pkg/util/config/config.go)

ZYecho:
as the title describe, also change some inappropriate code style.
fix #416

### DIFF
--- a/pkg/spec/types.go
+++ b/pkg/spec/types.go
@@ -37,7 +37,7 @@ type syncUserOperation int
 // Possible values for the sync user operation (removal of users is not supported yet)
 const (
 	PGSyncUserAdd = iota
-	PGsyncUserAlter
+	PGSyncUserAlter
 	PGSyncAlterSet // handle ALTER ROLE SET parameter = value
 )
 
--- a/pkg/util/config/config.go
+++ b/pkg/util/config/config.go
@@ -79,25 +79,25 @@ type Config struct {
 	// default name `operator` enables backward compatibility with the older ServiceAccountName field
 	PodServiceAccountName string `name:"pod_service_account_name" default:"operator"`
 	// value of this string must be valid JSON or YAML; see initPodServiceAccount
-	PodServiceAccountDefinition            string `name:"pod_service_account_definition" default:""`
-	PodServiceAccountRoleBindingDefinition string `name:"pod_service_account_role_binding_definition" default:""`
-	MasterPodMoveTimeout                   time.Duration `name:"master_pod_move_timeout" default:"20m"`
-	DbHostedZone                           string `name:"db_hosted_zone" default:"db.example.com"`
-	AWSRegion                              string `name:"aws_region" default:"eu-central-1"`
-	WALES3Bucket                           string `name:"wal_s3_bucket"`
-	LogS3Bucket                            string `name:"log_s3_bucket"`
-	KubeIAMRole                            string `name:"kube_iam_role"`
-	DebugLogging                           bool   `name:"debug_logging" default:"true"`
-	EnableDBAccess                         bool   `name:"enable_database_access" default:"true"`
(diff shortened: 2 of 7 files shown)


Jan-M:
Thanks for your PR, looks good to me.

@sdudoladov

erthalion:
@ZYecho  Thanks for the PR. Unfortunately, as you can see from Travis build, test doesn't pass:
```
Unable to detect operator namespace from within its pod due to:
open /var/run/secrets/kubernetes.io/serviceaccount/namespace: no such 
```

It happens because operator has `oauth_token_secret_name` in the config and tries to resolve namespace for it. Probably you can either remove it from the config, or somehow mock namespace functionality.

ZYecho:
> @ZYecho Thanks for the PR. Unfortunately, as you can see from Travis build, test doesn't pass:
> 
> ```
> Unable to detect operator namespace from within its pod due to:
> open /var/run/secrets/kubernetes.io/serviceaccount/namespace: no such 
> ```
> It happens because operator has `oauth_token_secret_name` in the config and tries to resolve namespace for it. Probably you can either remove it from the config, or somehow mock namespace functionality.

you mean 
`oauth_token_secret_name` not need now in the `config` struct?

erthalion:
> you mean `oauth_token_secret_name` not need now in the `config` struct?

I mean you can try to avoid it to reduce the scope of the test, or mock namespace functionality - otherwise you're test is always red. Please note, that a config values in `NewFromMap` are taken via reflection, taking into account corresponding default values - so you're not just passing them inside via function arguments.

ZYecho:
> > you mean `oauth_token_secret_name` not need now in the `config` struct?
> 
> I mean you can try to avoid it to reduce the scope of the test, or mock namespace functionality - otherwise you're test is always red. Please note, that a config values in `NewFromMap` are taken via reflection, taking into account corresponding default values - so you're not just passing them inside via function arguments.

I change the test scope to avoid it, PTAL, thanks.

pkg/spec/types.go
```diff
@@ -37,7 +37,7 @@ type syncUserOperation int
 // Possible values for the sync user operation (removal of users is not supported yet)
 const (
 	PGSyncUserAdd = iota
-	PGsyncUserAlter
+	PGSyncUserAlter
```
erthalion: Why do you need this change, it seems unrelated, no?

pkg/util/config/config.go
```diff
@@ -163,13 +163,13 @@ func Copy(c *Config) Config {
 	return cfg
 }
 
-func validate(cfg *Config) (err error) {
+func Validate(cfg *Config) error {
```
erthalion: Why to capitalize this and `ProcessField`? They're in the same package with the tests, so as far as I remember should be visible anyway.

pkg/util/config/util_test.go
```diff
@@ -0,0 +1,60 @@
...
+		mapFields[field.Name] = field
+	}
+
+	tests := []struct {
+		name    string
+		args    map[string]string
+		wantErr bool
+	}{
+		{
+			name: "AllFields-ok",
```
erthalion: Probably it's a good idea to test `validate` and `processField` separately, but then you test pretty small part of `processField` as far as I see. Maybe it makes sense to get all the possible fields from an empty config via `structFields` (in  this way we also can test it) and then just exclude "inconvenient for testing" fields, that will require some mocking to test them.

erthalion:
@ZYecho overall looks better, thanks. But I have few commentaries/questions, please take a look at the review.

PR #504, opened 2019-02-28
log: add warn log when resource request greater than limit

### LINKED ISSUE #418: Warn if memory request > memory limit
The operator should log a warning if a PG cluster's manifest defines  memory_request > memory_limit. In that case pods will no be created. That should cover the case where a manifest contains  *no* custom memory limit *and* custom memory request higher than the *default* memory limit.

ZYecho:
add warn log when memory resource request greater than limit
fix #418

### DIFF
--- a/pkg/controller/postgresql.go
+++ b/pkg/controller/postgresql.go
@@ -333,8 +333,12 @@ func (c *Controller) warnOnDeprecatedPostgreSQLSpecParameters(spec *acidv1.Postg
 		c.logger.Warningf("Parameter %q is deprecated. Consider setting %q instead", deprecated, replacement)
 	}
 
-	noeffect := func(param string, explanation string) {
+	noEffect := func(param string, explanation string) {
 		c.logger.Warningf("Parameter %q takes no effect. %s", param, explanation)
+	}
+
+	notAllowed := func(param string, explanation string) {
+		c.logger.Warningf("Parameter %q is not allowed. %s", param, explanation)
 	}
 
 	if spec.UseLoadBalancer != nil {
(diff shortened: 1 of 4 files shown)


pkg/controller/postgresql.go
```diff
@@ -352,6 +357,13 @@ func (c *Controller) warnOnDeprecatedPostgreSQLSpecParameters(spec *acidv1.Postg
 		(spec.EnableReplicaLoadBalancer != nil || spec.EnableMasterLoadBalancer != nil) {
 		c.logger.Warnf("Both old and new load balancer parameters are present in the manifest, ignoring old ones")
 	}
+
+	memoryRequest, err1 := strconv.Atoi(spec.ResourceRequests.Memory)
+	memoryLimit, err2 := strconv.Atoi(spec.ResourceLimits.Memory)
```
sdudoladov: are you sure this will work ? 
both limits are request have the following format: 
> ResourceDescription{CPU: "10m", Memory: "50Mi"},

pkg/controller/postgresql.go
```diff
@@ -337,6 +338,10 @@ func (c *Controller) warnOnDeprecatedPostgreSQLSpecParameters(spec *acidv1.Postg
 		c.logger.Warningf("Parameter %q takes no effect. %s", param, explanation)
 	}
 
+	notAllowed := func(param string, explanation string) {
```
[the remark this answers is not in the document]
ZYecho: i didn't get what you mean

pkg/controller/postgresql.go
ZYecho: fix it.

ZYecho:
> Please also add this warning to the code that configures sidecars and init containers as they are also affected. Note that a Scalyr sidecar is configured separately.

they already use RequestIsSmallerThanLimit to judge, what else do you want?

sdudoladov:
>> Please also add this warning to the code that configures sidecars and init containers as they are also affected. Note that a Scalyr sidecar is configured separately.

>they already use RequestIsSmallerThanLimit to judge and log a warn, what else do you want?

logging a warning when a resource request as seen by the operator is greater than the limit for all the containers that an operator can potentially create. This includes up to 4 container types:
1. Postgres container - already covered by this PR
2. Scalyr sidecar
3. init containers
4. other sidecars

the whole idea is to make visible in the logs a situation where a container  cannot start because the memory request > limit . Right now finding this is the case involves running `kubectl describe` on various k8s resources. Logging such misconfiguration shall improve the operational experience for the operator users.

pkg/controller/postgresql.go
```diff
@@ -345,13 +349,18 @@ func (c *Controller) warnOnDeprecatedPostgreSQLSpecParameters(spec *acidv1.Postg
...
 	}
 
 	if (spec.UseLoadBalancer != nil || spec.ReplicaLoadBalancer != nil) &&
 		(spec.EnableReplicaLoadBalancer != nil || spec.EnableMasterLoadBalancer != nil) {
 		c.logger.Warnf("Both old and new load balancer parameters are present in the manifest, ignoring old ones")
 	}
+
+	isSmaller, err := util.RequestIsSmallerThanLimit(spec.ResourceRequests.Memory, spec.ResourceLimits.Memory)
+	if err == nil && !isSmaller {
+		notAllowed("memoryRequest", "higher than resource limit")
```
sdudoladov: That won't work because that will reject the legitimate case of request being equal to limit.

In `RequestIsSmallerThanLimit`, the line 
```go
 return request.Cmp(limit) == -1, nil
```
is responsible for that . `request.Cmp(limit)` will return zero if quantities are equal. Please check Kubernetes source code / docs for that method.

PR #509, opened 2019-03-01
refactor: refactor set_memory_request_to_limit logic

### LINKED ISSUE #419: Refactor implementation of set_memory_request_to_limit
The code for this option in controller is OK, but the code in[ `generateStatefulSet`](https://github.com/zalando-incubator/postgres-operator/blob/master/pkg/cluster/k8sres.go#L648) (used during Sync of PG clusters) should be refactored.

ZYecho:
fix #419


ZYecho:
close it, repoen it future.

PR #525, opened 2019-03-28
Config option to allow Spilo container to run non-privileged.

### LINKED ISSUE #395: Option to run containers in non-privileged mode
As far as I can see from 19dfa0c2b025a6aa9d260df13b16b7754110d01c, privileged access is used for AWS volume resizing.

However, I don't want to give my database containers privileged access so it would be great if this could be configurable.

(cc @abh)

aaroniscode:
Runs non-privileged by default.

Fixes #395 

Happy to make any changes if the team prefers a different approach.

### DIFF
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -25,6 +25,7 @@ config:
   secret_name_template: '{username}.{cluster}.credentials'
   super_username: postgres
   enable_teams_api: "false"
+  spilo_privileged: "false"
   # set_memory_request_to_limit: "true"
   # postgres_superuser_teams: "postgres_superusers"
   # enable_team_superuser: "false"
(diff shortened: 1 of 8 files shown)


FxKu:
Thanks for your contribution. Could you also introduce the flag in the [ConfigMap](https://github.com/zalando/postgres-operator/blob/master/manifests/configmap.yaml)? This would make it very easy to toggle the privilege mode.

sdudoladov:
@aaroniscode we support two alternative ways to configure the operator : the operator's own CRD and a configmap , configuration option needs to be present in both

aaroniscode:
@sdudoladov I believe the PR supports both methods to configure the operator as I tested with both the ConfigMap AND the CRD.  Am I missing something?

sdudoladov:
@aaroniscode no,  sorry , my mistake.

please add an example config  to the [configmap](https://github.com/zalando/postgres-operator/blob/master/manifests/configmap.yaml), and we will merge the PR

aaroniscode:
@FxKu @sdudoladov added the ConfigMap as requested. Let me know if there is anything else. Thanks for reviewing.

FxKu:
Yes one more thing, sorry :)
Now that we got a Helm chart, the new flag should also be present in its [values.yaml](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values.yaml).

Aside from that, it's fine for me and I can merge today.

aaroniscode:
@FxKu no problem! I added the flag to the chart's values.yaml. I had to rebase and force push as my fork didn't have the helm chart.

Issue #530, opened 2019-04-02
config option to allow Spilo to run non-root


aaroniscode:
Spilo (https://github.com/zalando/spilo) does not currently have a configuration option to run as non-root. However, it can be configured to run as non-root with minimal changes.

See: https://github.com/aaroniscode/spilo-nonroot

The issue with the Postgres Operator is that the Persistent Volume mounted for the data directory is only accessible by the root user. Kubernetes has an option `SecurityContext: fsGroup` that enables non-root users to access data on PersistentVolumes.

More information:
* https://kubernetes.io/docs/tutorials/stateful-application/zookeeper/#configuring-a-non-privileged-user
* https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod

Working on a PR for a simple config option to allow configuring the SecurityContext's fsGroup.

aaroniscode:
Zalando team -- Can you share your thoughts on this approach?  I have a customer that requires containers run as non-root and looking for a long-term solution that doesn't require running a patched version of the operator. Happy to make changes to the design or approach.

abh:
The latest version from GitHub has this feature, I think

aaroniscode:
@abh I just searched the repo and couldn't find anything that would allow spilo to run as non-root. The key to enabling that is setting a SecurityContext with a fsGroup that allows the non-root user to access the PersistentVolume. The only file in the repo that deals with a SecurityContext is `k8sres.go` at line 384: https://github.com/zalando/postgres-operator/blob/15ec6a920d71282d28ce8239423c0239c7b28e02/pkg/cluster/k8sres.go#L384.

If there is something I'm missing, please let me know as this capability is important to our client. Thanks.

Jan-M:
Thumbs up for the idea and PR, obviously running things as non root in Docker is something that should be encouraged and possible, despite us not doing it.

We will have a look.

abh:
@aaroniscode Ah, yes -- I got it confused with the `spilo_privileged: false` option which was enough to get things going with the PodSecurityPolicies in my cluster.

aaroniscode:
@abh no worries. The two are related!

aaroniscode:
Closing as #531 merged. Thanks!

PR #531, opened 2019-04-02
StatefulSet fsGroup config option to allow non-root spilo

aaroniscode:
xref #530 

Interested in your thoughts around this approach and whether defaulting fsGroup to 0, which doesn't provide any securityContext to the pods in the StatefulSet is the preferred approach.


erthalion:
Thanks for the PR! It looks reasonable, I'll test in in the next few days and by the results of this test we can hopefully merge it.

erthalion:
While testing it with `spiloFSGroup = 0` for some reason I've got a strange error from patroni (when it tries to list pods), that I can't see without the change:

```
pods is forbidden: User "system:serviceaccount:default:postgres-operator" cannot list resource "pods"
```

While I'm checking, if I'm doing something wrong, @aaroniscode have you encountered such a problem?

aaroniscode:
@erthalion A quick review of the code and I can't see how a change to the securityContext of Spilo would impact a permissions capability. I would think that would be a RBAC rule. I never encountered this in my testing, but I'll take another look today.

Is it possible that the service account is just lacking the permissions due to some other configuration or issue?

erthalion:
> Is it possible that the service account is just lacking the permissions due to some other configuration or issue?

Yes, I guess something like that, otherwise it's hard to explain. I'll try to figure this out, just to make sure that it has nothing to do with the PR.

aaroniscode:
@erthalion Let me know if you aren't able to resolve the issue and it's looking like it's related to the PR. At the moment I can't see how it would be related, but maybe I'm missing something. I can try to track it down on my side if you aren't having any luck.

erthalion:
Yes, it's not related - looks like it was a minor hiccup with service accounts, after I've rebased the code on the latest master everything seems fine. I'll do few more tests with/without fs group, and then approve the pr.

aaroniscode:
Great news. Thanks for the update!

erthalion:
Thanks @aaroniscode, at the end it looks good, I only have a few questions:

* I couldn't find it in the K8S docs or the source code, but from my tests it looks like it's not possible to change `FSGroup` after creation, is it correct?
* This options depends on a spilo image, but one can change it in both operator and database configuration, while in this PR it's possible only on the operator level (so, globally). Is it worth it to allow change it on database level too?
* I would like to ask you to add this option to the configuration CRD and Help chart as well (as described here [1]), since operator can be configured via all of them. Could you please do this? 

[1]: https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters

aaroniscode:
> * I couldn't find it in the K8S docs or in the source code, but from my tests it looks like it's not possible to change `FSGroup` after creation, is it correct? (If yes, it's probably worth to document this)

@erthalion I just tested and confirmed FSGroup cannot be changed after creation.

> * This options depends on a spilo image, but one can change it in both operator and database configuration, while in this PR it's possible only on the operator level (so, globally). Is it worth it to allow change it on database level too?

I wasn't aware you could change it at the database configuration. I think it would be worth it to allow it to change on the database level too. I'll see how I can add that to the PR.

> * I would like to ask you to add this option to the configuration CRD and Help chart as well (as described here [1](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters)), since operator can be configured via all of them. Could you please do this?

Not a problem. I'll add that as well.

I'll work on this early next week. Hope the timing works for you.

aaroniscode:
Quick update -- I believe I've got the code updated to reflect the changes requested. I going to test some more tomorrow morning then push the changes by mid-day EDT (New York time).

aaroniscode:
Found an issue with the original implementation of using a `int64`. If the Operator used a non-root FSGroup like 103 which is the `postgres` group, there was no way for the Postgres CRD to use a root FSGroup of 0 because 0 is also the default value.

So switched it over to a `*int64` and that way `nil` represents the default.

I'm running up on a deadline as I am on vacation starting tomorrow so while I was able to do some quick testing, I didn't get to test as much as I wanted. But I think this is working as intended. Let me know if I missed anything or if you think this needs any design changes. Thanks!

aaroniscode:
I'm not sure why the build is failing. I confirmed the build works on my branch and I'm using go 1.12.4.

If there is something wrong on my side, just point me in the right direction and I'll fix it ASAP.

erthalion:
> I'm not sure why the build is failing. I confirmed the build works on my branch and I'm using go 1.12.4.

I'm confused myself, since for me locally everything is fine too. Investigating right now, but to manage the expectations - probably we will not be able to merge it today. Is it ok from your side?

aaroniscode:
> I'm confused myself, since for me locally everything is fine too. Investigating right now, but to manage the expectations - probably we will not be able to merge it today. Is it ok from your side?

No worries. I'll be available next week if you need any changes once you have a chance to review. Thanks for considering this PR.

erthalion:
Oh, of course - the tests are failing for reason. @aaroniscode there is another `generatePodTemplate` invocation in the master right now (came with logical backup feature) that should be updated with `spiloFSGroup`. And when Travis clones the PR it clones the merge commit (where this function call is present) and fails to build it. The solution would be just rebase and fix the error.

aaroniscode:
> Oh, of course - the tests are failing for reason. @aaroniscode there is another `generatePodTemplate` invocation in the master right now (came with logical backup feature) that should be updated with `spiloFSGroup`. And when Travis clones the PR it clones the merge commit (where this function call is present) and fails to build it. The solution would be just rebase and fix the error.

Well, that makes sense! So, I read through the logical backup PR. I think we have a number of options of how we want to proceed.

1. The logical backup CronJob is created without specifying the FSGroup.
2. The logical backup CronJob assumes the Spilo pod's FSGroup.
3. The logical backup CronJob gets it's own FSGroup because it can also override the docker image.

I implemented the quick fix using Option #1. I don't think Option #2 or #3 are needed as I don't believe the pod is assigned a Persistent Volume.

Just let me know if I'm missing something as I quickly skimmed the logical backup PR.

erthalion:
Yes, I believe option 1 is what we need.

aaroniscode:
@erthalion just wanted to check in and see if you need anything else on this PR?

I have already implemented Option 1 that I described above. One of the checks is failing because coverage decreased -- I think this is because of the generators creating a new function.

erthalion:
> @erthalion just wanted to check in and see if you need anything else on this PR?

I'm testing the PR right now. The only thing I wanted to ask is that looks like the documentation for this:

> I couldn't find it in the K8S docs or in the source code, but from my tests it looks like it's not possible to change FSGroup after creation, is it correct? (If yes, it's probably worth to document this)

is still missing. Would be nice to mention in a few words that this can't be changed on the fly.

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -23,6 +23,7 @@ configuration:
     secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"
     oauth_token_secret_name: postgresql-operator
     pod_role_label: spilo-role
+    # spilo_fsgroup: "103"
```
erthalion: I think this should be without quotes, otherwise `readUint64` would complain.

aaroniscode: Removed the quotes!

aaroniscode:
There were some conflicts with master so I had to rebase. I believe I addressed all the recent items in the last few days.

> > I couldn't find it in the K8S docs or in the source code, but from my tests it looks like it's not possible to change FSGroup after creation, is it correct? (If yes, it's probably worth to document this)
> 
> is still missing. Would be nice to mention in a few words that this can't be changed on the fly.

Added to the docs!

Thanks for testing the PR and reviewing.

erthalion:
> There were some conflicts with master so I had to rebase. I believe I addressed all the recent items in the last few days.

Great, thank you! I hope I can make a final review tomorrow.

erthalion:
:+1:

sdudoladov:
:+1:

sdudoladov:
@aaroniscode thanks for the contribution !

Issue #408, opened 2018-11-09
Implement /status subresource to show current PG cluster state


sdudoladov:
In certain cases the operator may alter PG cluster characteristics coming from manifests without changing the manifests itself, such as when the minimal number of nodes for a PG cluster is specified. Such divergence of the cluster state defined in the manifest and the actual cluster state is reflected in the operator logs, but is not immediately visible through the rest api or  k8s command line tools. To avoid confusing the users in such cases, we shall implement the `/status` subresource for the Postgresql crd that will always show the actual state of a PG cluster.

PR #533, opened 2019-04-03
change name of current status field of CRD

FxKu:
There's a plan to implement the `status` subresource for the Postgres CRD somewhen in the future #408.
It's up to decide what to do with the current status behavior, as it used for syncing. When keeping both, the name of the latter has to change. This PR does that: `PostgresClusterStatus`.

One immediate benefit could be, that other K8S frameworks (e.g. OperatorHub) do not misinterpret our custom status object as a K8s subresource. See [here](https://github.com/operator-framework/community-operators/pull/210)

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/marshal.go
+++ b/pkg/apis/acid.zalan.do/v1/marshal.go
@@ -81,7 +81,7 @@ func (p *Postgresql) UnmarshalJSON(data []byte) error {
 		}
 
 		tmp.Error = err.Error()
-		tmp.Status = ClusterStatusInvalid
+		tmp.PostgresClusterStatus = ClusterStatusInvalid
 
 		*p = Postgresql(tmp)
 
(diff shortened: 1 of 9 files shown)


FxKu:
Currently, the display of the status is somewhat strange, e.g.:

```
kubectl describe postgresql acid-minimal-cluster             
Name:                         acid-minimal-cluster
Namespace:                    default
Labels:                       <none>
Annotations:                  <none>
Postgres   Cluster   Status:  SyncFailed
API Version:                  acid.zalan.do/v1
Kind:                         postgresql
Metadata:
```

cbandy:
Perhaps `Cluster Status`? The resource (kind) is `postgresql`.

redbaron:
Technically  it is a breaking change and APIVersion of a CRD should change
and then operator should support both versions to enable upgrades of
operator.

FxKu:
Yes, it breaks the API. Will raise the version, if we are really going for it. For now, it's just an idea. After some reading about the subresource, I've also found something in the [code](https://github.com/zalando/postgres-operator/blob/15ec6a920d71282d28ce8239423c0239c7b28e02/pkg/apis/acid.zalan.do/v1/crds.go#L36).

Therefore, I went for another approach in PR #534.

FxKu:
Closing this PR in favor for #534

PR #534, opened 2019-04-04
Use /status subresource instead of plain manifest field

### LINKED ISSUE #408: Implement /status subresource to show current PG cluster state

FxKu:
This PR tries to solve the same issue like PR #533, but does not change the name of the `status` field. Instead it introduces a nested structure of `PostgresStatus`. Currently only with one string field.

Should fix #408


pkg/cluster/cluster.go
```diff
@@ -172,7 +173,7 @@ func (c *Cluster) setStatus(status acidv1.PostgresStatus) {
 }
 
 func (c *Cluster) isNewCluster() bool {
-	return c.Status == acidv1.ClusterStatusCreating
+	return c.Status == acidv1.PostgresStatus{PostgresClusterStatus: acidv1.ClusterStatusCreating}
```
cbandy: Is it appropriate to compare the entire `PostgresStatus` struct here? Should it instead be looking only at `PostgresClusterStatus`?

```suggestion
	return c.Status.PostgresClusterStatus == acidv1.ClusterStatusCreating
```

pkg/cluster/sync.go
```diff
@@ -28,7 +28,7 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 		if err != nil {
 			c.logger.Warningf("error while syncing cluster state: %v", err)
 			c.setStatus(acidv1.ClusterStatusSyncFailed)
-		} else if c.Status != acidv1.ClusterStatusRunning {
+		} else if (c.Status != acidv1.PostgresStatus{PostgresClusterStatus: acidv1.ClusterStatusRunning}) {
```
cbandy: Same question here.
```suggestion
		} else if c.Status.PostgresClusterStatus != acidv1.ClusterStatusRunning {
```

pkg/cluster/cluster.go
FxKu: Thanks for your review. Your are right, should be a simple string comparison.

pkg/cluster/cluster.go
```diff
@@ -172,7 +173,7 @@ func (c *Cluster) setStatus(status acidv1.PostgresStatus) {
 }
 
 func (c *Cluster) isNewCluster() bool {
```
erthalion: Maybe we could also introduce similar functions for all the statuses? So that long and tedious ```postgresStatus.PostgresClusterStatus != ClusterStatusSyncFailed``` will become ```!isSyncFailed()```?

pkg/apis/acid.zalan.do/v1/util_test.go
```diff
@@ -128,11 +128,11 @@ var unmarshalCluster = []struct {
...
 		Error: json.Unmarshal([]byte(`{"teamId": 0}`), &PostgresSpec{}).Error(),
 	},
-	[]byte(`{"kind":"Postgresql","apiVersion":"acid.zalan.do/v1","metadata":{"name":"acid-testcluster1","creationTimestamp":null},"spec":{"postgresql":{"version":"","parameters":null},"volume":{"size":"","storageClass":""},"patroni":{"initdb":null,"pg_hba":null,"ttl":0,"loop_wait":0,"retry_timeout":0,"maximum_lag_on_failover":0,"slots":null},"resources":{"requests":{"cpu":"","memory":""},"limits":{"cpu":"","memory":""}},"teamId":"","allowedSourceRanges":null,"numberOfInstances":0,"users":null,"clone":{}},"status":"Invalid"}`), nil},
+	[]byte(`{"kind":"Postgresql","apiVersion":"acid.zalan.do/v1","metadata":{"name":"acid-testcluster1","creationTimestamp":null},"spec":{"postgresql":{"version":"","parameters":null},"volume":{"size":"","storageClass":""},"patroni":{"initdb":null,"pg_hba":null,"ttl":0,"loop_wait":0,"retry_timeout":0,"maximum_lag_on_failover":0,"slots":null},"resources":{"requests":{"cpu":"","memory":""},"limits":{"cpu":"","memory":""}},"teamId":"","allowedSourceRanges":null,"numberOfInstances":0,"users":null,"clone":{}},"status":{"PostgresClusterStatus":"Invalid"}}`), nil},
```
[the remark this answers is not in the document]
FxKu: Hm, valid point. I'm still figuring out, how to provide better backwards compatibility.

redbaron:
In simple Go you could provide PosgresClusterStstus.UnmarshalJSON method which would be able to read .spec string from old json representation of resource in cluster, but I am not sure how it can be incorporated into auto-generated clients.

pkg/apis/acid.zalan.do/v1/util_test.go
```diff
@@ -494,12 +563,23 @@ func TestMarshal(t *testing.T) {
...
+		// Unmarshal and marshal example to capture api changes
+		var cluster Postgresql
+		err := cluster.UnmarshalJSON(tt.marshal)
+		if err != nil {
+			if tt.err == nil || err.Error() != tt.err.Error() {
+				t.Errorf("Unmarshal expected error: %v, got: %v", tt.err, err)
+			}
+			continue
+		}
+		expected, err := json.Marshal(cluster)
```
golangcibot: SA4006: this value of `err` is never used (from `staticcheck`)

Jan-M:
@FxKu @sdudoladov this looks merge ready to me?

pkg/apis/acid.zalan.do/v1/util.go
```diff
@@ -102,5 +102,5 @@ func (postgresStatus PostgresStatus) Creating() bool {
 }
 
 func (postgresStatus PostgresStatus) String() string {
-	return fmt.Sprintf(`"status":{"PostgresClusterStatus": %s}`, postgresStatus.PostgresClusterStatus)
+	return fmt.Sprintf(`{"status":{"PostgresClusterStatus": %s}}`, string(postgresStatus.PostgresClusterStatus))
```
Jan-M: Can we assume that string() does the proper quote/escape?

Jan-M: After short discussion we elaborate on going back to marshal as the proper solution to generate json of nested objects.

pkg/cluster/cluster.go
```diff
@@ -149,21 +148,24 @@ func (c *Cluster) setProcessName(procName string, args ...interface{}) {
...
+	if err != nil {
 		c.logger.Errorf("could not marshal status: %v", err)
 	}
 
-	patch := []byte(fmt.Sprintf(`{"status": %s}`, string(b)))
 	// we cannot do a full scale update here without fetching the previous manifest (as the resourceVersion may differ),
 	// however, we could do patch without it. In the future, once /status subresource is there (starting Kubernets 1.11)
 	// we should take advantage of it.
-	newspec, err := c.KubeClient.AcidV1ClientSet.AcidV1().Postgresqls(c.clusterNamespace()).Patch(c.Name, types.MergePatchType, patch)
+	newspec, err := c.KubeClient.AcidV1ClientSet.AcidV1().Postgresqls(c.clusterNamespace()).Patch(c.Name, types.MergePatchType, patch, "status")
```
sdudoladov: 1. the `updateStatus()` in k8s codebase generally  does `PUT` instead of `PATCH` ([example](https://github.com/kubernetes/client-go/blob/master/kubernetes/typed/extensions/v1beta1/deployment.go#L139)). Is this case the effect of patching `/status` always the same as completely replacing it with `PUT`? 
2.  actually our code generation should give us `updateStatus()` automatically. The code generation is adapted from etcd / Prometheus operators and the latter [definitely has this generated method](https://github.com/coreos/prometheus-operator/blob/0e22050ca3ebe6b7033f8c0f1cb1605e3abe63c5/pkg/client/versioned/typed/monitoring/v1/prometheus.go#L134)

pkg/cluster/cluster_test.go
```diff
@@ -318,3 +328,33 @@ func TestShouldDeleteSecret(t *testing.T) {
 		}
 	}
 }
+
+/*
+func TestSetStatus(t *testing.T) {
```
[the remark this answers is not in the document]
FxKu: removed

sdudoladov:
:+1:

FxKu:
:+1:

sdudoladov:
UPD this  PR assumes the `/status` subresource exists which may not be the case in older versions of the PG CRD , we are preparing the fix

Issue #198, opened 2017-12-31
Google Cloud Storage Credentials


michailbrynard:
After experimenting with the Patroni helm chart, I've decided to go with the postgres-operator instead. 

The Patroni helm chart allows Gcloud credentials to be stored as a secret which is then mounted as a volume referenced by the GOOGLE_APPLICATION_CREDENTIALS environmental variable. It looks like the operator makes no such provision. What would be the suggested way to inject the Google Cloud credentials for WALE backups into the Spillo pods using the operator?

Jan-M:
Sounds like the operator should take care of this, similar to it already specifying the IAM role in an AWS environment. We can probably take a quick look at this, as we know from the helm chart how this should look like and would make the operator more attractive to non AWS users.

michailbrynard:
Yes, that would be great for users running on GKE!

michailbrynard:
After a bit of experimentation on GKE, it seems that if the GOOGLE_APPLICATION_CREDENTIALS environmental variable is not set, the gcloud sdk in the spilo container defaults to using the gcloud service account associated with the kubernetes service account for the pod. This makes the feature to allow the mounting of the gcloud credentials less critical as the default behaviour is similar to `--aws-instance-profile` as long as the RBAC and Google Cloud service accounts are set up properly.

rporres:
We use the operator to deploy postgres in bare metal kubernetes. We're interested in this as we keep our infra backups in GCE. Unfortunately @michailbrynard's workaround won't work in our case. Any suggestions about how to proceed? We're happy to contribute if needed, but since we're starting using the operator guidance would be more than welcomed.

FxKu:
@rporres take a look the PR #535. Another user who runs the operator on bare metal and came up with a solution.

PR #481, opened 2019-02-04
Populate pod environment variables from secrets

kupson:
Fixes #480.

Example Secret:

```
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: minio-secret-test
data:
  access_key: dGVzdAo=
  secret_key: dGVzdAo=
```

Configuration:
```
apiVersion: acid.zalan.do/v1
configuration:
  ...
  kubernetes:
    ...
    pod_environment_secret_name: minio-secret-test
    pod_environment_secret_keys:
      access_key: AWS_ACCESS_KEY_ID
      secret_key: AWS_SECRET_ACCESS_KEY
```

Generated Pod config snippet:
```
  env:
    - name: AWS_ACCESS_KEY_ID
      valueFrom:
        secretKeyRef:
          key: access_key
          name: minio-secret-test
    - name: AWS_SECRET_ACCESS_KEY
      valueFrom:
        secretKeyRef:
          key: secret_key
          name: minio-secret-test
```

PR #535, opened 2019-04-04
Add secret mount to operator

Shinzu:
First initial commit for adding  cloud secrets to non cloud environment.

It ties to address #198

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -340,11 +340,11 @@ To access cloud resources like S3 from a cluster in a bare metal setup you can u
 `additional_secret_mount` and `additional_secret_mount_path` config parameters.
 With this you can provision cloud credentials to the containers in the pods of the StatefulSet.
 This works this way that it mounts a volume from the given secret in the pod and this can
-then accessed in the container over the configured mount path. Via [Custum Pod Environment Variables](## Custom Pod Environment Variables)
+then accessed in the container over the configured mount path. Via [Custum Pod Environment Variables](#custom-pod-environment-variables)
 you can then point the different cloud sdk's (aws, google etc.) to this mounted secret.
 With this credentials the cloud sdk can then access cloud resources to upload logs etc.
 
-This secrets can be pre provisioned in different ways:
+A secret can be pre provisioned in different ways:
 
 * Generic secret created via `kubectl create secret generic some-cloud-creds --from-file=some-cloud-credentials-file.json`
 
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -333,3 +333,19 @@ The operator can manage k8s cron jobs to run logical backups of Postgres cluster
 4. You may use your own image by overwriting the relevant field in the operator configuration. Any such image must ensure the logical backup is able to finish [in presence of pod restarts](https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#handling-pod-and-container-failures) and [simultaneous invocations](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-job-limitations) of the backup cron job.
 
 5. For that feature to work, your RBAC policy must enable operations on the `cronjobs` resource from the `batch` API group for the operator service account. See [example RBAC](../manifests/operator-service-account-rbac.yaml)
+
+## Access to cloud resources from clusters in non cloud environment
+
+To access cloud resources like S3 from a cluster in a bare metal setup you can use
+`additional_secret_mount` and `additional_secret_mount_path` config parameters.
+With this you can provision cloud credentials to the containers in the pods of the StatefulSet.
+This works this way that it mounts a volume from the given secret in the pod and this can
+then accessed in the container over the configured mount path. Via [Custum Pod Environment Variables](## Custom Pod Environment Variables)
+you can then point the different cloud sdk's (aws, google etc.) to this mounted secret.
+With this credentials the cloud sdk can then access cloud resources to upload logs etc.
+
+This secrets can be pre provisioned in different ways:
+
+* Generic secret created via `kubectl create secret generic some-cloud-creds --from-file=some-cloud-credentials-file.json`
+
+* Automaticly provisioned via a Controller like [kube-aws-iam-controller](https://github.com/mikkeloscar/kube-aws-iam-controller). This controller would then also rotate the credentials. Please visit the documention for more information.
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -407,6 +407,12 @@ yet officially supported.
 * **aws_region**
   AWS region used to store ESB volumes. The default is `eu-central-1`.
 
+* **additional_secret_mount**
+  Additional Secret (aws or gcp credentials) to mount in the the pod. The default is empty.
+
+* **additional_secret_mount_path**
+  Path to mount the above Secret in the filesystem of the container(s). The default is empty.
+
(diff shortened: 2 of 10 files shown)


redbaron:
Have you seen https://github.com/zalando/postgres-operator/pull/481 ?

Shinzu:
yes but i think this tackles an other problem

docs/administrator.md
```diff
@@ -346,3 +346,19 @@ To read the latter log entry with the escaped characters rendered, view it
...
 noise because the `PodTemplate` used by the operator is yet to be updated
 with the default values used internally in Kubernetes.
+
+## Access to cloud resources from clusters in non cloud environment
+
+To access cloud resources like S3 from a cluster in a bare metal setup you can use
+`additional_secret_mount` and `additional_secret_mount_path` config parameters.
+With this you can provision cloud credentials to the containers in the pods of the StatefulSet.
+This works this way that it mounts a volume from the given secret in the pod and this can
+then accessed in the container over the configured mount path. Via [Custum Pod Environment Variables](## Custom Pod Environment Variables)
```
FxKu: ... of the StatefulSet.
A volume is mounted from the given secret in the pod and can then be accessed in the container over the configured mount path. Via [Custom Pod Environment Variables](## Custom Pod Environment Variables) ... 

btw: link doesn't work for me

docs/administrator.md
```diff
@@ -346,3 +346,19 @@ To read the latter log entry with the escaped characters rendered, view it
...
+you can then point the different cloud sdk's (aws, google etc.) to this mounted secret.
+With this credentials the cloud sdk can then access cloud resources to upload logs etc.
+
+This secrets can be pre provisioned in different ways:
```
FxKu: Keep a singular: A secret can be ...

docs/reference/operator_parameters.md
```diff
@@ -366,6 +366,12 @@ either. In the CRD-based configuration those options are grouped under the
 * **aws_region**
   AWS region used to store ESB volumes. The default is `eu-central-1`.
 
+* **additional_secret_mount**
+  Additional Secret (aws or gcp credentials) to mount in the the pod. The default is empty.
```
FxKu: in the pod

docs/administrator.md
Shinzu: fixed

docs/reference/operator_parameters.md
Shinzu: fixed

docs/administrator.md
Shinzu: fixed

redbaron:
We keep poorly reinventing pod spec, when is the time to stop? https://github.com/zalando/postgres-operator/issues/479

Shinzu:
@redbaron i'm with you, in the end this is just convenience function for user who don't want to create a whole pod spec

erthalion:
As we discussed in #479, most likely we would need to have both "modes" - full spec and not full spec. In this sense I believe it makes sense to have this feature for the second mode, what do you think?

pkg/cluster/k8sres_test.go
```diff
@@ -129,3 +129,74 @@ func TestShmVolume(t *testing.T) {
...
+			if volumeMountName != additionalSecretMount {
+				t.Errorf("%s %s: Expected mount %s was not created, have %s instead",
+					testName, tt.subTest, additionalSecretMount, volumeMountName)
+			}
+		}
+
+		volsuma := len(tt.podSpec.Containers[0].VolumeMounts)
+
+		if volsuma != volsumb+1 {
+			t.Errorf("Got not expected number of VolumeMounts: got %v instead of %v",
```
FxKu: "Unexpected number ...

btw: I find `volsumb` a strange name. Maybe `numMounts` for b and `numMountsCheck` for a?

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -61,6 +61,8 @@ configuration:
     # log_s3_bucket: ""
     # kube_iam_role: ""
     aws_region: eu-central-1
+    # additional_secret_mount: 'some-secret-name'
+    # additional_secret_mount_path: '/some/dir'
```
FxKu: does it really work with single quotes? In the ConfigMap I can see other examples with them but not in the `OperatorConfiguration`. Other the other way around: Would `"` work for ConfigMap, too.? If yes, change it also in the `values.yaml` file for Helm chart. The Helm templates for ConfigMap and CRD can stay as they are.

pkg/cluster/k8sres_test.go
Shinzu: renamed variables

manifests/postgresql-operator-default-configuration.yaml
Shinzu: using now double quotes

Jan-M:
Looks good to me. @sdudoladov @FxKu

FxKu:
@Shinzu could you please rebase and resolve the conflicts?

pkg/cluster/k8sres.go
```diff
@@ -993,6 +1001,28 @@ func addShmVolume(podSpec *v1.PodSpec) {
...
+	volumes := append(podSpec.Volumes, v1.Volume{
+		Name: additionalSecretMount,
+		VolumeSource: v1.VolumeSource{
+			Secret: &v1.SecretVolumeSource{
+				SecretName: additionalSecretMount,
+			},
+		},
+	})
+
+	for i := range podSpec.Containers {
```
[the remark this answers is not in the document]
Shinzu: yes it is intentional

sdudoladov:
@Shinzu can you please rebase and then we will merge this PR

Shinzu:
rebase done

sdudoladov:
:+1:

FxKu:
hm, for some reason the config parameters are in single quotes again?

Shinzu:
@FxKu @sdudoladov re applied the lost changes

FxKu:
:+1:

sdudoladov:
:+1:

KarstenSiemer:
Hi!
Can somebody tell me how the secret has to look like exactly? There is no use to saying there is a secret with a json file in it you can attach but putting no information on how it has to look like.
The documentation of WAL-G is only point out env variables and spilo is equally silent about how to pass a json.
Thanks!

Shinzu:
hello, this file just looks like a normal credential file for aws or gcp eg https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html

KarstenSiemer:
i still do not completely see the full image for getting wal-e to work with baremetal clusters.
> Via Custom Pod Environment Variables you can point different cloud SDK's (AWS, GCP etc.) to this mounted secret, e.g. to access cloud resources for uploading logs etc.

>AWS, GCP etc.

So one mounts a file with the credentials, this is placed inside the pod but unused, unless i configure env vars to pick up that file.

But, when looking at the documentation of wal-e, only gcp has a variable ("GOOGLE_APPLICATION_CREDENTIALS") for that. AWS doesn't.
How does one pass the credentials file to the aws connector?

whereismyjetpack:
The behavior of this PR is slightly different than the behavior of injecting these values from the configmap. Mainly, the secret object must live in the namespace where the postgresql object is, and not the operator..

Ideally, for our usecase, we'd specify cloud credentials in the postgres-operator namespace, and the operator would add them to the existing secret that gets created, or to a new cloud credentials secret object.

PR #538, opened 2019-04-08
Add paragraph on how to extend operator configuration in dev docs

### LINKED ISSUE #532: explain adding a new option to the operator
to keep operator configuration consistent, we should update quite a few places within a code and docs. The task is to add a checklist for that to the developer documentation

FxKu:
Something to point contributors to when adding features. Focuses on extending the configuration. Am I missing something?

resolves #532

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -10,7 +10,8 @@ After the installation, issue
     $ minikube start
 ```
 
-Note: if you are running on a Mac, you may also use Docker for Mac Kubernetes instead of a docker-machine.
+Note: if you are running on a Mac, you may also use Docker for Mac Kubernetes
+instead of a docker-machine.
 
 Once you have it started successfully, use [the quickstart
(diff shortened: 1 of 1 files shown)


docs/developer.md
```diff
@@ -290,10 +304,38 @@ PASS
...
-It will automatically create an `acid-minimal-cluster` in the namespace `test`. Then you can for example check the Patroni logs:
+It will automatically create an `acid-minimal-cluster` in the namespace `test`.
+Then you can for example check the Patroni logs:
+
 ```
 kubectl logs acid-minimal-cluster-0
 ```
+
+## Introduce additional configuration parameters
+
```
sdudoladov: By convention if the same option is defined in the operator config and in the cluster manifest, the latter takes precedence

FxKu: Isn't this already documented somewhere? Thought it was [here](https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md). I can add a link, nevertheless.

PR #541, opened 2019-04-12
Fix CRD-based operator configuration

### LINKED ISSUE #540: OperatorConfiguration does not support pod_management_policy
[Support](https://github.com/zalando/postgres-operator/issues/514) for explicitly setting the pod management policy was recently added. However, this does not appear to be supported by the OperatorConfiguration (and no default value is assumed). As a result, clusters fail to be created with the following error:

`level=warning msg="error while syncing cluster state: could not sync statefulsets: could not create missing statefulset: could not generate statefulset: could not set the pod management policy to the unknown value: "`

sdudoladov:
fixes #540 and a few other misconfigurations. Note our Helm chart does not support CRD-based conf right now.

@FxKu thank you for the checklist for new params, that is helpful in this issue

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
+++ b/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
@@ -62,8 +62,9 @@ type KubernetesMetaConfiguration struct {
 	PodEnvironmentConfigMap    string        `json:"pod_environment_configmap,omitempty"`
 	PodPriorityClassName       string        `json:"pod_priority_class_name,omitempty"`
 	MasterPodMoveTimeout       time.Duration `json:"master_pod_move_timeout,omitempty"`
-	EnablePodAntiAffinity      bool          `json:"enable_pod_antiaffinity" default:"false"`
-	PodAntiAffinityTopologyKey string        `name:"pod_antiaffinity_topology_key" default:"kubernetes.io/hostname"`
+	EnablePodAntiAffinity      bool          `json:"enable_pod_antiaffinity,omitempty"`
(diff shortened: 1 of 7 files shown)


FxKu:
Can we use this PR also to align ConfigMap with CRD? I think there are also parts in the CRD which are not part of the ConfigMap, e.g. #459

sdudoladov:
FYI I also updated the golang to version 1.12. The `staticcheck` tool we use in `make tools` [supports only 2 last versions of golang](https://github.com/dominikh/go-tools/issues/442#issuecomment-482561255), and that currently breakes the build

To update your local golang installation you [can use this script](https://gist.github.com/nikhita/432436d570b89cab172dcf2894465753#gistcomment-2767808)

sdudoladov:
:+1:

FxKu:
:+1:

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -62,8 +62,9 @@ type KubernetesMetaConfiguration struct {
 	PodEnvironmentConfigMap    string        `json:"pod_environment_configmap,omitempty"`
 	PodPriorityClassName       string        `json:"pod_priority_class_name,omitempty"`
 	MasterPodMoveTimeout       time.Duration `json:"master_pod_move_timeout,omitempty"`
-	EnablePodAntiAffinity      bool          `json:"enable_pod_antiaffinity" default:"false"`
-	PodAntiAffinityTopologyKey string        `name:"pod_antiaffinity_topology_key" default:"kubernetes.io/hostname"`
```
erthalion: Where goes this "default" value?

Issue #148, opened 2017-10-23
consider switching from glide to dep package manager


mkabilov:

zerg-junior:
update:

1. `client-go` is [hard to update in projects that use `dep`](https://github.com/kubernetes/client-go/issues/78#issuecomment-387845808)
2. the [Go v1.11 modules are experimental](https://github.com/golang/go/wiki/Modules#go-111-modules)

So we probably should just stick with `glide` despite it not being developed since December/2017

sdudoladov:
[Go Modules in 2019](https://blog.golang.org/modules2019)
> Our aim is for Go 1.13, scheduled for August 2019, to enable module mode by default (that is, to change the default from auto to on) and deprecate GOPATH mode. In order to do that, we’ve been working on better tooling support along with better support for the open-source module ecosystem.

we should move to modules in the near future.

FxKu:
Go modules support got just merged :smiley:

PR #544, opened 2019-04-16
[WIP] An attempt to build with modules

erthalion:
Work in progress, but I guess it would be nice to have.


FxKu:
resolves #148

sdudoladov:
I found[ the original module proposal](https://research.swtch.com/vgo-intro#proposal) helpful to understand the modules' system while reviewing the operator's kubectl plugin (it switched from Glide to modules recently )

FxKu:
@erthalion could you rebase? Our test pipeline in the master branch is currently failing because the new kubectl-pg plugin needs `GO111MOUDLE=on`. Maybe this is a good moment for the operator to switch to Go modules, too.

FxKu:
A [change](https://github.com/kubernetes/code-generator/commit/5ac21ed6de7ac91d03e7917872c463de906d46a6) in the code-generator source code made me change the package/directory name from `acid.zalan.do` to `acidzalando`. However, the group name remains the same (with dots).

I created a [PR](https://github.com/kubernetes/code-generator/issues/77) in the codegen repo. If they consider changing it back we could also switch back and not break things.

sdudoladov:
> A change in the code-generator source code made me change the package/directory name from acid.zalan.do to acidzalando.

What impact did similar renaming have on other operators that depend on the code generator ?
if someone uses the renamed API the name change will break their code ?

sdudoladov:
FYI go 1.13 was released,  it changes the way [modules are configured](https://golang.org/doc/go1.13#modules)

FxKu:
Looks like they are going to [fix it](https://github.com/kubernetes/kubernetes/pull/82410), so we can leave the package name as is. Will update the PR, when it's merged on the code-generator side.

hack/update-codegen.sh
```diff
@@ -5,9 +5,9 @@ set -o nounset
 set -o pipefail
 
 SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
-CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ${GOPATH}/src/k8s.io/code-generator)}
+CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
```
sdudoladov: but there is no `code-generator` subdirectory in the operator top-level dir or ?

pkg/controller/postgresql.go
```diff
@@ -312,7 +312,7 @@ func (c *Controller) processClusterEventsQueue(idx int, stopCh <-chan struct{},
 	for {
 		obj, err := c.clusterEventQueues[idx].Pop(cache.PopProcessFunc(func(interface{}) error { return nil }))
 		if err != nil {
-			if err == cache.FIFOClosedError {
+			if err == cache.ErrFIFOClosed {
```
sdudoladov: The change is caused by renaming in the `cache` package ; fixes a lint violation there [reference](https://github.com/kubernetes/client-go/commit/2b8d87c0821de4a8eaa8da496845d8a04cf1eadf#diff-c74ce1b4d1c0c7c6f63898649343d81bL37)

Makefile
```diff
@@ -79,7 +79,7 @@ scm-source.json: .git
 
 tools:
 	@go get -u honnef.co/go/tools/cmd/staticcheck
-	@go get -u github.com/Masterminds/glide
+	@go get k8s.io/client-go@kubernetes-1.15.0
```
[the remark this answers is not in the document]
FxKu: valid point. I'll update it. Btw, I introduced it here, because I'm getting the following error when only relying on `go mod vendor`:
```# k8s.io/client-go/rest
../../../../pkg/mod/k8s.io/client-go@v11.0.0+incompatible/rest/request.go:598:31: not enough arguments in call to watch.NewStreamWatcher
	have (*versioned.Decoder)
	want (watch.Decoder, watch.Reporter)
```

Was also reported [here](https://github.com/kubernetes/client-go/issues/584).

hack/update-codegen.sh
FxKu: I'll change it back. Just copied it from [this](https://github.com/kubernetes/sample-apiserver/blob/f3d6226a155191940485b3f7a45742e1ca39c0b8/hack/update-codegen.sh#L22) example.

go.mod
```diff
@@ -0,0 +1,18 @@
...
+
+require (
+	github.com/Azure/go-autorest v11.1.2+incompatible // indirect
+	github.com/aws/aws-sdk-go v1.25.0
+	github.com/imdario/mergo v0.3.7 // indirect
+	github.com/lib/pq v1.2.0
+	github.com/motomux/pretty v0.0.0-20161209205251-b2aad2c9a95d
+	github.com/sirupsen/logrus v1.4.2
+	gopkg.in/yaml.v2 v2.2.2
+	k8s.io/api v0.0.0-20190927115716-5d581ce610b0
```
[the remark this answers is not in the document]
FxKu: I simply ran `go mod init` followed by `go mod vendor`. So this would give me the latest version of the dependencies. So much newer.

sdudoladov:
:+1:

FxKu:
:+1:

sdudoladov:
:+1:

Issue #475, opened 2019-01-31
Investigate a solution for the local testing of the operator


sdudoladov:
The current way of testing is to run minikube, either manually or with some tooling around it like `/run-operator_locally.sh` or Vagrant. This has at least three problems:

1. minikube is a single node cluster => unsuitable for testing vital functions such as pod migration between nodes.
2. minikube starts slowly; that prolongs local testing. 
3.  every contributor  needs to come up with their own solution for local testing

We need to come up with a better option which will enable us to conveniently and uniformly  run e2e tests locally / potentially in Travis CI.

A promising option is [kind](https://github.com/kubernetes-sigs/kind)

sdudoladov:
On the 1st  look through the docs `kind`  from k8s' own SIGTEST seems OK, or at least more stable/with better prospects than the [alternatives](https://github.com/kubernetes-sigs/kind#alternatives)

I will implement a single operator e2e test that  uses  operator-on-`kind` to create a PG DB to evaluate: 
1. start up costs
2. ability to add/label nodes (needed to test pod migration, affinity etc.)
3. necessary tooling. Ideally e2e tests should not require long custom bash scripts as `minikube`-based testing currently does.

sdudoladov:
@Jan-M @alexeyklyukin @erthalion  @redbaron do you have more ideas ?

Jan-M:
`kind` looks interesting to me if it can start 3 nodes locally and we can deploy/run our tests against this api similar to what we do internally.

alexeyklyukin:
I used to test the docker for Mac internal Kubernetes instead of minikube (on a Mac), as minikube was not particularly stable there, but `kind` looks promising indeed to test a multi-node migration.

sdudoladov:
FYI [actual k8s e2e tests' infra](https://kubernetes.io/blog/2019/03/22/kubernetes-end-to-end-testing-for-everyone/) overkill for our use case, but we might borrow ideas from it

sdudoladov:
one other dimension to consider is to add some mock of the S3 API; that may enable testing backups/clones etc. locally

FxKu:
So, #548 got merged by now. We should add a section in dev docs, though. Closing.

PR #548, opened 2019-04-24
Implement runner for e2e tests

sdudoladov:
early work on #475


sdudoladov:
Current state: we have the infra to start writing tests. `make e2e` will execute any tests it finds in `./e2e/tests` against a single `kind` cluster.

Questions:

1. Please review the existing project structure and CI setup
2. How shall we setup dependency management  for Python in this case ? 
3. Determine a structure of an individual test. I suggest we provide standard implementations for `setUp` and `tearDown` methods that deploy the operator with a single DB using examples form `/manifest` and delete all operator-related files respectively. It will be the responsibility of  an individual test to (un)do test-specific patches. Caveat: `tearDown` is always called , this means ii will delete important state after test failure. Such a failing test can be re-run manually without the `tearDown` method
4. Test cases:
    a. connect to a PG DB
    b. drain a node with a master and see a failover happens

e2e/run.sh
```diff
@@ -0,0 +1,25 @@
...
+
+kind create cluster --name ${cluster_name} --config ./e2e/kind-config-smoke-tests.yaml
+export KUBECONFIG="$(kind get kubeconfig-path --name=${cluster_name})"
+kubectl cluster-info
+
+image=$(docker images --filter=reference="registry.opensource.zalan.do/acid/postgres-operator" --format "{{.Repository}}:{{.Tag}}"  | head -1)
+kind load docker-image ${image} --name ${cluster_name}
+
+python3 -m unittest discover --start-directory e2e/tests/ -v &&
+kind delete cluster --name ${cluster_name}
```
FxKu: What about? `$KUBECONFIG is still set to use /home/<me>/.kube/kind-config-kind-smoke-test-postgres-operator even though that file has been deleted, remember to unset it`

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,146 @@
...
+                            {
+                              "name": "postgres-operator",
+                              "image": image
+                            }
+                        ]
+                    }
+                 }
+            }
+        }
+        k8s_api.apps_v1.patch_namespaced_deployment("postgres-operator", "default", body)
```
FxKu: Wouldn't it be better to create the deployment with latest image straight away instead of patching an older version? Now it's also sort of a backwards compatibility check, right? ;) Any reason not to the create [api] for deployments (https://github.com/kubernetes-client/python/blob/master/examples/deployment_examples.py). Can still be changed later, though.

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self):
+
+        # https://github.com/kubernetes-client/python/issues/309
+        warnings.simplefilter("ignore", ResourceWarning)
+
+        self.config = config.load_kube_config()
+        self.k8s_client = client.ApiClient()
+        self.core_v1 = client.CoreV1Api()
+        self.crd = client.CustomObjectsApi()
+        self.apps_v1 = client.AppsV1Api()
```
FxKu: We will need more APIs, don't we? Well, can be added when needed...

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,146 @@
...
+
+        body = {
+            "spec": {
+                "numberOfInstances": number_of_instances
+            }
+        }
+        _ = k8s_api.crd.patch_namespaced_custom_object("acid.zalan.do",
+                                                       "v1", "default", "postgresqls", "acid-minimal-cluster", body)
+
+        labels = 'version=acid-minimal-cluster'
```
FxKu: in other functions you've put the labels in the spec. I think it can also be done here. `'version=acid-minimal-cluster'` is also hard-coded in `test_scaling` from where `wait_for_pg_to_scale` is called

FxKu:
I'm fine with this setup and could start writing tests on my own. I think only then I'll get a better feeling if we have to adjust the current test strategy. One thing, I find a bit annoying is that test results can differ depending on the local go / python / docker / kind setup. Would be nice, if it's the same for all of us. Docker in Docker in Docker?

Jan-M:
Maybe one could use "kind" to run a test container? And just output the logs? That gives the container network and k8s api

e2e/tests/test_smoke.py
sdudoladov: that's exactly the idea behind this class: collect APIs in one place and add as needed

e2e/tests/test_smoke.py
sdudoladov: yes, it is evidently better to configure deployment completely before creating it. there however two ways to achieve this goal:
1. modify the deployment manifest. That approach requires shuffling the file around and that is exactly [what the `run_operator_locally.sh` does](https://github.com/zalando/postgres-operator/blob/master/run_operator_locally.sh#L128) I preferred standard k8s API to that approach.
2. upload the deployment definition in memory and modify it. That's the proper approach which we should use  once we see the tests we need actually fit into the proposed runner structure.

Jan-M: Not following completely, but is python so you can also just have nice template strings using

```
deployment = """"
...
""""
```

And use e.g. jinja2 templates to resolve.

e2e/run.sh
sdudoladov: i removed the code snippet with the `unset` for now cause it lead to some unexpected results on subsequent local e2e runs.  it was sth like that in `./e2e/run.sh`
```bash
if [[ -v KUBECONFIG ]] # bash >= 4.2
 then
  rm ${$KUBECONFIG}
  unset KUBECONFIG
 fi
```

sdudoladov:
> I find a bit annoying is that test results can differ depending on the local go / python / docker / kind setup. Would be nice, if it's the same for all of us. Docker in Docker in Docker?

imo too much Dockering here.  And the link from above points a discussion why such setup may [be bad for CI](http://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/)

> Maybe one could use "kind" to run a test container? And just output the logs?

Sounds better to me, but then we need to be careful when we start decommission nodes to test pod migration.

delivery.yaml
```diff
@@ -41,6 +41,15 @@ pipeline:
             export PATH=$PATH:$HOME/go/bin
             cd $OPERATOR_TOP_DIR/postgres-operator
             go test ./...
+        - desc: 'Run e2e tests'
+          cmd: |
+            export PATH=$PATH:$HOME/go/bin
+            cd $OPERATOR_TOP_DIR/postgres-operator
+            echo "INFO install kubectl to test 'kind' from client side"
+            (curl -LO --silent https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl &&
```
erthalion: It always looks scary, especially in tests. There is a apt.kubernetes.io repository, maybe it's better to install it this way?

e2e/requirements.txt
```diff
@@ -0,0 +1,5 @@
+kubernetes
```
erthalion: Python packages tend to break stuff between the versions, maybe it makes sense think about using fixed versions.

FxKu:
After writing my own test a few questions popped up. We discussed it already, so this is just for the records.

- It would make sense to group a series of tests in separate files. Can/Shall I reuse the start up method which is currently part of `test_smoke.py`? Or simply copying them?
- Would it make sense to put the cluster setup in it's own file (or part of make), and then run all tests against this setup. Then we might have to undo things we change with tests.

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,191 @@
...
+        self.assertEqual(3, Utils.count_pods_with_label(k8s, labels))
+
+        # TODO `kind` pods may stuck in the Terminating state for a few minutes; reproduce/file a bug report
+        Utils.wait_for_pg_to_scale(k8s, 2, self.RETRY_TIMEOUT_SEC)
+        self.assertEqual(2, Utils.count_pods_with_label(k8s, labels))
+
+    @timeout_decorator.timeout(TEST_TIMEOUT_SEC)
+    def test_taint_based_eviction(self):
+        """
+           Add taint "postgres=:NoExecute" to node with master.
```
sdudoladov: please briefly comment on the test scenario, e.g., what is expected when a taint is added.It basically tests the master failover  ; if I remember correctly replicas will simply die with the node and later recreated by the stateful set

PS we actually have 3 nodes in this setup, not sure if the master node is schedulable though

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,191 @@
...
+                        "key": "postgres"
+                    }
+                ]
+            }
+        }
+        podsList = k8s.core_v1.list_namespaced_pod("default", label_selector=labels)
+        for pod in podsList.items:
+            if ('spilo-role', 'master') in pod.metadata.labels.items():
+                master_pod_node = pod.spec.node_name
+            elif ('spilo-role', 'replica') in pod.metadata.labels.items():
```
sdudoladov: 1. that kind of assumes there is always one replica and 2 nodes ? may be ok for the test, just worth a comment
2. i'd also suggest moving `k8s.core_v1.patch_node(master_pod_node, body)`  closer to the `body` definition, makes it easier to read

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,191 @@
...
+
+        k8s.core_v1.patch_node(master_pod_node, body)
+        Utils.wait_for_master_failover(k8s, new_master_pod_node, self.RETRY_TIMEOUT_SEC)
+
+        self.assertTrue(master_pod_node != new_master_pod_node, "Master on {} did not fail over to {}".format(master_pod_node, new_master_pod_node))
+
+        # undo the tainting
+        body = {
+            "spec": {
+                "taints": []
```
sdudoladov: what about default taints ? are there any ?

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,191 @@
...
+
+    @staticmethod
+    def count_pods_with_label(k8s_api, labels):
+        return len(k8s_api.core_v1.list_namespaced_pod('default', label_selector=labels).items)
+
+    @staticmethod
+    def wait_for_master_failover(k8s_api, expected_master_pod_node, retry_timeout_sec):
+        pod_phase = 'Failing over'
+        new_master_pod_node = ''
+        while (pod_phase != 'Running') & (new_master_pod_node != expected_master_pod_node):
```
sdudoladov: `&` is `bitwise and` ; the standard `and` keyword shall be preferred for readability purposes

e2e/tests/test_smoke.py
```diff
@@ -0,0 +1,191 @@
...
+        for pod in podsList.items:
+            if ('spilo-role', 'master') in pod.metadata.labels.items():
+                master_pod_node = pod.spec.node_name
+            elif ('spilo-role', 'replica') in pod.metadata.labels.items():
+                new_master_pod_node = pod.spec.node_name
+
+        k8s.core_v1.patch_node(master_pod_node, body)
+        Utils.wait_for_master_failover(k8s, new_master_pod_node, self.RETRY_TIMEOUT_SEC)
+
+        self.assertTrue(master_pod_node != new_master_pod_node, "Master on {} did not fail over to {}".format(master_pod_node, new_master_pod_node))
```
sdudoladov: why is the `new_master_pod_node` not updated after failover ? at this point we still need to check that whatever runs on   
`new_master_pod_node` is actually the master of a given cluster. Otherwise assertion is redundant

e2e/tests/test_smoke.py
FxKu: nope. On normal test setups taints are `[]`

sdudoladov:
@CyberDem0n your review will be much appreciated

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+            elif ('spilo-role', 'replica') in pod.metadata.labels.items():
+                replica_pod_nodes.append(pod.spec.node_name)
+
+        return master_pod_node, replica_pod_nodes
+
+    def wait_for_operator_pod_start(self):
+        self. wait_for_pod_start("name=postgres-operator")
+        # HACK operator must register CRD / add existing PG clusters after pod start up
+        # for local execution ~ 10 seconds suffices
+        time.sleep(30)
```
sdudoladov: this is  sadly  not a  solution: [Travis build](https://travis-ci.org/zalando/postgres-operator/builds/536803211?utm_source=github_status&utm_medium=notification) fails but local one succeed.
does anyone has other ideas on how to circumvent the issue ?

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def __init__(self):
+        self.api = K8sApi()
+
+    def get_spilo_nodes(self, pod_labels, namespace='default'):
+        master_pod_node = ''
+        replica_pod_nodes = []
+        podsList = self.api.core_v1.list_namespaced_pod(namespace, label_selector=pod_labels)
+        for pod in podsList.items:
+            if ('spilo-role', 'master') in pod.metadata.labels.items():
```
CyberDem0n: `if pod.metadata.labels.get('spilo-role') == 'master'`

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+            }
+        }
+
+        # patch node and test if master is failing over to one of the expected nodes
+        k8s.api.core_v1.patch_node(current_master_node, body)
+        k8s.wait_for_master_failover(failover_targets)
+        k8s.wait_for_pod_start('spilo-role=replica')
+
+        new_master_node, new_replica_nodes = k8s.get_spilo_nodes(labels)
+        self.assertTrue(current_master_node != new_master_node,
```
CyberDem0n: assertNotEqual

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+                        "Master on {} did not fail over to one of {}".format(current_master_node, failover_targets))
+        self.assertTrue(num_replicas == len(new_replica_nodes),
```
CyberDem0n: assertEqual

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+               "enableLogicalBackup": True,
+               "logicalBackupSchedule": schedule
+            }
+        }
+        k8s.api.custom_objects_api.patch_namespaced_custom_object(
+            "acid.zalan.do", "v1", "default", "postgresqls", "acid-minimal-cluster", pg_patch_enable_backup)
+        k8s.wait_for_logical_backup_job_creation()
+
+        jobs = k8s.get_logical_backup_job().items
+        self.assertTrue(1 == len(jobs), "Expected 1 logical backup job, found {}".format(len(jobs)))
```
CyberDem0n: assertEqual

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+
+        job = jobs[0]
+        self.assertTrue(job.metadata.name == "logical-backup-acid-minimal-cluster",
```
CyberDem0n: assertEqual

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+                        "Expected job name {}, found {}"
+                        .format("logical-backup-acid-minimal-cluster", job.metadata.name))
+        self.assertTrue(job.spec.schedule == schedule,
```
CyberDem0n: assertEqual

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def wait_for_operator_pod_start(self):
+        self. wait_for_pod_start("name=postgres-operator")
+        # HACK operator must register CRD / add existing PG clusters after pod start up
+        # for local execution ~ 10 seconds suffices
+        time.sleep(30)
+
+    def wait_for_pod_start(self, pod_labels, namespace='default'):
+        pod_phase = 'No pod running'
+        while pod_phase != 'Running':
```
CyberDem0n: What about adding a sleep here in order to avoid hammering the api server?

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+            "spec": {
+                "numberOfInstances": number_of_instances
+            }
+        }
+        _ = self.api.custom_objects_api.patch_namespaced_custom_object(
+                    "acid.zalan.do", "v1", namespace, "postgresqls", "acid-minimal-cluster", body)
+
+        labels = 'version=acid-minimal-cluster'
+        while self.count_pods_with_label(labels) != number_of_instances:
+            pass
```
CyberDem0n: sleep?

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+
+    def count_pods_with_label(self, labels, namespace='default'):
+        return len(self.api.core_v1.list_namespaced_pod(namespace, label_selector=labels).items)
+
+    def wait_for_master_failover(self, expected_master_nodes, namespace='default'):
+        pod_phase = 'Failing over'
+        new_master_node = ''
+        labels = 'spilo-role=master,version=acid-minimal-cluster'
+
+        while (pod_phase != 'Running') or (new_master_node not in expected_master_nodes):
```
CyberDem0n: sleep?

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,323 @@
...
+            if pods:
+                new_master_node = pods[0].spec.node_name
+                pod_phase = pods[0].status.phase
+
+    def get_logical_backup_job(self, namespace='default'):
+        return self.api.batch_v1_beta1.list_namespaced_cron_job(namespace, label_selector="application=spilo")
+
+    def wait_for_logical_backup_job(self, expected_num_of_jobs):
+        while (len(self.get_logical_backup_job().items) != expected_num_of_jobs):
+            pass
```
CyberDem0n: sleep?

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,327 @@
...
+           Add taint "postgres=:NoExecute" to node with master. This must cause a failover.
+        """
+        k8s = self.k8s
+        labels = 'version=acid-minimal-cluster'
+
+        # get nodes of master and replica(s) (expected target of new master)
+        current_master_node, failover_targets = k8s.get_spilo_nodes(labels)
+        num_replicas = len(failover_targets)
+
+        # if all pods live on the same node, failover will happen to other worker(s)
```
sdudoladov: i'd expand this  comment to better explain why we can have `len(failover_targets) == 0`
1. single pod clusters (e.g. in test environments) - a new master pod must be started on another worker node meaning longer downtimes for such clusters
2. clusters that have master/replica on the same node due to unused pod anti-affinity -> a new replica must be started on another worker node for  a failover to take place. 
3. other clusters -> failover takes place / `failover_targets` not empty

e2e/tests/test_e2e.py
```diff
@@ -0,0 +1,327 @@
...
+    '''
+    Wraps around K8 api client and helper methods.
+    '''
+
+    RETRY_TIMEOUT_SEC = 5
+
+    def __init__(self):
+        self.api = K8sApi()
+
+    def get_spilo_nodes(self, pod_labels, namespace='default'):
```
sdudoladov: the signature imo would be better with sth like:
```python
def get_nodes_of_pg_cluster(self, cluster_name, cluster_namespace='default')
```
because
1. nodes are not namespaced objects in k8s -> `namespace` parameter alone is a bit confusing
2. `pod_labels` are not general labels but rather the `version` label that by convention in our code base means  the PG cluster name

sdudoladov:
:+1:

FxKu:
:+1:

PR #554, opened 2019-05-06
Add binries_path_template option.

disco-stu:
The operator expects that PostgreSQL binaries are located at
/usr/lib/postgresql/<pgVersion>/bin. While this is true for Debian and
Ubuntu based images, it doesn't work for images that uses another binary
location. One example is CentOS where binaries are located at
/usr/pgsql-<pgVersion>/bin.

This commit adds a configuration option to make the bin_path template
parameter configurable. If "binaries_path_location" is set it will be
used to generate the bin_path option. If binaries_path_location is
omitted it will still use the build-in default
/usr/lib/postgresql/<pgVersion>/bin.

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -16,6 +16,7 @@ data:
   super_username: postgres
   enable_teams_api: "false"
   spilo_privileged: "false"
+  binaries_path_template: "/usr/lib/postgresql/%s/bin"
   # custom_service_annotations:
   #   "keyx:valuez,keya:valuea"
   # set_memory_request_to_limit: "true"
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -26,6 +26,7 @@ config:
   super_username: postgres
   enable_teams_api: "false"
   spilo_privileged: "false"
+  binaries_path_template: "/usr/lib/postgresql/%s/bin"
   # set_memory_request_to_limit: "true"
   # postgres_superuser_teams: "postgres_superusers"
   # enable_team_superuser: "false"
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -175,6 +175,12 @@ explanation of `ttl` and `loop_wait` parameters.
 * **slots**
   permanent replication slots that Patroni preserves after failover by re-creating them on the new primary immediately after doing a promote. Slots could be reconfigured with the help of `patronictl edit-config`. It is the responsibility of a user to avoid clashes in names between replication slots automatically created by Patroni for cluster members and permanent replication slots. Optional.
 
+* **binaries_path_template**
+  template to use for generating the PostgreSQL binaries path. Patroni
+  uses this template in conjunction with the specified PostgreSQL
+  version to generate the bin_dir location used by patroni. The
+  default is "/usr/lib/postgresql/%s/bin".
+
 ## Postgres container resources
 
 Those parameters define [CPU and memory requests and
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -52,6 +52,7 @@ spec:
         type: logical
         database: foo
         plugin: pgoutput
+    binaries_path_template: "/usr/lib/postgresql/%s/bin"
     ttl: 30
     loop_wait: &loop_wait 10
     retry_timeout: 10
--- a/manifests/postgresql-operator-default-configuration.yaml
+++ b/manifests/postgresql-operator-default-configuration.yaml
@@ -24,6 +24,7 @@ configuration:
     oauth_token_secret_name: postgresql-operator
     pod_role_label: spilo-role
     spilo_privileged: false
+    binaries_path_template: "/usr/lib/postgresql/%s/bin"
     cluster_labels:
         application: spilo
     # inherited_labels:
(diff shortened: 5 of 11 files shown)


erthalion:
Good point, thanks for the PR! There are few tests that are failing due to the new field, could you fix them?

```
util_test.go:502: Marshal Postgresql 
    expected: "{\"kind\":\"Postgresql\",\"apiVersion\":\"acid.zalan.do/v1\",\"metadata\":{\"name\":\"acid-testcluster1\",\"creationTimestamp\":null},\"spec\":{\"postgresql\":{\"version\":\"\",\"parameters\":null},\"volume\":{\"size\":\"\",\"storageClass\":\"\"},\"patroni\":{\"initdb\":null,\"pg_hba\":null,\"ttl\":0,\"loop_wait\":0,\"retry_timeout\":0,\"maximum_lag_on_failover\":0,\"slots\":null},\"resources\":{\"requests\":{\"cpu\":\"\",\"memory\":\"\"},\"limits\":{\"cpu\":\"\",\"memory\":\"\"}},\"teamId\":\"\",\"allowedSourceRanges\":null,\"numberOfInstances\":0,\"users\":null,\"clone\":{}},\"status\":\"Invalid\"}", 
    got:      "{\"kind\":\"Postgresql\",\"apiVersion\":\"acid.zalan.do/v1\",\"metadata\":{\"name\":\"acid-testcluster1\",\"creationTimestamp\":null},\"spec\":{\"postgresql\":{\"version\":\"\",\"parameters\":null},\"volume\":{\"size\":\"\",\"storageClass\":\"\"},\"patroni\":{\"initdb\":null,\"pg_hba\":null,\"ttl\":0,\"loop_wait\":0,\"retry_timeout\":0,\"maximum_lag_on_failover\":0,\"slots\":null,\"binaries_path_template\":\"\"},\"resources\":{\"requests\":{\"cpu\":\"\",\"memory\":\"\"},\"limits\":{\"cpu\":\"\",\"memory\":\"\"}},\"teamId\":\"\",\"allowedSourceRanges\":null,\"numberOfInstances\":0,\"users\":null,\"clone\":{}},\"status\":\"Invalid\"}"
```

CyberDem0n:
The more appropriate way would be specifying the desired postgres version in the `PGVERSION` env variable.
Spilo already [supports](https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py#L823-L828) it for quite a while.

manifests/configmap.yaml
```diff
@@ -16,6 +16,7 @@ data:
   super_username: postgres
   enable_teams_api: "false"
   spilo_privileged: "false"
+  binaries_path_template: "/usr/lib/postgresql/%s/bin"
```
erthalion: I see that you've added also a new field to the operator configuration, but I'm afraid it's not used, since it's not in the `Config` data structure. It's totally make sense to have it on the operator level (in case if one wants to change the binary path for all the databases), but it needs to be present in `Config`, corresponding CRD for configuration and Helm chart (these are different ways of configuring the operator) taking into account that it can be overwritten on the database config level. Could you add it too?

erthalion:
> The more appropriate way would be specifying the desired postgres version in the PGVERSION env variable.
> Spilo already supports it for quite a while.

Yes, probably. Then we can reformulate this PR into getting rid of hardcoded binary path to provide a PGVERSION.

manifests/configmap.yaml
FxKu: Just fyi, [here](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters) is a list what has to be done when introducing new config parameters.

disco-stu:
Thanks for all the input. I'll come up with a updated pull request within the next days.

disco-stu:
I've updated the pull request according to the hints that @erthalion and @FxKu gave me. Now, all unit tests passes. However, I don't finished my own integration tests yet.

@CyberDem0n If I'm right: specifying PGVERSION isn't an option since /var/lib/postgresql/.../bin is hardcoded inside the operator. "configure_spilo.py" only replaces "%s" in the hardcoded bin_path. This doesn't provide enough flexibility to work with CentOS / RHEL based images or any other binary location except /var/lib/postgresql/%s/bin. Please correct me if I'm wrong. Any input on this would be great.

Please let me know if something else needs to be changed.

erthalion:
If I understood @CyberDem0n correctly, the idea with `PGVERSION` is that operator needs to provide only this version (just a number, so we will remove any binary path from the code completely), and the correct binary path can be specified in the docker image, that will be used.

Jan-M:
Let's try to not push this problem to the operator level rather than just providing the PGVERSION to the container and let the image figure it out. (as @erthalion @CyberDem0n already pointed out).

FxKu:
Can we close this PR then? @disco-stu does is work with the `env` variable?

disco-stu:
@FxKu: I solved my problem by creating a simple symlink from /usr/lib/postgresql/[...]/ to /usr/pgsq-[...]/. This is a rather ugly solution but it solves my problem for now.

Regarding this patch: AFAICT the operator always creates a "bin_dir" parameter (inside the environment variable SPILO_CONFIGURATION) pointing to the debian style version specific binary path (E.g. /usr/lib/postgresql/[...]/bin).

It's possible to define another bin_dir inside TEMPLATE provided by "configure_splio.py", but this value will be overwritten by the operator anyway.

AFAIK there are three sections that are capable of setting "bin_dir":
  1. The Operator
  2. configure_spilio.py (TEMPLATE)
  3. configure_spilo.py ([line 845](https://github.com/zalando/spilo/blob/ddab145098db8a0f2adeed6d4c7d2cdf00c6f37a/postgres-appliance/scripts/configure_spilo.py#L845))

Both lists (1. and 2.) are merged by [deep_update](https://github.com/zalando/spilo/blob/ddab145098db8a0f2adeed6d4c7d2cdf00c6f37a/postgres-appliance/scripts/configure_spilo.py#L843). Where a is the configuration provided by the operator and b is the config generate by spilo. Because values specified in a don't get overwritten by values specified in b, the value of bin_dir that is specified by the operator will have precedence over any other value.

Line 845 and following aren't even called if bin_dir is already set.

PR #558, opened 2019-05-07
patch CRD on operator update

FxKu:
Simply update does not work --> https://github.com/kubernetes/kubernetes/issues/70674
Now `createOperatorCRD` still returns errors, but maybe warnings are better? See #444


pkg/controller/util.go
```diff
@@ -52,7 +54,15 @@ func (c *Controller) createOperatorCRD(crd *apiextv1beta1.CustomResourceDefiniti
...
 			return fmt.Errorf("could not create customResourceDefinition: %v", err)
 		}
-		c.logger.Infof("customResourceDefinition %q is already registered", crd.Name)
+		c.logger.Infof("customResourceDefinition %q is already registered and will only be updated", crd.Name)
+
+		patch, err := json.Marshal(crd)
+		if err != nil {
+			return fmt.Errorf("could not marshal new customResourceDefintion: %v", err)
+		}
+		if _, err := c.KubeClient.CustomResourceDefinitions().Patch(crd.Name, types.MergePatchType, patch); err != nil {
```
sdudoladov: how does this patch affect existing Postgres clusters when it actually adds the `/status` subresource  ?

ref: [original problem](https://github.com/zalando/postgres-operator/pull/534#issuecomment-490040234)

FxKu: I've edited the deployment and replaced the image - this caused a rolling update and spilo failover.

sdudoladov:
:+1:

FxKu:
:+1:

Issue #545, opened 2019-04-18
Invalid value for enable_pod_antiaffinity in default CRD configuration


IlyaSemenov:
The default CRD configuration (https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql-operator-default-configuration.yaml) can't be loaded by the latest operator:

```
time="2019-04-18T07:11:37Z" level=fatal msg="unable to read operator configuration: could not get operator configuration object \"postgresql-operator-configuration\": v1.OperatorConfiguration.Configuration: v1.OperatorConfigurationData.Kubernetes: v1.KubernetesMetaConfiguration.EnablePodAntiAffinity: ReadBool: expect t or f, but found \", error found in #10 byte of ...|ffinity\":\"false\",\"oa|..., bigger context ...|_label\":\"cluster-name\",\"enable_pod_antiaffinity\":\"false\",\"oauth_token_secret_name\":\"postgresql-opera|..." pkg=controller
```

it should be `enable_pod_antiaffinity: false` not `enable_pod_antiaffinity: "false"`

FxKu:
Fixed by merging PR #559. Thanks for notifying.

PR #559, opened 2019-05-08
Add CRD configuration to Helm chart values.yaml

### LINKED ISSUE #539: Fix RBAC defined by the Helm chart
running the operator with our Helm chart currently produces the following error:

> time="2019-04-09T19:44:37Z" level=warning msg="Pods and/or Patroni may misfunction due to the lack of permissions: could not create role binding zalando-postgres-operator : cannot bind the pod service account \"postgres-operator\" defined in the config map to the cluster role in the \"resource-usage-collector\" namespace: clusterroles.rbac.authorization.k8s.io \"zalando-postgres-operator\" not found" pkg=controller

this happens because the example RBAC configuration we provide - and the one our operator expects for local execution - uses the "zalando-postgres-operator"  name,  while the Helm uses the simple "postgres-operator" phrase. The easiest way to adjust this is to change the name in Helm. 

Olaf Klischat, thank you for the detailed issue report.

@FxKu pls have a look

FxKu:
Open question:
- Quotes can be a problem, e.g. in CRD it's `false` and in ConfigMap it's `"false"`
- CRD can handle more nested config parameters. Can it be flattened?

fix #539


sdudoladov:
note the order of deployment is different with CRD : you first submit the operator deployment and then the crd object for the operator configuration. Otherwise it fails with:
```
Error: validation failed: unable to recognize "": no matches for kind "OperatorConfiguration" in version "acid.zalan.do/v1"
```

with the configmap it is the reverse: first submit the configmap, then the deployement

FxKu:
Found that there is a `crd-install` hook that can be used for CRD definitions. Then CRs can be used within the same chart. Haven't checked for parsing errors, yet.

sdudoladov:
:+1:

FxKu:
I gave it another try by removing the `configmap.yaml` from the templates folder and getting expected errors while parsing the OperatorConfiguration. But maybe this can be resolved easily in the `operatorconfiguration.yaml`.

we can also fix #545 with this PR. Makes me wonder if params such as `default_cpu_limit` should come with quotes...

FxKu:
There are still issues. For some reason the default service account is used in the Pod template:
```time="2019-05-15T09:51:05Z" level=warning msg="Pods and/or Patroni may misfunction due to the lack of permissions: could not create pod service account  : resource name may not be empty" pkg=controller
time="2019-05-15T09:51:05Z" level=info msg="creation of the cluster started" cluster-name=default/acid-minimal-cluster pkg=controller worker=0```

And the resulting error:
```2019-05-15 09:44:58,627 WARNING: Can not get cluster from dcs
2019-05-15 09:45:03,637 ERROR: get_cluster
  File "/usr/local/lib/python3.6/dist-packages/patroni/dcs/kubernetes.py", line 146, in _load_cluster
    response = self.retry(self._api.list_namespaced_pod, self._namespace, label_selector=self._label_selector)```

When using the ConfigMap, I got the same problem first, but now it the operator pod doesn't start at all:
```panic: strconv.ParseBool: parsing "false ": invalid syntax

goroutine 1 [running]:
github.com/zalando/postgres-operator/pkg/util/config.NewFromMap(0xc000233f20, 0xc000042010)
	/root/go/src/github.com/zalando/postgres-operator/pkg/util/config/config.go:146 +0x226
```

As discussed yesterday, we might drop ConfigMap template from Helm chart.

Jan-M:
:+1:

charts/postgres-operator/Chart.yaml
```diff
@@ -1,13 +1,18 @@
...
 description: Postgres operator creates and manages PostgreSQL clusters running in Kubernetes
 keywords:
 - postgres
 - operator
+- cloud-native
+- patroni
+- spilo
 maintainers:
+- name: Zalando
+  email: opensource@zalando.de
```
sdudoladov: I believe maintainers in this case are we aka  "Team ACID @ Zalando <team-acid@zalando.de>"

sdudoladov:
@FxKu running this with a CRD-based configuration still produces an error:
```
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"pods is forbidden: User \"system:serviceaccount:default:operator\" cannot list resource \"pods\" in API group \"\" in the namespace \"default\"","reason":"Forbidden","details":{"kind":"pods"},"code":403}
```

FxKu:
When using only the CRD parameters you have to set the serviceAccount name to `"operator"` in the [values.yaml](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values.yaml). Not very intuitive, I know. I've updated the docs to point users to this.

As an alternative, we could use the release name as the default name for the service account. This is how it's handled in the ConfigMap. Would also save some if blocks in the yaml files for the helm chart. Can this have side effects, as we have the default name "operator" hard-coded, e.g. [here](https://github.com/zalando/postgres-operator/blob/master/pkg/controller/controller.go#L168).

charts/postgres-operator/Chart.yaml
FxKu: should our team mail really be that visible? I think, it's fine this way.

sdudoladov:
:+1:

FxKu:
:+1:

PR #560, opened 2019-05-10
specify ReadOnlyRootFilesystem: false for pod security policies

knan-nrk:
spilo depends on ReadOnlyRootFilesystem: false, but the operator does
not explicitly specify it. And a default pod security policy may change
the default to ReadOnlyRootFilesystem: true.

Explicitly specify ReadOnlyRootFilesystem: false so kubernetes can pick
a less restrictive policy the operator has access to.

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -360,6 +360,8 @@ func generateSpiloContainer(
 	volumeMounts []v1.VolumeMount,
 	privilegedMode bool,
 ) *v1.Container {
+	falseBool := false
+
 	return &v1.Container{
 		Name:            name,
 		Image:           *dockerImage,
@@ -383,6 +385,7 @@ func generateSpiloContainer(
 		Env:          envVars,
 		SecurityContext: &v1.SecurityContext{
 			Privileged: &privilegedMode,
+			ReadOnlyRootFilesystem: &falseBool,
 		},
 	}
 }


knan-nrk:
Without this, the spilo pods gets assigned the default restrictive policy of our cluster and goes into CrashLoopBackoff state because of the read only root filesystem.

FxKu:
If your policy requires this parameter to be true wouldn't this PR then silently violate it. What is the error message that you are getting? Maybe it's actually a good thing that it fails :)

There's another PR to run Spilo as non-root user #531 - using a custom image though. Not sure if the `ReadOnlyRootFilesystem` can be true then.

knan-nrk:
Okay, let me re-explain. Our kubernetes cluster basically has two Pod Security Policies - restrictive and permissive. By default, service users only have access to the restrictive policy, which requires readonlyrootfilesystem = true and changes the default for pods to readonlyrootfilesystem = true.

With the default silently changed to readonlyrootfilesystem=true, the unpacking of a.tar.xz inside the image fails and the container goes into crashloopbackoff.

We have explicitly given the postgres-operator service user access to the permissive policy via a rolebinding, so it can use both policies. But which policy is chosen for use is dependent on what the pod/deployment/etc requests, and can not be specified directly. If the pod does not explicity request readonlyrootfilesystem=false like this PR does, the permissive policy is never chosen even though it is available, since it does not look like the pod needs it.

This only specifies more explicitly that readonlyrootfilesystem=false is required by the spilo pod, so kubernetes can choose correctly what to do. This is true both for master and #531 when the spilo image is built with compress=true.

Ping me on slack if you want more clarification?

knan-nrk:
Any more comments on this one? It should not break anything in existing installations or clusters.

Jan-M:
@knan-nrk @FxKu does #582 not help you? What are we missing to enable you? Will read the longer message in a few minutes.

FxKu:
@Jan-M I had the same question in the other PR :)
This here is for Spilo. The other one was for the operator container.

I was just wondering if we should define more hard-coded parameters regarding the [securityContext](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#securitycontext-v1-core). But, we could also start with this PR and and more in the future, if we or other users need it.

FxKu:
:+1:

Jan-M:
:+1:

Issue #521, opened 2019-03-15
Cluster domain hard-coded


mxdrenthe:
Problem:
On a barematel kubernetes installation cluster name isn't `cluster.local`
Because of hard-coded cluster name connecting to the instance is impossible.

Error message:
`level=error msg="could not connect to PostgreSQL database: dial tcp: lookup acid-minimal-cluster.default.svc.cluster.local on xxx.xxx.xxx.xxx:xxxx: no such host" cluster-name=default/acid-minimal-cluster pkg=cluster`

Spotted line:
https://github.com/zalando/postgres-operator/blob/5f6abfb7977abf0ba0c05f502a119e565cc6e1e6/pkg/cluster/database.go#L37

redbaron:
It shouldn't be hardcoded indeed. As workaround you can make in cluster DNS
to serve both cluster.local and your current cluster zone

mxdrenthe:
The workaround indeed solves the finding containers.
Noticed that the services created the selectors won't match the created pods.
label: `spilo-role: xxx` is missing

FxKu:
PR #561 got merged. Should be solved now. Open a new issue if there are still problems regarding services.

PR #561, opened 2019-05-10
database.go: remove hardcoded .svc.cluster.local dns suffix

knan-nrk:
Use the pod's configured dns search path, for clusters
where .svc.cluster.local is not correct.

### DIFF
--- a/pkg/cluster/database.go
+++ b/pkg/cluster/database.go
@@ -34,7 +34,7 @@ func (c *Cluster) pgConnectionString() string {
 	password := c.systemUsers[constants.SuperuserKeyName].Password
 
 	return fmt.Sprintf("host='%s' dbname=postgres sslmode=require user='%s' password='%s' connect_timeout='%d'",
-		fmt.Sprintf("%s.%s", c.Name, c.Namespace),
+		fmt.Sprintf("%s.%s.svc.%s", c.Name, c.Namespace, c.OpConfig.ClusterDomain),
 		c.systemUsers[constants.SuperuserKeyName].Name,
 		strings.Replace(password, "$", "\\$", -1),
 		constants.PostgresConnectTimeout/time.Second)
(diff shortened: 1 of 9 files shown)


erthalion:
Related to #521 I guess

pkg/cluster/database.go
```diff
@@ -34,7 +34,7 @@ func (c *Cluster) pgConnectionString() string {
 	password := c.systemUsers[constants.SuperuserKeyName].Password
 
 	return fmt.Sprintf("host='%s' dbname=postgres sslmode=require user='%s' password='%s' connect_timeout='%d'",
-		fmt.Sprintf("%s.%s.svc.cluster.local", c.Name, c.Namespace),
```
FxKu: While you are right, that we should not enforce `cluster.local`, we should still use a FQDN. Without the suffix we put more load on the DNS to figure out the full address itself. It's better to define the Cluster Domain for the operator and use it here.

knan-nrk: True. This works but will be slow at connecting because of ndots:5 and such, and I can see it causing problems down the line. It's rather annoying that Cluster Domain can't be queried from the kubernetes api.

I'll look into creating a config entry for it.

FxKu: I've [documented](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters) the required steps on how to introduce additional config parameters, recently. Hope it helps.

FxKu:
@knan-nrk any news? Did you manage to create a configuration parameter for the cluster domain?

knan-nrk:
@FxKu implemented a config parameter, as suggested.Tested manually with the config parameter set and unset.

(Had a look at the test code and it is impenetrable to me, being a go novice. No idea how to build a test for cluster:database:pgConnectionString.)

Jan-M:
Looks good to me? @erthalion @sdudoladov

sdudoladov:
:+1:

erthalion:
:

erthalion:
:+1:

sdudoladov:
@knan-nrk thanks for the yet another contribution :D

PR #507, opened 2019-02-28
Standby cluster rebase

redbaron:
Support for standby cluster mode

xref: https://github.com/zalando/postgres-operator/issues/428

PR #562, opened 2019-05-10
Adding the support for standby cluster

RafiaSabih:
This is to enable the standby clusters via postgres-operator on Kubernetes.

More about the standby clusters supported by patroni can be find [here](https://github.com/zalando/patroni/blob/bd2c54581abb42a7d3a3da551edf0b8732eefd27/docs/replica_bootstrap.rst#standby-cluster)


sdudoladov:
have you already seen #507 ?

sdudoladov:
@RafiaSabih 
1. what is the status of this PR ? 
2. Please add the administrator/user  documentation explaining what  standby clusters are and how they can be used with links to Patroni/Spilo docs when necessary

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -109,6 +110,13 @@ type Patroni struct {
 	Slots                map[string]map[string]string `json:"slots"`
 }
 
+//StandbyCluster
+type StandbyDescription struct {
+	ClusterName string `json:"cluster,omitempty"`
+	UID         string `json:"uid,omitempty"`
```
sdudoladov: is it the UID of the source cluster from where the standby cluster is streaming ?

docs/reference/cluster_manifest.md
```diff
@@ -175,6 +175,9 @@ explanation of `ttl` and `loop_wait` parameters.
 * **slots**
   permanent replication slots that Patroni preserves after failover by re-creating them on the new primary immediately after doing a promote. Slots could be reconfigured with the help of `patronictl edit-config`. It is the responsibility of a user to avoid clashes in names between replication slots automatically created by Patroni for cluster members and permanent replication slots. Optional.
 
+* **standby_cluster**
```
sdudoladov: please also extend the `manifests/complete-postgres-manifest.yaml` with a commented-out example of how this new field can be used

pkg/cluster/k8sres.go
```diff
@@ -1253,6 +1257,34 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
 	return result
 }
 
+func (c *Cluster) generateStandbyEnvironment(description *acidv1.StandbyDescription) []v1.EnvVar {
+	result := make([]v1.EnvVar, 0)
+
+	if description.S3WalPath == "" {
```
sdudoladov: does it mean standby that bypasses an S3 bucket with archives and streams directly from the source cluster ?

pkg/apis/acid.zalan.do/v1/util_test.go
```diff
@@ -328,7 +328,7 @@ var unmarshalCluster = []struct {
...
 		},
-		marshal: []byte(`{"kind":"Postgresql","apiVersion":"acid.zalan.do/v1","metadata":{"name":"teapot-testcluster1","creationTimestamp":null},"spec":{"postgresql":{"version":"","parameters":null},"volume":{"size":"","storageClass":""},"patroni":{"initdb":null,"pg_hba":null,"ttl":0,"loop_wait":0,"retry_timeout":0,"maximum_lag_on_failover":0,"slots":null},"resources":{"requests":{"cpu":"","memory":""},"limits":{"cpu":"","memory":""}},"teamId":"acid","allowedSourceRanges":null,"numberOfInstances":0,"users":null,"clone":{}},"status":{"PostgresClusterStatus":"Invalid"}}`),
+		marshal: []byte(`{"kind":"Postgresql","apiVersion":"acid.zalan.do/v1","metadata":{"name":"teapot-testcluster1","creationTimestamp":null},"spec":{"postgresql":{"version":"","parameters":null},"volume":{"size":"","storageClass":""},"patroni":{"initdb":null,"pg_hba":null,"ttl":0,"loop_wait":0,"retry_timeout":0,"maximum_lag_on_failover":0,"slots":null} ,"resources":{"requests":{"cpu":"","memory":""},"limits":{"cpu":"","memory":""}},"teamId":"acid","allowedSourceRanges":null,"numberOfInstances":0,"users":null,"clone":{},"standby_cluster":{}},"status":{"PostgresClusterStatus":"Invalid"}}`),
```
FxKu: could you add or edit one example where the `standby_cluster` parameter is not empty?

pkg/apis/acid.zalan.do/v1/postgresql_type.go
RafiaSabih: yes

docs/reference/cluster_manifest.md
RafiaSabih: Done.

pkg/apis/acid.zalan.do/v1/util_test.go
RafiaSabih: Well, tried now.

pkg/cluster/k8sres.go
RafiaSabih: well yes that is the intention.

docs/user.md
```diff
@@ -254,6 +254,18 @@ metadata:
 Note that timezone is required for `timestamp`. Otherwise, offset is relative
 to UTC, see [RFC 3339 section 5.6) 3339 section 5.6](https://www.ietf.org/rfc/rfc3339.txt).
 
+## Setting up a standby cluster
+
+Standby clusters are like normal cluster but they are streaming from a remote master. Patroni supports it, [read this](https://github.com/zalando/patroni/blob/bd2c54581abb42a7d3a3da551edf0b8732eefd27/docs/replica_bootstrap.rst#standby-cluster) to know more about them. Currently operator supports setting up standby only through an S3 bucket. For this you need to add a section standby in the YAML file as follows.
```
sdudoladov: > Currently operator supports setting up standby only through an S3 bucket.

This part needs more details. Earlier it is mentioned that a standby cluster streams from some remote cluster and then suddenly we are talking about an S3 bucket. For people not that familiar with our particular setup it may sound confusing.

docs/user.md
```diff
@@ -254,6 +254,18 @@ metadata:
...
+
+```yaml
+spec:
+  standby:
+    uid: "UID of the master cluster"
+    cluster: "Name of the master cluster"
```
sdudoladov: can a standby cluster have the same name as the master cluster ? i'd say if they are in the same k8s cluster definitely not, but if they are in different k8s clusters it should work.

manifests/complete-postgres-manifest.yaml
```diff
@@ -64,6 +64,11 @@ spec:
   #  cluster: "acid-batman"
   #  timestamp: "2017-12-19T12:40:33+01:00" # timezone required (offset relative to UTC, see RFC 3339 section 5.6)
   #  s3_wal_path: "s3://custom/path/to/bucket"
+  # Make this a standby cluster and mention the information of source cluster from where to do continuous streaming.
+  # standby:
+  #  uid: "uid of the master"
```
sdudoladov: k8s uuid of the master *cluster*,  this reads a bit confusing right now

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -109,6 +110,13 @@ type Patroni struct {
 	Slots                map[string]map[string]string `json:"slots"`
 }
 
+//StandbyCluster
+type StandbyDescription struct {
+	ClusterName string `json:"cluster,omitempty"`
```
sdudoladov: master cluster name ?

pkg/cluster/k8sres.go
sdudoladov: 1. that behavior should be documented then
2. is it really the case ? looks like if `S3WalPath` evaluates to the empty string the standby env is empty

pkg/cluster/k8sres.go
```diff
@@ -1253,6 +1257,34 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
...
+
+	msg = "Use custom parsed S3WalPath %s from the manifest"
+	c.logger.Warningf(msg, description.S3WalPath)
+
+		result = append(result, v1.EnvVar{
+			Name:  "STANDBY_WALE_S3_PREFIX",
+			Value: description.S3WalPath,
+		})
+
+	result = append(result, v1.EnvVar{Name: "STANDBY_METHOD", Value: "STANDBY_WITH_WALE"})
```
sdudoladov: 1. is there a plan/need to support  `STANDBY_WITH_WAL-G` ?
2. please document the env vars used here in [the Spilo environment doc ](https://github.com/zalando/spilo/blob/96813721e9b23cbe9d2511ffdf93db2a4d73f7ed/ENVIRONMENT.rst)

docs/reference/cluster_manifest.md
```diff
@@ -175,6 +175,9 @@ explanation of `ttl` and `loop_wait` parameters.
...
+  initializes cluster as a standby creating a cascading replication, where elected master is streaming from specified remote location
```
sdudoladov: elected master == [standby leader ](https://patroni.readthedocs.io/en/latest/replica_bootstrap.html)? let's use the uniform terminology everywhere

docs/user.md
```diff
@@ -254,6 +254,23 @@ metadata:
...
+
+```yaml
+spec:
+  standby:
+    s3_wal_path: "s3 bucket path to the master"
+```
+Things to note:
+
+- An empty string is provided in s3_wal_path of the standby cluster will result in error and no statefulset will be created.
+- If standby is not needed, simply remove the standby_cluster section from the patroni through patronictl edit-config.
```
sdudoladov: this mean the standby leader will be promoted to the normal master or ?

Jan-M: Let's remove this second part or make it an "active" sentence like:

To promote a standby cluster manually, use patronictl and remove config entry.

sdudoladov:
1. Do we want to impose restrictions on the PG manifest of standby ? The `clone` and `standby` section are clearly incompatible in the present implementation with each other for example. Definitions of new users / DBs also will not work until the standby leader is promoted; PG versions should match and so on.

2. do we need sth like `spilo-role=standby-leader` label on pods ?

3. The operator should warn about and ignore `UPDATE`/`SYNC` events that add a standby section to the manifest (the latter can happen if a manifest was updated while the operator was down )

4. some operator log samples with comments from the example run of this PR: 

>  level=info msg="Standby from S3 buckets3://
>  level=warning msg="Use custom parsed S3WalPath

these messages follow each other in the logs and have the identical content but different log levels. One info log record should be enough.

>  msg="error while syncing cluster state: could not sync statefulsets: could not generate statefulset: s3_wal_path is empty for standby cluster" 

 this msg is misplaced  though conceptually correct . What cannot be generated here is the Spilo container environment (hence pod template / stateful set) because of the invalid `s3_wal_path` . It will be better to place this around pod template generation and/or expend the actual message.
We can actually set the[`PostgresClusterStatus`](https://github.com/zalando/postgres-operator/blob/0fbfbb23bbd5a534bab7565498b6e48e8ffe54d2/pkg/apis/acid.zalan.do/v1/const.go#L4) to  `ClusterStatusAddFailed` or sth similar when the operator processes the `CREATE` event.

>  error while syncing cluster state: could not sync roles: could not init db connection: could not init db connection: pq: password authentication failed for user "postgres"

So  the actual password leaves within the standby leader, which fails a Sync for standby clusters. We need to  acknowledge this situation when we attempt to sync users/DBs. I wonder if the relevant secrets for standbys can be just manually created with the operator simply skipping their creation and logging a warning instead.

This presents problems for replication as well:
```
2019-06-13 14:27:28.354 UTC,"standby","",13240,"127.0.0.1:58218",5d025d50.33b8,2,"authentication",2019-06-13 14:27
:28 UTC,4/1985,0,FATAL,28000,"no pg_hba.conf entry for replication connection from host ""127.0.0.1"", user ""standby"", SSL off",,,,,,,,,""
2019-06-13 14:27:29.420 UTC,,,13359,,5d025d51.342f,1,,2019-06-13 14:27:29 UTC,,0,FATAL,XX000,"could not connect to the primary server: FATAL:  password authentication failed for user ""standby""
FATAL:  no pg_hba.conf entry for replication connection from host ""10.2.1.38"", user ""standby"", SSL off",,,,,,,,,""
```

```
2019-06-13 14:23:19.137 UTC,"standby","",28157,"10.2.1.38:57236",5d025c57.6dfd,2,"authentication",2019-06-13 14:23:19 UTC,5/4160,0,FATAL,28P01,"password authentication failed for user ""standby""","Password does not match for user ""standby"".
```

Jan-M:
I would look at the log output, that should be mostly "info" and also visible to the UI API if not already.

The non sync of users and this error reporting is fine, I see this as future work. Some toggle around the user sync when operator is aware of this being a standby cluster.

As for the adding and removing of the standby section, this is part of future investigation of how patroni behaves. We may want to toggle into standby cluster in the future on low disk space scenario (prevent writes).

Jan-M:
Replication password errors for now solvable by 1 pod standby deployment only. Rest is future work.

FxKu:
As @sdudoladov already mentioned, having a dedicated spilo-role label for `standy` would be nice. I would just call it `standby`.

To avoid confusion on creating a standby cluster, I would suggest to provide an additional minimal manifest, instead of hiding it in the complete example. It should not contain database related stuff because it's all shipped from WAL, right? And `numberOfInstances` should be 1.

docs/user.md
```diff
@@ -281,6 +281,25 @@ spec:
     s3_force_path_style: true
 ```
 
+## Setting up a standby cluster
+
+Standby clusters are like normal cluster but they are streaming from a remote cluster. Patroni supports it, [read this](https://github.com/zalando/patroni/blob/bd2c54581abb42a7d3a3da551edf0b8732eefd27/docs/replica_bootstrap.rst#standby-cluster) to know more about them.
```
redbaron: This MR builds standby cluster using WAL shipping  for both initial bootstrap and keeping it up to date, right? this is not how Patroni's `standby_cluster` works references here works, which configures a streaming replication.  Maybe mode of operation should be reflected somehow in the spec key name to avoid confusion?

FxKu: Hm, yes this is a bit confusing. Maybe this phrase can be moved somewhere to the end of this paragraph - or maybe just reuse the link for the next mentioning of standby_cluster. In the future we can indeed think about adding a `mode` key for different standby settings.

docs/reference/cluster_manifest.md
```diff
@@ -199,6 +199,9 @@ explanation of `ttl` and `loop_wait` parameters.
   automatically created by Patroni for cluster members and permanent replication
   slots. Optional.
 
+* **standby_cluster**
```
FxKu: just **standby** to align with manifests and Postgresql_type. Also in util_tests.

docs/user.md
redbaron: every mode would require own parameters, IMHO a bit of future proofing might help, such as:

```
standby:
  wal_shipping:
     s3_bucket: ....
```

then future streaming replication would look like:

```
stanby:
  streaming:
     host: ...
     slot: ...
```

pkg/cluster/k8sres.go
redbaron: where is implementation for these env vars? I can't find any mentions of it in https://github.com/zalando/spilo

FxKu:
:+1:

docs/user.md
```diff
@@ -281,6 +281,23 @@ spec:
     s3_force_path_style: true
 ```
 
+## Setting up a standby cluster
+
+Standby clusters are like normal cluster but they are streaming from a remote cluster. As the first version of this feature, the only scenario covered by operator is to stream from a wal archive of the master. Following the more popular infrastructure of using Amazon's S3 buckets, it is mentioned  as s3_wal_path here. To make a cluster as standby add a section standby in the YAML file as follows.
+
+```yaml
+spec:
+  standby:
```
redbaron: there would be multiple modes of standby operation and every mode requires own set of configs. IMHO a bit of future proofing might help, such as:
```
standby:
  wal_shipping:
     s3_bucket: ....
```
then future streaming replication would look like:

```
standby:
  streaming:
     host: ...
     slot: ...
```

RafiaSabih:
:+1:

pkg/cluster/k8sres.go
RafiaSabih: Here is the PR for doing the same --> https://github.com/zalando/spilo/pull/343 feel free to comment there.

redbaron:
@sdudoladov , @FxKu , @RafiaSabih ,

together with https://github.com/zalando/spilo/blob/master/postgres-appliance/scripts/configure_spilo.py I can see that this PR is another syntax for Patroni's `bootstrap.dcs.standby_cluster` .

May I ask why this approach was taken instead of exposing `bootstrap.dcs.standby_cluster` directly in Postgres CRD like it was done https://github.com/zalando/postgres-operator/pull/507 ? Asking because I need to update it to actually add support for streaming replication and resync with you ,so that we dont run on a fork forever.

RafiaSabih:
Well, following the requirements for our enterprise, we only have the case where we need standby which can stream from wal-path of a source cluster. Plus, our aim was to keep the manifest simple enough and it seemed too much to expose the details of bootstrap.dcs.standby_cluster in the postgres-manifest, specially when none of those settings needs to be altered, for our use-case atleast.

PR #563, opened 2019-05-13
fix statefulset sync

sdudoladov:
removes excessive syncing of stateful sets

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -8,7 +8,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/api/apps/v1beta1"
-	"k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


Jan-M:
How are we testing this in kind for example to make sure it is solving the problem?

sdudoladov:
```bash
kind start cluster --name kind-m-2 --config ~/.kube/kind-config-multikind.yaml --loglevel debug
kind load docker-image "registry.opensource.zalan.do/acid/postgres-operator:v1.1.0-23-gf29bdaf-dirty" --name kind-m
# submit operator as in quickstart but with the new image version in the deployment
kubectl create -f manifests/minimal-postgres-manifest.yaml
```
restart the operator pod to force a SYNC and see how the unnecessary Sync does not happen

sdudoladov:
:+1:

FxKu:
:+1:

PR #573, opened 2019-05-27
Improve cluster sidecar documentation

### LINKED ISSUE #413: Make default resources configurable for sidecars
At present all sidecars except Scalyr use [the same resource requirements/limits](https://github.com/zalando-incubator/postgres-operator/blob/96e3ea951178f385170d8930483a847b9bc30062/pkg/cluster/k8sres.go#L370) `defaultResources` as the Postgres container, which may be wasteful. We shall modify the [PostgreSQL spec/manifest](https://github.com/zalando-incubator/postgres-operator/blob/86ba92ad020f49751c7e60be92ea51ff97613b28/pkg/apis/acid.zalan.do/v1/postgresql_type.go#L52) to enable configuring defaults on the per-sidecar basis, similar to what we do for the Scalyr.

knan-nrk:
UPD closes #413

### DIFF
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -262,3 +262,36 @@ defined in the sidecar dictionary:
   a dictionary of environment variables. Use usual Kubernetes definition
   (https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/)
   for environment variables. Optional.
+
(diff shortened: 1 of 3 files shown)


sdudoladov:
:+1:

erthalion:
:+1:

sdudoladov:
@knan-nrk thank you for the contribution :)

Issue #578, opened 2019-06-04
Bug: Podserviceaccount and clusterrole cannot have different names


knan-nrk:
Currently, controller:initRoleBinding assumes that the clusterrole and pod serviceaccount have the same name, and overwrites roleRef.name with podserviceaccount.name. Which seems like a bug.

Do we want clusterrole name as a config parameter, perhaps?

I'll send a patch decoupling them, at least.

sdudoladov:
@FxKu please have a look

FxKu:
It got introduced with #529 to fix #528.

We have a hard-coded default to `zalando-postgres-operator` RoleRef. When using the Helm chart a random release name can be generated so we have to update the RoleRef to make the role binding work. But yes, if the clusterrole has a different name the binding will be wrong. The hard-coded default should instead use `c.PodServiceAccount.Name`.

The default is only used if `PodServiceAccountRoleBindingDefinition` is empty which - interestingly - is a configuration parameter already. So by adding this (admittedly complex) parameter to the config it should be possible to have a different name for the clusterrole.

FxKu:
PR #581 got merged.

PR #581, opened 2019-06-06
decouple clusterrole name and serviceaccount name

knan-nrk:
Improves the rolebinding name to `"<serviceaccount>:<clusterrole>"`, and
does not overwrite clusterrole reference with serviceaccount name.

Suggested patch for #578 

Tested working for us, using configmap for configuration, with the default zalando-postgres-operator clusterrole name, and a custom podserviceaccountname.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -230,8 +230,7 @@ func (c *Controller) initRoleBinding() {
 	default:
 		c.PodServiceAccountRoleBinding = obj.(*rbacv1beta1.RoleBinding)
 		c.PodServiceAccountRoleBinding.Namespace = ""
-		c.PodServiceAccountRoleBinding.ObjectMeta.Name = c.PodServiceAccount.Name
-		c.PodServiceAccountRoleBinding.RoleRef.Name = c.PodServiceAccount.Name
+		c.PodServiceAccountRoleBinding.ObjectMeta.Name = fmt.Sprintf("%s:%s", c.PodServiceAccount.Name, "zalando-postgres-operator")
 		c.PodServiceAccountRoleBinding.Subjects[0].Name = c.PodServiceAccount.Name
 		c.logger.Info("successfully parsed")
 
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -230,8 +230,7 @@ func (c *Controller) initRoleBinding() {
 	default:
 		c.PodServiceAccountRoleBinding = obj.(*rbacv1beta1.RoleBinding)
 		c.PodServiceAccountRoleBinding.Namespace = ""
-		c.PodServiceAccountRoleBinding.ObjectMeta.Name = c.PodServiceAccount.Name
-		c.PodServiceAccountRoleBinding.RoleRef.Name = c.PodServiceAccount.Name
+		c.PodServiceAccountRoleBinding.ObjectMeta.Name = fmt.Sprintf("%s:%s", c.PodServiceAccount.Name, "zalando-postgres-operator")
```
FxKu: Not sure if this will work with helm chart. Could you instead use `fmt.Sprintf` for the hard-coded `PodServiceAccountRoleBindingDefinition` and use `c.PodServiceAccount.Name` for `metadata.name` and `roleRef.name`? And then remove these two lines here.

knan-nrk: So use fmt.Sprintf in the default, remove these lines, and then define a custom rolebindingdefinition? That would also work, I guess, and be somewhat cleaner.

FxKu: yes. and if you don't define the rolebinding, clusterrole is the same as service account as it is right now

knan-nrk:
Updated as suggested. Seems to work with and without custom rolebinding definition.

FxKu:
Tested it locally on minikube:
- Updating existing cluster works
- Installing with helm works

So, I'm fine with this change.

FxKu:
:+1:

Jan-M:
:+1:

Jan-M:
:+1:

PR #583, opened 2019-06-07
Add PDB configuration toggle

### LINKED ISSUE #547: Make PodDisruptionBudget configurable
Unless I'm missing something here, the operator invariably installs for every created cluster a PodDisruptionBudget (pdb) that requires one master to be available at all times (minRequired=1). That prevents any evictions of the master pod, e.g. during node updates or reboots. Since Patroni should handle a master outage by electing another (assuming that clients can handle the short unavailability of write access to the cluster), I'd suggest making the PDB configurable or entirely optional.

FxKu:
This PR builds up on the work from #384. Instead of manifest parameter the PDB can be disabled from the configuration.

Closes #384 
Resolves #547


docs/reference/operator_parameters.md
```diff
@@ -161,6 +161,13 @@ configuration they are grouped under the `kubernetes` key.
   replaced by the cluster name. Only the `{cluster}` placeholders is allowed in
   the template.
 
+* **enable_pod_disruption_budget**
+  if disabled `MinAvailable` in the PDB's spec will be set to `0`. That means
+  the PDB will get created anyway. The PDB is also relaxed when scaling down the
+  Postgres cluster to `"numberOfInstances": 0`. When scaling up or enabling this
```
Jan-M: Why do we impl. the downscale part without the upscale working properly? Whats the win of having pdb at 0 for 0 pods?

FxKu: Pdb at 0 has been suggested in #384. I found out that PDB did not get synced on UPDATE event. Only on SYNC. Now it will be set to 1 as soon as the cluster is scaling up.

Jan-M:
:+1:

docs/reference/operator_parameters.md
```diff
@@ -161,6 +161,12 @@ configuration they are grouped under the `kubernetes` key.
   replaced by the cluster name. Only the `{cluster}` placeholders is allowed in
   the template.
 
+* **enable_pod_disruption_budget**
+  if disabled `MinAvailable` in the PDB's spec will be set to `0`. That means
+  the PDB will get created anyway. The PDB is also relaxed when scaling down the
+  Postgres cluster to `"numberOfInstances": 0`. By default PDB is enabled which
+  results in `"MinAvailable": 1`.
```
sdudoladov: i'd rephrase it to improve readability into sth like:

>  when enabled, sets `MinAvailable` in the PDB's spec to `1`.  Combined with the operator logic for moving pods away from unschedulable nodes,  that setting protects master pods form voluntarily disruptions and hence unwanted DB downtimes. When disabled, still creates the PDB but  sets `MinAvailable` to `0`. This implies a master pod may be terminated before the operator completes the failover, thus meaning longer downtimes. Default: enabled.

pkg/cluster/k8sres.go
```diff
@@ -1273,6 +1273,12 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
 
 func (c *Cluster) generatePodDisruptionBudget() *policybeta1.PodDisruptionBudget {
 	minAvailable := intstr.FromInt(1)
+	pdbEnabled := c.OpConfig.EnablePodDisruptionBudget
+
+	// Is PodDisruptionBudget is disabled or if there is no master, set the budget to 0.
+	if (pdbEnabled != nil && !*pdbEnabled) || c.Spec.NumberOfInstances <= 0 {
```
sdudoladov: `c.Spec.NumberOfInstances <= 0 ` technically means there are no pods in a DB cluster; there are other situations where the master does not exists (e.g. in standby  clusters)

pkg/cluster/sync.go
```diff
@@ -73,20 +73,6 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 		}
 	}
 
-	// create database objects unless we are running without pods or disabled that feature explicitly
```
sdudoladov: what caused  this change ? re-shuffling code by an editor ?

FxKu: Look at the logs from my comment in the other PDB PR: https://github.com/zalando/postgres-operator/pull/384#discussion_r287337006

Because of the error messages, first I thought that the sync code exited to early. That's why I moved PDB syncing before db syncing. But then I realized, I was looking at the logs of an UPDATE event, where PDB syncing was missing. I have introduced it there - also before the DB syncing. Therefore, I kept the change here to have the same syncing order. This way K8s resources are synced before DB stuff. Feels better to me this way.

docs/administrator.md
```diff
@@ -154,6 +154,20 @@ data:
...
+parameter of the PDB is set to `1` which makes it work for single-node clusters
+or in situations when there is no master (e.g. standby clusters).
+
+The PDB is only relaxed in two scenarios:
+* If a cluster is scaled down to `0` instances (e.g. for draining nodes)
+* If the PDB is disabled in the configuration (`enable_pod_disruption_budget`)
+
+The PDB is still in place having `MinAvailable` set to `0`. If enabled it will
+be automatically set to `1` on scale up.
+
```
sdudoladov: "Disabling PDBs helps avoiding blocking Kubernetes upgrades in managed k8s environments at the cost of prolonged DB downtimes;  see PR #384 for the use case"

docs/administrator.md
```diff
@@ -154,6 +154,20 @@ data:
   pod_antiaffinity_topology_key: "failure-domain.beta.kubernetes.io/zone"
 ```
 
+### Pod Disruption Budget
+
+By default the operator uses a PodDisruptionBudget (PDB) to protect the cluster
+from voluntarily disruptions and hence unwanted DB downtime. The `MinAvailable`
+parameter of the PDB is set to `1` which makes it work for single-node clusters
+or in situations when there is no master (e.g. standby clusters).
```
sdudoladov: > which makes it work for single-node clusters

it prevents killing masters in single pods clusters.

> situations when there is no master (e.g. standby clusters).

not really necessary here,  I meant with my previous review is that the comment in the code was unnecessarily generic.

sdudoladov:
:+1:

FxKu:
:+1:

PR #585, opened 2019-06-07
rbac: add user-facing clusterroles

### LINKED ISSUE #584: rbac: user facing clusterroles
After postgres-operator has been installed on a cluster, non-clusteradmin users of the k8s cluster cannot view or create databases by default.

```
kubectl get postgresqls
Error from server (Forbidden): postgresqls.acid.zalan.do is forbidden: User "Erik.Inge.Bolso@nrk.no" cannot list resource "postgresqls" in API group "acid.zalan.do" in the namespace "default"

```
Should probably have some example clusterroles in the manifests directory that can be added to give users of the k8s access to view and make postgreses. I'll make a pull request.

knan-nrk:
Fixes #584. Adds user facing clusterroles so kubectl get postgresql and kubectl create postgresql works for view and edit/admin default rolebindings.

### DIFF
--- a/manifests/user-facing-clusterroles.yaml
+++ b/manifests/user-facing-clusterroles.yaml
@@ -1,0 +1,34 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+  name: zalando-postgres-operator:users:edit
+rules:
+- apiGroups:
+  - acid.zalan.do
+  resources:
+  - postgresqls
+  - postgresqls/status
+  verbs:
+  - "*"
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+  name: zalando-postgres-operator:users:view
+rules:
+- apiGroups:
+  - acid.zalan.do
+  resources:
+  - postgresqls
+  - postgresqls/status
+  verbs:
+  - get
+  - list
+  - watch
+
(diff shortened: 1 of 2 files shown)


manifests/user-facing-clusterroles.yaml
```diff
@@ -0,0 +1,34 @@
...
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+  name: zalando-postgres-operator:users:edit
+rules:
+- apiGroups:
+  - acid.zalan.do
+  resources:
+  - postgresqls
+  - postgresqls/status
```
FxKu: `OperatorConfiguration` CRD is missing under `resources`. Any reason?

FxKu:
If we add these cluster roles. We need to document them somewhere. [User](https://github.com/zalando/postgres-operator/blob/master/docs/user.md) docs probably makes sense, as these roles are meant for users :)

manifests/user-facing-clusterroles.yaml
knan-nrk: Hmm. This is user-facing. I don't think users of the k8s cluster that want to create a database need to see the operator config in order to create a database, I consider that private to the operator and appropriate to stay at clusteradmin level permission to see.

knan-nrk:
Updated with a little bit of doc and went from two to three roles.

manifests/user-facing-clusterroles.yaml
FxKu: I thought about that, too. Fine for me.
I think, `postgresqls/status` could also be removed from the edit role. For admin and view it makes sense.

knan-nrk: Agreed. Updated. :)

FxKu:
:+1:

sdudoladov:
:+1:

sdudoladov:
@knan-nrk  thank you for yet another contribution :+1:

PR #586, opened 2019-06-07
docs: add storageclass to complete-postgres-manifest example

knan-nrk:
Small documentation update.

### DIFF
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -8,9 +8,21 @@ spec:
   - name: date
     image: busybox
     command: [ "/bin/date" ]
+  sidecars:
+  - name: shiny-sidecar
+    image: busybox
+    command: [ "/bin/sh", "-ec", "sleep 365d" ]
+      resources:
+        limits:
+          cpu: 250m
+          memory: 500Mi
+        requests:
+          cpu: 100m
+          memory: 100Mi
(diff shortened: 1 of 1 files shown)


manifests/complete-postgres-manifest.yaml
```diff
@@ -8,9 +8,21 @@ spec:
   - name: date
     image: busybox
     command: [ "/bin/date" ]
+  sidecars:
```
FxKu: We already have a sidecar example in the manifest.
Also keep in mind, that many users start testing the operator with the provided manifests.
Therefore, dummy code should be in comments.

knan-nrk: In that case, complete-postgres-manifest needs more things commented out, like the replication slots. And init_containers are not sidecars as such, IMO.

FxKu: I was referring to [this](https://github.com/zalando/postgres-operator/blob/master/manifests/complete-postgres-manifest.yaml#L75).

knan-nrk: Oh, right, never mind me then. I seem to be slightly behind the times. :)

I'll drop the sidecar bit.

knan-nrk:
Updated. :)

sdudoladov:
:+1:

FxKu:
:+1:

PR #595, opened 2019-06-21
align config map, operator config, helm chart values and templates

### LINKED ISSUE #591: configLogicalBackup is missing in helm's configmap
**.Values.configLogicalBackup** is not present in [configmap.yaml](https://github.com/zalando/postgres-operator/blob/93bfed3e7562f7d6784500d7b523190b6a352e9b/charts/postgres-operator/templates/configmap.yaml).

### LINKED ISSUE #594: camelCase for all manifest parameters
There is some inconsistency in naming of manifest parameters. Most are written in camelCase, except:

- init_containers
- pod_priority_class_name

For the latter one even the reference list it as `podPriorityClassName`. I'd suggest to change the parameters to camelCase style and use snake_case only for configuration parameters.

Thanks @knan-nrk for pointing via Slack.

FxKu:
ConfigMap and OperatorConfiguration are currently a bit out of sync and it's getting harder to read them as more and more parameters got introduced. Helm charts values and templates already have some small errors because of it. This PR tries to:

- sync the two config options (especially TeamsAPI in CRD-based config)
- sorts the parameters alphabetically so that they are easier to find and it's clearer for contributors where to put new ones
- splits `values.yaml` for helm chart to have a file for configMap and CRD config (`values-crd.yaml`)
- adds comments to parameters in `values.yaml` as a best practice for helm charts
- adds missing annotations in CRD template for helm chart
- updates doc reference (names and category)

Fixes #591 #606


pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -53,8 +53,8 @@ type PostgresSpec struct {
 	Databases             map[string]string    `json:"databases,omitempty"`
 	Tolerations           []v1.Toleration      `json:"tolerations,omitempty"`
 	Sidecars              []Sidecar            `json:"sidecars,omitempty"`
-	InitContainers        []v1.Container       `json:"init_containers,omitempty"`
-	PodPriorityClassName  string               `json:"pod_priority_class_name,omitempty"`
+	InitContainers        []v1.Container       `json:"initContainers,omitempty"`
+	PodPriorityClassName  string               `json:"podPriorityClassName,omitempty"`
```
FxKu: Tested with Postgres cluster using the old style and updated the image in the deployment. After rolling update both were still in place. However, when creating new clusters with snake_case styled parameters it is not taken into account. So, I could use [multiple JSON tags](https://stackoverflow.com/questions/18635671/how-to-define-multiple-name-tags-in-a-struct/18635910) for backwards compatibility.

FxKu: Simply having multiple `json` tags doesn't work. Only via another struct. So, lets keep it as it is for now and change on v2.0 release.

FxKu:
1. I reverted the shuffling of `set_memory_request_to_limit` parameter but moved it in the docs.
2. I have only parameters from TeamsAPI which where existing in ConfigMap but missing in CRD-config

Another open question would be `enableShmVolume` parameter, which is not supported in CRD config :(

sdudoladov:
> Another open question would be enableShmVolume parameter, which is not supported in CRD config

It should be supported

charts/postgres-operator/values-crd.yaml
```diff
@@ -0,0 +1,266 @@
...
+
+podAnnotations: {}
+podLabels: {}
+
+# general top-level configuration parameters
+configGeneral:
+  # etcd connection string for Patroni. Empty uses K8s-native DCS.
+  etcd_host: ""
+  # Spilo docker image
+  docker_image: registry.opensource.zalan.do/acid/spilo-11:1.5-p7
```
sdudoladov: `spilo-11:1.5-p9`

charts/postgres-operator/values.yaml
```diff
@@ -14,145 +14,202 @@ podAnnotations: {}
...
-repair_period: 5m
-resync_period: 5m
-spilo_privileged: false
-workers: 4
-
+configGeneral:
+  # etcd connection string for Patroni. Empty uses K8s-native DCS.
+  etcd_host: ""
+  # Spilo docker image
+  docker_image: registry.opensource.zalan.do/acid/spilo-11:1.5-p7
```
sdudoladov: `spilo-11:1.5-p9`

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -62,41 +63,41 @@ configuration:
...
     # additional_secret_mount: "some-secret-name"
     # additional_secret_mount_path: "/some/dir"
+    aws_region: eu-central-1
+    # kube_iam_role: ""
+    # log_s3_bucket: ""
+    # wal_s3_bucket: ""
+  logical_backup:
+    logical_backup_schedule: "30 00 * * *"
+    logical_backup_docker_image: "registry.opensource.zalan.do/acid/logical-backup"
+    logical_backup_s3_bucket: ""
```
sdudoladov: there is no default for that, people have to add their own, that is `logical_backup_s3_bucket: "your-bucket"`

manifests/complete-postgres-manifest.yaml
```diff
@@ -1,19 +1,19 @@
 apiVersion: "acid.zalan.do/v1"
 kind: postgresql
-
 metadata:
   name: acid-test-cluster
 spec:
+  dockerImage: registry.opensource.zalan.do/acid/spilo-11:1.5-p8
```
sdudoladov: `spilo-11:1.5-p9`

FxKu:
Added support for `enableShmVolume` with #605

manifests/configmap.yaml
```diff
@@ -3,62 +3,78 @@ kind: ConfigMap
...
-  # teams_api_url: http://fake-teams-api.default.svc.cluster.local
-  # team_api_role_configuration: "log_statement:all"
+  # enable_database_access: "true"
+  enable_master_load_balancer: "true"
+  # enable_pod_antiaffinity: "false"
+  # enable_pod_disruption_budget: "true"
+  enable_replica_load_balancer: "false"
+  # enable_shm_volume: "true"
+  # enable_team_superuser: "false"
+  enable_teams_api: "false"
```
RafiaSabih: Following the alphabetical order, api should come first

RafiaSabih:
:+1:

FxKu:
:+1:

PR #596, opened 2019-06-24
Enable promote/demote standby via operator

RafiaSabih:
Add the functionality of promoting the standby to master or demoting one to standby cluster through the changes in postgres-manifest. This is using REST API from Patroni to change the standby_cluster settings of patroni.


pkg/util/patroni/patroni.go
```diff
@@ -102,3 +103,14 @@ func (p *Patroni) SetPostgresParameters(server *v1.Pod, parameters map[string]st
 	}
 	return p.httpPostOrPatch(http.MethodPatch, apiURL(server)+configPath, buf)
 }
+
+//SetPostgresParameters sets Postgres options via Patroni patch API call.
+func (p *Patroni) EditStandby(server *v1.Pod, options map[string]string) error {
+	buf := &bytes.Buffer{}
+	err := json.NewEncoder(buf).Encode(map[string]map[string]interface{}{"standby_cluster": {"restore_command": options}})
```
redbaron: shouldn't it be `{"standby_cluster": options}` ?

redbaron: ping, is it a WIP PR?

RafiaSabih: Well, yes. I am sorry for the inconvenience it might have caused because of multiple commits. Will update the status as need review as soon as it is moves to the next step.

sdudoladov:
>  demoting one to standby cluster

Does Patroni actually support that ?

RafiaSabih:
> > demoting one to standby cluster
> 
> Does Patroni actually support that ?

Yup, with patch through its REST API.However, what happens is the cluster goes into the read-only mode, but not streaming from some other cluster.

 Here is the command, if you want to try.
curl -s -XPATCH -d         '{"standby_cluster": {"restore_command":"envdir \"/home/postgres/etc/wal-e.d/env-standby\"/scripts/restores_command.sh \"%f\"%p\" "}}'         http://localhost:8008/config | jq .

FxKu:
Closing in favor for #1035

PR #605, opened 2019-07-03
enable shmVolume setting in OperatorConfiguration

FxKu:
`shmVolume` option is currently missing in OperatorConfiguration CRD.
As the default is `true` we should use type `*bool` to tell when it's not set or set to false.
Plus, moved bool helper functions from `pkg.cluster.k8sres_test` to `pkg.util`


sdudoladov:
:+1:

FxKu:
:+1:

Issue #607, opened 2019-07-04
Use kustomize as alternative operator deployment


jakub-roman:
I'd like to suggest to have an alternative operator deployment using [kustomize](https://github.com/kubernetes-sigs/kustomize).

The reason is that we don't like to use helm (for various reasons) and having to apply yaml files manually is not optimal.

Kustomize would enable users to deploy the operator as easy as:
```
kubectl apply -k github.com/zalando/postgres-operator/manifests
```
while leaving the liberty of overriding properties such as `namespace` easily.

FxKu:
added a comment in [Quickstart](https://github.com/zalando/postgres-operator/blob/master/docs/quickstart.md#manual-deployment-setup). Thanks for the pointer.

PR #608, opened 2019-07-04
Kustomization

jakub-roman:
Add ability to install the operator via kustomization as proposed in #607

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -54,9 +54,13 @@ kubectl create -f manifests/operator-service-account-rbac.yaml  # identity and p
 kubectl create -f manifests/postgres-operator.yaml  # deployment
 ```
 
-When using kubectl 1.14 or newer the mentioned manifests could be also be
-bundled in one [Kustomization](https://github.com/kubernetes-sigs/kustomize)
-manifest.
+There is a [Kustomization](https://github.com/kubernetes-sigs/kustomize)
+manifest that [combines the mentioned resources](manifests/kustomization.yaml) 
+- it can be used with kubectl 1.14 or newer as easy as:
+
+```bash
+kubectl apply -k github.com/zalando/postgres-operator/manifests
+```
 
 For convenience, we have automated starting the operator with minikube using the
 `run_operator_locally` script. It applies the [`acid-minimal-cluster`](../manifests/minimal-postgres-manifest).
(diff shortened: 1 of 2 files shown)


FxKu:
Thanks for the contributions. Another K8s tool, I've never heard about ^^
I'm not sure on providing an example and put yet another section in the quickstart. Maybe it's enough to just mentioning it in one sentence for the manual setup.

Jan-M:
I like it. Seems pretty slim. Maybe it can be reduced to the instructions how to install with kubectl, incl. brief min version?

Maybe we can also slim down the Helm section to the most basic instructions and not too many words about helm.

yujunz:
> Thanks for the contributions. Another K8s tool, I've never heard about ^^
> I'm not sure on providing an example and put yet another section in the quickstart. Maybe it's enough to just mentioning it in one sentence for the manual setup.

It is a built-in plugin in recent Kubectl release. For quickstart, one line instruction on the version constraint and some link might be good enough.

yujunz:
By the way, the manifests is kind of out of date comparing to helm chart output

FxKu:
I've added a sentence about Kustomization in PR #609. So we could close this PR then.

@yujunz it can be that the original Helm chart author followed some best practices from the Helm docs. If you see any bigger differences, please open another issue to discuss it.

FxKu:
It's now mentioned in the [Quickstart](https://github.com/zalando/postgres-operator/blob/master/docs/quickstart.md#manual-deployment-setup). Thanks again @jakub-roman for pointing us to this feature :)

jakub-roman:
@FxKu the main value of this PR was the `manifests/kustomization.yaml` file - without it the kustomization won't work magically. I don't really care much about docs, just wanted to make it more verbose because kustomize is somehow young tool and people don't know much about it yet.

The idea was that you manage the kustomization file in this repository - you are the ones who know how to install the operator properly after all - so people doesn't have to know which manifests to use. 

The alternative is that people make their own kustomization files (which is already redundant), but then if you decide to add another manifest file necessary to install the operator everyone that uses kustomization will have to know about that change and make amendments in their kustomization files and that's just painful...

P.S. sorry I'm coming back to this so late, I didn't have access to internet last week

Jan-M:
We will add the file ;)

jakub-roman:
> We will add the file ;)

Should I just remove the quickstart docs since you already have a note about it so you can merge in this one?

FxKu:
Yes. Or you can add one more sentence with a link to the provided Kustomization file. Then i'll merge it.

FxKu:
:+1:

Jan-M:
:+1:

FxKu:
@jakub-roman thanks for your contribution :)

jakub-roman:
Thanks for including this, it makes the installation way easier for us :)

PR #621, opened 2019-07-16
change app label for logical backup pod

### LINKED ISSUE #593: Logical backup incompatible with the main Postgres pod anti-affinity rules
~~If `enable_pod_antiaffinity` is set to `true`, pods created by the logical backup cronjob are also  started with anti-affinity. That doesn't make sense, and prevents logical backups on clusters where number of Postgres pods equals to the total number of nodes in the cluster (e.g. 2 nodes and 2 Postgres pods).~~

### ~~Expected~~

~~Logical backup pods are always started without anti-affinity (or at least it's possible to remove anti-affinity for logical backup while keeping it for database cluster pods).~~

If `enable_pod_antiaffinity` is set to `true`, pods created by the logical backup cronjob can not be scheduled to the same nodes because of matching labels. That prevents logical backups on clusters where number of Postgres pods equals to the total number of nodes in the cluster (e.g. 2 nodes and 2 Postgres pods).

FxKu:
Label Logical Backup Pod with `application=spilo-logical-backup` to avoid interference with AntiAffitiny from StatefulSet.

Fixes #593

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -307,8 +307,7 @@ class K8s:
             time.sleep(self.RETRY_TIMEOUT_SEC)
 
     def get_logical_backup_job(self, namespace='default'):
-        return self.api.batch_v1_beta1.list_namespaced_cron_job(
-            namespace, label_selector="application=spilo-logical-backup")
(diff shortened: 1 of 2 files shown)


FxKu:
:+1:

RafiaSabih:
:+1:

Issue #616, opened 2019-07-12
Logical backups: ClusterRole binding and namespaced pod lookup


Bryji:
We encountered a couple of issues getting logical backup working, both around finding the pod to back up from:
- RBAC: the backup script attempts to look up the pod name at a cluster level, we needed to create a RoleBinding to the ClusterRole in each namespace to allow this to work. 
- We were deploying clusters of the same name in difference namespaces, backups were failing in one namespace because it was finding the Pod IP of a Spilo pod in a different namespace and unable to communicate with it.

Both these problems could be fixed if the namespace was provided to the backup job/pod (not available currently from what I can see), and if namespace scoped resource lookups were used.

Cluster-scoped resources:
GET /apis/GROUP/VERSION/RESOURCETYPE - return the collection of resources of the resource type
GET /apis/GROUP/VERSION/RESOURCETYPE/NAME - return the resource with NAME under the resource type

Namespace-scoped resources:
GET /apis/GROUP/VERSION/RESOURCETYPE - return the collection of all instances of the resource type across all namespaces
GET /apis/GROUP/VERSION/namespaces/NAMESPACE/RESOURCETYPE - return collection of all instances of the resource type in NAMESPACE
GET /apis/GROUP/VERSION/namespaces/NAMESPACE/RESOURCETYPE/NAME - return the instance of the resource type with NAME in NAMESPACE

Jan-M:
Thanks for reporting, the namespace issue should def. be resolve don the logical backup side.

We will look into this.

FxKu:
Should be fixed with 2c3c7fd. Can you check?

Bryji:
Deployed the operator with docker image tag: registry.opensource.zalan.do/acid/postgres-operator:v1.1.0-65-g1d45a6a which I think should have the latest code changes above.

The backup pod still seems to be being created without the POD_NAMESPACE env var though:
`/dump.sh: line 57: POD_NAMESPACE: unbound variable`. <-- backups are erroring out here since the commit on July 17.

Am I using the wrong docker image tag to try out the fix?

FxKu:
Yes, probably you are pulling the latest logical backup Docker image from our registry, where the fix is not merged yet. If you build the logical backup [image](https://github.com/zalando/postgres-operator/blob/master/docker/logical-backup/Dockerfile) yourself, it will work. Sorry, for not mentioning it.

FxKu:
[...]

PR #623, opened 2019-07-17
Query namespaced K8s API in logical backup script

FxKu:
export POD_NAMESPACE env var also for logical backup pod template and use it in `dump.sh`

adresses #616


Jan-M:
:+1:

FxKu:
:+1:

PR #626, opened 2019-07-19
Fix mkdocs.yaml and title display in ReadTheDocs sidebar

FxKu:
I introduced # H1 headers in each document, but turns out mkdocs doesn't like it.
Can be solved by using html `<h1>` tags.


RafiaSabih:
:+1:

FxKu:
:+1:

Issue #592, opened 2019-06-20
[logical backup] Ability to customize --endpoint-url and --sse


laevilgenius:
This would allow saving logical backups to a private s3.

FxKu:
You mean something similar to what we just added for cloning? #497

laevilgenius:
Yes. I can already inject **AWS_ACCESS_KEY_ID** and **AWS_SECRET_ACCESS_KEY** into job's pods, but there's no way to set _--endpoint-url_ and _--sse_ (which I need to be disabled) via env vars.

Jan-M:
At some point we need a strategy or vision on which global parameters come from where. Anything that is global config for every pod or Postgres cluster can also find its place in Spilo/Docker config.

Overwriting or extending the Spilo Docker image seems to scare a lot of users away asking for more and more features in the operator.

pierreozoux:
(Maybe outside the scope of this issue, but related to what you just said)
@Jan-M We plan to use this operator for our PG need, and the fact that [Gitlab is switching to it](https://gitlab.com/gitlab-org/omnibus-gitlab/issues/3752) gives even more credibility on this solution moving forward.
But, to be honest, as a new comer to the PG world, this is difficult to handle all these abstractions :)
At the bottom we have patroni, wal-e, wal-g. Then we have spilo that puts all together in a Dockerfile with bash and python scripts (plus some reference to sputs).
And finally, we have the operator that wraps it all.
I think somehow these layers have to be rationalized. I understand the history of the project, and I'm amazed by the work you put it. I just think that it is a bit scary for new comers to dive in all these abstractions ;)
(And I guess it would be difficult to maintain moving forward.)
So this looks like the problem space, I have no idea how to solve it in a better way. And I also understand the history of it, and if it is working for you, it would be difficult to spend resources to refactor things.

FxKu:
As #628 got merged, both endpoint and sse are configurable.

PR #628, opened 2019-07-22
Passing endpoint, access and secret key to logical-backup container

ermajn:
logical-backup container needs to upload dump results to S3 storage, and when using non-AWS custom s3, it is nice to have endpoint variable arround. 
Same goes for access and secret key.


FxKu:
Related to #592, but does not cover the `sse` option.

Could you also reflect these changes in the [logical backup script](https://github.com/zalando/postgres-operator/blob/master/docker/logical-backup/dump.sh)?

sdudoladov:
@ermajn [ AWS CLI docs](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) say the following:

> Precedence of options

> If you specify an option by using one of the environment variables described in this topic, it overrides any value loaded from a profile in the configuration file.

Isn't it the case  that setting `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to the empty string by default may override the existing conf and thus make backups unusable ?

ermajn:
> @ermajn [ AWS CLI docs](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) say the following:
> 
> > Precedence of options
> 
> > If you specify an option by using one of the environment variables described in this topic, it overrides any value loaded from a profile in the configuration file.
> 
> Isn't it the case that setting `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to the empty string by default may override the existing conf and thus make backups unusable ?

Probablly yes, acording to the documentation, you are right.
Than, default and empty string should not set environment variables. Does that seems ok for you?

muff1nman:
Any way to help get this PR merged?

ermajn:
> Any way to help get this PR merged?

As far as I can see, it is market to be in v1.3.
If I can do anything to merge it before, let me know.

muff1nman:
The one awkward aspect of this is that this is storing secrets in a configmap, rather than a secret. In addition, the secret then gets propagated to any other namespace in the form of env values on the cronjobs. Not pretty.

Spitballing here but would it instead make sense to move the logical backup configuration to the postgresqls.acid.zalan.do namespaced objects? The idea would be that you could create a secret in the namespace for holding the access/secret keys and reference it from the postgresqls.acid.zalan.do. This would be a significant departure, but it might make more sense.

ermajn:
> The one awkward aspect of this is that this is storing aws creds in a configmap, rather than in a secret. In addition, the secret then gets propagated to any other namespace in the form of env values on the cronjobs. Not pretty.
> 
> Spitballing here but would it instead make sense to move the logical backup configuration to the postgresqls.acid.zalan.do namespaced objects? The idea would be that you could create a secret in the namespace for holding the access/secret keys and reference it from the postgresqls.acid.zalan.do. This would be a significant change, but it might make more sense.

This pretty much make sense to me. 
Only in that case, reference would look up in secrets in namespace where other postgresqls.acid.zalan.do namespaced objects objects are, that means, no single secret can be used as it is happening now if credentials are stored in configmap.

FxKu:
@ermajn could you add the parameters also to the default CRD-based [configuration file](https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql-operator-default-configuration.yaml)? Use the same order as in my latest commit for the configMap. Please, inherit the alphabetical order also in the two values files of the helm chart. Makes finding stuff a bit easier in the end. Thanks :)

pkg/cluster/k8sres.go
```diff
@@ -1571,6 +1571,14 @@ func (c *Cluster) generateLogicalBackupPodEnvVars() []v1.EnvVar {
 			Name:  "LOGICAL_BACKUP_S3_BUCKET",
 			Value: c.OpConfig.LogicalBackup.LogicalBackupS3Bucket,
 		},
+		{
+			Name:  "LOGICAL_BACKUP_S3_ENDPOINT",
+			Value: c.OpConfig.LogicalBackup.LogicalBackupS3Endpoint,
+		},
+		{
+			Name:  "AWS_SSE",
```
FxKu: any reason why this env variable doesn't have the `LOGICAL_BACKUP_` prefix?

ermajn: no, that was my mistake. Dump.sh is already expecting LOGICAL_BACKUP_S3_SSE. I've fixed that.

FxKu:
:+1:

sdudoladov:
:+1:

sdudoladov:
@ermajn thank you for the contribution :)

vitobotta:
@ermajn Hi! I'm using Scaleway object storage and uploads of the logical backups fail unless I pass the "--region" argument to the aws command. Not sure if this is only required with some storage services but it is definitely required with Scaleway. Would it be possible for you to do a new PR to add the setting for the region? I could try myself but I wouldn't know where to start and it would take a while for me to understand how to do it :) Thanks!

vitobotta:
#812

vitobotta:
Actually it wasn't difficult to replicate this PR for the region setting, so I created a new PR for that.

PR #644, opened 2019-08-07
Create kustomize manifests for common usage

yujunz:
* deploy the operator with UI
* provision a cluster
* clone a cluster
* point-in-time-recovery


FxKu:
Thanks for this contribution. Quite many files, so will take time to test everything through.

For UI, why not using the manifest folder in the UI directory? And for the other use cases, I could also think of creating folders for use cases directly under /manifests, without the postgres-operator folder.

yujunz:
> For UI, why not using the manifest folder in the UI directory? And for the other use cases, I could also think of creating folders for use cases directly under /manifests, without the postgres-operator folder.

There are some hard coded values in current manifests, such as team acid, namespace and ingress. And UI needs to point to operator API as https://github.com/zalando/postgres-operator/pull/644/files#diff-80bb0ffe5793c1547700e246284ed4b8R23

PR #652, opened 2019-08-15
Allow disabling of CRD creation

deitmerit:
Signed-off-by: Stefan Frye <frye@deitmer-it.de>

In certain scenarios (e.g., automated testing) it is helpful to be able to switch off CRD creation by the Helm chart. As CRDs are not automatically deleted on `helm delete`, we prefer to manage them separately.


FxKu:
makes sense to me. I have always deleted CRDs manually before reinstalling :smiley:

FxKu:
:+1:

sdudoladov:
:+1:

PR #654, opened 2019-08-21
Avoid orphaned objects on delete

### LINKED ISSUE #218: Incomplete deletion if resources upon delete postgresql

### LINKED ISSUE #551: quick create-delete leaves orphaned  objects
Orphaned pods/endpoints are left in the cluster if the `ADD` event is followed by `DELETE` within a short time period for the same cluster.

This error manifests as 
```
time="2019-04-29T12:17:52Z" level=info msg="\"ADD\" event has been queued" cluster-name=default/acid-minimal-cluster pkg=controller worker=0
...
time="2019-04-29T12:17:53Z" level=info msg="waiting for the cluster being ready" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
time="2019-04-29T12:17:56Z" level=debug msg="Waiting for 2 pods to become ready" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
time="2019-04-29T12:18:00Z" level=info msg="\"DELETE\" event has been queued" cluster-name=default/acid-minimal-cluster pkg=controller worker=0
...
time="2019-04-29T12:18:29Z" level=info msg="statefulset \"default/acid-minimal-cluster\" has been deleted" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
time="2019-04-29T12:18:29Z" level=debug msg="deleting pods" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
time="2019-04-29T12:18:29Z" level=debug msg="no pods to delete" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
...
time="2019-04-29T12:18:29Z" level=debug msg="deleting PVCs" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
time="2019-04-29T12:18:29Z" level=debug msg="no PVCs to delete" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
...
time="2019-04-29T12:18:32Z" level=debug msg="removing leftover Patroni objects (endpoints or configmaps)" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
time="2019-04-29T12:18:32Z" level=warning msg="could not remove leftover patroni objects; could not fetch Patroni Endpoint \"/\": an empty namespace may not be set when a resource name is provided" cluster-name=default/acid-minimal-cluster pkg=cluster worker=0
time="2019-04-29T12:18:32Z" level=info msg="cluster has been deleted" cluster-name=default/acid-minimal-cluster pkg=controller worker=0
```

reproducible both with `kind` and actual `k8s`

this issue also prevents creating a new cluster with the same name afterwards

FxKu:
If a cluster is deleted immediately after creation, fields of the `cluster` struct are empty when processing the DELETE event. The same might happen when a cluster doesn't get into the `Running` state at all.

fixes #551

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -164,11 +164,13 @@ func (c *Cluster) setStatus(status string) {
 	}
 
 	// we cannot do a full scale update here without fetching the previous manifest (as the resourceVersion may differ),
-	// however, we could do patch without it. In the future, once /status subresource is there (starting Kubernets 1.11)
+	// however, we could do patch without it. In the future, once /status subresource is there (starting Kubernetes 1.11)
 	// we should take advantage of it.
 	newspec, err := c.KubeClient.AcidV1ClientSet.AcidV1().Postgresqls(c.clusterNamespace()).Patch(c.Name, types.MergePatchType, patch, "status")
 	if err != nil {
 		c.logger.Errorf("could not update status: %v", err)
+		// return as newspec is empty, see PR654
+		return
 	}
 	// update the spec, maintaining the new resourceVersion.
 	c.setSpec(newspec)
--- a/pkg/cluster/util.go
+++ b/pkg/cluster/util.go
@@ -365,7 +365,7 @@ func (c *Cluster) waitStatefulsetPodsReady() error {
 	c.setProcessName("waiting for the pods of the statefulset")
 	// TODO: wait for the first Pod only
 	if err := c.waitStatefulsetReady(); err != nil {
-		return fmt.Errorf("statuful set error: %v", err)
+		return fmt.Errorf("stateful set error: %v", err)
 	}
 
 	// TODO: wait only for master
--- a/pkg/controller/postgresql.go
+++ b/pkg/controller/postgresql.go
@@ -258,14 +258,6 @@ func (c *Controller) processEvent(event ClusterEvent) {
 		}
 		lg.Infoln("deletion of the cluster started")
 
-		// HACK if a delete happens too soon, cl fields are empty
-		// thus, pods, logical backup jobs and headless service will not get deleted
-		// see issues 551, 218
-		if cl.Name == "" {
-			cl.Name = clusterName.Name
-			cl.Namespace = clusterName.Namespace
-		}
-
 		teamName := strings.ToLower(cl.Spec.TeamID)
 
 		c.curWorkerCluster.Store(event.WorkerID, cl)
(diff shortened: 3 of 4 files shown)


sdudoladov:
#218 is a totally different issue that is caused by an early return when an error happens during deletion
I'd suggest handling it in a separate PR

FxKu:
Turns out that #218 has been resolved (partly) by PR #295. Updated the description.

sdudoladov:
@FxKu can you post the content of `cl` returned at [line 177](https://github.com/zalando/postgres-operator/blob/c7fc4bf9a489a9373db7ce2ce21387b5d4f87b63/pkg/controller/postgresql.go#L177) for that case of quick deletion?

sdudoladov:
After some experiments I think the synchronization of `CREATE`/ `DELETE` processing is correct.  It is the `setSpec()` method that causes the problem. When [the Patch call](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L169) fails, `newspec` contains zero values of respective types  and thus sets certain parts of the Postgres spec to empty values.

The following sequence of events leads to that situation:
1. Operator starts creating a clusters; that takes time
2. A `DELETE` event  happens during the processing of `CREATE` . At that point k8s  has already deleted  the manifest. 
3. Updating the cluster status in the (no longer existing) manifest as the very last step of  `CREATE` fails, causing `setSpec()` to nullify the Postgres spec. 
4. Operator attempts to process `DELETE` but is unable to find certain objects for deletion because `Name` and `Namespace` at the spec are  empty strings

The practical consequence of that bug is that certain resources - for example pods -  stay around forever, preventing rolling updates or creation of the cluster with the  same name.  

 Deletions during `Update` and `Sync` events should have the same behaviour.

@FxKu the solution I found is to add `return`  at the end [of the error processing code of `setSpec()`](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L171)

FxKu:
Thanks @sdudoladov for finding the source of this problem. Changed the code accordingly.

Btw. the problem also occurred when the cluster failed to get into a running state (e.g. by specifying non-existing Docker image) and a user then tries to delete the cluster. So, you're probably right with the `Sync` and `Update`.

FxKu:
:+1:

sdudoladov:
:+1:

PR #655, opened 2019-08-26
fix relative links to manifest files

phobot:
[minimal](../manifests/minimal-postgres-manifest.yaml)
and the
[complete](../manifests/complete-postgres-manifest.yaml)
Links don't work as the manifests directory is higher in the structure.


FxKu:
Good catch. Thanks for fixing it.

FxKu:
:+1:

sdudoladov:
:+1:

PR #657, opened 2019-08-29
Custom annotations 329

### LINKED ISSUE #329: Allow for custom annotations
We are running [Istio](https://istio.io/) in Kubernetes with automatic sidecar injection. For some services like Postgres we do not use Istio injection. This is configured through pod template annotations. It would be good if it is possible to add some custom annotations on Postgres StatefulSet and pod template in cluster spec.

runyontr:
Deploying the operator with

```yaml
apiVersion: v1
data:
...
  custom_pod_annotations: global:foobar
...
kind: ConfigMap
metadata:
  name: crds-postgres-operator
```

and a cluster with 

```yaml
apiVersion: acid.zalan.do/v1
kind: postgresql
metadata:
  name: dataeng-postgres
  namespace: default
spec:
...
  podAnnotations:
    local: foo
...
```

Yields pods that have annotations:

```
apiVersion: v1
kind: Pod
metadata:
  annotations:
    global: foobar
    local: foo
    ...
  name: dataeng-postgres-0
  namespace: default
```

Closes #329

### DIFF
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -53,9 +53,11 @@ configKubernetes:
   cluster_domain: cluster.local
   # additional labels assigned to the cluster objects
   cluster_labels:
-      application: spilo
+    application: spilo
   # label assigned to Kubernetes objects created by the operator
   cluster_name_label: cluster-name
+  # additional annotations to add to every database pod
+  custom_pod_annotations:
   # toggles pod anti affinity on the Postgres pods
   enable_pod_antiaffinity: false
   # toggles PDB to set to MinAvailabe 0 or 1
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -68,7 +68,7 @@ configKubernetes:
   # list of labels that can be inherited from the cluster manifest
   # inherited_labels:
   # - application
-  # - app
+  # - environment
 
   # timeout for successful migration of master pods from unschedulable node
   # master_pod_move_timeout: 20m
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -55,7 +55,7 @@ configKubernetes:
   # label assigned to Kubernetes objects created by the operator
   cluster_name_label: version
   # annotations attached to each database pod
-  custom_pod_annotations: global:foobar
+  # custom_pod_annotations: keya:valuea
   # toggles pod anti affinity on the Postgres pods
   enable_pod_antiaffinity: "false"
   # toggles PDB to set to MinAvailabe 0 or 1
@@ -129,8 +129,7 @@ configLoadBalancer:
   # DNS zone for cluster DNS name when load balancer is configured for cluster
   db_hosted_zone: db.example.com
   # annotations to apply to service when load balancing is enabled
-  # custom_service_annotations:
-  #   "keyx:valuez,keya:valuea"
+  # custom_service_annotations: "keyx:valuez,keya:valuea"
 
   # toggles service type load balancer pointing to the master pod of the cluster
   enable_master_load_balancer: "true"
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -54,6 +54,8 @@ configKubernetes:
   cluster_labels: application:spilo
   # label assigned to Kubernetes objects created by the operator
   cluster_name_label: version
+  # annotations attached to each database pod
+  custom_pod_annotations: global:foobar
   # toggles pod anti affinity on the Postgres pods
   enable_pod_antiaffinity: "false"
   # toggles PDB to set to MinAvailabe 0 or 1
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -64,7 +64,7 @@ configKubernetes:
   # infrastructure_roles_secret_name: postgresql-infrastructure-roles
 
   # list of labels that can be inherited from the cluster manifest
-  # inherited_labels: ""
+  # inherited_labels: application,environment
 
   # timeout for successful migration of master pods from unschedulable node
   # master_pod_move_timeout: 20m
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -118,6 +118,10 @@ These parameters are grouped directly under  the `spec` key in the manifest.
    then the default priority class is taken. The priority class itself must be
    defined in advance. Optional.
 
+* **podAnnotations**
+  This map of strings gets attached as an annotation to each pod created for the database.
+
+
 * **enableShmVolume**
   Start a database pod without limitations on shm memory. By default docker
   limit `/dev/shm` to `64M` (see e.g. the [docker
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -119,7 +119,8 @@ These parameters are grouped directly under  the `spec` key in the manifest.
    defined in advance. Optional.
 
 * **podAnnotations**
-  This map of strings gets attached as an annotation to each pod created for the database.
+  A map of key value pairs that gets attached as [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
+  to each pod created for the database.
 
 
 * **enableShmVolume**
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -31,7 +31,7 @@ data:
   enable_teams_api: "false"
   # etcd_host: ""
   # infrastructure_roles_secret_name: postgresql-infrastructure-roles
-  # inherited_labels: ""
+  # inherited_labels: application,environment
   # kube_iam_role: ""
   # log_s3_bucket: ""
   # logical_backup_docker_image: "registry.opensource.zalan.do/acid/logical-backup"
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -77,5 +77,5 @@ data:
   # teams_api_url: http://fake-teams-api.default.svc.cluster.local
   # toleration: ""
   # wal_s3_bucket: ""
-  watched_namespace: "*" # listen to all namespaces
+  watched_namespace: "*"  # listen to all namespaces
   workers: "4"
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -13,6 +13,7 @@ data:
   cluster_name_label: version
   # custom_service_annotations:
   #   "keyx:valuez,keya:valuea"
+  custom_pod_annotations: global:foobar
   db_hosted_zone: db.example.com
   debug_logging: "true"
   # default_cpu_limit: "3"
@@ -37,7 +38,7 @@ data:
   # logical_backup_docker_image: "registry.opensource.zalan.do/acid/logical-backup"
   # logical_backup_s3_bucket: "my-bucket-url"
   # logical_backup_schedule: "30 00 * * *"
-  master_dns_name_format: '{cluster}.{team}.staging.{hostedzone}'
+  master_dns_name_format: "{cluster}.{team}.staging.{hostedzone}"
   # master_pod_move_timeout: 10m
   # max_instances: "-1"
   # min_instances: "-1"
@@ -60,13 +61,13 @@ data:
   ready_wait_interval: 3s
   ready_wait_timeout: 30s
   repair_period: 5m
-  replica_dns_name_format: '{cluster}-repl.{team}.staging.{hostedzone}'
+  replica_dns_name_format: "{cluster}-repl.{team}.staging.{hostedzone}"
   replication_username: standby
   resource_check_interval: 3s
   resource_check_timeout: 10m
   resync_period: 5m
   ring_log_lines: "100"
-  secret_name_template: '{username}.{cluster}.credentials'
+  secret_name_template: "{username}.{cluster}.credentials"
   # sidecar_docker_images: ""
   # set_memory_request_to_limit: "false"
   spilo_privileged: "false"
@@ -76,5 +77,5 @@ data:
   # teams_api_url: http://fake-teams-api.default.svc.cluster.local
   # toleration: ""
   # wal_s3_bucket: ""
-  watched_namespace: "*"  # listen to all namespaces
+  watched_namespace: "*" # listen to all namespaces
   workers: "4"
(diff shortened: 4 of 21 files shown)


FxKu:
Thanks for this contribution. Could enable this option also for the ConfigMap? And would be cool, if this field is covered in tests as explained [here](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters).

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -170,6 +171,7 @@ type OperatorConfigurationData struct {
 	LoggingRESTAPI             LoggingRESTAPIConfiguration        `json:"logging_rest_api"`
 	Scalyr                     ScalyrConfiguration                `json:"scalyr"`
 	LogicalBackup              OperatorLogicalBackupConfiguration `json:"logical_backup"`
+	PodAnnotations             map[string]string                  `json:"podAnnotations"`
```
FxKu: Why is there a field for `PodAnnotations` when you've already have `CustomPodAnnotations`?

FxKu:
Thanks for the update. I'm still missing an entry in the [ConfigMap](https://github.com/zalando/postgres-operator/blob/master/manifests/configmap.yaml) and please also update the [values.yaml](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values.yaml) file.

Use example values so that users immediately understand what to insert  in these files. You can comment the lines so that the dummy annotations are not set by default.

charts/postgres-operator/values-crd.yaml
```diff
@@ -53,9 +53,11 @@ configKubernetes:
   cluster_domain: cluster.local
   # additional labels assigned to the cluster objects
   cluster_labels:
-      application: spilo
+    application: spilo
   # label assigned to Kubernetes objects created by the operator
   cluster_name_label: cluster-name
+  # additional annotations to add to every database pod
+  custom_pod_annotations:
```
FxKu:
```suggestion
  # custom_pod_annotations: 
  #   keyb: valueb
  #   keyz: valuez
```

charts/postgres-operator/values.yaml
```diff
@@ -54,6 +54,8 @@ configKubernetes:
   cluster_labels: application:spilo
   # label assigned to Kubernetes objects created by the operator
   cluster_name_label: version
+  # annotations attached to each database pod
+  custom_pod_annotations: global:foobar
```
```suggestion
  # custom_pod_annotations:
  #     "keyb:valueb,keyz:valuez"

```

manifests/configmap.yaml
```diff
@@ -13,6 +13,7 @@ data:
   cluster_name_label: version
   # custom_service_annotations:
   #   "keyx:valuez,keya:valuea"
+  custom_pod_annotations: global:foobar
```
FxKu:
```suggestion
  # custom_pod_annotations:
  #   "keyb:valueb,keyz:valuez"
```

manifests/configmap.yaml
```diff
@@ -76,5 +77,5 @@ data:
   # teams_api_url: http://fake-teams-api.default.svc.cluster.local
   # toleration: ""
   # wal_s3_bucket: ""
-  watched_namespace: "*"  # listen to all namespaces
+  watched_namespace: "*" # listen to all namespaces
```
FxKu: IIRC the two spaces are there because of `helm lint`. Please leave it like this.

docs/reference/cluster_manifest.md
```diff
@@ -118,6 +118,10 @@ These parameters are grouped directly under  the `spec` key in the manifest.
    then the default priority class is taken. The priority class itself must be
    defined in advance. Optional.
 
+* **podAnnotations**
+  This map of strings gets attached as an annotation to each pod created for the database.
```
FxKu:
```suggestion
  A map of key value pairs that gets attached as [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
  to each pod created for the database.
```

charts/postgres-operator/values-crd.yaml
FxKu: I forgot to mention that there is also an example [manifest](https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql-operator-default-configuration.yaml) for an `OperatorConfiguration` resource. Could you add the field there, too? Thanks. Same layout as I suggested here.

pkg/cluster/k8sres.go
```diff
@@ -1462,10 +1481,13 @@ func (c *Cluster) generateLogicalBackupJob() (*batchv1beta1.CronJob, error) {
 			},
 		}}
 
+	annotations := c.generatePodAnnotations(nil)
```
andersosthus: This would cause LogicalBackupJobs to only get annotations from the Operator config, not the Cluster spec.
I think you can change it to c.generatePodAnnotations(c.Spec) to make it pick up annotations from the Cluster spec.

runyontr: Good point, thanks!

FxKu:
Tested and works. Just wondering if it really closes #329 because @angapov was also talking about the StatefulSet. What if someone wants to have custom annotations for all objects of a Postgres cluster (similar to `inheritedLabels`)? We should not aim for adding more and more manifests parameters.

So my question would be if defining custom annotations is a feature only really relevant for Pods? Or does it only solve one part of the problem when intergating the operator to different environments.

Bryji:
> So my question would be if defining custom annotations is a feature only really relevant for Pods? Or does it only solve one part of the problem when intergating the operator to different environments.
Agree with both points here - the annotation should go on the statefulset and hence get inherited by the pods and it would be good to be consistent with "inherited_labels" --> "inherited_annotations"

hpurmann:
We're seeing
```
time="2019-10-01T08:17:16Z" level=debug msg="making PATCH http request: http://10.16.17.6:8008/config" cluster-name=staging/acid-postgres-cluster pkg=cluster
time="2019-10-01T08:17:16Z" level=warning msg="could not patch postgres parameters with a pod staging/acid-postgres-cluster-1: could not make request: Patch http://10.16.17.6:8008/config: read tcp 10.16.21.18:43620->10.16.17.6:8008: read: connection reset by peer" cluster-name=staging/acid-postgres-cluster pkg=cluster
```
istio is enabled  in this namespace. We manually modified the `StatefulSet` to add a pod annotation to disable it. Afterwards, we didn't see this error anymore. Of course, this is gone the next time the operator kicks in. Looking forward to seeing this feature. Has anybody else seen problems with istio and postgres operator?

FxKu:
Ok, in order to move forward with this PR I had a look again on how `inheritedLabels` and `customServiceAnnotations` are implemented. `inheritedAnnotations` is definitely something for another PR. It should be implemented similiar to the current [`labelsSet`](https://github.com/zalando/postgres-operator/blob/f0e29060b1fa0884fcf338e87492cd8e90bfd9e6/pkg/cluster/util.go#L382) function that is called for all the derived resources. Such a feature could exist next to `CustomPodAnnotations`.

Compared to the `CustomServiceAnnotations` option, `CustomPodAnnotations` is also available from the manifest. I wonder if this is really necessary?

garfieldcoked:
Thanks @hpurmann your comment helped me to get this working, I was having the same issue. I also tried suggestion the [istio guide](https://istio.io/docs/setup/additional-setup/sidecar-injection/#more-control-adding-exceptions)  on adding exceptions to ignore where label `application=spilo` is present but I couldnt get it to work.

@runyontr, any idea when your PR will be merged?

runyontr:
@garfieldcoked I think the failed build on the PR is an issue with the PR system, not the code.  I don't think they're waiting on me for anything.  @FxKu Any update?

garfieldcoked:
@hpurmann I managed to get Istio to ignore the postgres pods. I created an override values file that I supply during the installation of Istio

```
# values.yaml 

# Update sidecar never inject selector to ignore resources where the following labels exist
sidecarInjectorWebhook:
  neverInjectSelector:
    - matchExpressions:
      - {key: "application", operator: In, values: [spilo]}
```

After installation, I verified the config map was updated
```
$ kubectl -n istio-system get configmap istio-sidecar-injector -o jsonpath='{.data.config}' | grep -A5 neverInjectSelector:
```

Which displays:
```
neverInjectSelector:
  - matchExpressions:
    - key: application
      operator: In
      values:
      - spilo
```

I then deleted the sidecar injector pod so the configmap is picked up then deleted postgres pod.


Of course, this won't be needed once this PR is merged but needed a stable fix until then :)

FxKu:
@runyontr I think you only need to rebase and tests shall pass.

Regarding this PR in general, you haven't commented on my latest questions. I don't think adding podAnnotations as a manifest parameter is necessary. Having only the configuration should solve the issue with e.g. istio. Other opinions @sdudoladov @Jan-M ?

runyontr:
@FxKu It might not be necessary for the istio solution, but in our use case, we need database specific info injected into the annotation so (e.g.) our datadog agent knows which databases to pull data for.

Rebased

Bryji:
We also need this feature (annotations) for Datadog integration

hpurmann:
@garfieldcoked Thanks for confirming the problem with istio and your described quick fix!

mc1711:
> We're seeing
> 
> ```
> time="2019-10-01T08:17:16Z" level=debug msg="making PATCH http request: http://10.16.17.6:8008/config" cluster-name=staging/acid-postgres-cluster pkg=cluster
> time="2019-10-01T08:17:16Z" level=warning msg="could not patch postgres parameters with a pod staging/acid-postgres-cluster-1: could not make request: Patch http://10.16.17.6:8008/config: read tcp 10.16.21.18:43620->10.16.17.6:8008: read: connection reset by peer" cluster-name=staging/acid-postgres-cluster pkg=cluster
> ```
> 
> istio is enabled in this namespace. We manually modified the `StatefulSet` to add a pod annotation to disable it. Afterwards, we didn't see this error anymore. Of course, this is gone the next time the operator kicks in. Looking forward to seeing this feature. Has anybody else seen problems with istio and postgres operator?

@hpurmann  we are having similar issue, with istio and postgres operator. 

We have a connection reset by peer error as well. 

`level=error msg="could not connect to PostgreSQL database: read tcp 10.244.0.87:44830->10.111.214.22:5432: read: connection reset by peer" cluster-name=default/at-db pkg=cluster`


We moved to istio 1.3 and since then we managed to have the istio-proxy sidecar running, by the way the postgres operator is still not able to update the spilo-role label. In our configuration this results in having the Endpoints not generated



```
$kubectl get pod                                           

NAME                                               READY     STATUS    RESTARTS   AGE
at-db-0                                          2/2       Running   0          13m
at-db-1                                          2/2       Running   0          3h
at-db-2                                          2/2       Running   0          3h


$kubectl get endpoints                        

NAME                              ENDPOINTS                                               AGE
at-db                           <none>                                                  4d
at-db-repl                      <none>                                                  11d

$kubectl get svc
NAME                              TYPE           CLUSTER-IP       EXTERNAL-IP   PORT(S)              AGE
at-db                           LoadBalancer   10.111.214.22    <pending>     5432:31383/TCP       4d
at-db-repl                      ClusterIP      10.99.140.84     <none>        5432/TCP             11d
```

FxKu:
@runyontr any updates on rebasing the branch?

runyontr:
I'm showing it as properly rebased:

```shell
$ git rebase origin/master
Current branch custom-annotations-329 is up to date.
```

FxKu:
Could you sync your master: https://github.com/runyontr/postgres-operator/tree/master and rebase your branch again? If travis is still failing then, I'll have a closer look again.

runyontr:
@FxKu Still doesn't seem to help.

pkg/cluster/cluster.go
```diff
@@ -329,6 +329,10 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 		match = false
 		reasons = append(reasons, "new statefulset's annotations doesn't match the current one")
 	}
+	if !reflect.DeepEqual(c.Statefulset.Spec.Template.Annotations, statefulSet.Spec.Template.Annotations) {
+		match = false
+		reasons = append(reasons, "new statefulset's pod's annotations don't match the current one")
+	}
```
FxKu: This can be removed as it is already checked a few lines [later](https://github.com/zalando/postgres-operator/blob/a337d3ca7ee19e8d5f9769c99715293a70fc2993/pkg/cluster/cluster.go#L383)

pkg/cluster/k8sres.go
```diff
@@ -949,6 +954,20 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
+	annotations := make(map[string]string)
+	for k, v := range c.OpConfig.CustomPodAnnotations {
+		annotations[k] = v
+	}
+	if spec == nil {
+		return annotations
+	}
+	for k, v := range spec.PodAnnotations {
+		annotations[k] = v
+	}
```
FxKu: This function should return `nil` when no annotations are set. Otherwise this leads to rolling updates during each sync as the operator finds a mismatch between the current emtpy annotations map `map[string]string(nil)` and the new `map[string]string{}`. Change it to something like:
```
	if spec != nil {
		for k, v := range spec.PodAnnotations {
			annotations[k] = v
		}
	}
	if len(annotations) == 0 {
		return nil
	}
```

pkg/cluster/k8sres.go
```diff
@@ -485,13 +486,14 @@ func generatePodTemplate(
...
-			Namespace: namespace,
+			Labels:      labels,
+			Namespace:   namespace,
+			Annotations: annotations,
 		},
 		Spec: podSpec,
 	}
 	if kubeIAMRole != "" {
-		template.Annotations = map[string]string{constants.KubeIAmAnnotation: kubeIAMRole}
+		template.Annotations[constants.KubeIAmAnnotation] = kubeIAMRole
```
iand: This should check whether Annotations is nil since generatePodAnnotations may return nil

pkg/cluster/cluster.go
FxKu: @runyontr want to address this feedback? I think, it's redundant code.

runyontr: This check is for the annotations on the pod.  the one starting on line 328 is for the annotations on the statefulset.

FxKu: The check on line 328 is fine, but here (232) you are comparing the template which is already done in line 383 (what I have linked)

redbaron:
I am linking issues/PR all of which can be addressed by single https://github.com/zalando/postgres-operator/issues/479

erthalion:
:

erthalion:
:+1:

FxKu:
:+1:

FxKu:
Thanks again @runyontr for your contribution and also reflecting all the feedback :)

PR #659, opened 2019-09-02
Limit go test to pkg folder

FxKu:
The kubectl-pg dependencies are currently not downloaded which leads to an error when calling `go test ./...`. As unit tests only live in the `pkg` directory at the moment we can also call `go test ./pkg/...`.

Adding `GO111MODULE=on` would also solve this problem, but we can leave it to PR #544 .


sdudoladov:
:+1:

FxKu:
:+1:

PR #671, opened 2019-09-19
Handle IPv6 k8s pods in Patroni URLs

optix2000:
Encapsulate IPv6 addresses in brackets.

Fixes `could not patch postgres parameters with a pod namespace/pod-0: could not make request: Patch http://ffff:ffff::1009:8008/config: invalid URL port \"ffff:1009:8008\`

Changing patroni configs now works properly.
```
time="2019-09-19T07:36:06Z" level=debug msg="calling Patroni API on a pod namespace/pod-0 to set the following Postgres options: map[max_connections:2000]" cluster-name=namespace/pod pkg=cluster
time="2019-09-19T07:36:06Z" level=debug msg="making PATCH http request: http://[ffff:ffff::1009]:8008/config" cluster-name=namespace/pod pkg=cluster
time="2019-09-19T07:36:06Z" level=debug msg="performing rolling update" cluster-name=namespace/pod pkg=cluster
```

### DIFF
--- a/pkg/util/patroni/patroni.go
+++ b/pkg/util/patroni/patroni.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"time"
 
@@ -43,8 +44,23 @@ func New(logger *logrus.Entry) *Patroni {
 	}
 }
 
-func apiURL(masterPod *v1.Pod) string {
-	return fmt.Sprintf("http://%s:%d", masterPod.Status.PodIP, apiPort)
+func apiURL(masterPod *v1.Pod) (string, error) {
+	ip := net.ParseIP(masterPod.Status.PodIP)
+	if ip == nil {
+		return "", fmt.Errorf("%s is not a valid IP", masterPod.Status.PodIP)
+	}
+	var ipString string
+	// If IPv6, add braces
+	if ip.To4() == nil {
+		if ip.To16() == nil {
+			// Shouldn't ever get here, but library states it's possible.
+			return "", fmt.Errorf("%s is not a valid IPv4/IPv6 address", ip.String())
+		}
+		ipString = fmt.Sprintf("[%s]", ip.String())
+	} else {
+		ipString = ip.String()
+	}
+	return fmt.Sprintf("http://%s:%d", ipString, apiPort), nil
 }
 
 func (p *Patroni) httpPostOrPatch(method string, url string, body *bytes.Buffer) (err error) {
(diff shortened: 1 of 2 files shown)


pkg/util/patroni/patroni.go
```diff
@@ -43,8 +44,23 @@ func New(logger *logrus.Entry) *Patroni {
...
+		return "", fmt.Errorf("%s is not a valid IP", masterPod.Status.PodIP)
+	}
+	var ipString string
+	// If IPv6, add braces
+	if ip.To4() == nil {
+		if ip.To16() == nil {
+			// Shouldn't ever get here, but library states it's possible.
+			return "", fmt.Errorf("%s is not a valid IPv4/IPv6 address", ip.String())
+		}
+		ipString = fmt.Sprintf("[%s]", ip.String())
```
erthalion: Probably we can even use [JoinHostPort](https://golang.org/pkg/net/#JoinHostPort) here, as it will take care of brackets.

erthalion:
@optix2000 Thanks for the PR, looks good. Could you please also add few unit tests for `apiURL` function? In the meantime we will try to check why travis is complaining.

pkg/util/patroni/patroni.go
optix2000: Good catch. Do you think it's worth keeping the IP validation since `JoinHostPort` handles both?

erthalion: Good point. Unfortunately I can't find any direct confirmation in K8S docs or code, that `PodIP` always contain correct `IP` string. Looks like in some cases `IP` could be not assigned and `PodIP` would contain an empty string, but I'm not sure if we can face such cases. Maybe you have some references, that proves correctness of IP address? Otherwise probably it makes sense to keep some validation just in case.

Jan-M:
I like IPv6 support :+1:

erthalion:
@optix2000 thanks for the update. So, could you please rebase your changed on top of the master, so we can check the tests? Plus my request about a few unit tests for apiURL function is still there.

optix2000:
Any good examples to look at for unit tests? Want to make sure I'm following the same convention

pkg/util/patroni/patroni_test.go
```diff
@@ -0,0 +1,71 @@
...
+			errors.New("::: is not a valid IP"),
+		},
+	}
+	for _, test := range testTable {
+		resp, err := apiURL(newMockPod(test.podIP))
+		if resp != test.expectedResponse {
+			t.Errorf("expected response %v does not match the actual %v", test.expectedResponse, resp)
+		}
+		if err != test.expectedError {
+			if err != nil && test.expectedError != nil && err.Error() != test.expectedError.Error() {
```
erthalion: I'm not sure, why this condition is necessary? If an actual error value is different from expected, then I guess it's a fail even if one of them is `nil`, is that correct?

erthalion:
> Any good examples to look at for unit tests? Want to make sure I'm following the same convention

Thanks, tests look good. I've just left one commentary, and then I think we can merge.

pkg/util/patroni/patroni_test.go
optix2000: Yeah it's for catching the case where an error was thrown, but none was expected or vice versa.

erthalion: Maybe I'm missing something, but isn't in this case if we expect no errors (`test.expectedError == nil`), but something wrong happened (`err != nil`), then we will still not fail the test, because `(err != nil && test.expectedError != nil) == false`? So that the nested condition is false and `t.Errorf` is not executed?

erthalion:
:+1:

sdudoladov:
:+1:

erthalion:
Great, @optix2000, thanks for the contribution!

PR #679, opened 2019-10-04
fixed chart default values to use OperatorConfigurationCRD

excavador:


FxKu:
There is a separate [values.yaml](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values-crd.yaml) to be used for CRD-based configuration. Have you seen it? Do you require the default values file to work for CRD configuration?

excavador:
> There is a separate [values.yaml](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values-crd.yaml) to be used for CRD-based configuration. Have you seen it? Do you require the default values file to work for CRD configuration?

I did. But how it supposed to be used? Copy-paste to your project and use for your release?

FxKu:
Yes, and you can specify the values file during chart installation:

```helm install --values ./charts/postgres-operator/values-crd.yaml` ...```

Any reason, you can't go with the default ConfigMap configuration?

excavador:
Thank you for explanation

Issue #681, opened 2019-10-10
Logical backup AWS access keys/custom AWS endpoint


IlyaSemenov:
The logical backup command runs as follows:

```
aws s3 cp - s3://foo-postgres-backups/spilo/foo-db/e223d869-7a53-11e9-8f3a-5254005c0982/logical_backups/1570692867.sql.gz --debug --expected-size 19594950 --sse=AES256
```

It's not passing neither AWS credentials nor custom AWS endpoint to the pod:

https://github.com/zalando/postgres-operator/blob/f0e29060b1fa0884fcf338e87492cd8e90bfd9e6/pkg/cluster/k8sres.go#L1540-L1548

Can it be extended to do so?

PR #682, opened 2019-10-10
Support custom AWS endpoint in logical backup container

IlyaSemenov:
Add support for `AWS_ENDPOINT` env variable in logical backup image. See #681.


FxKu:
Thanks for another PR :smiley: Seems to address the same issue like #628, which is going even further by making endpoint, sse etc. configurable.

IlyaSemenov:
Oh, right. I didn't find that one. Let me close my tickets then.

PR #689, opened 2019-10-16
Add support for nodeSelector

Bryji:
Add support for nodeSelector specification in the cluster manifest.

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -8,7 +8,6 @@ import yaml
 
 from kubernetes import client, config
 
-
 class EndToEndTestCase(unittest.TestCase):
     '''
     Test interaction of the operator with multiple K8s components.
@@ -28,7 +27,6 @@ class EndToEndTestCase(unittest.TestCase):
         In the case of test failure the cluster will stay to enable manual examination;
         next invocation of "make test" will re-create it.
         '''
-
         # set a single K8s wrapper for all tests
         k8s = cls.k8s = K8s()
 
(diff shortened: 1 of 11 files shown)


FxKu:
Thanks for this contribution. Would be super great if you could also write an [e2e-test](https://github.com/zalando/postgres-operator/blob/master/e2e/tests/test_e2e.py) for this. Testing the NodeSelector with our [kind](https://kind.sigs.k8s.io) e2e setup should be relatively straight forward. You just have to familiarize with the Python K8S API. In the meantime, I'll have a look at the code.

Bryji:
Ok, sounds like a good learning opportunity, I'll look into that.

FxKu:
Tested it in the meantime and can verify that it works.

Btw: With kind I have also tested it in combination with enabled pod anti affinity and only one of two nodes labeled which led to one unscheduleable Pod.

Bryji:
We also have the change the deployed in our dev environment, so confirmed working there also. I've partially implemented the test, just at the point where I can do the final assertion but I had to pause and look at other work. 

I've just tried it manually in kind 3 times - first time I saw same results as you - one node pending due to podAntiAffinity on the statefulset, 2nd and 3rd time there was no anti-affinity and both pods scheduled to the same node. "enable_pod_antiaffinity" must have somehow been true in my first run. If this is false (default) for the test I should be able to assert that both scheduled to the same node.

redbaron:
This is yet another case where https://github.com/zalando/postgres-operator/issues/479 would solve these kind of problems once and for all

e2e/tests/test_e2e.py
```diff
@@ -212,6 +220,90 @@ def test_logical_backup_cron_job(self):
...
+        target_node = k8s.api.core_v1.read_node(name='postgres-operator-e2e-tests-worker2')
+        pg_patch_env_special = {
+            "metadata": {
+                "labels": {
+                    "environment": "special"
+                }
+            }
+        }
+        target_node_name = target_node.metadata.name
+        print ("Patching node [%s] with 'environment:special' label" %(target_node_name))
```
FxKu: my flake 8 linter tells me, it doesn't like the space after print. Put it after the % operator and it's happy:

```suggestion
        print("Patching node [%s] with 'environment:special' label" % (target_node_name))
```

FxKu:
Just ran the e2e test locally and it worked. Not sure about the printed messages. If they would show the progress it would be fine. But they appear after the tests finish - so a bit useless I guess.

I'm also not sure if the nodeSelector test itself needs to be so "complicated". All you got to do is:

1. label one node,
2. patch the current manifest
3. delete the replica pod
4. check if it appears on the node with a `nodeSelector` defined
5. then label the other node, too, so that you don't get into trouble in subsequent tests (e.g. tainting - i get you force the order with `zz_`, but might not be necessary)

Btw. you could add a commented example in the complete manifest, to show users how to use this feature.

Bryji:
Thanks for the feedback @FxKu . I'll look at addressing your comments. I agree the test feels a bit clunky as is, I'll try your approach.

Bryji:
Implementing the test your way has actually turned up an issue. Looks like the operator doesn't detect that a new statefulset is required so nodeSelector doesn't get applied. Seems to only work on new deploys right now.

FxKu:
Hm, yesterday I've added the nodeSelector to a manifest during the e2e tests. During the scaling test, I saw that the new Pod was scheduled to the labeled node and had the nodeSelector defined in its manifest. Haven't looked at the statefulSet though.

Bryji:
It was working on a new deploy but not detecting changes to existing. Fixed in the last push and e2e test simplified as per suggestion.

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -58,16 +58,17 @@ type KubernetesMetaConfiguration struct {
 	ClusterLabels                          map[string]string     `json:"cluster_labels,omitempty"`
 	InheritedLabels                        []string              `json:"inherited_labels,omitempty"`
 	ClusterNameLabel                       string                `json:"cluster_name_label,omitempty"`
-	NodeReadinessLabel                     map[string]string     `json:"node_readiness_label,omitempty"`
+	NodeReadinessLabel        			   map[string]string	 `json:"node_readiness_label,omitempty"`
```
FxKu: What IDE or editor are you using? Please, use a Go and Python linter to format the code correctly :)

e2e/tests/test_e2e.py
```diff
@@ -212,6 +220,61 @@ def test_logical_backup_cron_job(self):
         self.assertEqual(0, len(jobs),
                          "Expected 0 logical backup jobs, found {}".format(len(jobs)))
 
+    @timeout_decorator.timeout(TEST_TIMEOUT_SEC)
+    def test_node_selector(self):
+        k8s = self.k8s
+
+        # 1. Add a label to a node
+        target_node = k8s.api.core_v1.read_node(name='postgres-operator-e2e-tests-worker2')
```
FxKu: I would rather see the current master node been labeled here, and not hard-coded by name (see taint test)

e2e/tests/test_e2e.py
```diff
@@ -212,6 +220,61 @@ def test_logical_backup_cron_job(self):
...
+            version="v1",
+            namespace="default",
+            plural="postgresqls",
+            name="acid-minimal-cluster",
+            body=pg_patch_custom_resource)
+
+        k8s.api.core_v1.delete_namespaced_pod("acid-minimal-cluster-0", "default")
+        k8s.api.core_v1.delete_namespaced_pod("acid-minimal-cluster-1", "default")
+        # Unfortunately necessary to allow time for sync to take place
+        time.sleep(90)
```
FxKu: should be sufficient to delete only the replica pod and use the wait function. Do not kill the master like this.

e2e/tests/test_e2e.py
```diff
@@ -212,6 +220,61 @@ def test_logical_backup_cron_job(self):
...
+
+        # Assertions:
+        # Sanity check, we expected 2 pods to be deployed
```
FxKu: I would rather check, if you find the replica on the labeled node and if it has the nodeSelector defined in it's `spec`.

e2e/tests/test_e2e.py
```diff
@@ -212,6 +220,61 @@ def test_logical_backup_cron_job(self):
...
+                         "Sanity check: expected {} replicas, found {}".format(2, len(pods_list.items)))
+
+        # Core test: both pods should now be deployed in target_node (since anti-affinity defaults to off)
+        for pod in pods_list.items:
+            print ("Found a pod on node [%s]" % (pod.spec.node_name))
+            self.assertEqual(target_node_name, pod.spec.node_name,
+                             "Expected {} replicas, found {}".format(2, pod.spec.node_name))
+
+        # Add label to the other nodes so other tests can run normally
+        k8s.api.core_v1.patch_node('postgres-operator-e2e-tests-worker', pg_patch_env_special)
```
FxKu: and this shouldn't be hard-coded either

pkg/util/config/config.go
```diff
@@ -43,6 +43,7 @@ type Resources struct {
 	MaxInstances            int32             `name:"max_instances" default:"-1"`
 	MinInstances            int32             `name:"min_instances" default:"-1"`
 	ShmVolume               *bool             `name:"enable_shm_volume" default:"true"`
+	NodeSelector            map[string]string `name:"node_selector" default:""`
```
FxKu: You introduced a new field to the configuration, but it's not reflected in other [files](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters). On the other hand, we already have the `NodeReadinesLabel` do define where Pods should be scheduled.

pkg/cluster/k8sres.go
```diff
@@ -329,6 +329,20 @@ func tolerations(tolerationsSpec *[]v1.Toleration, podToleration map[string]stri
 	return []v1.Toleration{}
 }
 
+func nodeSelector(clusterManifestSelector map[string]string, operatorSelectorConfig map[string]string) map[string]string {
```
FxKu: we already have the func `nodeAffinity` - you could extend it to also take the `nodeSelector` entries from the manifest. Then generateStatefulset and generatePodTemplate don't need more arguments. That means for the e2e test you have to look `nodeAffinity` and not `nodeSelector`.

pkg/cluster/k8sres_test.go
```diff
@@ -451,3 +452,66 @@ func TestSecretVolume(t *testing.T) {
...
+
+	out := acidv1.PostgresSpec{}
+	in.DeepCopyInto(&out)
+
+	if out.NodeSelector["env"] != env {
+		t.Errorf("TestNodeSelectorDeepCopy: NodeSelector deep copy env value [%s] did not match expected [%s]", out.NodeSelector["env"],  env)
+	}
+}
+
+
```
FxKu: all the blank lines will go away when linting

FxKu:
After taken a closer look, I think you've implemented a cluster-wise pendent to the `NodeReadinessLabel` from the global configuration. Maybe it makes more sense to also name the manifest parameter like this, although the name `NodeSelector` is more fitting.

And please, enable code linting.

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
Bryji: Using Intellij, I thought I was tidying that up!

FxKu:
Thanks again, @Bryji! Node affintiy support was just merged with #1166 .

Issue #633, opened 2019-07-28
Specify TLS / SSL certificate from secret


abh:
I have a TLS certificate in a "tls secret" in kubernetes with a valid certificate. I'd like postgres to use this instead of the self-signed certificate. How can I specify this in the cluster manifest?

FxKu:
TLS support is merged. So closing this issue. Thanks for pushing it.

PR #690, opened 2019-10-17
Custom TLS secrets

zimbatm:
Hi,

I am looking for early feedback on this PR. The goal right now is to talk about the design and if this is something Zalando would be interested in merging, plus an associated roadmap of things that would need to happen. If we could do that it would be much appreciated!

The main motivation is to be able to specify a custom TLS certificate from k8s secrets ( #633 ). Right now it only supports static secrets but down the road I would also like to support reloading the secrets whenever they change to allow proper key rotation and integration with cert-manager.

Cheers,
z


zimbatm:
Hey @Jan-M, thanks for taking a look :)

On our side we have the [cert-manager](https://docs.cert-manager.io/en/latest/) operator deployed to handles the certificates. It introduces new types of resources such as the `Issuer` resource that represents a CA and handles ACME, Vault, self-signed, ... and then the `Certificate` resource that represents a TLS certificate. The `Certificate` then generates a k8s secret with the same shape as `kubectl create secret tls`.

The idea of this PR at the moment is to add a new `tls.secretName` on the `postgresql` resource so that each PG cluster can get their own certificate. This makes the extension compatible with both cert-manager and `kubectl create secret tls`.

If I understand properly what you mentioned, your idea would be to have the postgres-operator trigger the certificate generation. The upside would be that ceritifcates would be generate on a per-host basis instead of on a per-cluster basis. But the downside is that now the postgres-operator has to integrate more tightly to either cert-manager or another operator.

Jan-M:
Nope, I do not think the operator should start dealing with TLS cert and issueing. So in favor of supporting necessary things to inject and or provide TLS certs to Postgres pods.

I just ment we had an idea once about a "CA" for local clusters.

Can we not have the secret name as a format string in the config and then some global and per cluster enable disable?

zimbatm:
Okay this PR is working now. I still need to handle certificate rotation somehow, and add more tests and documentation.

> Can we not have the secret name as a format string in the config and then some global and per cluster enable disable?

I'm not sure what the advantage would be. Since both the Certificate and Postgresql resources are created by the user, he is in the best position to define a name and make the connection explicit. Otherwise he would have to ask the administrator what secretName mapping has been configured on the operator.

zimbatm:
For certificate rotation, this could be handled automatically in the spilo container: https://github.com/zalando/spilo/pull/373

charts/postgres-operator/values-crd.yaml
```diff
@@ -93,10 +93,9 @@ configKubernetes:
   # template for database user secrets generated by the operator
   secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"
   # group ID with write-access to volumes (required to run Spilo as non-root process)
-  # spilo_fsgroup: 103
-
+  spilo_fsgroup: "103"
   # whether the Spilo container should run in privileged mode
-  spilo_privileged: false
+  spilo_privileged: "false"
```
abh: this change seem unnecessary / just cluttering the PR.

charts/postgres-operator/values.yaml
```diff
@@ -89,7 +89,7 @@ configKubernetes:
   # template for database user secrets generated by the operator
   secret_name_template: '{username}.{cluster}.credentials'
   # group ID with write-access to volumes (required to run Spilo as non-root process)
-  # spilo_fsgroup: "103"
+  spilo_fsgroup: "103"
```
abh: same comment as above

docs/user.md
```diff
@@ -425,3 +425,35 @@ monitoring is outside the scope of operator responsibilities. See
 [configuration reference](reference/cluster_manifest.md) and
 [administrator documentation](administrator.md) for details on how backups are
 executed.
+
+## TLS configuration
+
+By default self-signed certificates are generated when the postgres container
+starts up. But it's also possible to provide your own secrets like so:
+
+Generate a new cert:
```
abh: I don't think you need to document how to create the cert, just how to add it to kubernetes with `kubectl create secret` (as you do below) and how to configure the cluster.

go.sum
```diff
@@ -49,12 +49,14 @@ github.com/docker/docker v0.7.3-0.20190327010347-be7ac8be2ae0/go.mod h1:eEKB0N0r
 github.com/docker/go-units v0.3.3/go.mod h1:fgPhTUdO+D/Jk86RDLlptpiXQzgHJF7gydDDbaIK4Dk=
 github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96 h1:cenwrSVm+Z7QLSV/BsnenAOcDXdX4cMv4wP0B/5QbPg=
 github.com/docker/spdystream v0.0.0-20160310174837-449fdfce4d96/go.mod h1:Qh8CwZgvJUkLughtfhJv5dyTYa91l1fOUCrgjqmcifM=
+github.com/elazarl/goproxy v0.0.0-20170405201442-c4fc26588b6e h1:p1yVGRW3nmb85p1Sh1ZJSDm4A4iKLS5QNbvUHMgGu/M=
```
abh: are these necessary changes or just because your environment is different?

manifests/configmap.yaml
```diff
@@ -69,6 +69,7 @@ data:
   secret_name_template: '{username}.{cluster}.credentials'
   # sidecar_docker_images: ""
   # set_memory_request_to_limit: "false"
+  spilo_fsgroup: "103"
```
abh: is this required? (same question as above, I guess :-) )

charts/postgres-operator/values-crd.yaml
```diff
@@ -93,10 +93,9 @@ configKubernetes:
   # template for database user secrets generated by the operator
   secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"
   # group ID with write-access to volumes (required to run Spilo as non-root process)
-  # spilo_fsgroup: 103
-
+  spilo_fsgroup: "103"
```
abh: if the fsgroup is required for the TLS feature to work, it should be spelled out.

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -124,6 +125,12 @@ type StandbyDescription struct {
 	S3WalPath string `json:"s3_wal_path,omitempty"`
 }
 
+type TLSDescription struct {
+	SecretName string `json:"secretName,omitempty"`
+	CertificateFile string `json:"certificateFile,omitempty"`
+	PrivateKeyFile string `json:"privateKeyFile,omitempty"`
+}
+
```
abh: run `go fmt` ?

charts/postgres-operator/values-crd.yaml
zimbatm: yes please ignore. I am building on top of https://github.com/zalando/postgres-operator/pull/704 where we had similar conversations. I will rebase this PR when the other gets merged.

go.sum
zimbatm: They get changed automatically when doing a few operations. Once https://github.com/zalando/postgres-operator/pull/705 gets merged I will rebase this PR on top and it will disappear.

charts/postgres-operator/values-crd.yaml
zimbatm: good idea. I already have a check in the code but I will add a note in the documentation as well.

docs/user.md
zimbatm: ok, removed

go.sum
zimbatm: I removed that commit

charts/postgres-operator/values-crd.yaml
zimbatm: done!

pkg/apis/acid.zalan.do/v1/postgresql_type.go
zimbatm: fixed

zimbatm:
Alright, this is ready for another review

zimbatm:
now with tests!

zimbatm:
once https://github.com/zalando/spilo/pull/373 is merged I also want to add support for CA certificates

FxKu:
@zimbatm so we merged your PR in Spilo. Can you update this PR - so that we might include it in the next release.

zimbatm:
thanks @FxKu!

It looks like coverall wants me to write even more tests. Do you have a recommendation at what place I can write tests for pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go ?

FxKu:
At the moment, I see some overlap with #736 and also with the current additional secret mount [implementation](https://github.com/zalando/postgres-operator/blob/13e436276b49fea50f39135d6649d1645bd5bb6f/pkg/cluster/k8sres.go#L1130). If it would be possible to specify an array additional volumes, I guess both features could be covered with it. On the other hand, we also aim for low complexity of the cluster manifest. Just one flag `enableTLS` would be nice. Cert-file and private key file template could be part of the operator configuration.

zimbatm:
okay I will take a look tomorrow. Today I also added configuration to pass a CA file (not yet pushed) that will make the implementation more complete and also allow for checking client certificates.

zimbatm:
This PR is ready to me. It's tested and it's working for me (aside from the master build failures).

I think the other PR is nice but also opens the deployment to much more variance. Here the nice this is that the TLS setup is delegated to the controller and has a specific implementation, whereas the other PR allows to build a custom TLS scheme on top with a initContainer, custom Volume and maybe a sidecar container to handle key rotation. So the out-of-the-box experience is probably worse as the user has to figure out how to tie everything together. I believe that using a secret volume is the right call to store the TLS certificates.

abh:
Any chance this can be merged?

erthalion:
> Any chance this can be merged?

Sorry for leaving it hanging for so long. I believe we can take a look at this soon (at the beginning of January), but in general I agree with the summary from @zimbatm . I intent to merge another volume related PR, so maybe we can include this one too.

zimbatm:
Is there anything I can do to get this merged?

erthalion:
> Is there anything I can do to get this merged?

I'm afraid no, unfortunately it take so long on our side. But I have this PR in my plan.

zimbatm:
Well that's life. As long as you're not blocked by me :)

erthalion:
Btw, @zimbatm could you merge master once into this PR? Or make it available to push to this branch for operator maintainers. We can start fixing travis tests here.

zimbatm:
@erthalion: rebased! I am looking into how to give you write access to the branch.

erthalion:
> I am looking into how to give you write access to the branch.

Apparently there should be some [toggle](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork) already

zimbatm:
It looks like this option is disabled when the PR is coming from an org instead of an account. Do you want me to open a new PR coming from my account?

zimbatm:
just went ahead and opened #798

erthalion:
> It looks like this option is disabled when the PR is coming from an org instead of an account.

Interesting. Nope, let's keep it like it is right now. If I would like to suggest something, I can fork a PR fork :)

Jan-M:
Closing as superseded by newer PR

PR #692, opened 2019-10-19
Fix parameter expansion when performing logical backup.

OlleLarsson:
This pull request is a minor fix for the function  `get_config_resource` in `dump.sh`. 
Previously paramater expansion was not being performed and the string litteral value `"$1"` was used in the select statement.


FxKu:
Thanks for your PR. In the meantime I realized that I could remove querying the config reference from the deployment and instead inject the cluster name label in the logical backup pod. So this fix is then obsolete.

PR #599, opened 2019-06-26
Add CRD validation

FxKu:
added CRD manifests with validation for `postgresql` and `OperatorConfiguration`

closes #388
closes #269 
closes #72  
addresses #537

PR #693, opened 2019-10-21
Defaults to operator configuration.

pierreozoux:
In the doc, this is now thee recommendation.


FxKu:
It requires more changes to switch to CRD-based config as the default. Currently, the deployment registers the `OperatorConfiguration` CRD. So this has to happen first before applying the configuration manifest. Unfortunately, when we reference the configuration manifest already in the deployment it causes the operator pod to fail until we apply the configuration yaml. Providing a CRD yaml that one can apply first would solve this issue.

On the other hand using the CRD-based config over the ConfigMap only makes a difference if we would also provide CRD validation. There's already #599 for this. The PR also contains a yaml for both CRDs the operator is using. I've already included your idea there :)

pierreozoux:
Great! Then I close mine ;)

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -53,7 +53,7 @@ configuration:
     # pod_priority_class_name: ""
     pod_role_label: spilo-role
     # pod_service_account_definition: ""
-    pod_service_account_name: postgres-pod
+    pod_service_account_name: zalando-postgres-operator
```
FxKu:
```suggestion
    pod_service_account_name: postgres-pod
```

mikkeloscar:
:+1:

PR #694, opened 2019-10-22
charts: Add pods/exec permission

hasegeli:
Without this it fails with: `could not sync persistent volumes: could not sync volumes: could not resize the filesystem on pod \"testns/testdb-0\": could not get device and type for the postgres filesystem: could not execute: pods \"testdb-0\" is forbidden: User \"system:serviceaccount:postgres-operator:postgres-operator\" cannot create resource \"pods/exec\" in API group \"\" in the namespace \"testns\"`

### DIFF
--- a/charts/postgres-operator/templates/clusterrole.yaml
+++ b/charts/postgres-operator/templates/clusterrole.yaml
@@ -90,6 +90,12 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - pods/exec
+  verbs:
+  - create
+- apiGroups:
+  - ""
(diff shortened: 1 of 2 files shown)


FxKu:
Thanks for finding it. Could you also insert the permission in [this](https://github.com/zalando/postgres-operator/blob/master/manifests/operator-service-account-rbac.yaml) file, please? :)

hasegeli:
Done

FxKu:
:+1:

CyberDem0n:
:+1:

PR #696, opened 2019-10-29
Switch to CRD-based config as the default

FxKu:
With CRD validation for `OperatorConfiguration` in place, this configuration option can be regarded superior to ConfigMap one.

Also suggested by @pierreozoux in #693


pierreozoux:
Can we get this moving?

If I can help?

(It is just that I need to maintain a fork for the kustomize to work, and it would be awesome if this is upstream)

danke!

FxKu:
Closing this PR as it has to many conflicts and helm chart values file is already merged

PR #704, opened 2019-11-03
spilo_fsgroup: set to 103 by default

zimbatm:
Setting the FSGroup to match the postgres GID is a requirement for postgres to load the TLS certificates mounted on a secret volume (which will be handled by #690).

So now the question is, why not always set the FSGroup to match the postgres GID?


charts/postgres-operator/values-crd.yaml
```diff
@@ -93,10 +93,9 @@ configKubernetes:
   # template for database user secrets generated by the operator
   secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"
   # group ID with write-access to volumes (required to run Spilo as non-root process)
-  # spilo_fsgroup: 103
-
+  spilo_fsgroup: "103"
   # whether the Spilo container should run in privileged mode
-  spilo_privileged: false
+  spilo_privileged: "false"
```
FxKu: IIRC, CRD-config does not like quotes for bools. Please remove them.

charts/postgres-operator/values-crd.yaml
```diff
@@ -93,10 +93,9 @@ configKubernetes:
   # template for database user secrets generated by the operator
   secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"
   # group ID with write-access to volumes (required to run Spilo as non-root process)
-  # spilo_fsgroup: 103
```
FxKu: Regarding [this](https://github.com/zalando/postgres-operator/pull/531#discussion_r289402865) comment, quotes should be removed.

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -38,7 +38,7 @@ configuration:
     pod_service_account_name: operator
     pod_terminate_grace_period: 5m
     secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"
-    # spilo_fsgroup: 103
```
FxKu: samg here, remove the quotes

FxKu:
Please, add the default for SpiloFSGroup also in the internal [struct](https://github.com/zalando/postgres-operator/blob/340dc4aa3dc6c0ff9d575ca58bc6a7040c959171/pkg/util/config/config.go#L28).

Not sure if I remember the discussion from #531 correctly, does this work out-of-the-box or is it required to patch the Spilo image? @erthalion

charts/postgres-operator/values-crd.yaml
zimbatm: Are you sure about this?

I tested the changes in `manifests/configmap.yaml` where it fails when the values are not quoted. I assume the rest of the YAML files are using the same parser/schema?

zimbatm:
I have a local `kind` setup to test these changes. I only tested that the master and slave come up properly. Let me know if you want me to do some more testing.

charts/postgres-operator/values-crd.yaml
FxKu: For the ConfigMap it's indeed necessary to quote such flags. That's why I only commented on the CRD-baed config changes ;)

zimbatm:
@FxKu the PR is ready for another review

zimbatm:
rebased

erthalion:
@zimbatm Could there be any other use cases for this except mentioned in #690 ? The change by itself is fine, but can come with compatibility issues if someone would try to use an image different from Spilo (e.g. one of it's modifications) and system gid for postgres could be different from 103.

zimbatm:
How I see things, the operator is tightly coupled with the spilo image and if the user does something custom they would have to override the default gid as well. Most likely their custom spilo image would be a fork from the official spilo project and inherit the gid.

The motivation of this change is to make the TLS setup a tiny bit easier. I don't have other scenarios in mind that could also benefit from this change. Although I suspect that any other volume that is mounted in the spilo pod would probably want to be readable from the postgres group.

erthalion:
> How I see things, the operator is tightly coupled with the spilo image

Sort of, but tight coupling doesn't mean it's not possible to use some modifications, and how far those modifications are coming is up to an author. I would also mention that this is a pod-level parameter, which means all sidecars will also have it. Overall since all this is only for one feature, changing the default value seems questionable for me.

But probably it makes sense to go other way around. I don't see why we can't say in #690 that if we have custom TLS section defined, then also set `spilo_fsgroup` to some predefined default value, which for now can be just a constant suitable for the original spilo image. What do you think?

zimbatm:
Alright. Thanks for looking into this and replying to my questions. I was hoping that this change would simplify the TLS setup and ultimately remove one if-condition branch in the whole system. But like you said, the security context is applied to the whole pod unfortunately and I don't see another route. #690 already checks to make sure that the FSGroup is being set so I will push that to the TLS configuration documentation instead.

erthalion:
Thank you.

>  I will push that to the TLS configuration documentation instead.

But actually I've meant that we can set `spilo_fsgroup` in this case automatically, without requiring to modify operator configuration. At the end of the day operator generates a pod manifest, so why not set `securityContext.FSGroup` there?

zimbatm:
Okay

PR #706, opened 2019-11-04
remove scalyr sidecar

FxKu:
because in most cases current Postgres log setup is sufficient


Jan-M:
No viable right now to remove code part. Scalyr implementation may change in the future.

PR #710, opened 2019-11-05
configurable resources for logical backup pod template

### LINKED ISSUE #688: Logical backup resource limits
Hey folks -- thanks for such an awesome piece of open source software!

I'm running some very big resource limits for my PGs to power a memory and IOPS hungry analytics system. We're very sensitive to how much these huge instances cost, so we don't really have much spare memory or CPU of the size we need to run the instances. Because the logical backup pods use the same resource requests as the databases, this means our logical backups are unschedule. We'd rather not overprovision our cluster to have more of these expensive instances, and the backup is streaming so it shouldn't need the crazy CPU / memory the db nodes themselves have.

I took a look at the source and it seems like there isn't a way to adjust the resource limits the backup pod requests, and it seems like that might be on [purpose](https://github.com/zalando/postgres-operator/blob/f0e29060b1fa0884fcf338e87492cd8e90bfd9e6/pkg/cluster/k8sres.go#L1429).

Would you be open to allowing operators to adjust the logical backup resource requests and make them smaller for situations like mine? For the time being I can just manage the cronjob resource outside the operator, but it'd be nice to be able to have it just work!

FxKu:
If one has configured quite hungry pods in the manifest, one still has to consider it for logical backup pods. There's no way to configure it. ~By relying only on the default resources it's up to the user to choose adequate requests and limits.~

Edit 2023: PR has been updated to provide resource config options for logical backup pod template. When not configured default postgres pod resources will be used. In the future we can also provide manifest fields and use the config options as defaults.

fixes #688


FxKu:
I talked with @sdudoladov back when I opened this PR. I think he was more in favor for adding another config parameter. I could also think of taking the resources of the Postgres cluster and dividing them by a configurable factor.

Makc2032 (approved):

ToGoBananas (approved):

jopadi:
👍

FxKu:
:+1:

PR #715, opened 2019-11-12
disable load balancers by default in examples

FxKu:
load balancers services should not be the default in our examples


Jan-M:
:+1:

FxKu:
:+1:

PR #716, opened 2019-11-12
fix errors when changing service type

### LINKED ISSUE #713: Service creation fails at first sync after enabling/disabling Load Balancer type
I have a bare-metal k8s cluster, which doesn't have a LoadBalancer. I only use postgresql for cluster database connection so load balancer is not needed.

postgres-operator tries to deploy a load balancer and it stuck in "Load balancer is being provisioned". So far there's no problem but it makes me worry something might happen. I tried to change the service to ClusterIP but the operator changes it back after a while.

FxKu:
when service type is changed, operator deletes the current service and tries to create a new one immediately, which fails as deletion doesn't happen that fast. This PR reuses the same retry/wait logic from StatefulSet deletion. 

After this part of `updateService`, creation order of endpoint and service is also reversed to avoid connection problems while syncing roles in the next step

fixes #713

### DIFF
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -13,7 +13,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/zalando/postgres-operator/pkg/util"
-	"github.com/zalando/postgres-operator/pkg/util/constants"
 	"github.com/zalando/postgres-operator/pkg/util/k8sutil"
 	"github.com/zalando/postgres-operator/pkg/util/retryutil"
 )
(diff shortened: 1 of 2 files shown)


erthalion:
:+1:

pkg/cluster/resources.go
```diff
@@ -278,19 +277,25 @@ func (c *Cluster) replaceStatefulSet(newStatefulSet *appsv1.StatefulSet) error {
...
+	if err != nil {
 		return fmt.Errorf("could not delete statefulset %q: %v", statefulSetName, err)
 	}
 	// make sure we clear the stored statefulset status if the subsequent create fails.
 	c.Statefulset = nil
 	// wait until the statefulset is truly deleted
 	c.logger.Debugf("waiting for the statefulset to be deleted")
 
-	err := retryutil.Retry(constants.StatefulsetDeletionInterval, constants.StatefulsetDeletionTimeout,
+	err = retryutil.Retry(c.OpConfig.ResourceCheckInterval, c.OpConfig.ResourceCheckTimeout,
```
sdudoladov: per operator docs:
```
Since in the CRD-based case the operator needs to create a CRD first, which is controlled by the resource_check_interval and resource_check_timeout parameters, those parameters have no effect and are replaced by the CRD_READY_WAIT_INTERVAL and CRD_READY_WAIT_TIMEOUT environment variables. 
```
does it work with the operator CRD ?

FxKu: This is confusing. These parameters are also used in other places. Tested now with CRD-based configuration and it works. I think, the docs are only referring to the [creation](https://github.com/zalando/postgres-operator/blob/f9487e41c1cdfc27ef461fb35fe35e0df2f0e121/pkg/controller/util.go#L71) of the CRD itself.

erthalion: Lets then update the docs to be less WFTish.

sdudoladov:
:+1:

PR #718, opened 2019-11-14
feat: use controller namespace for logical backup API calls

bchrobot:
Currently, the controller must be deployed in the default namespace in order for logical backups to
work. Logical backups should work regardless of the namespace the controller is deployed in.

### DIFF
--- a/docker/logical-backup/dump.sh
+++ b/docker/logical-backup/dump.sh
@@ -68,7 +68,7 @@ declare -a search_strategy=(
 )
 
 function get_config_resource() {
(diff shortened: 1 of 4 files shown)


FxKu:
Thanks for finding this issue. This should work for all namespaces, indeed. But, I would not add a config parameter to store the deployment namespace as it's very easy to make them inconsistent to each other. You could simply create the deployment in a namespace different to the config.

What about querying the K8s REST API for find the deployment (search for `postgres-operator` in --all-namespaces)? Still, people could name the deployment differently, of course. So, maybe the controller could pass it's namespace as an `env` variable to a Postgres cluster resource.

PR #724, opened 2019-11-18
add get for deployments to operator RBAC

### LINKED ISSUE #720: Service account cannot access deployments under /apis/apps/v1
Hello,

I've been trying to get logical backups to run, but I keep getting a 403. The function [get_config_resource](https://github.com/zalando/postgres-operator/blob/master/docker/logical-backup/dump.sh#L70) in dump.sh is trying to access /apis/apps/v1/..., but k8s throws an error that says: service account cannot get resource \"deployments\" in API group \"apps\".

I deployed the operator using the helm chart, and looking at the clusterrole.yaml file, I don't see a section for apps/deployments. If I add:

```
- apiGroups:
  - "apps"
  resources:
  - deployments
  verbs:
  - get
```
then it works. So, is this missing from the configuration?

FxKu:
Fixes #720 

Thanks @alejom99 for finding this out!

### DIFF
--- a/manifests/operator-service-account-rbac.yaml
+++ b/manifests/operator-service-account-rbac.yaml
@@ -106,6 +106,12 @@ rules:
 - apiGroups:
   - apps
   resources:
+  - deployments
+  verbs:
+  - get
+- apiGroups:
+  - apps
+  resources:
   - statefulsets
   verbs:
   - create


FxKu:
:+1:

Jan-M:
:+1:

PR #642, opened 2019-08-05
Fix label selector in logical backup script

yujunz:
Closes #641

PR #725, opened 2019-11-18
inject cluster name label into logical backup pod

FxKu:
Why complicated querying for the cluster name label when we can inject it into the logical backup pod?

Closes: #718 
Fixes: #692 
Fixes: #642 

Thanks to
@bchrobot pointing out the limitation for calling resources only on the default namespace,
@OlleLarsson for checking config call and
@yujunz for raising the awareness of not using a hard-coded cluster name label


FxKu:
:+1:

sdudoladov:
:+1:

PR #729, opened 2019-11-20
Set logical backup for Azure

gmldba:
Added parameters to manifests.
Added azure cli to the logical backup dockerfile
Added function to upload logical backup to azure storage
Added conditional to only upload according to the present settings.
Added documentation to explain how to set backups for Azure.


hoangelos:
I assume this was never merged?  We're considering using Zalando operator, but really hate that it's tied to AWS, since we use Azure.

PR #731, opened 2019-11-22
Enforce minimum cpu and memory limits

### LINKED ISSUE #135: No support for volumes less than 1GB
The operator uses gigabytes internally to evaluate the disk space when resizing volumes, so the cluster is always created with at least 1Gi volume, however, cluster updates compare the existing volume size with the actual volume size in the manifest and if the latter is less than 1Gi they fail.

This is clearly visible since our test manifest uses 512Mi volumes.

FxKu:
CRD validation does not catch if someone specified too small CPU and memory limits. This PR suggest to error out whenever resources are set too low on create and update (if edited).


docs/user.md
```diff
@@ -40,6 +40,10 @@ you can find this example also in the manifests folder:
 kubectl create -f manifests/minimal-postgres-manifest.yaml
 ```
 
+Note, that the minimum volume size to properly run the `postgresql` resource is
+`1Gi`. If a lower value is set in the manifest the operator will cancel ADD,
```
sdudoladov: it is the minimum  size of EBS volumes.

docs/user.md
```diff
@@ -182,6 +186,32 @@ See [infrastructure roles secret](../manifests/infrastructure-roles.yaml)
...
+    requests:
+      cpu: 10m
+      memory: 100Mi
+    limits:
+      cpu: 300m
+      memory: 300Mi
+```
+
+The minimum limit to properly run the `postgresql` resource is `256m` for `cpu`
+and `256Mi` for `memory`. If a lower value is set in the manifest the operator
```
sdudoladov: Does the quickstart guide run successfully with such values ?

docs/user.md
```diff
@@ -182,6 +186,32 @@ See [infrastructure roles secret](../manifests/infrastructure-roles.yaml)
...
+will cancel ADD, UPDATE or SYNC events on this resource with an error. If no
+resources are defined in the manifest the operator will obtain the configured
+[default requests](reference/operator_parameters.md#kubernetes-resource-requests).
```
sdudoladov: what will happen if defaults are lower than `256m` / `256Mi`?

pkg/cluster/sync.go
```diff
@@ -34,6 +35,16 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 		}
 	}()
 
+	if err = c.validateResources(&c.Spec); err != nil {
```
sdudoladov: I assume the min resource requirement affects only `postgres` pods and not Scalyr and other sidecars ?

pkg/cluster/cluster.go
```diff
@@ -491,6 +495,53 @@ func compareResourcesAssumeFirstNotNil(a *v1.ResourceRequirements, b *v1.Resourc
 
 }
 
+func (c *Cluster) validateResources(spec *acidv1.PostgresSpec) error {
+
+	const (
+		cpuMinimum     = "256m"
```
sdudoladov: The enforcement of such limits and  their values should imo be configurable.

docs/user.md
FxKu: ah, good to know. I'll add that info.

docs/user.md
FxKu: it does

pkg/cluster/sync.go
FxKu: yes. Is it not clear from the code?

pkg/cluster/cluster.go
FxKu: I think in this case it's better if we define what's good for the operator. Note, that only the `spec` is validated. You could still configure lower default values and they will be used if you do not specify requests/limits in the cluster manifest.

sdudoladov:
>  #599, but it will not catch if someone specified too small resource limits or volume size.

to give some context to these decisions:   The operator relies on Patroni for failovers. This daemon runs within the `postgres` container together with the PG server. Setting too low memory limit for the container may cause Patroni or PG to be killed due to OOM. The too low CPU limit may in theory starve Patroni thus causing unnecessary failovers due to TTL expiration although I do not recall observing this effect in practice (unlike OOM kills).

pkg/cluster/sync.go
```diff
@@ -34,6 +35,16 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
...
+		err = fmt.Errorf("insufficient resources specified: %v", err)
+		if oldStatus.Running() {
```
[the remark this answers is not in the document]
FxKu: Updated the docs. When everything is running fine, we shouldn't cancel the event. I'm fine doing the same for UPDATE ... just thought about the case when somebody updates the limits to lower than the current values in the spec. Maybe I could only check on that and cancel the event.

pkg/cluster/cluster.go
```diff
@@ -491,6 +495,53 @@ func compareResourcesAssumeFirstNotNil(a *v1.ResourceRequirements, b *v1.Resourc
...
+		}
+	}
+
+	storageSize := spec.Volume.Size
+	isSmaller, err = util.IsSmallerQuantity(storageSize, storageMinimum)
+	if err != nil {
+		return fmt.Errorf("error validating volume size: %v", err)
+	}
+	if isSmaller {
+		return fmt.Errorf("defined volume size %s is below required minimum %s to properly run postgresql resource", storageSize, storageMinimum)
```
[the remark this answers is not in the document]
FxKu: Good point. I'm thinking about dropping the enforcement for volume size from this PR

pkg/cluster/cluster.go
```diff
@@ -512,6 +554,21 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
 		}
 	}()
 
+	// check if pod resources were edited below the enforced minimum limits
+	if err := c.validateResources(&newSpec.Spec); err != nil {
+		err = fmt.Errorf("insufficient resource limits specified: %v", err)
+
+		isCPULimitSmaller, err2 := util.IsSmallerQuantity(newSpec.Spec.Resources.ResourceLimits.CPU, oldSpec.Spec.Resources.ResourceLimits.CPU)
```
sdudoladov: why check resource limits again ? isn't it caught by `err := c.validateResources(&newSpec.Spec);` at the line above ?

pkg/cluster/cluster.go
```diff
@@ -491,6 +495,43 @@ func compareResourcesAssumeFirstNotNil(a *v1.ResourceRequirements, b *v1.Resourc
 
 }
 
+func (c *Cluster) validateResources(spec *acidv1.PostgresSpec) error {
+
+	const (
+		cpuMinLimit    = "256m"
```
sdudoladov: the limits are worth a comment justifying their existence. sth like
```
setting too low limits causes unnecessary evictions / OOM kills
```

pkg/cluster/cluster.go
FxKu: moved the comment here and explain my intention a bit better

sdudoladov:
:+1:

FxKu:
:+1:

PR #735, opened 2019-11-25
WIP: transition from Helm V2 crd-install hook to Helm v3 crds/

rektide:
Hello. I tried installing postgres-operator with helm v3.

When I did the helm install, it gave me two of this warning:

> manifest_sorter.go:175: info: skipping unknown hook: "crd-install"

Diving into the [Helm hooks docs](https://v3.helm.sh/docs/topics/charts_hooks/), I found this information on the crd-install hook:

>  crd-install: In Helm 2, this executes after templates are rendered, but before the regular installation has been run. In Helm 3, this has been deprecated in favor of the crds/ directory.

In response to this, I moved the `template/customresourcedefinitions.yaml`  files into `crds/operatorconfigurations.yaml` and `crds/postgresqls.yaml` files. I then got some errors with the [templating in the labels](https://github.com/zalando/postgres-operator/compare/master...rektide:v3-helm#diff-63ecfcdb64b48f68d5685a8f34f19b30R6-R9), and created another commit to remove these labels, under the assumption that Helm v3 adds the right labels automatically.

This is still WIP, having some issues getting clean logs at startup, but I think this is a start on getting the chart to be Helm v3 compatible. I don't know what a good practice is for supporting Helm v2 AND Helm v3 users in this repo. Perhaps we need a charts/postgres-operator-v2 and charts/postgres-operator-v3 split? How are other projects handling Helm v3?

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -3,10 +3,8 @@ kind: CustomResourceDefinition
 metadata:
   name: operatorconfigurations.acid.zalan.do
   labels:
-    app.kubernetes.io/name: {{ template "postgres-operator.name" . }}
-    helm.sh/chart: {{ template "postgres-operator.chart" . }}
-    app.kubernetes.io/managed-by: {{ .Release.Service }}
-    app.kubernetes.io/instance: {{ .Release.Name }}
+    #app.kubernetes.io/managed-by: {{ .Release.Service }}
+    #app.kubernetes.io/instance: {{ .Release.Name }}
 spec:
   group: acid.zalan.do
   names:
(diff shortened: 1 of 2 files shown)


FxKu:
Thanks for notifying us about this change. I could also think of getting rid of the crd manifests and let only the operator deployment do the job. That would make code maintenance easier.

jparal:
re: how other projects deal with helm v3, check out prometheus-oprator where they made the CRDs backward compatible with helm v2:
https://github.com/helm/charts/commit/89b233eef6dbc1b6fac418bde3a5a6f4e14406d4#diff-e4d188f136b67765a471cbb46268cbd2

brunsgaard:
@jparal I followed the same direction as the prometheus operator helm chart in #738

FxKu:
Closing this PR in favor for #738 

Nevertheless, thanks @rektide for brining up the topic!

PR #744, opened 2019-11-30
include status subresource in validation

FxKu:
otherwise editing manifests is a bit tedious.

EDIT: When running Postgres clusters since before the v1.2.0 release, it could be that the status is still in the old format, which is `Status: string`. This happens because during sync we only update the status to the new format, if the status [was not running](https://github.com/zalando/postgres-operator/blob/551564011139d5a6cb9f635d091b2b54f56f125d/pkg/cluster/sync.go#L32). The validation does not affect running clusters. When you create new ones, you will not specify the status as this is up to the operator to define. You would only run into an error when manually editing the manifest, with status not being `PostgresClusterStatus: Running`. But then you would simply delete that line, and validation would succeed. That's actually the scenario this PR want's to solve for the new format :smiley:


charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -298,3 +298,7 @@ spec:
                   pattern: '^(\d+(e\d+)?|\d+(\.\d+)?(e\d+)?[EPTGMK]i?)$'
                 scalyr_server_url:
                   type: string
+        status:
+          type: object
+          additionalProperties:
+            type: string
```
brunsgaard: Just to start the flamewar in yet another Github repo. You are missing newlines in the end of some of your files. 😈

FxKu: Thanks. Used my private Laptop whose editor lacks the go linting.

sdudoladov:
@FxKu for documentation purposes please add a comment describing why this validation should not affect older `postgresql` manifests where the status was a string and not a  subresource

sdudoladov:
:+1:

FxKu:
:+1:

PR #746, opened 2019-12-02
move to apiextensions from v1beta1 to v1

FxKu:
v1beta1 is marked as deprecated in K8s 1.16. CRD structure changes from:

```
spec:
  additionalPrinterColumns: {}
  subresources:
    status: {}
  version: v1
  validation:
    openAPIV3Schema: {}
```

```
spec:
  versions:
    name: v1
    additionalPrinterColumns: {}
    subresources:
      status: {}
    schema:
      openAPIV3Schema: {}
```

EDIT: v1 for CRDs is not available before v1.16 (see [this](https://github.com/kubernetes/kubernetes/issues/82443) thread) which results into this error:
```
no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"
```

So this PR has to wait. When creating CRDs on v1.16 clusters, they will get transformed into the new schema, for what we saw in tests.


Hades32:
I'm getting errors in Helm about this now. Seems it has moved to a hard-deprecation...

Could you re-base this PR?

sdudoladov:
:+1:

FxKu:
:+1:

Issue #460, opened 2019-01-19
Switch to having Kubernetes do (EBS) volume management / resizing


frittentheke:
Right now the trusty postgres-operator does use the AWS API directly to resize EBS volumes and requires a call of of resize2fs inside the running pod after the volume has been grown - (https://github.com/zalando-incubator/postgres-operator/blob/master/docs/user.md#increase-volume-size).

Starting with 1.11 Kubernetes got more capable of managing persistent volumes and also does resizing of them now, see: https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/

Snapshotting volumes also comes along (currently in alpha) - https://kubernetes.io/blog/2019/01/17/update-on-volume-snapshot-alpha-for-kubernetes/ if that would ever be sensible to use (instead of using pg_basebackup to clone an instance).

My point here is to consider switching to only using Kubernetes for volume management / resizing as this will make the postgres-operator even more agnostic to the cloud provider / volume types and frees the operator from having to do yet another job.

hjacobs:
:+1:, I know that @Jan-M was already commenting on one or more Kubernetes issues (can't find them now) regarding resize/snapshot.

erthalion:
> Starting with 1.11 Kubernetes got more capable of managing persistent volumes and also does resizing of them now, see: https://kubernetes.io/blog/2018/07/12/resizing-persistent-volumes-using-kubernetes/

Looks cool, thank you. We indeed need to implement something like that, but probably only as an alternative for now, because as I can see from the docs, it requires a pod to be restarted, and to resize online we need to have [ExpandInUsePersistentVolumes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#resizing-an-in-use-persistentvolumeclaim) enabled (which is in alpha so far).

Jan-M:
We want to rely as much as possible on Kubernetes to solve our problems, thus we will look into supporting this. Due to the lack of online resize we will continue to support it the way we do right now for AWS to avoid the pod restart.

Jan-M:
As for EBS snapshots, I had some uncertainty in the past on how well this works for Postgres, but @CyberDem0n thinks this should be fine to use as a method so maybe a use case can be specced out.

We will spend a bit of work on logical backup support, this can be more useful due to smaller size on s3 or similar.

frittentheke:
[...]

PR #747, opened 2019-12-03
Allow preprovisioned persistent volumes / fix backup docker image

fwagner-smf:
With the `useExistingClaim` it is possible to use preprovisioned persistent volume claims for the stateful set.

Fix of `dump.sh` script in the logical_backup docker image. The `aws` command was not correctly quoted when using a custom endpoint url.


FxKu:
@fwagner-smf thanks for the PR. Could you move the fix on `dump.sh` to a separate PR?

fwagner-smf:
Hi @FxKu -> done with addtional PR #750

fwagner-smf:
Hi @FxKu, any chance we get this PR merged?
Thx.

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -571,6 +571,9 @@ var PostgresCRDResourceValidation = apiextv1beta1.CustomResourceValidation{
 							"storageClass": {
 								Type: "string",
 							},
+							"exitingVolumeClaim": {
```
FxKu: should be `useExistingClaim`

fwagner-smf: thanks for the finding. fixed.

hengchu:
I would also appreciate the addition of this feature. I have a small k8s cluster on AWS with one instance group per AZ. I have noticed that sometimes postgres-operator would try to provision and mount an EBS volume in (for example) us-east-2c to a node on us-east-2a, resulting in scheduling failures.

With this feature, I can lock down the AZ of EBS volumes in sync with the AZ of my nodes, and will avoid such issues in the future.

Thanks!

frittentheke:
@hengchu actually provisioning of the (backing) persistent volumes is not done by the operator, but rather the CSI driver (or provisioner for that matter) or if you are running an older version of Kubernetes by kubelet itself (AWS SDK included).

I believe the approach to use existing volume or dynamically provision them is about setting a storage class or not in the PersistentVolumeClaim which supports Dynamic Provisioning (https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#enabling-dynamic-provisioning).
Otherwise the PVC will be pending a matching persistent volume that has to be created "manually".

I proposed switching to letting Kubernetes (or rather CSI) also handle the resizing of volumes in issue:  https://github.com/zalando/postgres-operator/issues/460

fwagner-smf:
@frittentheke but this doesn't speak against merging this PR, or? This is quite a common way of handling existingClaims (at least in helm charts)

frittentheke:
Quickly looking at the code, I do not understand why you would not want the PVC to be created?
The "claim" merely tells Kubernetes what kind of persistent volume your pods would require to be "happy". Via the storage class you already can enable or disable dynamic provisioning and then handle the actual volumes and their properties yourself. Is there anything special you would configure in the PersistentVolumeClaims themselves that the operator cannot do?

fwagner-smf:
AFAIK if I want to manually provide a specifically named PVC, which I want to explicitly use in a deployment, then I have to be able to provide that PVC with the here added `useExistingClaim` parameter.

frittentheke:
Well yes, if you want to be in control of that PVCs name you certainly have to be able to configure it somehow. I was more getting at on why you care about the PVCs name (or rather the contents).

Well Gitlab Helm Chart does have a mechanism like this: https://docs.gitlab.com/charts/charts/gitlab/gitaly/#git-repository-persistence

There the reason to manually create a PVC (with an expected name) is that you might want to configure something within that PVC that the chart does not allow - like pointing it to a volume by name. 

If that is your use case, the patch could very well be to allow configuration of the PVC itself in greater detail, like being able to set a volume name.

hengchu:
> @hengchu actually provisioning of the (backing) persistent volumes is not done by the operator, but rather the CSI driver (or provisioner for that matter) or if you are running an older version of Kubernetes by kubelet itself (AWS SDK included). That component also has to take into account where the volume is needed and (topology awareness provided) Kubernetes needs to schedule the Pod accordingly.
> 
> I believe the approach to use existing volume or dynamically provision them is about setting a storage class or not in the PersistentVolumeClaim which supports Dynamic Provisioning (https://kubernetes.io/docs/concepts/storage/dynamic-provisioning/#enabling-dynamic-provisioning).
> Otherwise the PVC will be pending a matching persistent volume that has to be created "manually".
> 
> I proposed switching to letting Kubernetes (or rather CSI) also handle the resizing of volumes in issue: #460

Thanks for the tip! That worked for me :)

fwagner-smf:
> So I'd rather use that approach of skipping the PVC creation, but to still expect a PVC with a name by convention to exist.

Ok, that would be a valid approach as as well.
If I see it correctly, that should be possible already w/o code change... Then I overlooked that option.

frittentheke:
> > So I'd rather use that approach of skipping the PVC creation, but to still expect a PVC with a name by convention to exist.
> 
> Ok, that would be a valid approach as as well.
> If I see it correctly, that should be possible already w/o code change... Then I overlooked that option.

If so - great, but then this should be documented then in the users documentation for using the operator to not remain a "secret side effect" we discovered discussing this PR ;-)

FxKu:
@fwagner-smf so all you want is to provision the persistent volumes yourself? Then specifying a storage class under the `volume` field (that does not provide automatic provisioning) sounds like a viable option and we could close this PR. I agree, that such a case could be documented either in the admin docs or with more detail in the reference

PR #757, opened 2019-12-10
Fix resource validation

FxKu:
use the same pattern matching then from the configuration for the resource section in manifest. Otherwise something like:
```yaml
cpu:
  limit: "3"
``` 
will not be allowed


Jan-M:
:+1:

sdudoladov:
:+1:

PR #754, opened 2019-12-10
go.mod: fix dependencies

zimbatm:
Running `go test -v ./...` on master currently fails with:

    go/pkg/mod/k8s.io/client-go@v11.0.0+incompatible/rest/request.go:598:31: not enough arguments in call to watch.NewStreamWatcher

Follow
https://github.com/kubernetes/client-go/blob/master/INSTALL.md#go-modules
and use the latest release of kubernetes instead. Fixed with:

    export GO111MODULE=on
    go get k8s.io/client-go@kubernetes-1.16.3
    go mod tidy

There are also failing tests that are not fixed by this PR.

PR #766, opened 2019-12-16
fix build

sdudoladov:
At present e2e tests fail with 

```
Error response from daemon: manifest for registry.opensource.zalan.do/acid/postgres-operator-e2e-tests:latest not 
found: manifest unknown: manifest unknown 
```
the cause of this are modifications to go.mod /go.sum  during build that make `git status` add the  [`dirty` suffix to an image tag](https://github.com/zalando/postgres-operator/blob/master/e2e/run.sh#L16)

changes are the result of [installing kind via go modules](https://github.com/zalando/postgres-operator/blob/master/e2e/Makefile#L45) ([PR](https://github.com/zalando/postgres-operator/pull/742)); They were reverted in a[ later PR](https://github.com/zalando/postgres-operator/pull/754/files) despite the PR's author correctly raising a warning that certain tests exhibit failures.

UPD looks like some form of dep conflict with `kind` ; the proposed changes from this PR actually break unit tests


sdudoladov:
@zimbatm  already [identified the issue](https://github.com/zalando/postgres-operator/pull/754#issuecomment-564128321)

for now we may switch back  to installing kind as a binary to prevent `kind` installation from overwriting operator dependencies

zimbatm:
One trick I used in other projects is to create a `tools` package that contains a file like this:
```go
// +build tools
package tools

// put all the tools here
import (
  _ "sigs.k8s.io/kind"
)
```
and a bunch of wrapper scripts like this:
```sh
#!/bin/sh
exec go run sigs.8s.io/kind "$@"
```

Thanks to the `+build tools` build pragma, the package never gets built, but each dependency still gets listed in `go.sum`.

In this case it wouldn't solve the dependency conflict but it has the benefit of making it explicit and avoid unnecessary churn in the `go.sum` file.

PR #767, opened 2019-12-16
add missing fields to OperatorConfiguration CRD validation

FxKu:
Spotted some last missing fields which were yet not covered by CRD validation. Plus, some polishing on Postgres manifests and removing dead code from `operator_configuration_type.go`.


sdudoladov:
:+1:

FxKu:
:+1:

PR #782, opened 2020-01-07
use cluster-name as default label everywhere

FxKu:
and replace code were the label "version" was hard-coded


e2e/tests/test_e2e.py
```diff
@@ -292,7 +292,7 @@ def count_pods_with_label(self, labels, namespace='default'):
     def wait_for_master_failover(self, expected_master_nodes, namespace='default'):
         pod_phase = 'Failing over'
         new_master_node = ''
-        labels = 'spilo-role=master,version=acid-minimal-cluster'
+        labels = 'spilo-role=master,cluster-name=acid-minimal-cluster'
```
Jan-M: I dont like that this is hard coded. We should have the end to end test covering that the label can change to make sure config variable is picked up.

FxKu: Yeah, but there are even more selectors hard coded here: spilo-role, name etc.
But true, I could add a test on changing such labels

Jan-M:
:+1:

FxKu:
:+1:

RafiaSabih:
:+1:

Bryji:
This was a breaking change for us. The operator was no longer able to update statefulsets as Kubernetes does not allow you to change the label on an existing sts:
`time="2020-02-28T18:50:54Z" level=warning msg="error while syncing cluster state: could not sync statefulsets: could not update statefulset: could not patch statefulset spec \"postgres-operator/platops-postgres-operator-platypusdb\": StatefulSet.apps \"platops-postgres-operator-platypusdb\" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden" cluster-name=postgres-operator/platops-postgres-operator-platypusdb pkg=cluster
`
Fix was to set the cluster_name_label back to "version" in our base values file
```
  configKubernetes:
    cluster_name_label: "version"
```

Posting the comment here for any others looking for the root cause/fix.

PR #786, opened 2020-01-09
create namespaced role instead of using cluster role

FxKu:
In each namespace operator currently creates an additional ServiceAccount and binds it to the `postgres-pod` ClusterRole. This PR adds code to create a Role instead and let the RoleBinding reference it.

Next to `pod_service_account_definition` and `pod_service_account_role_binding_definition`, a third configuration parameter is added to allow users to configure the role definition as well.

Yet, this PR has now sync strategy for roles. What if you want to change the role's privileges? With one ClusterRole it's easy. Just edit it. But how to sync all roles the operator would create?


docs/administrator.md
```diff
@@ -127,14 +109,20 @@ kubectl create -f manifests/postgres-operator.yaml
...
+resources can be [configured](reference/operator_parameters.md#kubernetes-resources).
+Note, that the operator performs **no** further syncing of them.
 
-Note that the service account is named `zalando-postgres-operator`. You may have
-to change the `service_account_name` in the operator ConfigMap and
-`serviceAccountName` in the `postgres-operator` deployment appropriately. This
-is done intentionally to avoid breaking those setups that already work with the
-default `operator` account. In the future the operator should ideally be run
-under the `zalando-postgres-operator` service account.
+Previous releases created RoleBindings to the operator ClusterRole. This can
```
erthalion: It sounds a bit vague, since we're saying "previous releases created RoleBinding to the operator ClusterRole", but do not mention current state of things, just "This service account is bound to a namespaced Role via RoleBinding". Which Role and what is the difference is not clear. Do I read it right, or I'm missing something?

pkg/controller/controller.go
```diff
@@ -190,6 +192,86 @@ func (c *Controller) initPodServiceAccount() {
 	// actual service accounts are deployed at the time of Postgres/Spilo cluster creation
 }
 
+func (c *Controller) initRole() {
+
+	// service account on its own lacks any rights starting with k8s v1.8
+	// operator binds it to the namespaced role with sufficient privileges
+	if c.opConfig.PodServiceAccountRoleDefinition == "" {
+		c.opConfig.PodServiceAccountRoleDefinition = fmt.Sprintf(`
+		{
```
erthalion: Is it possible to not store this in the code file, maybe just as some sort of a resource?

Jan-M: We should make this a seperate issue, to see where we have ugly text in the Go code to be replaced by code or templates.

FxKu: I've checked, and aside from tests this is the only place where it's used. I agree, we should do it in either go code or files, but I thought the code addition would be easier to review if it comes in the same style as the `ServiceAccountDefinition` and `ServiceAccountRoleBindingDefinition`.

FxKu:
Please tell, if the default name `postgres-operator-patroni` for ServiceAccount, RoleBinding and Role is fine for you. I could also think of `postgres-operator-db-pods`, `postgres-operator-spilo` ...

docs/administrator.md
FxKu: I've updated this paragraph saying that until now the clusterrole is referenced by default. Who knows, maybe some are relying on this.

machine424:
We're really interested in this feature.

FxKu:
@machine424 thanks for your interest. I've rebased the branch now, so that you can continue working on it. The big open TODO here is the syncing part. What if you want to change the role? How are the existing roles updated then? That's why we went with another clusterrole instead, because it's one edit there and you are done.

machine424:
Ok, I'll give it a try, thanks.

caniko:
I really need this, it won't happen any time soon, right?

rgarcia89:
I would be interested in a namespaced deployment way too. I do have some use-cases where I do not have permission to create clusterroles due to policy. Roles and RoleBindings however would work to at least deploy the clusters into the same cluster or the namespaces.

PR #794, opened 2020-01-15
add service for exposing operator api

siku4:
I added a k8s service (Helm template) for exposing the operator api. I needed it for running the UI as a separate pod.

Issue #722, opened 2019-11-18
Add UI to the helm chart


Flowkap:
Hey guys. First great work!! 

I'd be very happy if the UI could be added to the setup via helm. We're maintaining a big system and have a good tooling around managing helm installs. Using the manifests is possible but it'll be great to have an UI option on the helm chart to install everything consistently. 

BR Flowkap

FxKu:
I think, I would prefer a separate chart to allow users to go with or without the UI. Feel free to create a PR.

PR #795, opened 2020-01-15
add first version of ui helm chart

siku4:
for an own project I created a separate UI helm chart. Related to #722 it could be also helpful for others

#794 provides operator API service, that is needed by the UI

### DIFF
--- a/charts/postgres-operator-ui/Chart.yaml
+++ b/charts/postgres-operator-ui/Chart.yaml
@@ -1,0 +1,19 @@
+apiVersion: v1
+name: postgres-operator-ui
+version: 0.1.0
+appVersion: 1.2.0
+home: https://github.com/zalando/postgres-operator
+description: Postgres Operator UI provides a graphical interface for a convenient database-as-a-service user experience
+keywords:
+- postgres
+- operator
+- ui
+- cloud-native
+- patroni
+- spilo
+maintainers:
+- name: Zalando
+  email: opensource@zalando.de
+sources:
+- https://github.com/zalando/postgres-operator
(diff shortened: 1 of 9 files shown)


FxKu:
Thanks for this contribution! Guess this will make more people use our UI :smiley: 
I will do some tests to give it proper review. Might take some time, but on a first look it already looks quite complete.

charts/postgres-operator-ui/Chart.yaml
```diff
@@ -0,0 +1,19 @@
...
+keywords:
+- postgres
+- operator
+- ui
+- cloud-native
+- patroni
+- spilo
+maintainers:
+- name: Zalando
+  email: opensource@zalando.de
```
FxKu: You can also add your GitHub user handle here to get some credit

charts/postgres-operator-ui/templates/deployment.yaml
```diff
@@ -0,0 +1,67 @@
...
+                  "databases_visible": true,
+                  "master_load_balancer_visible": true,
+                  "nat_gateways_visible": false,
+                  "replica_load_balancer_visible": true,
+                  "resources_visible": true,
+                  "users_visible": true,
+                  "postgresql_versions": [
+                    "11",
+                    "10",
+                    "9.6"
```
FxKu: should add "12" and at least "9.5" as it's not EOL yet. Spilo also supports 9.4 and 9.3, but this would be too much (and also not recommended)

FxKu:
:+1:

Jan-M:
:+1:

PR #797, opened 2020-01-16
standby clusters can only have 1 pod for now

FxKu:
If `min_instances` is configured to be > 1 (in case you always want a replica), standby cluster end up with two pods, which is currently not properly supported. Workaround would be to copy secrets from the source cluster.


Jan-M:
:+1:

FxKu:
:+1:

PR #798, opened 2020-01-17
add support for custom TLS certificates

zimbatm:
re-opened #690 from my account to allow pushes from maintainers.


erthalion:
Well, tests are already better, since they do not fail with some obscure error, but rather with timeout. I'll check why.

FxKu:
@zimbatm since we enabled CRD validation all new manifest fields must be covered by it. That's why the test is failing - validation doesn't know `tls`.

Have a look at the [CRD](https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql.crd.yaml#L18) and add the field there (alphabetical order).
Reflect it also in [go](https://github.com/zalando/postgres-operator/blob/master/pkg/apis/acid.zalan.do/v1/crds.go) code.

zimbatm:
thanks @FxKu. I just pushed updated to both schema files. Let's see what CI is saying now.

zimbatm:
it looks like coveralls is unhappy. any idea how I could add test coverage for `pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go` ?

cazter:
stoked on this PR as it will be really helpful for what we have going.

but testing with the current spilo, and spilo (or maybe patroni), raises issue with the permissions:

> 2020-02-05T22:04:32.602243923Z 2020-02-05 22:04:32 UTC [31471]: [1-1] 5e3b3bf0.7aef 0     FATAL:  data directory "/home/postgres/pgdata/pgroot/data" has invalid permissions
> 2020-02-05T22:04:32.602296208Z 2020-02-05 22:04:32 UTC [31471]: [2-1] 5e3b3bf0.7aef 0     DETAIL:  Permissions should be u=rwx (0700) or u=rwx,g=rx (0750).

cazter:
Disregard my comment. Issue is related to the operator, not this PR. Opened issue https://github.com/zalando/postgres-operator/issues/821

abh:
This will fix #633 (mentioning it here just so it gets linked to that issue)

zimbatm:
Rebased to fix the merge conflicts.

@FxKu is it possible to schedule 30min next week to get this merged? You can reserve a slot at https://calendly.com/zimbatm/

pkg/cluster/k8sres.go
```diff
@@ -840,21 +847,71 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
+	// determine the FSGroup for the spilo pod
+	effectiveFSGroup := c.OpConfig.Resources.SpiloFSGroup
+	if spec.SpiloFSGroup != nil {
+		effectiveFSGroup = spec.SpiloFSGroup
+	}
+
 	volumeMounts := generateVolumeMounts(spec.Volume)
 
+	// configure TLS with a custom secret volume
+	if spec.TLS != nil && spec.TLS.SecretName != "" {
```
FxKu: I'm confused what would happen here, if one uses your manifest [example](https://github.com/zalando/postgres-operator/blob/d152ea9a4a57e853c7ac20f55eb27189fb547b09/manifests/complete-postgres-manifest.yaml#L106) where the secret name is empty. Because, I've followed your documentation and created a secret, but then only used the complete manifest to see that nothing happens to the stateful set

zimbatm: I think I assumed that file was used to display the defaults. In my mind it wasn't going to be used as an example to setup TLS. Should I add a comment next to the empty value to say that it's disabled?

FxKu: It's [this](https://github.com/zalando/postgres-operator/blob/d152ea9a4a57e853c7ac20f55eb27189fb547b09/manifests/complete-postgres-manifest.yaml#L104) comment, that confuses me. To me it sounds like: "If you leave it blank, the TLS secret will be generated and used by the pods"

Similar to e.g. service accounts in our helm [chart](https://github.com/zalando/postgres-operator/blob/51909204fd7085f5b890cc4689221fece54ccecc/charts/postgres-operator/values.yaml#L261).

So yeah adding a comment behind the field in the manifest sounds good + in the [reference docs](https://github.com/zalando/postgres-operator/blob/master/docs/reference/cluster_manifest.md). Btw, just noticed that the reference does not cover the TLS section at all. Please add a short explanation for all fields.

pkg/cluster/k8sres_test.go
```diff
@@ -451,3 +452,64 @@ func TestSecretVolume(t *testing.T) {
 		}
 	}
 }
+
+func TestTLSConfiguration(t *testing.T) {
```
FxKu: hm, can you rename it to `TestTLS`. Configuration sounds like you're testing parameters from the config map / CRD config

docs/user.md
```diff
@@ -454,3 +454,41 @@ monitoring is outside the scope of operator responsibilities. See
 [configuration reference](reference/cluster_manifest.md) and
 [administrator documentation](administrator.md) for details on how backups are
 executed.
+
+## TLS configuration
```
FxKu:
```suggestion
## Custom TLS certificates
```

Configuration sounds like, it can be globally configured in the ConfigMap / OperatorConfiguration which isn't possible at the moment.

pkg/cluster/k8sres.go
zimbatm: Ok I see :+1: . Fix incoming.

zimbatm: Ok I added some extra documentation, let me know what you think.

docs/reference/cluster_manifest.md
```diff
@@ -359,3 +363,28 @@ CPU and memory limits for the sidecar container.
...
+
+By setting the `secretName` value in this section, the cluster will switch to
+load the given Kubernetes Secret into the container as a volume and uses that
+as the certificate instead. It is up to the user to create and manage the
+Kubernetes Secret either by hand or using a tool like the CertManager
+operator.
+
+When `secretName` is set, the `spiloFSGroup` value will also be set to `103`
+by default unless defined otherwise by the user. This is necessary so that the
+postgres instance can read from the secret volume.
```
FxKu: This redundant to the user docs and can be removed I think.

docs/reference/cluster_manifest.md
```diff
@@ -359,3 +363,28 @@ CPU and memory limits for the sidecar container.
...
+
+By default, the spilo image generates its own TLS certificate during startup.
+This certificate is not secure since it cannot be verified and thus doesn't
+protect from active MITM attacks.
+
+By setting the `secretName` value in this section, the cluster will switch to
+load the given Kubernetes Secret into the container as a volume and uses that
+as the certificate instead. It is up to the user to create and manage the
+Kubernetes Secret either by hand or using a tool like the CertManager
+operator.
```
FxKu: Can you move this section under to the **secretName** explanation?

docs/reference/cluster_manifest.md
```diff
@@ -359,3 +363,28 @@ CPU and memory limits for the sidecar container.
 * **memory**
   memory limits for the sidecar container. Optional, overrides the
   `default_memory_limits` operator configuration parameter. Optional.
+
+## Custom TLS certificates
+
+By default, the spilo image generates its own TLS certificate during startup.
+This certificate is not secure since it cannot be verified and thus doesn't
+protect from active MITM attacks.
```
FxKu: I would add: "A custom certificate can be specified under the `tls` top-level key."

Then it's aligned to the other groups. I would also remove the first mention of the `tls` field under top-level parameters. We also don't do it for other groups such as `patroni` etc.

pkg/cluster/k8sres.go
FxKu: Thank you :)
Just some minor comments and I think it's done then

manifests/complete-postgres-manifest.yaml
```diff
@@ -100,3 +100,10 @@ spec:
 #      env:
 #        - name: "USEFUL_VAR"
 #          value: "perhaps-true"
+
+# Custom TLS configuration. Disabled by default unless tls.secretName has a
+# value.
```
FxKu:
```suggestion
# Custom TLS certificate. Disabled unless tls.secretName has a value.
```

manifests/complete-postgres-manifest.yaml
```diff
@@ -100,3 +100,10 @@ spec:
...
+  tls:
+    secretName: "" # should correspond to a Kubernetes Secret resource to load
```
FxKu:
```suggestion
    secretName: ""  # should correspond to a Kubernetes Secret resource to load
```

FxKu: I think some yaml linter want two spaces before an inline comment

docs/reference/cluster_manifest.md
zimbatm: Up to the title?

FxKu: No just the section I've marked (lines 379-381)

zimbatm:
@FxKu ok, let me know if that's better!

docs/reference/cluster_manifest.md
```diff
@@ -359,3 +359,24 @@ CPU and memory limits for the sidecar container.
 * **memory**
   memory limits for the sidecar container. Optional, overrides the
   `default_memory_limits` operator configuration parameter. Optional.
+
+## Custom TLS certificates
+
+Those parameters are grouped under the `tls` top-level key.
+
+By default, the spilo image generates its own TLS certificate during startup.
```
erthalion: I believe this part doesn't belong to the operator documentation, rather to the spilo docs. I see the point of having some rationale why one would want to use this, but since it's just reference section, maybe it's better just to get rid of it. What do you think?

pkg/cluster/k8sres.go
```diff
@@ -840,21 +847,71 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
+		volumeMounts = append(volumeMounts, v1.VolumeMount{
+			MountPath: mountPath,
+			Name:      "tls-secret",
+			ReadOnly:  true,
+		})
+
+		// use the same filenames as cert-manager by default
+		certFile := ensurePath(spec.TLS.CertificateFile, mountPath, "tls.crt")
+		privateKeyFile := ensurePath(spec.TLS.PrivateKeyFile, mountPath, "tls.key")
+		caFile := ensurePath(spec.TLS.CAFile, mountPath, "ca.crt")
```
FxKu: Just found another issue. In one of your documented examples `caFile` is left out. But still it is appended as `env` variable and Postgres tries to look it up, but:
```
FATAL:  could not load root certificate file "/tls/ca.crt": No such file or directory
```

docs/reference/cluster_manifest.md
FxKu: You mean, starting from "By default ...", I guess

pkg/cluster/k8sres.go
FxKu: Maybe the unit test could also be extended then to check that the env variable is not created when the CA is left out

zimbatm: When I did my testing, Postgres complains but keeps going. Is that not the case for you?

I will add documentation for the cafile.

zimbatm: make the CAFile optional and documented it a bit in 662e7e68e90ed199b6fd6df1ac0388a1bb40407c

zimbatm: I wanted both secret types to work out of the box but I understand if the log line is undesirable.

docs/reference/cluster_manifest.md
```diff
@@ -359,3 +359,28 @@ CPU and memory limits for the sidecar container.
...
+
+* **certificateFile**
+  Filename of the certificate. Defaults to "tls.crt".
+
+* **privateKeyFile**
+  Filename of the private key. Defaults to "tls.key".
+
+* **caFile**
+  Optional filename to the CA certificate. Useful when the client connects
+  with `sslmode=verify-ca` or `sslmode=verify-full`.
```
FxKu: `Defaults to "ca.crt".` missing (?)

docs/user.md
```diff
@@ -454,3 +454,42 @@ monitoring is outside the scope of operator responsibilities. See
...
+```yaml
+apiVersion: "acid.zalan.do/v1"
+kind: postgresql
+
+metadata:
+  name: acid-test-cluster
+spec:
+  tls:
+    secretName: "pg-tls"
+    caFile: "ca.crt" # add this if the secret is configured with a CA
```
FxKu: hm, can't it be read from the secret if it's in there?

pkg/cluster/k8sres.go
```diff
@@ -840,21 +847,76 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
+		certFile := ensurePath(spec.TLS.CertificateFile, mountPath, "tls.crt")
+		privateKeyFile := ensurePath(spec.TLS.PrivateKeyFile, mountPath, "tls.key")
+		spiloEnvVars = append(
+			spiloEnvVars,
+			v1.EnvVar{Name: "SSL_CERTIFICATE_FILE", Value: certFile},
+			v1.EnvVar{Name: "SSL_PRIVATE_KEY_FILE", Value: privateKeyFile},
+		)
+
+		if spec.TLS.CAFile != "" {
+			caFile := ensurePath(spec.TLS.CAFile, mountPath, "")
```
```suggestion
			caFile := ensurePath(spec.TLS.CAFile, mountPath, "ca.crt")
```

docs/reference/cluster_manifest.md
zimbatm: Not anymore, I had to remove it above.

docs/user.md
zimbatm: The secret resource is not inspectable during schema generation. The secret resource might not exist, or might be also changed at later points in time. This would require to add another hook in the operator to listen for changes on that secret and update the schema accordingly. I don't know if it's worth the extra complexity.

Another solution would be to change the spilo image to unset the environment variable if the file doesn't exist. This would hide errors if the file really is supposed to exist and is not there.

That's why initially I just passed the env to the spilo image and let postgres complain. It's unfortunate that the log line contains "FATAL" because it didn't seem to affect the running instance the last time I tested it.

FxKu: I see. Thanks for the explanation.

FxKu:
:+1:

docs/reference/cluster_manifest.md
zimbatm: I merged this and improved the user documentation in 0ef214aa316eb23cf058821d916c45d64fa35047

zimbatm:
@FxKu last commit to improve the user documentation

docs/user.md
```diff
@@ -457,8 +457,15 @@ executed.
 
 ## Custom TLS certificates
 
-By default self-signed certificates are generated when the postgres container
-starts up. But it's also possible to provide your own secrets like so:
+By default, the spilo image generates its own TLS certificate during startup.
+This certificate is not secure since it cannot be verified and thus doesn't
+protect from active MITM attacks. In this section we show how a Kubernete
+Secret resources can be loaded with a custom TLS certificate.
```
FxKu:
```suggestion
However, this certificate cannot be verified and thus doesn't protect from
active MITM attacks. In this section we show how to specify a custom TLS
certificate which is be mounted in the database pods via a K8s Secret.
```

docs/user.md
```diff
@@ -489,7 +499,10 @@ spec:
...
-Before applying these changes, the operator must also be configured with the
-`spilo_fsgroup` set to the GID matching the postgres user group. If the value
-is not provided, the cluster will default to `103` which is the GID from the
-default spilo image.
+Certificate rotation is handled in the spilo image which checks every 5
+minutes if the certificates have changed and reloads postgres accordingly.
+
+Known issues:
+
+* Existing clusters don't handle changes to the `spilo_fsgroup` well.
```
FxKu: That's not very precise. Are changes ignored? Does it brake the cluster?

zimbatm:
I think this is ready

zimbatm:
rebased and squashed

FxKu:
Thanks @zimbatm. Can you address my last comment? We are already telling in the reference docs that changing the `spilo_fsgroup` requires a Pod restart.

zimbatm:
fair enough, I will just remove that section then.

docs/user.md
zimbatm: It inherits the same issue that changing `spilo_fsgroup` has where the pod needs to be re-created.

I will remove that section for now.

zimbatm: Once the user reads-up on `spilo_fsgroup` they will probably find the issue.

zimbatm:
it looks like the e2e test failures are also there in master

FxKu:
:+1:

FxKu:
yes that's one part in the e2e test which probably needs more waiting time, since it's passing at times and sometimes not :-/

erthalion:
:+1:

FxKu:
Thanks @zimbatm for adding this cool feature and your patience in the review process.

zimbatm:
yesssssss!! :tada: 

thanks to you too for helping me out on this PR!

ReSearchITEng:
@FxKu @zimbatm 
I tried to use this functionality with latest image opr image: v1.4.0-15-gba9cf68-dirty, but it does not seem to work.
In the request I have added this section:
```
spec:
   tls:
      secretName: "pg-tls"
```
The statefulSet does not get the mount point for the secret. Nothing seems to happen, and no errors in the opr logs.

Am I missing something?

zimbatm:
It's hard to say without more details. I haven't used openshift before so I don't know what changes from a vanilla k8s installation. The first thing I would do is look at the event log to see if anything stands out. If it's an existing pg cluster, try creating a fresh one as pods don't really like the fsGroup to be changed.

ReSearchITEng:
it seems to be the fsGroup that is defaulted to "103":
```
  Warning  FailedCreate      23s (x21 over 3m)  statefulset-controller  create Pod acid-tls-1 in StatefulSet acid-tls-1 failed error: pods "acid-tls-1-0" is forbidden: unable to validate against any security context constraint: [spec.volumes[0]: Invalid value: "persistentVolumeClaim": persistentVolumeClaim volumes are not allowed to be used fsGroup: Invalid value: []int64{103}: 103 is not an allowed group]```

PR #799, opened 2020-01-22
[WIP] Connection pooler

erthalion:
The idea is to support connection pooler creation via operator. There is no strict dependencies, one can specify which type of pooler is required, but at the beginning probably only pgbouncer would be implemented.

Suggested changes to the manifest assume a new section:

```yaml
connectionPool:
  type: pgbouncer
  instancesNumber: 1
  schema: pgbouncer
  user: pgbouncer
  mode: session
  resources:
  ...
```

Schema & user describe into which schema a lookup function will be installed for all available databases, and which user pooler will use for connection. A set of reasonable default values must be provided with the possibility to simply specify a full pod template for deployment.


pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -150,6 +150,20 @@ type ScalyrConfiguration struct {
 	ScalyrMemoryLimit   string `json:"scalyr_memory_limit,omitempty"`
 }
 
+// Defines default configuration for connection pool
+type ConnectionPoolConfiguration struct {
+	NumberOfInstances    *int32 `json:"connection_pool_instances_number,omitempty"`
```
Jan-M: Given we use `instancesNumber` we should use the same in the "pool" section": `numberOfInstances` although this is so long and I dont like the word instance anymore. Maybe `numberOfPods`

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -153,3 +155,15 @@ type UserFlags []string
...
 }
+
+// Options for connection pooler
+type ConnectionPool struct {
+	NumberOfInstances *int32              `json:"instancesNumber,omitempty"`
+	Schema            *string             `json:"schema,omitempty"`
+	User              *string             `json:"user,omitempty"`
+	Type              *string             `json:"type,omitempty"`
+	Mode              *string             `json:"mode,omitempty"`
+	PodTemplate       *v1.PodTemplateSpec `json:"podTemplate,omitempty"`
```
Jan-M: Especially pod template in the user manifest opens a gate to misuse.

Jan-M:
I would slim down some Postgres manifest options in favor of a slim user configuration.

If one is not happy with the number of fields we offer, it is easy to deploy pgbouncer similar to our config manually.

I also think the type can probably be hidden behind the supplied Docker image given we have good env vars specced.

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -185,6 +199,7 @@ type OperatorConfigurationData struct {
 	LoggingRESTAPI             LoggingRESTAPIConfiguration        `json:"logging_rest_api"`
 	Scalyr                     ScalyrConfiguration                `json:"scalyr"`
 	LogicalBackup              OperatorLogicalBackupConfiguration `json:"logical_backup"`
+	ConnectionPool             ConnectionPoolConfiguration        `json:"connection_pool"`
```
Jan-M: What is the idea here?

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -153,3 +155,15 @@ type UserFlags []string
 type PostgresStatus struct {
 	PostgresClusterStatus string `json:"PostgresClusterStatus"`
 }
+
+// Options for connection pooler
+type ConnectionPool struct {
+	NumberOfInstances *int32              `json:"instancesNumber,omitempty"`
```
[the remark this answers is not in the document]
erthalion: That's why most of them are optional.
I wanted originally call it "pooling", but it's too broad and not clear, what kind of "pool" are we talking about. `NumberOfPods` is good.

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
erthalion: What's an idea behind operator default values for pool configs, or what?

pkg/cluster/k8sres.go
```diff
@@ -1651,3 +1712,202 @@ func (c *Cluster) generateLogicalBackupPodEnvVars() []v1.EnvVar {
...
+
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        c.connPoolName(),
+			Namespace:   c.Namespace,
+			Labels:      c.labelsSet(true),
+			Annotations: map[string]string{},
+			// make StatefulSet object its owner to represent the dependency.
+			// By itself StatefulSet is being deleted with "Ophaned"
+			// propagation policy, which means that it's deletion will not
```
RafiaSabih: did you mean "orphaned"?

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -152,6 +152,19 @@ type ScalyrConfiguration struct {
...
+type ConnectionPoolConfiguration struct {
+	NumberOfInstances    *int32 `json:"connection_pool_number_of_instances,omitempty"`
+	Schema               string `json:"connection_pool_schema,omitempty"`
+	User                 string `json:"connection_pool_user,omitempty"`
+	Image                string `json:"connection_pool_image,omitempty"`
+	Mode                 string `json:"connection_pool_mode,omitempty"`
+	DefaultCPURequest    string `name:"connection_pool_default_cpu_request,omitempty"`
+	DefaultMemoryRequest string `name:"connection_pool_default_memory_request,omitempty"`
+	DefaultCPULimit      string `name:"connection_pool_default_cpu_limit,omitempty"`
+	DefaultMemoryLimit   string `name:"connection_pool_default_memory_limit,omitempty"`
```
FxKu: `json:"...

docs/reference/cluster_manifest.md
```diff
@@ -359,3 +364,25 @@ CPU and memory limits for the sidecar container.
...
   `default_memory_limits` operator configuration parameter. Optional.
+
+## Connection pool
+
+Parameters are grouped under the `connectionPool` top-level key and specify
+configuration for connection pool. If this section is not empty, a connection
+pool will be created for a database even if `enableConnectionPool` is not
+present.
+
+* **number_of_instances**
```
FxKu: we have this convention to make manifest fields CamelCase and the config options snake_case. Any reason `dockerImage` isn't listed here?

manifests/postgresql.crd.yaml
```diff
@@ -162,6 +208,22 @@ spec:
                     type: string
             replicaLoadBalancer:  # deprecated
               type: boolean
+            connectionPool:
```
FxKu: This part is duplicate, except for the enum part under `mode`. Can you copy the latter to the position above?
I would like to have alphabetical sorting in the validation because it's so huge already and I think things are easier to find later.

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -176,6 +176,65 @@ var PostgresCRDResourceValidation = apiextv1beta1.CustomResourceValidation{
...
 						},
 					},
+					"connectionPool": {
+						Type: "object",
+						Properties: map[string]apiextv1beta1.JSONSchemaProps{
+							"dockerImage": {
+								Type: "string",
+							},
+							"mode": {
+								Type: "string",
```
FxKu:
```suggestion
								Type: "string",
								Enum: []apiextv1beta1.JSON{
									{
										Raw: []byte(`"session"`),
									},
									{
										Raw: []byte(`"transaction"`),
									},
								},
```

pkg/apis/acid.zalan.do/v1/register.go
```diff
@@ -10,7 +10,8 @@ import (
 
 // APIVersion of the `postgresql` and `operator` CRDs
 const (
-	APIVersion = "v1"
+	APIVersion     = "v1"
+	PostgresqlKind = "postgresql"
```
FxKu: don't get this change if it's only used in one place. I prefer using a const though but when we should do the same with `postgresqlList` and `OperatorConfiguration`.

pkg/cluster/cluster.go
```diff
@@ -1137,3 +1200,29 @@ func (c *Cluster) deletePatroniClusterConfigMaps() error {
...
 }
+
+// Test if two connection pool configuration needs to be synced. For simplicity
+// compare not the actual K8S objects, but the configuration itself and request
+// sync if there is any difference.
+func (c *Cluster) needSyncConnPoolDeployments(oldSpec, newSpec *acidv1.ConnectionPool) (sync bool, reasons []string) {
+	reasons = []string{}
+	sync = false
+
+	changelog, err := diff.Diff(oldSpec, newSpec)
```
FxKu: would the diff function from our [util](https://github.com/zalando/postgres-operator/blob/master/pkg/util/util.go#L69) pkg also work, instead of introducing another dependency?

pkg/cluster/cluster.go
```diff
@@ -48,11 +49,17 @@ type Config struct {
 	PodServiceAccountRoleBinding *rbacv1beta1.RoleBinding
 }
 
+type ConnectionPoolResources struct {
```
FxKu: maybe `ConnectionPoolChildResources` to not mix it up with cpu/memory resources?

pkg/util/constants/pooler.go
```diff
@@ -0,0 +1,13 @@
...
+
+// Connection pool specific constants
+const (
+	ConnectionPoolUserName             = "pooler"
+	ConnectionPoolSchemaName           = "pooler"
+	ConnectionPoolDefaultType          = "pgbouncer"
+	ConnectionPoolDefaultMode          = "transition"
+	ConnectionPoolDefaultCpuRequest    = "100m"
+	ConnectionPoolDefaultCpuLimit      = "100m"
+	ConnectionPoolDefaultMemoryRequest = "100M"
```
FxKu: `Mi`? or is there no difference?

pkg/cluster/util.go
```diff
@@ -483,3 +495,7 @@ func (c *Cluster) GetSpec() (*acidv1.Postgresql, error) {
 func (c *Cluster) patroniUsesKubernetes() bool {
 	return c.OpConfig.EtcdHost == ""
 }
+
+func (c *Cluster) needConnectionPool() bool {
+	return c.Spec.ConnectionPool != nil || c.Spec.EnableConnectionPool == true
```
FxKu: So, when I define a `connectionPool` field, but set `enableConnectionPool` to `false` it will still get created?

erthalion:
> > @@ -483,3 +495,7 @@ func (c *Cluster) GetSpec() (*acidv1.Postgresql, error) {
>  func (c *Cluster) patroniUsesKubernetes() bool {
>  	return c.OpConfig.EtcdHost == ""
>  }
> +
> +func (c *Cluster) needConnectionPool() bool {
> +	return c.Spec.ConnectionPool != nil || c.Spec.EnableConnectionPool == true
>
> So, when I define a `connectionPool` field, but set `enableConnectionPool` to `false` it will still get created?

At the moment yes. The idea is that both ways are independent, but I see
what you mean - in the case if someone has both `enableConnectionPool`
and `connectionPool` and suddenly decide to turn it off (maybe
temporary), but keep configuration, it could be confusing. Maybe we can
make `enableConnectionPool` of `*bool`, and check `connectionPool` only
if `enableConnectionPool` is not specified, i.e. nil.

pkg/cluster/util.go
erthalion: At the moment yes. The idea is that both ways are independent, but I see
what you mean - in the case if someone has both `enableConnectionPool`
and `connectionPool` and suddenly decide to turn it off (maybe
temporary), but keep configuration, it could be confusing. Maybe we can
make `enableConnectionPool` of `*bool`, and check `connectionPool` only
if `enableConnectionPool` is not specified, i.e. nil.

pkg/cluster/cluster.go
erthalion: Hm, probably event `ConnectionPoolObjects` or something.

pkg/cluster/cluster.go
erthalion: Unfortunately no, already existing `diff` is just for visual comparison of objects. The introduced dependency will find a difference between two arbitrary objects and return a list of what, where and how changed.

pkg/apis/acid.zalan.do/v1/register.go
erthalion: This is a leftover from the implementation where I wanted to attach new objects to `postgresql` instead of statefulset. Indeed, we can probably just get rid of this.

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -154,3 +157,23 @@ type UserFlags []string
...
+// TODO: prepared snippets of configuration, one can choose via type, e.g.
+// pgbouncer-large (with higher resources) or odyssey-small (with smaller
+// resources)
+// Type              string `json:"type,omitempty"`
+//
+// TODO: figure out what other important parameters of the connection pool it
+// makes sense to expose. E.g. pool size (min/max boundaries), max client
+// connections etc.
+type ConnectionPool struct {
+	NumberOfInstances *int32 `json:"number_of_instances,omitempty"`
```
Jan-M: `numberOfInstances`

pkg/util/constants/pooler.go
```diff
@@ -0,0 +1,13 @@
+package constants
+
+// Connection pool specific constants
+const (
+	ConnectionPoolUserName             = "pooler"
+	ConnectionPoolSchemaName           = "pooler"
+	ConnectionPoolDefaultType          = "pgbouncer"
+	ConnectionPoolDefaultMode          = "transition"
```
Jan-M: `transaction`

pkg/util/constants/pooler.go
erthalion: Well, it's M for decimal SI and Mi for binary SI. But everywhere else we do Mi, so we can change it here.

pkg/cluster/resources.go
```diff
@@ -721,3 +849,34 @@ func (c *Cluster) GetStatefulSet() *appsv1.StatefulSet {
...
+		return nil, fmt.Errorf("there is no connection pool in the cluster")
+	}
+
+	patchData, err := specPatch(newDeployment.Spec)
+	if err != nil {
+		return nil, fmt.Errorf("could not form patch for the deployment: %v", err)
+	}
+
+	// An update probably requires RetryOnConflict, but since only one operator
+	// worker at one time will try to update it changes of conflicts are
```
sdudoladov: changes -> chances of conflicts ?

docs/user.md
```diff
@@ -454,3 +454,56 @@ monitoring is outside the scope of operator responsibilities. See
 [configuration reference](reference/cluster_manifest.md) and
 [administrator documentation](administrator.md) for details on how backups are
 executed.
+
+## Connection pool
+
+The operator can create a database side connection pool for those applications,
```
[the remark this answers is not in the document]
erthalion: It's database side as an opposite for application side.

docs/reference/cluster_manifest.md
```diff
@@ -359,3 +364,28 @@ CPU and memory limits for the sidecar container.
...
+Parameters are grouped under the `connectionPool` top-level key and specify
+configuration for connection pool. If this section is not empty, a connection
+pool will be created for a database even if `enableConnectionPool` is not
+present.
+
+* **numberOfInstances**
+  How many instances of connection pool to create.
+
+* **mode**
+  In which mode to run connection pool, transaction or section.
```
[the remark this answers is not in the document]
erthalion: Pool mode is not pgbouncer specific: https://github.com/yandex/odyssey/blob/master/documentation/configuration.md#pool-string

charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -318,6 +318,43 @@ spec:
...
+                  pattern: '^(\d+m|\d+(\.\d{1,3})?)$'
+                  #default: "1"
+                connection_pool_default_cpu_request:
+                  type: string
+                  pattern: '^(\d+m|\d+(\.\d{1,3})?)$'
+                  #default: "1"
+                connection_pool_default_memory_limit:
+                  type: string
+                  pattern: '^(\d+(e\d+)?|\d+(\.\d+)?(e\d+)?[EPTGMK]i?)$'
+                  #default: "100m"
```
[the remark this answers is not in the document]
erthalion: Sure, but since pgbouncer should not be that memory intensive as e.g. a database itself, it makes sense to be able to configure both request and limit, maybe even to different values (we need perform couple of experiments first).

charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -318,6 +318,43 @@ spec:
...
+                  #default: "pooler"
+                connection_pool_user:
+                  type: string
+                  #default: "pooler"
+                connection_pool_number_of_instances:
+                  type: integer
+                  #default: 1
+                connection_pool_image:
+                  type: string
+                  #default: "registry.opensource.zalan.do/acid/pgbouncer:1.0.0"
```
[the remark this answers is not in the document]
erthalion: There is no such thing as vanilla pgbouncer docker image, https://github.com/pgbouncer/pgbouncer/pull/324

pkg/cluster/k8sres.go
```diff
@@ -1669,3 +1731,290 @@ func (c *Cluster) generateLogicalBackupPodEnvVars() []v1.EnvVar {
...
+				ContainerPort: pgPort,
+				Protocol:      v1.ProtocolTCP,
+			},
+		},
+		Env: envVars,
+	}
+
+	podTemplate := &v1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:      c.connPoolLabelsSelector().MatchLabels,
```
FxKu: I think, it makes sense to include all [labels](https://github.com/zalando/postgres-operator/blob/51909204fd7085f5b890cc4689221fece54ccecc/pkg/cluster/util.go#L382) we use for other resources. Cluster name, application, inherited labels etc.

erthalion: Not really. I see that we need an application probably, but other than that unfortunately right when operator deletes pods e.g. for updating the cluster it deletes everything by label, which means it also deletes pooler pod.

pkg/util/config/config.go
```diff
@@ -83,13 +83,28 @@ type LogicalBackup struct {
...
 }
 
+// Operator options for connection pooler
+type ConnectionPool struct {
+	NumberOfInstances            *int32 `name:"connection_pool_number_of_instances" default:"1"`
+	Schema                       string `name:"connection_pool_schema" default:"pooler"`
+	User                         string `name:"connection_pool_user" default:"pooler"`
+	Image                        string `name:"connection_pool_image" default:"registry.opensource.zalan.do/acid/pgbouncer:master-5"`
+	Mode                         string `name:"connection_pool_mode" default:"transaction"`
+	MaxDBConnections             *int32 `name:"connection_pool_max_db_connections" default:"60"`
```
FxKu: What if someone configures max connection to be below 60? Should we really define a default here? Maybe this parameter should rather define a ratio to configured max_connections instead, e.g. between 0 and 1, where 1 is all connections minus reserved ones. And default would be 0.6 etc.

pkg/cluster/database.go
```diff
@@ -243,3 +264,67 @@ func makeUserFlags(rolsuper, rolinherit, rolcreaterole, rolcreatedb, rolcanlogin
...
+		msg := "could not get databases to install pool lookup function: %v"
+		return fmt.Errorf(msg, err)
+	}
+
+	templater := template.Must(template.New("sql").Parse(connectionPoolLookup))
+
+	for dbname, _ := range currentDatabases {
+		c.logger.Infof("Install pool lookup function into %s", dbname)
+
+		params := TemplateParams{
```
FxKu: why define an extra type when you can do `map[string]interface{}{"pool_schema" ...`?

pkg/util/config/config.go
erthalion: 60 is actually more like `last level defaults`, which would be almost always overridden in the operator/db cluster configuration. But I believe it's better to have an independent limit here, since max db connections depends mostly on how many resources are available, and max client connections (on the target database side, not on the pooler) depends mostly on estimations how many clients could be there.

pkg/cluster/k8sres.go
erthalion: I was a bit confused with the explanation. Actually it's not about updates, but about cluster deletes, when operator deletes all the pods, and only then goes to delete the pool deployment. So it's not a bit issue, but causes an extra unexpected round of deletes which I want to avoid.

pkg/cluster/resources_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func mockInstallLookupFunction(schema string, user string) error {
+	return nil
+}
+
+func boolToPointer(value bool) *bool {
```
sdudoladov: Why is this function here and not in `utils` ? It is actually reused by multiple test packages.

pkg/cluster/sync_test.go
```diff
@@ -0,0 +1,212 @@
...
+	acidv1 "github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1"
+	"github.com/zalando/postgres-operator/pkg/util/config"
+	"github.com/zalando/postgres-operator/pkg/util/k8sutil"
+
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func int32ToPointer(value int32) *int32 {
```
sdudoladov: Is it a  duplicate of [pkg/controller/operator_config.go:func int32ToPointer](https://github.com/zalando/postgres-operator/pull/799/files#diff-ab4a7db75ac2e59f513f47a9d540726dR25) ?

docs/reference/operator_parameters.md
```diff
@@ -595,3 +595,34 @@ scalyr sidecar. In the CRD-based configuration they are grouped under the
...
+* **connection_pool_max_db_connections**
+  How many connections the pooler can max hold.
+
+* **connection_pool_mode**
+  Default pool mode, session or transaction.
+
+* **connection_pool_default_cpu_request**
+  **connection_pool_default_memory_reques**
+  **connection_pool_default_cpu_limit**
+  **connection_pool_default_memory_limit**
```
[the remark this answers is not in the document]
erthalion: No, but I don't think they need to - probably it makes sense to make those always equal just by default, since connection pooler resource requirements are usually rather stable.

e2e/tests/test_e2e.py
```diff
@@ -494,12 +590,39 @@ def wait_for_operator_pod_start(self):
         # for local execution ~ 10 seconds suffices
         time.sleep(60)
 
+    def get_operator_pod(self):
+        pods = self.api.core_v1.list_namespaced_pod(
+            'default', label_selector='name=postgres-operator'
+        ).items
+
+        if pods:
+            return pods[0]
```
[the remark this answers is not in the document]
erthalion: Right, but this function `get_operator_pod` is used only in `get_operator_logs` when a timeout exception happened on other type of pods, i.e. when we already have a running operator pod. Maybe it makes sense to express this implicit dependency explicitly via a comment or so.

manifests/configmap.yaml
```diff
@@ -11,6 +11,16 @@ data:
   cluster_history_entries: "1000"
   cluster_labels: application:spilo
   cluster_name_label: cluster-name
+  # connection_pool_default_cpu_limit: "1"
+  # connection_pool_default_cpu_request: "1"
+  # connection_pool_default_memory_limit: 100m
+  # connection_pool_default_memory_request: "100Mi"
+  connection_pool_image: "registry.opensource.zalan.do/acid/pgbouncer:master-5"
```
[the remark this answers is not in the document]
erthalion: It's needed to make it work for e2e tests.

pkg/cluster/util.go
```diff
@@ -408,7 +408,32 @@ func (c *Cluster) labelsSet(shouldAddExtraLabels bool) labels.Set {
...
+// inherit most of the labels from the cluster itself, but at the same time
+// have e.g. different `application` label, so that recreatePod operation will
+// not interfere with it (it lists all the pods via labels, and if there would
+// be no difference, it will recreate also pooler pods).
+func (c *Cluster) connPoolLabelsSelector() *metav1.LabelSelector {
+	connPoolLabels := labels.Set(map[string]string{})
+
+	extraLabels := labels.Set(map[string]string{
+		"connection-pool": c.connPoolName(),
+		"application":     "connection-pool",
```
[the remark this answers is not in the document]
erthalion: `spilo-connection-pool` can make someone think that there is some special connection pooler that comes with spilo, rather than something already existing like pgbouncer. `db-connection-pool`?

pkg/cluster/cluster.go
```diff
@@ -1138,3 +1221,119 @@ func (c *Cluster) deletePatroniClusterConfigMaps() error {
...
+
+	if err != nil {
+		c.logger.Warningf("Cannot generate expected resources, %v", err)
+	}
+
+	for _, env := range poolContainer.Env {
+		if env.Name == "PGUSER" {
+			ref := env.ValueFrom.SecretKeyRef.LocalObjectReference
+
+			if ref.Name != c.credentialSecretName(config.User) {
```
FxKu: I think, one check is missing here. Only apply the change if the user in the spec is empty. Same for schema.

pkg/cluster/cluster.go
```diff
@@ -1138,3 +1221,119 @@ func (c *Cluster) deletePatroniClusterConfigMaps() error {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+// Check if we need to synchronize connection pool deployment due to new
+// defaults, that are different from what we see in the DeploymentSpec
+func (c *Cluster) needSyncConnPoolDefaults(
```
FxKu: `maxDBConnections` is missing here (?)

erthalion: It's mostly on purpose. Options mentioned there are reasonable to change and relatively easy to track. Not sure if there would be ever need to change `maxDBConnections` (e.g. pgbouncer in all my tests was always keeping it substantially lower than this limit).

pkg/cluster/database.go
```diff
@@ -28,13 +30,37 @@ const (
...
+		RETURNS record AS $$
+		BEGIN
+			SELECT usename, passwd FROM pg_catalog.pg_shadow
+			WHERE usename = i_username INTO uname, phash;
+			RETURN;
+		END;
+		$$ LANGUAGE plpgsql SECURITY DEFINER;
+
+		REVOKE ALL ON FUNCTION {{.pool_schema}}.user_lookup(text)
+			FROM public, {{.pool_schema}};
```
FxKu:
```suggestion
			FROM public, {{.pool_user}};
```
If you don't choose `pooler` as your user, it will not get created which leads to an `role does not exist` error here

pkg/cluster/sync.go
```diff
@@ -600,3 +618,163 @@ func (c *Cluster) syncLogicalBackupJob() error {
...
+		// Try to sync in any case. If we didn't needed connection pool before,
+		// it means we want to create it. If it was already present, still sync
+		// since it could happen that there is no difference in specs, and all
+		// the resources are remembered, but the deployment was manualy deleted
+		// in between
+		c.logger.Debug("syncing connection pool")
+
+		// in this case also do not forget to install lookup function as for
+		// creating cluster
+		if !oldNeedConnPool || !c.ConnectionPool.LookupFunction {
```
FxKu: Doesn't this need to be `&&`? On each sync I see the LookupFunction logs ("Installing..."). `oldNeedConnPool` should be `false` on sync but `ConnectionPool.LookupFunction` is set to `true` within the lookup function. Or is it intended?

pkg/cluster/k8sres.go
```diff
@@ -1772,6 +1834,306 @@ func (c *Cluster) getLogicalBackupJobName() (jobName string) {
...
+	numberOfInstances := spec.ConnectionPool.NumberOfInstances
+	if numberOfInstances == nil {
+		numberOfInstances = util.CoalesceInt32(
+			c.OpConfig.ConnectionPool.NumberOfInstances,
+			k8sutil.Int32ToPointer(1))
+	}
+
+	if *numberOfInstances < constants.ConnPoolMinInstances {
+		msg := "Adjusted number of connection pool instances from %d to %d"
+		c.logger.Warningf(msg, numberOfInstances, constants.ConnPoolMinInstances)
```
FxKu: something like coalesce(numberOfInstances, c.OpConfig.ConnectionPool.NumberOfInstances) would prevent logs to look like: `Adjusted number of connection pool instances from 824638671720 to 2`

pkg/cluster/k8sres.go
FxKu: typo is still there

pkg/cluster/database.go
```diff
@@ -243,3 +273,78 @@ func makeUserFlags(rolsuper, rolinherit, rolcreaterole, rolcreatedb, rolcanlogin
...
+func (c *Cluster) installLookupFunction(poolSchema, poolUser string) error {
+	var stmtBytes bytes.Buffer
+	c.logger.Info("Installing lookup function")
+
+	if err := c.initDbConn(); err != nil {
+		return fmt.Errorf("could not init database connection")
+	}
+	defer func() {
+		// in case if everything went fine this can generate a warning about
+		// trying to close an empty connection.
```
FxKu: seen this, too. Can't you simply remove the `closeDBConn` block at the end of this function?

pkg/cluster/sync.go
```diff
@@ -600,3 +618,163 @@ func (c *Cluster) syncLogicalBackupJob() error {
...
+			newConnPool := newSpec.Spec.ConnectionPool
+
+			specSchema := ""
+			specUser := ""
+
+			if newConnPool != nil {
+				specSchema = newConnPool.Schema
+				specUser = newConnPool.Schema
```
FxKu:
```suggestion
				specUser = newConnPool.User
```

pkg/cluster/cluster.go
```diff
@@ -1138,3 +1221,119 @@ func (c *Cluster) deletePatroniClusterConfigMaps() error {
...
+	config := c.OpConfig.ConnectionPool
+	podTemplate := deployment.Spec.Template
+	poolContainer := podTemplate.Spec.Containers[constants.ConnPoolContainer]
+
+	if spec == nil {
+		spec = &acidv1.ConnectionPool{}
+	}
+
+	if spec.NumberOfInstances == nil &&
+		*deployment.Spec.Replicas != *config.NumberOfInstances {
```
FxKu: this should also check if `deployment.Spec.Replicas` is smaller than `ConnPoolMinInstances`. Otherwise, it could be set to 1 here and then raised to 2 again.

pkg/util/constants/pooler.go
```diff
@@ -0,0 +1,18 @@
...
+	ConnectionPoolDefaultMode          = "transaction"
+	ConnectionPoolDefaultCpuRequest    = "1"
+	ConnectionPoolDefaultCpuLimit      = "1"
+	ConnectionPoolDefaultMemoryRequest = "100Mi"
+	ConnectionPoolDefaultMemoryLimit   = "100Mi"
+
+	ConnPoolContainer            = 0
+	ConnPoolMaxDBConnections     = 60
+	ConnPoolMaxClientConnections = 10000
+	ConnPoolMinInstances         = 2
```
FxKu: Number for min instances should be the same as the internal default for Connection Pool's `numberOfInstances`. So raise it to two everywhere?

pkg/cluster/k8sres.go
erthalion: Nope, it's just a pointer needs to be dereferenced.

pkg/cluster/cluster.go
erthalion: Nope, it needs to be done on the config level (e.g. when it's created).

pkg/cluster/k8sres.go
```diff
@@ -1772,6 +1834,306 @@ func (c *Cluster) getLogicalBackupJobName() (jobName string) {
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	deployment := &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        c.connPoolName(),
+			Namespace:   c.Namespace,
+			Labels:      c.labelsSet(true),
```
FxKu: should this be `connPoolLabelsSelector().MatchLabels` instead, like in the pod template of this deployment?

pkg/cluster/k8sres.go
```diff
@@ -1772,6 +1834,306 @@ func (c *Cluster) getLogicalBackupJobName() (jobName string) {
...
+		Selector: map[string]string{
+			"connection-pool": c.connPoolName(),
+		},
+	}
+
+	service := &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:        c.connPoolName(),
+			Namespace:   c.Namespace,
+			Labels:      c.labelsSet(true),
```
FxKu: Same with deployment. Use `connPoolLabelsSelector().MatchLabels` instead?

FxKu:
:+1:

erthalion:
:+1:

Issue #421, opened 2018-11-23
Detach a Postgres cluster from the operator temporarily


Jan-M:
We were facing a situation where we needed to edit the statefulset of one cluster manually. This does not work while the operator is running and syncing objects.

We solved this by scaling operator to 0, but this is not ideal.

Suggest we look into the manifest to add something like:

sync: false
detach: true
maintenance: true

Or similar.

Jan-M:
This change may need to be observed out of order, not sure how this could be done. But ideally it does not first apply all kind of pending changes on this cluster but stops sync immediately.

In first round though this could always be achieved with manual operator restart.

valer-cara:
:+1: Similar issue here. I have a 3-node cluster, and the first one is down because I got some EBS errors on an AWS M5 machine which rendered the volume read-only ([dmesg here](https://gist.github.com/valer-cara/e33a7920ee913f073361391f42b7d571)).

I wanted to unmount the volume and do a `fsck` on it, however to do that i'll need to stop the operator, delete the sts with `--cascade=false` to avoid pod re-creating, delete the affected pod & manually start debugging the volume. Finally restarting the operator would recreate the STS & the downed pod.

A maintenance feature per cluster would be most useful.

----

PS: for issues that don't require unmounting volumes one can simply use `patronictl -c ./postgres.yaml pause` to stop patroni temporarily and do whatever's needed. [more here](https://github.com/zalando/patroni/blob/master/docs/pause.rst)

valer-cara:
So maybe there are multiple levels of maintenance that can be expressed in the cluster manifest? 

```
maintenance:
  mode: paused
```

- `pause`: where patroni is paused on the cluster
- `detach`: where the operator doesn't monitor k8s resources anylonger
- `full`: all the above?

Not sure if this is the best approach or naming, just trying to underline the types of "detachment" out there.

sdudoladov:
one special case to watch out is the incomplete rolling update

1. The operator starts the rolling update of the cluster 
2. the first replica gets restarted but for whatever reason cannot join the cluster again
3. this blocks the rolling upgrade

FxKu:
With #802 merged this could be realized via `acid.zalan.do/conrtoller` annotations. So far, it's manual process. Closing for now.

PR #802, opened 2020-01-24
define ownership between operator and clusters via annotation

FxKu:
to run more than one operator next to each other without interfering

could also be used to [detach](https://github.com/zalando/postgres-operator/issues/421) a cluster from the operator


pkg/controller/postgresql.go
```diff
@@ -455,41 +470,48 @@ func (c *Controller) queueClusterEvent(informerOldSpec, informerNewSpec *acidv1.
...
 	}
 
-	// We will not get multiple Add events for the same cluster
-	c.queueClusterEvent(nil, pg, EventAdd)
+	return
 }
 
 func (c *Controller) postgresqlUpdate(prev, cur interface{}) {
-	pgOld, ok := prev.(*acidv1.Postgresql)
-	if !ok {
```
Jan-M: where did this error handling end up?

FxKu: The code is moved to `postgresqlCheck` method. There the error is only logged. Same behavior as before.

pkg/controller/controller.go
```diff
@@ -68,6 +70,7 @@ func NewController(controllerConfig *spec.ControllerConfig) *Controller {
 		config:           *controllerConfig,
 		opConfig:         &config.Config{},
 		logger:           logger.WithField("pkg", "controller"),
+		controllerID:     os.Getenv("CONTROLLER_ID"),
```
sdudoladov: Because of the way [`os.Getenv` works](https://golang.org/pkg/os/#Getenv), setting `CONTROLLER_ID` to the empty string (e.g. due to a mistake in a CI/CD pipeline) will have the same effect as not setting it at all. That might lead to  a hard-to-spot misconfiguration with undefined consequences. using [LookupEnv](https://golang.org/pkg/os/#LookupEnv) with a warning for such a case may be slightly better

pkg/controller/postgresql.go
```diff
@@ -40,12 +40,27 @@ func (c *Controller) clusterResync(stopCh <-chan struct{}, wg *sync.WaitGroup) {
...
+
 	// TODO: use the SharedInformer cache instead of quering Kubernetes API directly.
 	list, err := c.KubeClient.AcidV1ClientSet.AcidV1().Postgresqls(c.opConfig.WatchedNamespace).List(options)
 	if err != nil {
 		c.logger.Errorf("could not list postgresql objects: %v", err)
 	}
-	return list, err
+	if c.controllerID != "" {
+		c.logger.Debugf("watch only clusters with controllerID %q", c.controllerID)
+	}
```
sdudoladov: 1. the `else` branch of this condition - i.e. watching all PG resources without explicit annotations  - is also worth a logging message just in case someone does  a misconfiguration
2. why is this message logged here and not at the controller (operator pod) start up ?

pkg/controller/controller.go
```diff
@@ -411,3 +415,16 @@ func (c *Controller) getEffectiveNamespace(namespaceFromEnvironment, namespaceFr
...
+// Whether it's owner is determined by the value of 'acid.zalan.do/controller'
+// annotation. If the value matches the controllerID then it owns it, or if the
+// controllerID is "" and there's no annotation set.
+func (c *Controller) hasOwnership(postgresql *acidv1.Postgresql) bool {
+	if postgresql.Annotations != nil {
+		if owner, ok := postgresql.Annotations[constants.PostgresqlControllerAnnotationKey]; ok {
+			return owner == c.controllerID
+		}
+	}
+	return c.controllerID == ""
```
sdudoladov: 1. so the operator with `return c.controllerID == ""` will also be responsible for all PG clusters with non-empty `PostgresqlControllerAnnotationKey` that are not managed by other operator ? For example for the orphan ones ?
2. because of its importance, this method needs a couple of unit tests

sdudoladov:
the `Namespaces` section of the operator doc should also mention that there are now 2 complementary ways to isolate operators: namespaces and ownership annotations

docs/administrator.md
```diff
@@ -113,6 +113,33 @@ If your cluster uses a DNS domain other than the default `cluster.local`, this
...
 
+## Operators with defined ownership on certain Postgres clusters
+
+By default, multiple operators can only run together in Kubernetes when isolated
+into their [own namespaces](administrator.md#specify-the-namespace-to-watch).
+But, it is also possible to define ownership between operator instances and
+Postgres clusters running all in the same namespace without interfering.
+
+First, define the [`CONTROLLER_ID`](../../manifests/postgres-operator.yaml#L38)
+environment variable in the operator deployment manifest. Then specify the ID
```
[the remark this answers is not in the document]
FxKu: Good question. You could run multiple operators sharing the same config map. Then you would need to overwrite the controller ID with an env variable like here. I could add another configuration option, but will there be any benefit from it?

pkg/controller/postgresql.go
FxKu: 1. The `else` part reflects the current behavior people are used to. I'm not sure if adding another log line would raise more confusion or trigger users to add annotations to their manifests.
2. Where you think is a good place in the controller? Maybe it's good to log it at each iteration of listClusters to remind people that this operator only listens to certain clusters.

pkg/controller/controller.go
FxKu: 1. There can indeed be orphaned Postres clusters if the operator with the corresponding controller ID is gone. The operator with `c.controllerID == ""` will not take over. Not sure, if he should.
2. definitely. Will add them.

docs/administrator.md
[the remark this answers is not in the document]
sdudoladov: > You could run multiple operators sharing the same config map.

I totally missed that. It makes this feature even more interesting :)

pkg/controller/controller.go
sdudoladov: please add unit tests to avoid regressions in the future

sdudoladov:
also one needs to carefully configure separate operator instances to use different CRD / configmaps ; when the ownership annotations are used

cmd/main.go
```diff
@@ -19,6 +19,7 @@ var (
 	outOfCluster   bool
 	version        string
 	config         spec.ControllerConfig
+	controllerID   string
```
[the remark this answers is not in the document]
FxKu: hm, looks like a leftover

pkg/controller/controller.go
```diff
@@ -239,6 +242,7 @@ func (c *Controller) initRoleBinding() {
 
 func (c *Controller) initController() {
 	c.initClients()
+	c.controllerID = os.Getenv("CONTROLLER_ID")
```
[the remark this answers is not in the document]
FxKu: I chose to add the controller ID as an argument for new controller

FxKu:
> 1. Why does this PR use annotations and not labels ? From [Annotations docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects):
> 

From our K8s experts: "Labels are for humans, annotation for robots. The whole class etc patterns are based on annotations in all controllers!"

FxKu:
> 1. Why does this PR use annotations and not labels ? From [Annotations docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#attaching-metadata-to-objects):
> 

From our K8s experts: "Labels are for humans, annotation for robots. The whole class etc patterns are based on annotations in all controllers!"

FxKu:
> also one needs to carefully configure separate operator instances to use different CRD / configmaps ; when the ownership annotations are used

That's up to the user. Multiple operators can share one ConfigMap/CRD or use their own dedicated configuration.

FxKu:
> the `Namespaces` section of the operator doc should also mention that there are now 2 complementary ways to isolate operators: namespaces and ownership annotations

Hm, I mean the two paragraphs are next to each other. I think, it's fine. I did not want to advertise the feature too much as it should be used with care.

sdudoladov:
:+1:

FxKu:
:+1:

PR #803, opened 2020-01-27
Add ServiceAnnotations cluster config

jonathanbeber:
The [operator parameters][1] already support the `custom_service_annotations` config. With this parameter is possible to define custom annotations that will be used on the services created by the operator. The `custom_service_annotations` as all the other [operator parameters][1] are defined on the operator level and do not allow customization on the cluster level. A cluster may require different service annotations, as for example, set up different cloud load balancers timeouts, different ingress annotations, and/or enable more customizable environments.

This commit introduces a new parameter on the cluster level, called `serviceAnnotations`, responsible for defining custom annotations just for the services created by the operator to the specifically defined cluster. It allows a mix of configuration between `custom_service_annotations` and `serviceAnnotations` where the latest one will have priority. In order to allow custom service annotations to be used on services without LoadBalancers (as for example, service mesh services annotations) both `custom_service_annotations` and `serviceAnnotations` are applied independently of load-balancing configuration. For retro-compatibility purposes, `custom_service_annotations` is still under [Load balancer related options][2]. The two default annotations when using LoadBalancer services, `external-dns.alpha.kubernetes.io/hostname` and `service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout` are still defined by the operator. `service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout` can be overridden by `custom_service_annotations` or `serviceAnnotations`, allowing a more customizable environment. `external-dns.alpha.kubernetes.io/hostname` can not be overridden once there is no differentiation between custom service annotations for replicas and masters.

Also, it updates the documentation and creates the necessary unit and e2e tests to the above-described feature.

[1]: [https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md](https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md)

[2]: [https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md#load-balancer-related-options](https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md#load-balancer-related-options)

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -266,6 +266,10 @@ spec:
                       pattern: '^(\d+(e\d+)?|\d+(\.\d+)?(e\d+)?[EPTGMK]i?)$'
                       # Note: the value specified here must not be zero or be higher
                       # than the corresponding limit.
+            serviceAnnotations:
+              type: object
+              additionalProperties:
+                type: string
             sidecars:
               type: array
               nullable: true
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -122,6 +122,11 @@ These parameters are grouped directly under  the `spec` key in the manifest.
   A map of key value pairs that gets attached as [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
   to each pod created for the database.
 
+* **serviceAnnotations**
+  A map of key value pairs that gets attached as [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
+  to the services created for the database cluster. Check the
+  [administrator docs](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#load-balancers-and-allowed-ip-ranges)
+  for more information regarding default values and overwrite rules.
 
 * **enableShmVolume**
   Start a database pod without limitations on shm memory. By default Docker
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -380,8 +380,9 @@ In the CRD-based configuration they are grouped under the `load_balancer` key.
   `false`.
 
 * **custom_service_annotations**
-  when load balancing is enabled, LoadBalancer service is created and
-  this parameter takes service annotations that are applied to service.
+  This key/value map provides a list of annotations that get attached to each
+  service of a cluster created by the operator. If the annotation key is also
+  provided by the cluster definition, the manifest value is used.
   Optional.
 
 * **master_dns_name_format** defines the DNS name string template for the
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -32,6 +32,8 @@ spec:
 #  spiloFSGroup: 103
 #  podAnnotations:
 #    annotation.key: value
+#  serviceAnnotations:
+#    annotation.key: value
 #  podPriorityClassName: "spilo-pod-priority"
 #  tolerations:
 #  - key: postgres
--- a/manifests/postgresql.crd.yaml
+++ b/manifests/postgresql.crd.yaml
@@ -230,6 +230,10 @@ spec:
                       pattern: '^(\d+(e\d+)?|\d+(\.\d+)?(e\d+)?[EPTGMK]i?)$'
                       # Note: the value specified here must not be zero or be higher
                       # than the corresponding limit.
+            serviceAnnotations:
+              type: object
+              additionalProperties:
+                type: string
             sidecars:
               type: array
               nullable: true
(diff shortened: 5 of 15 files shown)


charts/postgres-operator/crds/postgresqls.yaml
```diff
@@ -173,6 +173,10 @@ spec:
               type: object
               additionalProperties:
                 type: string
+            serviceAnnotations:
```
FxKu: I've applied alphabetical sorting here to make it easier to find parameters because we have so many now. Can you move it down, please?

docs/reference/operator_parameters.md
```diff
@@ -380,8 +380,9 @@ In the CRD-based configuration they are grouped under the `load_balancer` key.
   `false`.
 
 * **custom_service_annotations**
-  when load balancing is enabled, LoadBalancer service is created and
-  this parameter takes service annotations that are applied to service.
+  This key/value map provides a list of annotations that get attached to each
+  service of a cluster created by the operator. If the annotation key is also
+  provided by the cluster definition, the cluster definition value is used.
```
FxKu:
```suggestion
  provided by the cluster definition, the manifest value is used.
```

manifests/complete-postgres-manifest.yaml
```diff
@@ -32,6 +32,8 @@ spec:
 #  spiloFSGroup: 103
 #  podAnnotations:
 #    annotation.key: value
+  serviceAnnotations:
```
FxKu: please use comments here, like we did for `podAnnotations`

manifests/postgresql.crd.yaml
```diff
@@ -137,6 +137,10 @@ spec:
               type: object
               additionalProperties:
                 type: string
+            serviceAnnotations:
```
FxKu: alphabetical order violated, see comment above

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -294,6 +294,14 @@ var PostgresCRDResourceValidation = apiextv1beta1.CustomResourceValidation{
 							},
 						},
 					},
+					"serviceAnnotations": {
```
FxKu: alphabetical order violated, see comment above

docs/reference/cluster_manifest.md
```diff
@@ -122,6 +122,19 @@ These parameters are grouped directly under  the `spec` key in the manifest.
   A map of key value pairs that gets attached as [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
   to each pod created for the database.
 
+* **serviceAnnotations**
+  A map of key value pairs that gets attached as [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
+  to the services created for the database cluster. If
+  `enableMasterLoadBalancer` parameter (or the operator config
```
FxKu: I think this part about service annotations when enabling load balancers better fits into [this](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#load-balancers-and-allowed-ip-ranges) section in the admin docs. Still, you might give a hint and link to this paragraph.

manifests/complete-postgres-manifest.yaml
jonathanbeber: I use this value on the e2e test. I'll move this to a new manifest specific for `serviceAnnotations` and comment it.

FxKu: ah I see. Another approach could be to patch a running `postgresql` and see if the `serviceAnnotations` appear on the services

jonathanbeber: indeed I like your solution but it'd not work since [`sameService`](https://github.com/zalando/postgres-operator/blob/master/pkg/util/k8sutil/k8sutil.go#L122) is not prepared to update service with annotations change. That's a known problem and will work on it in a separate PR soon. I will also add an e2e test for it as you described.

jonathanbeber:
@FxKu thanks for your review. I addressed these points.

FxKu:
Thanks, tested it myself and worked. Not sure, if I really like the extra manifest. But if you are planning another PR, where it can be removed again, it's fine with me.

Just one more thing: Could you revert the changes in go.mod and go.sum. They probably got introduced due to running e2e tests. That's something we should fix soon.

jonathanbeber:
> Thanks, tested it myself and worked. Not sure, if I really like the extra manifest. But if you are planning another PR, where it can be removed again, it's fine with me.

Yes, definitely we can remove it soon on my next PR.

> Just one more thing: Could you revert the changes in go.mod and go.sum. They probably got introduced due to running e2e tests. That's something we should fix soon.

Done.

thanks again for your review @FxKu

FxKu:
:+1:

jonathanbeber:
:+1:

PR #808, opened 2020-01-31
make minimum limits boundaries configurable

FxKu:
Do not simply error out, but raise limits to configured minimum limits.
Lower limits for operator pod and default limits for pods.

### DIFF
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -281,8 +281,8 @@ resources:
     cpu: 500m
     memory: 500Mi
   requests:
-   cpu: 100m
-   memory: 250Mi
+    cpu: 100m
+    memory: 250Mi
 
 # Affinity for pod assignment
 # Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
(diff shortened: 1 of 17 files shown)


charts/postgres-operator/values-crd.yaml
```diff
@@ -115,13 +115,17 @@ configKubernetes:
...
-  default_cpu_limit: "3"
-  # cpu request value for the postgres containers
+  default_cpu_limit: "1"
+  # CPU request value for the postgres containers
   default_cpu_request: 100m
   # memory limits for the postgres containers
-  default_memory_limit: 1Gi
+  default_memory_limit: 500Mi
   # memory request value for the postgres containers
   default_memory_request: 100Mi
```
Jan-M: Here a sound value should also be like 256MB from our expectations?

pkg/cluster/cluster.go
```diff
@@ -616,13 +599,22 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
...
+
 		oldSs, err := c.generateStatefulSet(&oldSpec.Spec)
 		if err != nil {
 			c.logger.Errorf("could not generate old statefulset spec: %v", err)
 			updateFailed = true
 			return
 		}
 
+		// update newSpec to for latter comparison with oldSpec
+		c.validateResources(&newSpec.Spec)
```
Jan-M: A bit ugly function name for one that modifies values.

Jan-M:
:+1:

Jan-M:
:+1:

FxKu:
:+1:

PR #813, opened 2020-02-03
Add region setting for logical backups to non-AWS storage

vitobotta:
Hi! Following the example of #628 I've added a setting for the region when configuring logical backups. With some S3-compatible services like Scaleway you need to specify the region as well as the endpoint in the aws upload command, otherwise the upload fails. I found this while investigating with the dump.sh script and found that I had to add `--region <region>` to the aws command for the upload to succeed. Tested many times.Hope this is OK. Thanks!

### DIFF
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -204,7 +204,7 @@ configLogicalBackup:
   logical_backup_s3_access_key_id: ""
   # S3 bucket to store backup results
   logical_backup_s3_bucket: "my-bucket-url"
-  # S3 region
+  # S3 region of bucket
   logical_backup_s3_region: ""
   # S3 endpoint url when not using AWS
   logical_backup_s3_endpoint: ""
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -204,6 +204,8 @@ configLogicalBackup:
   logical_backup_s3_access_key_id: ""
   # S3 bucket to store backup results
   logical_backup_s3_bucket: "my-bucket-url"
+  # S3 region
+  logical_backup_s3_region: ""
   # S3 endpoint url when not using AWS
   logical_backup_s3_endpoint: ""
   # S3 Secret Access Key
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -195,7 +195,7 @@ configLogicalBackup:
   logical_backup_s3_access_key_id: ""
   # S3 bucket to store backup results
   logical_backup_s3_bucket: "my-bucket-url"
-  # S3 region
+  # S3 region of bucket
   logical_backup_s3_region: ""
   # S3 endpoint url when not using AWS
   logical_backup_s3_endpoint: ""
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -195,6 +195,8 @@ configLogicalBackup:
   logical_backup_s3_access_key_id: ""
   # S3 bucket to store backup results
   logical_backup_s3_bucket: "my-bucket-url"
+  # S3 region
+  logical_backup_s3_region: ""
   # S3 endpoint url when not using AWS
   logical_backup_s3_endpoint: ""
   # S3 Secret Access Key
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -462,7 +462,7 @@ grouped under the `logical_backup` key.
   accessible by Postgres pods. Default: empty.
 
 * **logical_backup_s3_region**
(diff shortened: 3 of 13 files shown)


FxKu:
Thanks for this PR.
This new option is not yet covered by [CRD validation](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters). Please add it (in alphabetical order).

I think at some point we should switch to a map to pass all these arguments. Cannot create a new field every time a user needs another s3 cp [option](https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html).

vitobotta:
Hi @FxKu ! I've added the missing bit. Hopefully it's OK now. :) I'm still new to this stuff. Thanks

docs/reference/operator_parameters.md
```diff
@@ -461,6 +461,9 @@ grouped under the `logical_backup` key.
   S3 bucket to store backup results. The bucket has to be present and
   accessible by Postgres pods. Default: empty.
 
+* **logical_backup_s3_region**
+  Required with some non-AWS S3 storage services.
```
FxKu:
```suggestion
  Specifies the regions of the bucket which is required with some non-AWS S3 storage services. The default is empty.
```

docs/reference/operator_parameters.md
```diff
@@ -461,6 +461,9 @@ grouped under the `logical_backup` key.
...
+
 * **logical_backup_s3_endpoint**
   When using non-AWS S3 storage, endpoint can be set as a ENV variable.
```
FxKu: Can you add this info, too? :)
```suggestion
  When using non-AWS S3 storage, endpoint can be set as a ENV variable. The default is empty.
```

FxKu:
Validation is still missing. Let me point you to the correct places (insert before the line I've linked):
[CRD manifest](https://github.com/zalando/postgres-operator/blob/8794e4f9acad87c8acbb0e4dd13b952fcd38f89e/manifests/operatorconfiguration.crd.yaml#L222)
Manifest [copy](https://github.com/zalando/postgres-operator/blob/8794e4f9acad87c8acbb0e4dd13b952fcd38f89e/charts/postgres-operator/crds/operatorconfigurations.yaml#L246) for Helm chart
CRD validation in [Go code](https://github.com/zalando/postgres-operator/blob/master/pkg/apis/acid.zalan.do/v1/crds.go#L912)

vitobotta:
Hi @FxKu is it OK now or is there anything else missing? Thank you for your patience :)

vitobotta:
Looks like the last commit failed. Something else missing perhaps? Sorry, I'm not familiar with Go etc yet.

charts/postgres-operator/values-crd.yaml
```diff
@@ -204,6 +204,8 @@ configLogicalBackup:
   logical_backup_s3_access_key_id: ""
   # S3 bucket to store backup results
   logical_backup_s3_bucket: "my-bucket-url"
+  # S3 region
```
FxKu:
```suggestion
  # S3 region of bucket
```

charts/postgres-operator/values.yaml
```diff
@@ -195,6 +195,8 @@ configLogicalBackup:
   logical_backup_s3_access_key_id: ""
   # S3 bucket to store backup results
   logical_backup_s3_bucket: "my-bucket-url"
+  # S3 region
```
```suggestion
  # S3 region of bucket
```

docs/reference/operator_parameters.md
```diff
@@ -461,8 +461,11 @@ grouped under the `logical_backup` key.
   S3 bucket to store backup results. The bucket has to be present and
   accessible by Postgres pods. Default: empty.
 
+* **logical_backup_s3_region**
+  Specifies the regions of the bucket which is required with some non-AWS S3 storage services. The default is empty.
```
FxKu:
```suggestion
  Specifies the region of the bucket which is required with some non-AWS S3 storage services. The default is empty.
```

FxKu:
Retrigger the travis build and now it succeeds. Can you update my last suggestions. Sorry missed them before. Than it's done :smiley:

vitobotta:
@FxKu Done. Hopefully it's OK this time :D

FxKu:
:+1:

hjacobs:
:+1:

hjacobs:
@vitobotta thanks for your blog post!

vitobotta:
@hjacobs Thank you guys for open sourcing this! :)

Jan-M:
Just FYI on your blog post: We do intentionally not use wal-g for shipping backups despite the fact that it is significantly fast. We do though use it for restoring.

vitobotta:
Hi @Jan-M ! I must have missed that that. Are there any problems with using wal-g for backups as well? Do you recommend that I change my clusters because of this?

vitobotta:
Thanks for merging!

PR #816, opened 2020-02-05
Fix MasterPodMoveTimeout field that cannot be unmarshalled

f2cognite:
I tried using the default CRD-based operator configuration from [here](https://github.com/zalando/postgres-operator/blob/master/manifests/operatorconfiguration.crd.yaml), but it failed:

```
time="2020-02-05T13:48:02Z" level=fatal msg="unable to read operator configuration: could not get operator configuration object \"postgres-operator-configuration\": v1.OperatorConfiguration.Configuration: v1.OperatorConfigurationData.Kubernetes: v1.KubernetesMetaConfiguration.MasterPodMoveTimeout: readUint64: unexpected character: \xff, error found in #10 byte of ...|timeout\":\"20m\",\"oaut|..., bigger context ...|\"enable_sidecars\":true,\"master_pod_move_timeout\":\"20m\",\"oauth_token_secret_name\":\"postgresql-operato|..." pkg=controller
```

From what I can tell,  attempting to unmarshal the `MasterPodMoveTimeout` field from a CRD-based operator configuration fails because the `time.Duration` type is not unmarshallable.

The solution to this seems to be to use the `Duration` type instead:

https://github.com/zalando/postgres-operator/blob/8794e4f9acad87c8acbb0e4dd13b952fcd38f89e/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go#L192-L193

That type seems to be what's consistently used for other fields, and it is unmarshallable:

https://github.com/zalando/postgres-operator/blob/8794e4f9acad87c8acbb0e4dd13b952fcd38f89e/pkg/apis/acid.zalan.do/v1/marshal.go#L128-L152

I've tested locally and with this change the error disappeared.

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
+++ b/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
@@ -67,7 +67,7 @@ type KubernetesMetaConfiguration struct {
 	// TODO: use namespacedname
 	PodEnvironmentConfigMap    string        `json:"pod_environment_configmap,omitempty"`
 	PodPriorityClassName       string        `json:"pod_priority_class_name,omitempty"`
-	MasterPodMoveTimeout       time.Duration `json:"master_pod_move_timeout,omitempty"`
+	MasterPodMoveTimeout       Duration      `json:"master_pod_move_timeout,omitempty"`
 	EnablePodAntiAffinity      bool          `json:"enable_pod_antiaffinity,omitempty"`
 	PodAntiAffinityTopologyKey string        `json:"pod_antiaffinity_topology_key,omitempty"`
 	PodManagementPolicy        string        `json:"pod_management_policy,omitempty"`
--- a/pkg/controller/operator_config.go
+++ b/pkg/controller/operator_config.go
@@ -66,7 +66,7 @@ func (c *Controller) importConfigurationFromCRD(fromCRD *acidv1.OperatorConfigur
 	result.NodeReadinessLabel = fromCRD.Kubernetes.NodeReadinessLabel
 	result.PodPriorityClassName = fromCRD.Kubernetes.PodPriorityClassName
 	result.PodManagementPolicy = fromCRD.Kubernetes.PodManagementPolicy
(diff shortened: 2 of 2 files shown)


erthalion:
Good catch! I've checked, and looks like you're right. I'm afraid the same problem is with the configuration via configmap, although it's sort of deprecated by now.

erthalion:
:+1:

Jan-M:
:+1:

pcornelissen:
FYI: this is not fixed. I just tried it ad the "... readUint64: unexpected character:" still happens with the  postgresql-operator-default-configuration.yaml from a few minutes ago

tclass:
is there some quick fix for this?

pcornelissen:
I just removed the value, but I don't know which value is used in that case ;)

erthalion:
> FYI: this is not fixed. I just tried it ad the "... readUint64: unexpected character:" still happens with the postgresql-operator-default-configuration.yaml from a few minutes ago:

Strange, I'll check. @pcornelissen just to make sure, you build the operator from 00f00af2e84786eb4a87a35f2b987f5f946af3c2 or higher?

pcornelissen:
I checked out the repo on the 12th and used the files in Master, so the commit should be included.
The image used is: registry.opensource.zalan.do/acid/postgres-operator:v1.3.1

erthalion:
> I checked out the repo on the 12th and used the files in Master, so the commit should be included.
The image used is: registry.opensource.zalan.do/acid/postgres-operator:v1.3.1

I'm confused, do you use a prebuild image `postgres-operator:v1.3.1` (which doesn't include this fix yet), or build your own from the master branch?

pcornelissen:
I checked out the master and used the config files from there, which use the container image above. I don't know when the code is built, so I assumed that when this is fixed, that the corresponding image would also be updated.
If that is not the case, could you trigger an image rebuild (and update the yamls), so other people don't fall into this trap as well?

FxKu:
@pcornelissen included within the new v1.4.0 release

PR #818, opened 2020-02-06
Allow services update when changing annotations

jonathanbeber:
The current implementations for `pkg.util.k8sutil.SameService` considers
only service annotations change on the default annotations created by the
operator. Custom annotations are not compared and consequently not
applied after the first service creation.

This commit introduces a complete annotations comparison between the
current service created by the operator and the new one generated based on
the configs. Also, it adds tests on the above-mentioned function.


jonathanbeber:
@FxKu this PR addresses the problem mentioned on https://github.com/zalando/postgres-operator/pull/803#discussion_r371744729.

It doesn't remove the extra manifest for e2e, I'll remove it as soon as both PRs (this and #803) are merged. Once it's done, I'll fix the e2e test to remove the extra manifest and perform tests on services changes.

FxKu:
Thanks @jonathanbeber. Great to see more tests. Any reason you don't want to print out the values when they differ?

jonathanbeber:
@FxKu you're welcome. 

> Any reason you don't want to print out the values when they differ?
Actually I didn't think all the way through it, my first idea was that it might pollute the logs. The same way we don't print the source ranges.

Do you think would it be better to print all the new and old annotations? Should we apply the same for source ranges? For example, currently, for source ranges, we do not display the old and one, just the new one when [defining it](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L1253).

FxKu:
Source ranges we can keep as is, but annotation diffs I would print. IIRC we change annotations regularly only for Statefulsets, not Services. So I will not pollute the logs too much.

jonathanbeber:
@FxKu I added some logic to return the reason with the diff on service annotations changes. Also, once  #803 is merged, I removed the extra manifest mentioned on https://github.com/zalando/postgres-operator/pull/803#discussion_r371744729.

FxKu:
I merged #803 yesterday :) So could you reflect the change in this PR here?

jonathanbeber:
:+1:

FxKu:
:+1:

PR #823, opened 2020-02-10
specify cluster in e2e taint test

FxKu:
otherwise it will not wait for the right replica
plus raise min memory limit to fasten tests (decompressing image)


FxKu:
:+1:

RafiaSabih:
:+1:

PR #824, opened 2020-02-11
patch/update services on type change

FxKu:
PR #716 introduced waiting for a service to be deleted before recreating it, because on error endpoint was left nil and operator could not connect. However, now when endpoint generation fails (if it already exists), service is not created which is even worse. First recreate the service then handle endpoint


Jan-M:
:+1:

FxKu:
:+1:

PR #829, opened 2020-02-14
switch to rbac/v1

FxKu:
TL;DR: first change RBAC apiVersion, then enable namespaced roles

Deploying the operator with namespaced roles support (see #786) creates the `postgres-pod` role in each namespace. If we want to bind to a clusterrole instead we must then overwrite the [internal definition](https://github.com/zalando/postgres-operator/blob/3b10dc645dd3f4112df7d15042f335b86c01b3fc/pkg/controller/controller.go#L199) by setting the `pod_service_account_role_binding_definition` in the configmap. So we should first configure the operator and then deploy.

And here's the problem: When changing the RBAC version within the same PR, the apiVersion in the role definition must be of version `v1` do be supported by the new operator. But the current operator only understands `v1beta1`. Everything could be done in one go, but it's safer to do this step here first. Everyone using these definition options must switch to `v1` then, but they would need to do this anyway at some point.

### DIFF
--- a/charts/postgres-operator-ui/templates/serviceaccount.yaml
+++ b/charts/postgres-operator-ui/templates/serviceaccount.yaml
@@ -9,7 +9,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
 
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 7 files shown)


FxKu:
:+1:

Jan-M:
:+1:

Issue #828, opened 2020-02-14
Postgres Operator should not assume it is the only operator


flickerfly:
It would be nice if this operator didn't build up like it was the only operator in town. Using a systemaccount called "operator" and being configured by a CRD of kind "OperatorConfiguration" is a bit awkward when implemented among other operators. It would save questions about what operator is associated with OperatorConfiguration. I do support the config by CR rather than ConfigMap concept though.

I wouldn't mind so much if it assumed it was the only postgres operator. That seems reasonable, but not sure why even that confusion couldn't be avoided.

FxKu:
I agree, that the name `OperatorConfiguration` is too generic. You have too see that the project started about three years ago, where operators in K8s were pretty new. I think that's why the original authors chose this name.

The operator service account will be eliminated with #786

flickerfly:
Yeah, totally get that history. :-) Cool to hear the serviceaccount is
already on the block.

PR #832, opened 2020-02-18
postgres-pod clusterrole

FxKu:
Currently #786 does not update/sync namespaced roles. By using a separate cluster role instead, changing privileges for all stateful sets and pods remains easy.

This PR strips out some parts of #786 not related to roles, like changing the default pod service account name from `operator` to `postgres-pod`. It also aligns the content from the RBAC [manifest](https://github.com/zalando/postgres-operator/blob/master/manifests/operator-service-account-rbac.yaml) with the charts [templates](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/templates/clusterrole.yaml).

Addresses #828


Jan-M:
:+1:

FxKu:
:+1:

PR #841, opened 2020-02-21
fix validation for S3ForcePathStyle

### LINKED ISSUE #838: Invalid type of value
Hi, I try to recreate my cluster from s3 and I currently fight with the following issue:
In my clone section of cluster manifest I configure the value: `s3_force_path_style: true ` among the others needed to restore from a backup.
If I don't use quotes, this is what I get after trying to apply via kubectl:
```
The postgresql "test-postgres" is invalid: spec.clone.s3_force_path_style: Invalid value: "boolean": spec.clone.s3_force_path_style in body must be of type string: "boolean"
```

So I tried to use quotes(`s3_force_path_style: "true" `), but then again operator cannot work with it:
```
time="2020-02-21T14:52:53Z" level=debug msg="skipping \"ADD\" event for the invalid cluster: json: cannot unmarshal string into Go struct field CloneDescription.s3_force_path_style of type bool" cluster-name=kamil/test-postgres pkg=controller
```

FxKu:
fixes #838

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -94,7 +94,7 @@ spec:
                 s3_secret_access_key:
                   type: string
                 s3_force_path_style:
-                  type: string
+                  type: boolean
                 s3_wal_path:
                   type: string
                 timestamp:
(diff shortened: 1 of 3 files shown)


FxKu:
:+1:

sdudoladov:
:+1:

FxKu:
Thanks @thepotatocannon for finding it.

PR #842, opened 2020-02-24
be more permissive with standbys

FxKu:
If somebody is already running a standby with more than one instance, the operator should not terminate the replicas, but give a warning instead that we recommend `1`.

The docs on standby clusters should also be more clear now on what is possible and how to set it up.

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -110,8 +110,9 @@ Those are top-level keys, containing both leaf keys and groups.
 
 * **min_instances**
   operator will run at least the number of instances for any given Postgres
-  cluster equal to the value of this parameter. When `-1` is specified, no
-  limits are applied. The default is `-1`.
+  cluster equal to the value of this parameter, except for standby clusters
+  which run with one pod if `numberOfInstances` is set to 1. When `-1` is
+  specified for `min_instances`, no limits are applied. The default is `-1`.
 
 * **resync_period**
   period between consecutive sync requests. The default is `30m`.
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -110,9 +110,10 @@ Those are top-level keys, containing both leaf keys and groups.
 
 * **min_instances**
   operator will run at least the number of instances for any given Postgres
-  cluster equal to the value of this parameter, except for standby clusters
-  which run with one pod if `numberOfInstances` is set to 1. When `-1` is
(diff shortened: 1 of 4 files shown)


docs/reference/operator_parameters.md
```diff
@@ -110,8 +110,9 @@ Those are top-level keys, containing both leaf keys and groups.
 
 * **min_instances**
   operator will run at least the number of instances for any given Postgres
-  cluster equal to the value of this parameter. When `-1` is specified, no
-  limits are applied. The default is `-1`.
+  cluster equal to the value of this parameter, except for standby clusters
+  which run with one pod if `numberOfInstances` is set to 1. When `-1` is
```
sdudoladov: this exception in the context of this doc looks somewhat arbitrary, the credentials' limitation is  worth mentioning here

pkg/cluster/k8sres.go
```diff
@@ -1048,11 +1048,13 @@ func (c *Cluster) getNumberOfInstances(spec *acidv1.PostgresSpec) int32 {
 	cur := spec.NumberOfInstances
 	newcur := cur
 
-	/* Limit the max number of pods to one, if this is standby-cluster */
 	if spec.StandbyCluster != nil {
-		c.logger.Info("Standby cluster can have maximum of 1 pod")
-		min = 1
-		max = 1
+		if newcur > 1 {
+			c.logger.Warningf("operator only supports standby clusters with 1 pod")
```
sdudoladov: better to mention credentials's problem in the log message for readability and comment the code

pkg/cluster/k8sres.go
```diff
@@ -1048,11 +1048,13 @@ func (c *Cluster) getNumberOfInstances(spec *acidv1.PostgresSpec) int32 {
...
+		} else {
+			min = newcur
+			max = newcur
```
sdudoladov: What is the idea behind these assignments ? This change effectively enables shutting down standby clusters (`newcur == 0` ) in production by circumventing the restriction imposed by the `min_instances` param ; that shutting down  might as well happen by mistake

FxKu: fair point. Now, I'm only setting it to the the spec's value if it's one.

pkg/cluster/k8sres.go
FxKu: that's another story and not related to any pod count. There will be enough error logs already to notice and with improved docs, people now know what to do.

sdudoladov:
:+1:

FxKu:
:+1:

PR #843, opened 2020-02-24
Bootstrapped databases with best practice role setup

FxKu:
At the moment, users are left alone with setting up read and write privileges in PostgreSQL. A best practice setup could look like this:

- three NOLOGIN roles: a DB owner, a DB reader and DB writer
- [default access privileges](https://www.postgresql.org/docs/12/sql-alterdefaultprivileges.html) for these three roles
- three (optional) LOGIN roles which inherit the privileges from the NOLOGIN roles and can be granted to application users

This PR adds a new section `preparedDatabases` to the manifest as the old `databases` field can not be changed without breaking changes. The whole role setup is hidden from the user. Application roles would still be specified under the `users` field. A few more features are currently included, but can be removed as it adds quite some complexity:

- If the preparedDatabases section is empty, create a database which is named like the Postgres cluster manifest
- Allow users to define database schemas for each bootstrapped database (default schema would be called `data`)
- Optional NOLOGIN and LOGIN roles per schema
- For each preparedDatabase allow a string map to specify extensions and which schema they should run (addresses #490)

As a core principle, I tried to only use go maps of complex objects to allow for non-breaking future additions. But in general, we do not aim for adding more complexity. The operator should not become a database migration tool where you specify any DDL stuff. E.g. schemas and extension might already be too much.


manifests/minimal-postgres-manifest.yaml
```diff
@@ -15,5 +15,7 @@ spec:
     foo_user: []  # role for application foo
   databases:
     foo: zalando  # dbname: owner
+  preparedDatabases:
+    bar: {}
```
sdudoladov: does this have the effect of creating the bar DB owned by `db_owner` ?

pkg/cluster/sync.go
```diff
@@ -536,6 +555,109 @@ func (c *Cluster) syncDatabases() error {
...
+
+	var schemas []string
+
+	for schema := range preparedSchemas {
+		schemas = append(schemas, schema)
+	}
+
+	if createPreparedSchemas, equal := util.SubstractStringSlices(schemas, currentSchemas); !equal {
+		for _, schemaName := range createPreparedSchemas {
+			owner := "_owner"
```
sdudoladov: these `_owner` / `_writer` etc.  suffixes look like constants everywhere in the code

manifests/minimal-postgres-manifest.yaml
FxKu: The roles are always called like the DB. So it would be `bar_owner` etc.
Plus, there will be a schema called `data`

pkg/cluster/database.go
```diff
@@ -224,6 +250,104 @@ func (c *Cluster) databaseNameOwnerValid(datname, owner string) bool {
...
+		schemaName, rolePrefix+"_writer", // sequences
+		schemaName, rolePrefix+"_reader", rolePrefix+"_writer", // types
+		schemaName, rolePrefix+"_reader", rolePrefix+"_writer")); err != nil { // functions
+		return fmt.Errorf("could not alter default privileges for database schema %s: %v", schemaName, err)
+	}
+
+	return nil
+}
+
+func (c *Cluster) execAlterGlobalDefaultPrivileges(owner, rolePrefix string) error {
```
sdudoladov: what if a role with the same name but different set of default privileges already exists ?

pkg/cluster/database.go
```diff
@@ -224,6 +250,104 @@ func (c *Cluster) databaseNameOwnerValid(datname, owner string) bool {
...
+	return true
+}
+
+func (c *Cluster) execAlterSchemaDefaultPrivileges(schemaName, owner, rolePrefix string) error {
+	if _, err := c.pgDb.Exec(fmt.Sprintf(schemaDefaultPrivilegesSQL, owner,
+		schemaName, rolePrefix+"_writer", rolePrefix+"_reader", // schema
+		schemaName, rolePrefix+"_reader", // tables
+		schemaName, rolePrefix+"_reader", // sequences
+		schemaName, rolePrefix+"_writer", // tables
+		schemaName, rolePrefix+"_writer", // sequences
```
sdudoladov: comments mention `// sequences` twice

pkg/util/users/users.go
```diff
@@ -74,25 +74,43 @@ func (strategy DefaultUserSyncStrategy) ProduceSyncRequests(dbUsers spec.PgUserM
 
 // ExecuteSyncRequests makes actual database changes from the requests passed in its arguments.
 func (strategy DefaultUserSyncStrategy) ExecuteSyncRequests(reqs []spec.PgSyncUserRequest, db *sql.DB) error {
+	var rr []spec.PgSyncUserRequest
```
sdudoladov: does  `rr` stand for  `request retries` ?

pkg/util/users/users.go
```diff
@@ -74,25 +74,43 @@ func (strategy DefaultUserSyncStrategy) ProduceSyncRequests(dbUsers spec.PgUserM
...
 			return fmt.Errorf("unrecognized operation: %v", r.Kind)
 		}
 
 	}
+
+	// creating roles might fail if group role members are created before the parent role
+	// retry adding roles as long as the number of failed attempts is shrinking
+	if len(rr) > 0 {
+		if len(rr) < len(reqs) {
+			if err := strategy.ExecuteSyncRequests(rr, db); err != nil {
```
sdudoladov: that recursive call to `ExecuteSyncRequests` is definitely worth a unit test

sdudoladov:
> Application roles would still be specified under the users field

With this change, what would be the preferred way to create application roles  ? With the old `users` section or with the new login roles created in this PR ?

pkg/cluster/database.go
FxKu: have a look at the SQL command. First SELECT on tables and sequences is granted to the `_reader` role, than the rest to the `_writer` role

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -77,6 +78,19 @@ type PostgresqlList struct {
...
+// PreparedDatabase describes elements to be bootstrapped
+type PreparedDatabase struct {
+	PreparedSchemas map[string]PreparedSchema `json:"schemas,omitempty"`
+	DefaultUsers    bool                      `json:"defaultUsers,omitempty" defaults:"false"`
+	Extensions      map[string]string         `json:"extensions,omitempty"`
+}
+
+// PreparedSchema describes elements to be bootstrapped in the schema
+type PreparedSchema struct {
+	DefaultRoles *bool `json:"defaultRoles,omitempty" defaults:"true"`
```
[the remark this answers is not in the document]
FxKu: Roles = NOLOGIN roles = safe
Users = LOGIN roles = should be enabled explicitly

manifests/complete-postgres-manifest.yaml
```diff
@@ -21,6 +21,17 @@ spec:
...
   databases:
     foo: zalando
+  preparedDatabases:
+    bar:
+      defaultUsers: true
+      extensions:
+        pg_partman: public
+        pgcrypto: public
+      schemas:
+        data: {}
```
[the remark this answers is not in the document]
FxKu: yes, it's not necessary here. But we also use default in configmap etc. so it's just to show the users how to write schemas

pkg/cluster/database.go
FxKu: I guess Postgres will overwrite existing privileges as it's an ALTER command

pkg/cluster/sync.go
```diff
@@ -541,26 +547,138 @@ func (c *Cluster) syncDatabases() error {
 		return fmt.Errorf("could not get current databases: %v", err)
 	}
 
-	for datname, newOwner := range c.Spec.Databases {
-		currentOwner, exists := currentDatabases[datname]
+	// if no prepared databases are specified create a database named like the cluster
+	preparedDatabases := c.Spec.PreparedDatabases
+	if preparedDatabases == nil || len(preparedDatabases) == 0 { // TODO: add option to disable creating such a default DB
```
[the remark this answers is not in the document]
FxKu: should be fixed now

pkg/cluster/sync.go
```diff
@@ -541,26 +547,138 @@ func (c *Cluster) syncDatabases() error {
...
+			return fmt.Errorf("could not init connection to database %s: %v", preparedDbName, err)
+		}
+		defer func() {
+			if err := c.closeDbConn(); err != nil {
+				c.logger.Errorf("could not close database connection: %v", err)
+			}
+		}()
+
+		// first, set default privileges for prepared database
+		c.execAlterGlobalDefaultPrivileges(preparedDbName+constants.OwnerRoleNameSuffix, preparedDbName)
```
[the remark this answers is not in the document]
FxKu: The operator can change owners of DBs listed under `databases`. I think, that already qualifies as revoking privileges. Altering default privileges works additive. Existing privileges won't be revoked.

Note, I moved this part into the function which creates the prepared databases, so that it's not executed on every snyc.

FxKu:
1. fixed
2. logs should be clearer now. There was indeed a bug, that roles were not created on UPDATE. Putting an existing database under the prepared section will fail if the owner is different. But, the owner can be changed under `databases` to fix this. Might be worth another paragraph in the docs.
3. I removed the word `bootstrapped` from the docs.

docs/user.md
```diff
@@ -216,6 +219,129 @@ to choose superusers, group roles, [PAM configuration](https://github.com/CyberD
...
+
+The roles described in the previous paragraph can be granted to LOGIN roles from
+the `users` section in the manifest. Optionally, the Postgres Operator can also
+bootstrap default LOGIN roles for the database an each schema individually.
+These roles will get the `_user` suffix and they inherit all right from their
+NOLOGIN counterparts.
+
+| Role name           | Member of      | Admin         |
+| ------------------- | -------------- | ------------- |
+| foo_owner_user      | foo_owner      | admin         |
```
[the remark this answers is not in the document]
Jan-M: Yeah that is a bit difficult. Ideally not the case. But then often by mistake objects are ultimately created and owned by the login role. @CyberDem0n any thoughts?

sdudoladov:
> But, the owner can be changed under databases to fix this. Might be worth another paragraph in the docs.

As a further note, everything Sync touches after prepared DBs (currently connection pooler), will not be Synced until permission issues are resolved

sdudoladov:
:+1:

FxKu:
:+1:

PR #846, opened 2020-02-25
add e2e test for node readiness label

FxKu:
add test case for setting node readiness label + improve log output of `moveMasterPodsOffNode` function

sort all e2e tests in the order how they are executed: alphabetically

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -211,6 +211,7 @@ class EndToEndTestCase(unittest.TestCase):
         k8s.api.custom_objects_api.patch_namespaced_custom_object(
             "acid.zalan.do", "v1", "default", "postgresqls", "acid-minimal-cluster", pg_patch_resources)
         k8s.wait_for_pod_failover(failover_targets, labels)
+        k8s.wait_for_pod_start('spilo-role=replica')
 
         pods = k8s.api.core_v1.list_namespaced_pod(
             'default', label_selector=labels).items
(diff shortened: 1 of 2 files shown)


e2e/tests/test_e2e.py
```diff
@@ -175,136 +242,62 @@ def test_multi_namespace_support(self):
...
-
-        k8s.wait_for_pg_to_scale(2)
-        self.assertEqual(2, k8s.count_pods_with_label(labels))
-        self.assert_master_is_unique()
-
-    @timeout_decorator.timeout(TEST_TIMEOUT_SEC)
-    def test_taint_based_eviction(self):
-        '''
-           Add taint "postgres=:NoExecute" to node with master. This must cause a failover.
+           Remove node readiness label from master node. This must cause a failover.
```
sdudoladov: ... on condition there is a healthy replica.

e2e/tests/test_e2e.py
```diff
@@ -175,136 +242,62 @@ def test_multi_namespace_support(self):
...
             }
         }
-        k8s.api.core_v1.patch_node(new_master_node, body)
+        k8s.update_config(patch_readiness_label_config)
+        new_master_node, new_replica_nodes = self.assert_failover(
+            current_master_node, num_replicas, failover_targets, cluster_label)
+
+        # patch also master node
+        k8s.api.core_v1.patch_node(current_master_node, patch_readiness_label)
+        # toggle pod anti affinity to move replica away from master node
```
sdudoladov: why ? at this point the test itself (checking the revocation of the readiness label) is already complete

e2e/tests/test_e2e.py
FxKu: well yeah... I take that for granted in all the tests

e2e/tests/test_e2e.py
FxKu: I made the comment more clear now. I'm patching the old master node in order to spread the pods again on two nodes for subsequent tests

sdudoladov:
:+1:

FxKu:
:+1:

PR #848, opened 2020-02-27
Change `logging_rest_api.api_port` to `8080` instead of `8008`

hengchu:
The documentation states that the default operator REST service is at port `8080`, but the current default CRD based configuration is `8008`. Changing the default config to match documentation.


FxKu:
:+1:

erthalion:
:

erthalion:
:+1:

Issue #850, opened 2020-02-28
cluster_domain default value not being assumed when omitted from OperatorConfiguration


grantlanglois:
When the cluster_domain parameter is omitted from the operator configuration the default value of "cluster.local" should be assumed. Instead the empty string is being assigned.

FxKu:
Closing. Thanks @grantlanglois for providing a PR.

PR #851, opened 2020-02-28
add json:omitempty option to ClusterDomain

grantlanglois:
ClusterDomain should have json:omitempty option so that the default value of cluster.local is assumed when the parameter is omitted from the OperatorConfiguration

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
+++ b/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
@@ -53,7 +53,7 @@ type KubernetesMetaConfiguration struct {
 	EnableInitContainers                   *bool                 `json:"enable_init_containers,omitempty"`
 	EnableSidecars                         *bool                 `json:"enable_sidecars,omitempty"`
 	SecretNameTemplate                     config.StringTemplate `json:"secret_name_template,omitempty"`
-	ClusterDomain                          string                `json:"cluster_domain"`
+	ClusterDomain                          string                `json:"cluster_domain,omitempty"`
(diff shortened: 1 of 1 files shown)


grantlanglois:
Fixes #850

FxKu:
just wondering, if it would still be possible to set: `cluster_domain: ""`

erthalion:
Thanks @grantlanglois good catch. Unfortunately this fix would be not enough, since defaults are not picked up if a configuration transformed from CRD. I can't push into your branch, so can you add this into the PR? This should be enough and we can merge it, since it would also prevent any empty values (what @FxKu mentioned).

```diff
diff --git a/pkg/controller/operator_config.go b/pkg/controller/operator_config.go
@@ -50,7 +51,7 @@ func (c *Controller) importConfigurationFromCRD(fromCRD *acidv1.OperatorConfigur
        result.PodTerminateGracePeriod = time.Duration(fromCRD.Kubernetes.PodTerminateGracePeriod)
        result.SpiloPrivileged = fromCRD.Kubernetes.SpiloPrivileged
        result.SpiloFSGroup = fromCRD.Kubernetes.SpiloFSGroup
-       result.ClusterDomain = fromCRD.Kubernetes.ClusterDomain
+       result.ClusterDomain = util.Coalesce(fromCRD.Kubernetes.ClusterDomain, "cluster.local")
        result.WatchedNamespace = fromCRD.Kubernetes.WatchedNamespace
        result.PDBNameFormat = fromCRD.Kubernetes.PDBNameFormat
        result.EnablePodDisruptionBudget = fromCRD.Kubernetes.EnablePodDisruptionBudget

```

FxKu:
:+1:

erthalion:
:+1:

PR #854, opened 2020-03-02
Use cryptographically secure password generation

f2cognite:
The current password generation algorithm is extremely deterministic, due to being based on the standard random number generator with a deterministic seed based on the current Unix timestamp (in seconds).

This can lead to a number of security issues, including:
* The same passwords being used in different Kubernetes clusters if the operator is deployed in parallel. (This issue was discovered because of four deployments having the same generated passwords due to automatically being deployed in parallel.)
* The passwords being easily guessable based on the time the operator pod started when the database was created. (This would typically be present in logs, metrics, etc., that may typically be accessible to more people than should have database access.)

This PR attempts to fix this issue by replacing the current randomness source with `crypto/rand`, which should produce cryptographically secure random data that is virtually unguessable. This will avoid both of the above problems as each deployment will be guaranteed to have unique, indeterministic passwords.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -2,8 +2,10 @@ package util
 
 import (
 	"crypto/md5" // #nosec we need it to for PostgreSQL md5 passwords
+	cryptoRand "crypto/rand"
 	"encoding/hex"
 	"fmt"
+	"math/big"
 	"math/rand"
 	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


erthalion:
Thank you for the PR, it looks reasonable and after a short testing I don't see any issues. The only commentary from me is that operator will obviously not synchronize any user on already existing cluster with a new stronger password, so if it's necessary to apply it in this case one have to rotate secrets. We probably need to mention it somewhere in change log or such.

erthalion:
:+1:

FxKu:
:+1:

PR #859, opened 2020-03-09
Lazy upgrade of the Spilo image

sdudoladov:
In certain situation (think node rotation) we know pods re-start for reason independent of the operator. When doing planned updates of the Spilo image, we piggyback on this observation and  skip the rolling update after the image update in the statefulset. That save us some downtime.


sdudoladov:
I welcome suggestions on how forcing the rolling upgrade should behave. The PR allows an administrator to force an update for the entire k8s cluster by changing the operator's configuration and restarting the operator pod. To support such behavior, the operator checks at every `Sync`  that all pods run the image specified in their stateful set. When it is not the case, the operator triggers the rolling update. That solution comes with a cost of extra k8s API calls and unneeded switchovers for clusters that have at least one "old" replica pod.

Currently the operator triggers the rolling update based on the results of comparison the actual and desired statefulssets' Specs. So it will not automatically catch pods running old images when the lazy update is disable because that action alone leaves the statefulset Spec intact.

Things to consider:

1. We can implement `syncPods` method that would - as part of `syncStatefulSet()` - simply restart replicas with the old image and only do switchovers when the master also runs the old image.
2. Do we need a toggle to force a rolling update on for individual PG clusters (i.e. a field in a PG manifest) ?

manifests/configmap.yaml
```diff
@@ -24,7 +24,7 @@ data:
   # enable_crd_validation: "true"
   # enable_database_access: "true"
   # enable_init_containers: "true"
-  # enable_lazy_image_upgrade: "true"
+  # enable_lazy_spilo_upgrade: "true"
```
FxKu: when the internal default is `false`, it shouldn't be true here. Same for CRD-based configuration.

pkg/cluster/sync.go
```diff
@@ -310,6 +336,19 @@ func (c *Cluster) syncStatefulSet() error {
 				}
 			}
 		}
+
+		if !podsRollingUpdateRequired {
```
FxKu:
```suggestion
		if c.OpConfig.EnableLazySpiloUpgrade && !podsRollingUpdateRequired {
```

I would move the check for lazy update here instead of doing it within `mustUpdatePodsAfterLazyUpdate

FxKu: I wonder if this part here is really necessary? When lazy update is active it would create another API call for every StatefulSet. Wouldn't the next sync fix outdated images?

e2e/tests/test_e2e.py
```diff
@@ -118,6 +118,61 @@ def test_enable_load_balancer(self):
...
+        "onDelete" pod update policy for stateful sets.
+
+        The test covers:
+        1) enabling lazy upgrade in existing operator deployment
+        2) forcing the normal rolling upgrade by changing the operator configmap and restarting its pod
+        '''
+
+        k8s = self.k8s
+        pod0 = "acid-minimal-cluster-0"
+        pod1 = "acid-minimal-cluster-1"
```
FxKu: hm, maybe we should use a constant for this minimal cluster, because we hard code it in so many tests. Then I would simply pass CONSTANT+0/1 to the other methods instead of defining variables, but that's just my taste

e2e/tests/test_e2e.py
```diff
@@ -118,6 +118,61 @@ def test_enable_load_balancer(self):
...
+
+        # enable lazy update
+        patch_lazy_spilo_upgrade = {
+            "data": {
+                "enable_lazy_spilo_upgrade": "true",
+                "docker_image": "registry.opensource.zalan.do/acid/spilo-cdp-12:1.6-p16"
```
FxKu: put this image into a variable to check if is used by the pod you delete

e2e/tests/test_e2e.py
```diff
@@ -118,6 +118,61 @@ def test_enable_load_balancer(self):
...
+            }
+        }
+        k8s.update_config(patch_lazy_spilo_upgrade)
+
+        # wait for sts update
+        time.sleep(60)
+
+        # restart the pod to get a container with the new image 
+        k8s.api.core_v1.delete_namespaced_pod(pod0, "default")
```
FxKu: can we make sure we delete the current replica pod and not simply the `-0` instance?

e2e/tests/test_e2e.py
```diff
@@ -118,6 +118,61 @@ def test_enable_load_balancer(self):
...
+        unpatch_lazy_spilo_upgrade = {
+            "data": {
+                "enable_lazy_spilo_upgrade": "false",
+            }
+        }
+        k8s.update_config(unpatch_lazy_spilo_upgrade)
+
+        # at this point operator will complete the normal rolling update
+        # so we additonally test if disabling the lazy update (forcing the normal rolling update) works
+        time.sleep(60)
```
FxKu: you could either use `wait_for_pod_start` here twice, or go for testing `assert_failover` as we do in other tests

pkg/cluster/sync.go
sdudoladov: > I would move the check for lazy update here instead of doing it within mustUpdatePodsAfterLazyUpdate

Changed. Note the condition is `!c.OpConfig.EnableLazySpiloUpgrade` , that is,  we only need to do it when  lazy update is disabled.

>  Wouldn't the next sync fix outdated images?

No. Sync() relies on stateful set spec to figure out if rolling update is needed. With the lazy upgrade, stateful set itself will be up-to-date (pod template will contain the most recent image),  but some of the pods may still be running old images just because they were not yet restarted. Operator will look in the stateful set spec, see the most recent image and happily skip the rolling upgrade for this cluster even though some pods run outdated images.

e2e/tests/test_e2e.py
sdudoladov: let's leave it as is right now and instead start a separate issue for tidying up e2e tests. I want to avoid cluttering this important PR with unrelated changes

e2e/tests/test_e2e.py
sdudoladov: it does not matter here, if we re-start master or replica. We simply need one newly launched pod to check that it indeed runs a new image as required by the semantics of lazy update

sdudoladov:
:+1:

FxKu:
:+1:

pkg/cluster/cluster.go
```diff
@@ -470,6 +470,14 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 		}
 	}
 
+	// lazy Spilo update: modify the image in the statefulset itself but let its pods run with the old image
+	// until they are re-created for other reasons, for example node rotation
+	if c.OpConfig.EnableLazySpiloUpgrade && !reflect.DeepEqual(c.Statefulset.Spec.Template.Spec.Containers[0].Image, statefulSet.Spec.Template.Spec.Containers[0].Image) {
+		needsReplace = true
+		needsRollUpdate = false
```
Jan-M, line 477: `needsRollUpdate = !c.OpConfig.EnableLazySpiloUpgrade` would also have been enough here.

The two codes parts just for the error message I am not sure are great. But done is done.

PR #860, opened 2020-03-09
omit PgVersion diff on sync

FxKu:
Changing the Postgres version can currently break the cluster. It is caught on `Update` but not on `Sync`.

Retrieving the current PG version on sync is only possible through the `bin_dir` setting in the `SPILO_CONFIGURATION` env variable of the `postgres` container. Extracting the version is done with `Sscanf` to make it work even when the bin path template is configurable, as suggested in #554.

This PR also adds `PostgresParam` where `PgVersion` field is read

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	pgBinariesLocationTemplate       = "/usr/lib/postgresql/%s/bin"
+	pgBinariesLocationTemplate       = "/usr/lib/postgresql/%v/bin"
 	patroniPGBinariesParameterName   = "bin_dir"
 	patroniPGParametersParameterName = "parameters"
 	patroniPGHBAConfParameterName    = "pg_hba"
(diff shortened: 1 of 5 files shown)


pkg/cluster/sync.go
```diff
@@ -23,6 +23,7 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	currentPgVersion := c.Spec.PostgresqlParam.PgVersion
```
FxKu: So this already returns the new version in the manifest :(

pkg/cluster/k8sres.go
```diff
@@ -716,6 +716,55 @@ func makeResources(cpuRequest, memoryRequest, cpuLimit, memoryLimit string) acid
 	}
 }
 
+func extractPgVersionFromBinPath(binPath string, template string) (string, error) {
+	var pgVersion float32
```
erthalion: Shouldn't it be just an integer?

FxKu: when there can be `9.6`?

erthalion: Yeah, forgot that for a couple of years it still would be a thing.

FxKu:
:+1:

erthalion:
:+1:

PR #862, opened 2020-03-11
e2e: wait for pods after disabling anti affinity

FxKu:
do not leave the pod distribution test with a cluster failing over, but wait for both instances to be ready


FxKu:
:+1:

erthalion:
:+1:

Issue #505, opened 2019-02-28
patroni boostrap options synchronous_mode (and others)


ReSearchITEng:
This might fall in a feature request list.

I am trying to setup synchronous_mode. For now the operator does not expose this patroni bootstrap option. This is similar to: #428

Initially I expected that entire spec/patroni gets passed to patroni boostrap, but I was wrong.
If this direction is not possible, will there be an option to pass another map of params from the cluster definition directly to patroni boostrap, without requiring code changes in the operator to interpreting every field?
Or, are there any other options to control patroni from my cluster request?

jecho:
piggybacking off this question. 

I’d like to request adding synchronous_mode type fields. I’ve exposed both synchronous_mode and synchronous_mode_strict in my branch that seems to work and I'd like to contribute. However, I’m still relatively new to how _patroni_ works and maybe this is why it has been excluded?

siku4:
I am are also very interested in this issue. 

@jecho I took a look at your fork (this [commit](https://github.com/jecho/postgres-operator-1/commit/337d018382548ed72677643ce79fdc7ddd1bb253), right?) and it looks good. Have you already tested it in production and made a pull request?

FxKu:
@jecho +1 for opening a PR. When adding manifest fields they have to be covered by CRD validation as well, but I will guide you then.

jecho:
@FxKu @siku4 cool, will make the PR in the next day or so

PR #869, opened 2020-03-18
Allow configuration of patroni's replication mode

lalbers:
Related to issue https://github.com/zalando/postgres-operator/issues/505
______
Patroni offers the possibility of a synchronous replication. This ensures consistency across a cluster by confirming that writes are written to a secondary before returning to the connecting client with a success.

It would be great, if this option is adjustable via postgresql.crd

Patroni replication modes:
[zalando/patroni - replication_modes.rst](https://github.com/zalando/patroni/blob/master/docs/replication_modes.rst)


manifests/complete-postgres-manifest.yaml
```diff
@@ -64,6 +64,7 @@ spec:
     ttl: 30
     loop_wait: &loop_wait 10
     retry_timeout: 10
+    synchronous_mode: true
```
FxKu: better set it to false here

docs/reference/cluster_manifest.md
```diff
@@ -212,6 +212,8 @@ explanation of `ttl` and `loop_wait` parameters.
   automatically created by Patroni for cluster members and permanent replication
   slots. Optional.
 
+* **synchronous_mode**
+  Patroni `synchronous_mode` parameter value. The default is set to `false`. Optional.
```
FxKu:
```suggestion
  Patroni `synchronous_mode` parameter value. Optional.

```

FxKu: There's no default specified in the go code and generally we don't do it for manifest fields. Just say it's optional as we do with other flags e.g. `enableMasterLoadBalancer` (and add a blank line between next header)

pkg/cluster/k8sres.go
```diff
@@ -222,6 +223,9 @@ PatroniInitDBParams:
 	if patroni.Slots != nil {
 		config.Bootstrap.DCS.Slots = patroni.Slots
 	}
+	if patroni.SynchronousMode != false {
```
```suggestion
	if patroni.SynchronousMode {
```

As there is no default, this field will be `nil` when left out. So we should only enter this code block when `true`. Right now, I'm not 100% sure if we have to check `patroni.SynchronousMode != nil`, too

manifests/complete-postgres-manifest.yaml
lalbers: I agree. Fixed :heavy_check_mark:  https://github.com/zalando/postgres-operator/pull/869/commits/59d567edd0bc0ecf1d17d87566c18935dd21ec1f

docs/reference/cluster_manifest.md
lalbers: Does the same but is shorter. So we can customize it if you like.

FxKu:
:+1:

FxKu:
@lalbers as Alex suggested could you add the same for `synchronous_mode_strict`, too?

lalbers:
> @lalbers as Alex suggested could you add the same for `synchronous_mode_strict`, too?

sounds reasonable, i will add that option :rocket:

FxKu:
:+1:

sdudoladov:
:+1:

Issue #781, opened 2020-01-06
Allow configuration of global_sidecars / sidecar_docker_images


frittentheke:
It's awesome to be able to "simply" inject sidecars globally via sidecar_docker_images of the operator configuration (https://github.com/zalando/postgres-operator/pull/331). 

Unfortunately there is no way to globally configure those sidecars yet ...

1.  While there is the pod_environment_configmap setting, the referenced config map has to exist in the namespace the postgresql cluster resides in (https://postgres-operator.readthedocs.io/en/latest/reference/operator_parameters/#kubernetes-resources)
2. The variables in this config map are only propagated to the postgresql container (not actually the whole pod / all containers) as the name of the variable would suggest: https://github.com/zalando/postgres-operator/issues/550

Certainly there are many use cases for globally applied sidecars, but mostly it's about applying common monitoring / logshipping, like discussed here https://github.com/zalando/postgres-operator/issues/264. 

I just ran into the same issue of being able to globally add the sidecars, but being unable to apply some common environment variables to configure those (see comment https://github.com/zalando/postgres-operator/issues/264#issuecomment-393478150)

Easiest would likely be to allow configuration of global sidecar containers just like individual postgresql custom resources (https://postgres-operator.readthedocs.io/en/latest/reference/cluster_manifest/#sidecar-definitions).

While fully exposing all PodSpec settings (https://github.com/zalando/postgres-operator/issues/479) does work as well, it is harder to validate against the postgresql CRD.

FxKu:
I think propagating the content of the pod_environment_configmap to all sidecars should be relatively easy to implement. Have nothing against this idea. As you said, the name already sounds like it would be available to all containers. Feel free to open PR.

frittentheke:
@FxKu  thanks for looking into this. 

While propagation of pod_environment_configmap to all containers makes sense in itself, this still means such a configmap has to exist in each namespace to "solve " the issue about configuration of the global sidecars. Do you have any thoughts on that?

FxKu:
[...]

PR #870, opened 2020-03-19
Change type of pod environment config map to NamespacedName

FxKu:
Change type of `PodEnvironmentConfigMap` to `NamespacedName` to allow this config map in a namespace other than where the operator runs. If the ConfigMap isn't found in the configured default namespace than the operator looks up it's own namespace (if different) to have the same behavior as before. 

addresses #781


pkg/cluster/k8sres.go
```diff
@@ -833,59 +834,65 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
-	if c.OpConfig.PodEnvironmentConfigMap != "" {
+	if c.OpConfig.PodEnvironmentConfigMap != (spec.NamespacedName{}) {
 		var cm *v1.ConfigMap
-		cm, err = c.KubeClient.ConfigMaps(c.Namespace).Get(c.OpConfig.PodEnvironmentConfigMap, metav1.GetOptions{})
+		cm, err = c.KubeClient.ConfigMaps(c.OpConfig.PodEnvironmentConfigMap.Namespace).Get(c.OpConfig.PodEnvironmentConfigMap.Name, metav1.GetOptions{})
 		if err != nil {
-			return nil, fmt.Errorf("could not read PodEnvironmentConfigMap: %v", err)
+			// if not found, try again using the operator namespace (old behavior)
+			if k8sutil.ResourceNotFound(err) && c.Namespace != c.OpConfig.PodEnvironmentConfigMap.Namespace {
+				cm, err = c.KubeClient.ConfigMaps(c.Namespace).Get(c.OpConfig.PodEnvironmentConfigMap.Name, metav1.GetOptions{})
```
sdudoladov: so if the namespace of a configmap with additional env vars is is not specified, do we look in the operator namespace or the PG cluster namespace ? `c.Namespace` is the latter, but the comment states the former

sdudoladov:
:+1:

frittentheke:
@FxKu this is very cool thanks for changing this so quickly.

May I kindly ask if you plan on adding the propagation of those ENV variables from the configmap to all containers as well? (https://github.com/zalando/postgres-operator/issues/781)

FxKu:
@frittentheke no concrete plans, yet. It's not on our current to do list. Shouldn't be too hard I guess. You still have the chance to open a PR :)

sdudoladov:
:+1:

FxKu:
:+1:

PR #874, opened 2020-03-24
Delete volumes for scaled down pods

### LINKED ISSUE #443: Delete PVC /Volumes during down-scaling
Right now Persistent volume claims (PVCs) are removed only on cluster deletion.  In the case the number of pods in a PG cluster is simply reduced, PVCs/volumes of no longer running pods remain. We may consider removing them altogether.

sdudoladov:
1. There is supposedly no standard way to find unused PVCs via k8s API, e.g. `kubectl describe pvc` [explicitly enumerates all pods/pvc](https://github.com/kubernetes/kubernetes/pull/65837/files#diff-b6046cf9a49c5cec609fe53ba861940dR1306) to provide `Mounted-By` field . This PR relies on the operator-specific naming scheme instead.
2.  `Sync()` in this PR also deletes PVCs for clusters scaled down before this change was implemented

Closes #443


hjacobs:
I did this in a similar way in `kube-janitor` (https://github.com/hjacobs/kube-janitor/issues/62), but I don't delete PVCs for scaled-down StatefulSets as I don't know whether they would still be needed. That's why it's good if the Postgres Operator handles it as it knows more about the state of the database and whether data can be restored (from replica or backup).

CyberDem0n:
Is there a switch to turn it off?

sdudoladov:
> Is there a switch to turn it off?

Not yet, I only started.  A very rough plan is

- [ ] a k8s cluster-wide switch
- [ ] an e2e test
- [ ] docs
- [ ] probably a better way to find unused PVC

charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -64,6 +64,8 @@ spec:
               type: boolean
             enable_shm_volume:
               type: boolean
+            should_delete_unused_pvc:
```
FxKu: would keep the pattern of naming flags `enable...`, maybe `enable_unused_pvc_deletion` or something (plus, mind alphabetical order)

FxKu:
I wonder, if the current code could be put into an `syncPersistentVolumeClaim` method that is called from `Update` and `Sync`. Sure you don't have the oldSpec on sync, but this can be handled.

pkg/cluster/sync.go
```diff
@@ -110,6 +111,24 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 		}
 	}
 
+	// remove unused PVCs in case deleting them during scale down failed; see Update()
+	// the last pvc stays until the cluster is explicitly deleted as opposed to being scaled down to 0 pods
+	if c.OpConfig.EnableUnusedPVCDeletion && c.getNumberOfInstances(&c.Spec) > 0 {
+
+		// XXX that also deletes PVC of pods shut down before this change is deployed
```
FxKu: Triple xXx?

sdudoladov: It marks the code that requires special attention, like `FIXME`, `HACK`, and other similar markers

Jan-M:
@sdudoladov Can you please take into consideration a broken cluster and user error.

E.g. we have a three pod cluster with master on -2. -0 and -1 pod are broken. User scales down to two pods. We delete volume -2.

May be worth e.g. asking patroni if there is a master running and is running on lower nodes.

pkg/cluster/volumes.go
```diff
@@ -52,6 +53,33 @@ func (c *Cluster) deletePersistentVolumeClaims() error {
...
+
+	// Scaling down to 0 replicas is not cluster deletion so keep the last pvc.
+	// Operator will remove it only when explicit "kubectl pg delete" is issued
+	if c.getNumberOfInstances(&c.Spec) == 0 {
+		c.logger.Info("cluster scaled down to 0 pods; skipping deletion of the last pvc")
+		return
+	}
+
+	// XXX that also deletes PVC of pods shut down before this change is deployed
+	for i := c.getNumberOfInstances(&c.Spec); ; i++ {
```
Jan-M, line 68: Make clear either in code or comment that PVC for pod-0 is never deleted.

pkg/cluster/volumes.go
```diff
@@ -52,6 +53,33 @@ func (c *Cluster) deletePersistentVolumeClaims() error {
 	return nil
 }
 
+func (c *Cluster) deleteUnusedPersistentVolumeClaims() {
```
Jan-M, line 56: Imho needs guard to only deleted PVCs >= running master (see other comment)

sdudoladov:
TODO 
[ ] add an option to the manifest to allow users to disable this feature
[ ] make the manipulation with code / indices more explicit when deleting the PVCs (i.g.e the cluster with 1 pod has the pod with the index 0 , so deleting PVC -1 is safe)

pkg/cluster/volumes.go
```diff
@@ -52,6 +53,33 @@ func (c *Cluster) deletePersistentVolumeClaims() error {
...
+	if c.getNumberOfInstances(&c.Spec) == 0 {
+		c.logger.Info("cluster scaled down to 0 pods; skipping deletion of the last pvc")
+		return
+	}
+
+	// XXX that also deletes PVC of pods shut down before this change is deployed
+	for i := c.getNumberOfInstances(&c.Spec); ; i++ {
+		podIndex := strconv.Itoa(int(i))
+		pvcName := "pgdata-" + c.Name + "-" + podIndex
+		if err := c.KubeClient.PersistentVolumeClaims(c.Namespace).Delete(context.TODO(), pvcName, c.deleteOptions); err != nil {
```
Jan-M, line 71: May delete 4, fail on 5. Next loop will not delete 5 ever again.

PR #876, opened 2020-03-24
enable controllerID for chart and allow configurable pod cluster role

### LINKED ISSUE #875: Deploying multiple operator instances fails
This appears to be caused by #832.

Deploying the first release of the operator using chart version `1.4.0` works. However, when deploying a second instance, helm returns this error:

```
Error: UPGRADE FAILED: kind ClusterRole with the name "postgres-pod" already exists in the cluster and wasn't defined in the previous release. Before upgrading, please either delete the resource from the cluster or remove it from the chart
```

I tried setting `rbac.create` to `false` but this causes other necessary RBAC resources to not get created. The chart also does not offer any way to customize the name of the `postgres-pod` ClusterRole. If this customization were available, multiple instances could be deployed without issue.

FxKu:
The PR fixes #875 and also adds the controller ID feature introduced with #802.
The pod service account name will be `postgres-pod` by default, but it can also be generated by helm. The controller ID is not defined by default.

### DIFF
--- a/charts/postgres-operator/templates/_helpers.tpl
+++ b/charts/postgres-operator/templates/_helpers.tpl
@@ -32,6 +32,20 @@ Create a service account name.
 {{- end -}}
 
 {{/*
+Create a pod service account name.
+*/}}
+{{- define "postgres-pod.serviceAccountName" -}}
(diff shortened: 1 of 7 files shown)


sdudoladov:
:+1:

FxKu:
:+1:

PR #882, opened 2020-03-27
Move operator to go 1.14

FxKu:
haven't been updated for a while now


sdudoladov:
:+1:

sdudoladov:
actually one already sees from the build pipeline that it does not help

FxKu:
```
/root/go/pkg/mod/k8s.io/apimachinery@v0.18.0/pkg/util/errors/errors.go:99:10: undefined: errors.Is
note: module requires Go 1.13
```
got this error also locally

sdudoladov:
[similar issue in another k8s project](https://github.com/kubernetes/kubernetes/issues/89148#issuecomment-600188383)

sdudoladov:
I have looked into this `Context` change. It provides the [foundation to future k8s features](https://github.com/kubernetes/enhancements/blob/master/keps/sig-api-machinery/20200123-client-go-ctx.md#motivation) (distributed tracing of API calls / request cancellation) and removes some API inconsistencies. I think for now we do not need to explicitly manipulate contexts, so using an empty context `context.TODO()` is fine

Summary from k8s 1.18 changelog: 
```
Signatures on methods in generated clientsets, dynamic, metadata, and scale clients have been 
modified to accept context.Context as a first argument. 

Signatures of Create, Update, and Patch  methods have been updated to accept CreateOptions, UpdateOptions and PatchOptions respectively. 

Signatures of Delete and DeleteCollection methods now accept DeleteOptions by value instead of by reference. 

Generated clientsets with the previous interface have been added in new "deprecated" packages to allow incremental migration to the new APIs. 
The deprecated packages will be removed in the 1.21 release. A tool is available at http://sigs.k8s.io/clientgofix to rewrite
method invocations to the new signatures.
```

FxKu:
:+1:

sdudoladov:
:+1:

PR #885, opened 2020-03-27
TLS - add OpenShift compatibility

ReSearchITEng:
Adds OpenShift compatibility by removing forced injection of fsGroup.
In it's current format, there is no workaround to use TLS in non-priviledged, non-root OpenShift, where fsGroup **must not be set** , as it's dynamically allocated at creation time.
Solves: https://github.com/zalando/postgres-operator/pull/798#issuecomment-605201260
@zimbatm and @FxKu - please review.

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -572,12 +572,15 @@ However, this certificate cannot be verified and thus doesn't protect from
 active MITM attacks. In this section we show how to specify a custom TLS
 certificate which is mounted in the database pods via a K8s Secret.
 
-Before applying these changes, in k8s (not in OpenShift) the operator must also 
-be configured with the `spilo_fsgroup` set to the GID matching the postgres user 
-group. If you don't know the value, use `103` which is the GID from the default 
-spilo image (`spilo_fsgroup=103` in the cluster request spec).
-If the value is not provided and you are in k8s (not OpenShift), the certificates 
-will not have proper permissions and the server will not work as desired.
+Before applying these changes, in k8s the operator must also be configured with
+the `spilo_fsgroup` set to the GID matching the postgres user group. If you
+don't know the value, use `103` which is the GID from the default spilo image
+(`spilo_fsgroup=103` in the cluster request spec).
+
+OpenShift allocates the users and groups dynamically (based on scc), and their
+range is different in every namespace. Due to this dynamic behaviour, it's not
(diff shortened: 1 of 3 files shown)


docs/user.md
```diff
@@ -572,10 +572,12 @@ However, this certificate cannot be verified and thus doesn't protect from
...
-Before applying these changes, the operator must also be configured with the
-`spilo_fsgroup` set to the GID matching the postgres user group. If the value
-is not provided, the cluster will default to `103` which is the GID from the
-default spilo image.
+Before applying these changes, in k8s (not in OpenShift) the operator must also 
+be configured with the `spilo_fsgroup` set to the GID matching the postgres user 
+group. If you don't know the value, use `103` which is the GID from the default 
+spilo image (`spilo_fsgroup=103` in the cluster request spec).
+If the value is not provided and you are in k8s (not OpenShift), the certificates 
+will not have proper permissions and the server will not work as desired.
```
FxKu: Instead of mentioning OpenShift in brackets, I would add a proper sentence at the end of this paragraph which explains why you cannot set it when using OpenShift.

FxKu:
I'm fine with this change and agree that it's better to remove the hardcoded spiloFsGroup and let you user configure it. It's documented and should be fine this way

docs/user.md
ReSearchITEng: documentation updated as requested.

pkg/cluster/k8sres.go
```diff
@@ -983,13 +980,8 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
 	if spec.TLS != nil && spec.TLS.SecretName != "" {
-		if effectiveFSGroup == nil {
-			c.logger.Warnf("Setting the default FSGroup to satisfy the TLS configuration")
-			fsGroup := int64(spiloPostgresGID)
-			effectiveFSGroup = &fsGroup
-		}
-		// this is combined with the FSGroup above to give read access to the
-		// postgres user
+		// this is combined with the FSGroup in the section above will
+		// give read access to the postgres user
```
```suggestion
		// this is combined with the FSGroup in the section above
		// will give read access to the postgres user
```

FxKu:
:+1:

Jan-M:
:+1:

Issue #855, opened 2020-03-05
master service gets no selectors on OpenShift


dejwsz:
Service for masters gets no selectors at all - they must be added manually. Without this cluster is not able to sync properly. Tested with operator 1.3.1 and spilo image: registry.opensource.zalan.do/acid/spilo-cdp-12:1.6-p16.
I needed to patch my "test-minimal-cluster" service dedicated to master PODs adding such 
```
selectors:
  selector:
    application: spilo
    cluster-name: test-minimal-cluster
    spilo-role: master
```
Replica service has selectors in place so it is fine. This is probably a bug?

CyberDem0n:
If Patroni is configured to use endpoints for leader race it will put the IP of the master pod into the subsets.
This is our recommended way to run the whole stack on K8s.
OpenShift is a different story. It doesn't allow putting IPs from the pods subnet into subsets, so label selector is required. But we never announced the official support of OpenShift anyway.

dejwsz:
I can see this: https://github.com/zalando/patroni/tree/master/kubernetes/openshift-example

dejwsz:
OK I found it. I updated the config map pointed by "pod_environment_configmap" like this:
  DEBUG: "true"
  KUBERNETES_USE_CONFIGMAPS: "true"
  PATRONI_KUBERNETES_ROLE_LABEL: spilo-role
and now master service gets its selectors properly. Of course I have value of "pod_role_label" also set to "spilo-role".

dejwsz:
To be sure it really worked I did it again after cleaning up everything carefully. I can confirm master service gets now proper selectors.

dejwsz:
This is also interesting regarding setting PATRONI_KUBERNETES_POD_IP, seems possible in openshift:
https://github.com/zalando/patroni/blob/master/kubernetes/openshift-example/templates/template_patroni_persistent.yaml

CyberDem0n:
Effectively it does nothing, because PATRONI_KUBERNETES_USE_ENDPOINTS isn't set.

dejwsz:
Yeah, I can see now as I could not repeat it today again, somehow I got these selectors yesterday.

ReSearchITEng:
I confirm that this issue is there on OpenShift 4.2 as well (even with KUBERNETES_USE_CONFIGMAPS: "true" ).
I confirm that once this section is added, cluster works fine.
[...]

PR #887, opened 2020-03-28
kubernetes_use_configmap

ReSearchITEng:
fixes https://github.com/zalando/postgres-operator/issues/855 bug, which causes missing selectors in master service, when they are actually required.
This way, this PR aligns operator with Patroni's DSC method: k8s with configMaps, by solving the bug in this "if condition": https://github.com/zalando/postgres-operator/blob/07c5da35e3572b8302f33e3dd8fbf5e8ca846229/pkg/cluster/k8sres.go#L1339
 
Because the operator was mainly tested in k8s with the default "endpoints" option, this use case (configMaps) option was hidden till tests were done in OpenShift.
When the kubernetes_use_configmaps parameter is set to true, this PR will also export the required **patroni** environment variable with the same name:  KUBERNETES_USE_CONFIGMAPS=true (as per patroni docs).


pkg/util/config/config.go
```diff
@@ -108,6 +108,7 @@ type Config struct {
 	ConnectionPool
 
 	WatchedNamespace      string            `name:"watched_namespace"`    // special values: "*" means 'watch all namespaces', the empty string "" means 'watch a namespace where operator is deployed to'
+	KubernetesUseConfigMaps                bool              `name:"kubernetes_use_configmaps" default:"false"`
```
FxKu: please `go.fmt` all your code. Usually, your IDE should auto-align here.

charts/postgres-operator/values.yaml
```diff
@@ -23,6 +23,8 @@ configGeneral:
   enable_shm_volume: "true"
   # etcd connection string for Patroni. Empty uses K8s-native DCS.
   etcd_host: ""
+  # Select if setup uses endpoints (default), or configmaps to manage leader (DCS=k8s)
+  # kubernetes_use_configmaps: false
```
```suggestion
  # kubernetes_use_configmaps: "false"
```
For configmap we have to quote the bools.

manifests/configmap.yaml
```diff
@@ -43,6 +43,7 @@ data:
   # enable_team_superuser: "false"
   enable_teams_api: "false"
   # etcd_host: ""
+  # kubernetes_use_configmaps: false
```
FxKu:
```suggestion
  # kubernetes_use_configmaps: "false"
```

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
 configuration:
   # enable_crd_validation: true
   etcd_host: ""
+  # kubernetes_use_configmaps: "False"
```
FxKu:
```suggestion
  # kubernetes_use_configmaps: false
```
here it's the opposite :smiley:

pkg/util/config/config.go
ReSearchITEng: oh, right, not sure why it didn't.
Did it now for all of them.

Jan-M:
:+1:

FxKu:
:+1:

PR #896, opened 2020-04-01
Emit K8S events to the postgresql CR as feedback to the requestor / user

frittentheke:
Currently the only "feedback" the user or creator of a _postgresql_ custom resource receives is the status. While the operator does produce quite a bit of log, this is not suitable to point users to in order for them to find issues with their particular cluster.

This PR shall enable to operator to (also) send events which are promoted by the Kubernetes documentation (https://kubernetes.io/docs/tasks/debug-application-cluster/#example-debugging-pending-pods) to help debugging issues with a resource and to search for messages from controllers handling this resource.

As this shall be a first cut, I might not have added enough events or even have added events that are too noisy. My intention was to report back certain aspects of the life-cycle:

- Creation with the individual steps / resources being created
- (Warnings in case resource limits are being adjusted)
- Sync in case anything goes wrong
- Deletion (might take a while or even run into problems - there seems to be not a lot of error handling / reporting there yet)
- Switchover or Rolling-Restarts initiated by the Operator

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -22,8 +22,10 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/tools/reference"
 
 	acidv1 "github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1"
+	"github.com/zalando/postgres-operator/pkg/generated/clientset/versioned/scheme"
 	"github.com/zalando/postgres-operator/pkg/spec"
 	"github.com/zalando/postgres-operator/pkg/util"
 	"github.com/zalando/postgres-operator/pkg/util/config"
@@ -143,7 +145,6 @@ func New(cfg Config, kubeClient k8sutil.KubernetesClient, pgSpec acidv1.Postgres
 	cluster.oauthTokenGetter = newSecretOauthTokenGetter(&kubeClient, cfg.OpConfig.OAuthTokenSecretName)
 	cluster.patroni = patroni.New(cluster.logger)
 	cluster.eventRecorder = eventRecorder
-
 	return cluster
 }
 
@@ -167,6 +168,16 @@ func (c *Cluster) setProcessName(procName string, args ...interface{}) {
 		Name:      fmt.Sprintf(procName, args...),
 		StartTime: time.Now(),
 	}
+}
+
+// GetReference of Postgres CR object
+// i.e. required to emit events to this resource
+func (c *Cluster) GetReference() *v1.ObjectReference {
+	ref, err := reference.GetReference(scheme.Scheme, &c.Postgresql)
+	if err != nil {
+		c.logger.Errorf("could not get reference for Postgresql CR %v/%v: %v", c.Postgresql.Namespace, c.Postgresql.Name, err)
+	}
+	return ref
 }
 
 // SetStatus of Postgres cluster
@@ -248,7 +259,7 @@ func (c *Cluster) Create() error {
 	}()
 
 	c.setStatus(acidv1.ClusterStatusCreating)
-	c.eventRecorder.Event(&c.Postgresql, v1.EventTypeNormal, "Create", "Started creation of new cluster resources")
+	c.eventRecorder.Event(c.GetReference(), v1.EventTypeNormal, "Create", "Started creation of new cluster resources")
 
 	if err = c.enforceMinResourceLimits(&c.Spec); err != nil {
 		return fmt.Errorf("could not enforce minimum resource limits: %v", err)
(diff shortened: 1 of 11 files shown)


FxKu:
Thanks for this interesting PR. There was already an internal request to extend the status, but events would also improve the user experience, I guess.

So far it doesn't work properly for me. Getting this error:
```Could not construct reference to: '&v1.Postgresql{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""},```

Could you also update the clusterrole [here](https://github.com/zalando/postgres-operator/blob/master/manifests/operator-service-account-rbac.yaml).

**Comment by frittentheke** [comment]
> Thanks for this interesting PR. There was already an internal request to extend the status, but events would also improve the user experience, I guess.
> 
> So far it doesn't work properly for me. Getting this error:
> `Could not construct reference to: '&v1.Postgresql{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""},`

@FxKu In general this works for me, but with the Postgresql reference passed along the various functions quite a lot there might be a call to the event function that does not get the input it expects...

Since events are logged (by the operator) along with the other logs, can you narrow this down to a certain action ?


> Could you also update the clusterrole [here](https://github.com/zalando/postgres-operator/blob/master/manifests/operator-service-account-rbac.yaml).

Done. I melted this into last commit.

**Comment by FxKu** [comment]
it's for all the event :smiley: 

at the end of each error output I see also:
```no kind is registered for the type v1.Postgresql in scheme "k8s.io/client-go/kubernetes/scheme/register.go:69"'```

**Comment by frittentheke** [comment]
> it's for all the events smiley
> 
> at the end of each error output I see also:
> `no kind is registered for the type v1.Postgresql in scheme "k8s.io/client-go/kubernetes/scheme/register.go:69"'`
> 
> Btw: Unit tests in `pkg/cluster/cluster_test.go` and `pkg/cluster/k8sres_test.go` are failinng because the event recorder is missing in the constructor

@FxKu  Urgh sorry about that, I might have missed /messed-up something when I cherry-picked some commits  from my local dev branch to this PR - I'll look into it.

**Comment by frittentheke** [comment]
@FxKu  

1. I did push a new commit to may branch that should fix the unit tests, sorry about missing on that initially. I also moved to the current HEAD of master.
2.   I then just built the operator locally and ran it via minikube (using the provided `./run_operator_locally.sh` script) and I received events just fine:

```
  Normal  Update      2m24s  postgres-operator  Performing rolling update
  Normal  Switchover  118s   postgres-operator  Switching over from "acid-minimal-cluster-0" to "default/acid-minimal-cluster-1"
  Normal  Switchover  117s   postgres-operator  Successfully switched over from "acid-minimal-cluster-0" to "default/acid-minimal-cluster-1"
  Normal  Switchover  117s   postgres-operator  Switchover from "acid-minimal-cluster-0" to "default/acid-minimal-cluster-1" FAILED: <nil>
  Normal  Update      99s    postgres-operator  Rolling update done - pods have been recreated
```


As I said, I expect that there might be more cases that events should be generated for (to properly allow users to follow the life-cycle of their clusters and there might still be bugs). But you not seeing any events at all seems strange.

**Comment by FxKu** [comment]
Thanks for the update. It's interesting: When updating my operator deployment I can see events for running clusters. But when I create a new cluster I get these `Could not construct reference to ...` errors for everything - add, update and delete events - until the first sync I guess. Seems that new clusters are not registered correctly. Hmm...

**Comment by frittentheke** [comment]
@FxKu yeah I observe the same errors now. Very strange that it complains about 

`no kind is registered for the type v1.Postgresql in scheme "k8s.io/client-go/kubernetes/scheme/register.go:69`

for new clusters.

**Comment by frittentheke** [comment]
@FxKu  apparently one has to get a proper reference to use when sending events.
Why and how this worked for some of those events - I don't know.

I added this with commit https://github.com/zalando/postgres-operator/pull/896/commits/633e001ee842f158e614bfc258d36dd6773e6c1c now ... and heureka, we have many many more events:

```
[...]
      superuser
      createdb
  Volume:
    Size:  1Gi
Status:
  Postgres Cluster Status:  Running
Events:
  Type    Reason       Age    From               Message
  ----    ------       ----   ----               -------
  Normal  Create       4m16s  postgres-operator  Started creation of new cluster resources
  Normal  Endpoints    4m16s  postgres-operator  Endpoint "default/acid-minimal-cluster" has been successfully created
  Normal  Services     4m16s  postgres-operator  The service "default/acid-minimal-cluster" for role master has been successfully created
  Normal  Services     4m16s  postgres-operator  The service "default/acid-minimal-cluster-repl" for role replica has been successfully created
  Normal  Secrets      4m14s  postgres-operator  The secrets have been successfully created
  Normal  StatefulSet  4m14s  postgres-operator  Statefulset "default/acid-minimal-cluster" has been successfully created
  Normal  StatefulSet  3m26s  postgres-operator  Pods are ready

```

**Comment by FxKu** [comment]
Very nice. Tested it and it works now. Nice new feature. Thanks a lot @frittentheke

**Comment by FxKu** [comment]
:+1:

**Comment by sdudoladov** [comment]
:+1:

File: pkg/cluster/cluster.go [hunk 1]
```diff
@@ -1136,6 +1163,7 @@ func (c *Cluster) Switchover(curMaster *v1.Pod, candidate spec.NamespacedName) e
 	// close the label waiting channel no sooner than the waiting goroutine terminates.
 	close(podLabelErr)
 
+	c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Switchover", "Switchover from %q to %q FAILED: %v", curMaster.Name, candidate, err)
```
mwpeterson, line 1166: It seems this should be wrapped in `if err != nil {}` 
`FAILED: <nil>` is _technically_ correct, but unexpected when the switchover was successful

mwpeterson: Opened issue #1134

PR #901, opened 2020-04-06
Fix val docs

ReSearchITEng:
fix config map values in values.yaml
documentation fixes

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -584,7 +584,10 @@ don't know the value, use `103` which is the GID from the default spilo image
 OpenShift allocates the users and groups dynamically (based on scc), and their
 range is different in every namespace. Due to this dynamic behaviour, it's not
 trivial to know at deploy time the uid/gid of the user in the cluster.
-This way, in OpenShift, you may want to skip the spilo_fsgroup setting.
+This way, in OpenShift, you want to skip the spilo_fsgroup setting.
+
+Note: instead of spilo_fsgroup in the Operator configmMap, we can use 
+`spiloFSGroup` which is defined per pg cluster deployment.
 
 Upload the cert as a kubernetes secret:
 ```sh
--- a/docs/user.md
+++ b/docs/user.md
@@ -584,10 +584,8 @@ don't know the value, use `103` which is the GID from the default spilo image
 OpenShift allocates the users and groups dynamically (based on scc), and their
 range is different in every namespace. Due to this dynamic behaviour, it's not
 trivial to know at deploy time the uid/gid of the user in the cluster.
-This way, in OpenShift, you want to skip the spilo_fsgroup setting.
-
-Note: instead of spilo_fsgroup in the Operator configmMap, we can use 
-`spiloFSGroup` which is defined per pg cluster deployment.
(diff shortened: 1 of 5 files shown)


ReSearchITEng:
@FxKu this is fixing some pooler params as well as adding promised clarifications in the docs.

FxKu:
Thanks @ReSearchITEng. Do these values really need to be quoted? We also don't do it for other numeric fields. But maybe connectionPooler fields differ here. Need to check. The rest is fine.

ReSearchITEng:
> Thanks @ReSearchITEng. Do these values really need to be quoted? We also don't do it for other numeric fields. But maybe connectionPooler fields differ here. Need to check. The rest is fine.

Yes, helm3 install fails due to unquoted values in these 2 places.
Rest of the helmchart is fine, only these newly added values broke the installation.

docs/user.md
```diff
@@ -584,7 +584,10 @@ don't know the value, use `103` which is the GID from the default spilo image
 OpenShift allocates the users and groups dynamically (based on scc), and their
 range is different in every namespace. Due to this dynamic behaviour, it's not
 trivial to know at deploy time the uid/gid of the user in the cluster.
-This way, in OpenShift, you may want to skip the spilo_fsgroup setting.
+This way, in OpenShift, you want to skip the spilo_fsgroup setting.
+
+Note: instead of spilo_fsgroup in the Operator configmMap, we can use 
+`spiloFSGroup` which is defined per pg cluster deployment.
```
FxKu, line 590:
```suggestion
Therefore, instead of using a global `spilo_fsgroup` setting, use the `spiloFSGroup` field
per Postgres cluster.```

File: manifests/complete-postgres-manifest.yaml [hunk 2]
```diff
@@ -125,5 +125,11 @@ spec:
...
     privateKeyFile: "tls.key"
     caFile: ""  # optionally configure Postgres with a CA certificate
+# file names can be also defined with absolute path, and will no longer be relative
+# to the "/tls/" path where the secret is being mounted by default.
 # When TLS is enabled, also set spiloFSGroup parameter above to the relevant value.
 # if unknown, set it to 103 which is the usual value in the default spilo images.
+# Instead of spiloFSGroup, one can define in operator configMap spilo_fsgroup, 
+# automating it for all pg clusters deployments.
+# In Openshift, there is no need to set spiloFSGroup/spilo_fsgroup, as it's being
+# set dynamically at deploytime.
```
FxKu, line 135: I think you can remove this part to not make this manifest too long. It's documented nicely in the user docs now.

FxKu:
Ah, because these fields are pointers. Good to know and thanks for finding it. I added two minor comments. After that we can merge it.

manifests/complete-postgres-manifest.yaml
ReSearchITEng: Left only one short note. Rest was removed.

ReSearchITEng:
fixes #911

FxKu:
:+1:

sdudoladov:
:+1:

FxKu:
Thanks @ReSearchITEng for another contribution :smiley:

PR #903, opened 2020-04-07
Re-create pods only if all replicas are running

sdudoladov:
Operator currently kills pods too eagerly, for example, when re-init is still running.

Here a sequence of events leading to the problem:

1. Re-init starts on the replica
2. Rolling update comes 
3. Operator kills a replica pod being re-initialized almost immediately because it selects for recreation [all pods belonging to a PG cluster](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/util.go#L388) (called from [here](https://github.com/zalando/postgres-operator/blob/66f2cda87fb041a6cc0151d2b442aad2c9755075/pkg/cluster/pod.go#L299)) and later skips [only the master](https://github.com/zalando/postgres-operator/blob/66f2cda87fb041a6cc0151d2b442aad2c9755075/pkg/cluster/pod.go#L319)

Re-doing re-init on the same pod later suffers from the same problem because the rolling upgrade may not complete. For instance, when a replica on pod -0 is being constantly killed but pod-1 is the yet-to-be-updated master. 

The current workaround for this problem is to manually shut down the operator, let the re-init complete and then start the operator again.

The PR fixes this by asking Patroni to confirm there is no PG replicas in  in the `creating-replica` state before deleting any pod. If that check fails, pod re-creation is postponed until next `Sync` .

@FxKu can you please link related issues ?


sdudoladov:
As of commit `84dc06bfb755482c41ee1ca096f52436b92e0ac7`, this patch prevents incorrect termination of replicas in the following scenario (tested manually for now):

1. Get some data to make re-init long enough
  `pgbench -i -s 1000 -n &> pgbench.log # approx.  17 GB`
2. start  re-init on the replica 
 `patronictl reinit $SCOPE replica-name`
3. Cause the rolling upgrade, for example  by updating resources

The patched operator under such circumstances will report update success without actually re-creating the pods. The 1st Sync after reinit terminates will find out incomplete rolling update thanks  to annotations and complete it.

Note Patroni [returns the `503` code](https://github.com/zalando/patroni/blob/cf0c0f8e7c74119e5115e8d75c1cf0ed9fb60842/patroni/api.py#L95) for GET on `replica_ip:8008`, so presence of  
```
 http get response: &{Status:503 Service Unavailable StatusCode:503 Proto:HTTP/1.0 
```
in the logs is not an error in this case.

FxKu:
:+1:

sdudoladov:
:+1:

PR #906, opened 2020-04-07
Prevent superuser from being a connection pool user

erthalion:
It's not supported, neither it's a best practice. Also fix potential
null pointer access.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -882,7 +882,11 @@ func (c *Cluster) initSystemUsers() {
 		// and second it's a bad practice.
 		username := c.OpConfig.ConnectionPooler.User
 
-		if c.Spec.ConnectionPooler.User != c.OpConfig.SuperUsername {
+		isSuperUser := c.Spec.ConnectionPooler.User == c.OpConfig.SuperUsername
+		isProtectedUser := c.shouldAvoidProtectedOrSystemRole(
+			c.Spec.ConnectionPooler.User, "connection pool role")
+
+		if !isSuperUser && !isProtectedUser {
 			username = util.Coalesce(
 				c.Spec.ConnectionPooler.User,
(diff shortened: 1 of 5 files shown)


pkg/controller/operator_config.go
```diff
@@ -169,6 +169,11 @@ func (c *Controller) importConfigurationFromCRD(fromCRD *acidv1.OperatorConfigur
 		fromCRD.ConnectionPooler.User,
 		constants.ConnectionPoolerUserName)
 
+	if result.ConnectionPooler.User == result.SuperUsername {
+		msg := "Connection pool user is not allowed to be the same as super user, username: %s"
+		panic(fmt.Errorf(msg, result.ConnectionPooler.User))
+	}
```
FxKu, line 175: isn't it validated in `config.go` anyway?

FxKu:
:+1:

pkg/cluster/cluster.go
```diff
@@ -877,9 +877,16 @@ func (c *Cluster) initSystemUsers() {
...
 
-		username := util.Coalesce(
-			c.Spec.ConnectionPooler.User,
-			c.OpConfig.ConnectionPooler.User)
+		// Using superuser as pooler user is not a good idea. First of all it's
+		// not going to be synced correctly with the current implementation,
+		// and second it's a bad practice.
+		username := c.OpConfig.ConnectionPooler.User
+
+		if c.Spec.ConnectionPooler.User != c.OpConfig.SuperUsername {
```
sdudoladov, line 885: Does it make sense to avoid names from `protected_role_names` and `replication_username` as well ?

erthalion: Good point. Yes, I believe it makes sense.

sdudoladov:
:+1:

FxKu:
:+1:

Issue #907, opened 2020-04-07
Disable WAL archiving on cluster


spohner:
We regularly create temporary database clusters to do computations and data processing. It would be great to turn off WAL archiving and uploading to S3 for these types of workloads. We do not need backups for these and disabling it will reduce costs.

Will submit a PR adding a disableWALArchiving option in the cluster-spec allowing to override the WAL_S3_BUCKET environment variable and setting it to blank. This will tell Spilo to disable the WAL archiving.

spohner:
Related PR: https://github.com/zalando/postgres-operator/pull/908

Jan-M:
If you need this already, you could build a new Spilo Docker Image with setting the WAL_S3_Bucket path to empty and using this in the manifest.

We will look into the PR nontheless. Also in regards to preventable errors.

spohner:
Maintaining a fork with this change atm, and seems to work just fine. Anyway thanks for the tip, @Jan-M!

PR #908, opened 2020-04-07
Disable WAL archiving

spohner:
This PR adds a disableWALArchiving option in the cluster-spec allowing to override the WAL_S3_BUCKET environment variable and setting it to blank. This will tell Spilo to disable the WAL archiving.

Relevant Spilo code:
https://github.com/zalando/spilo/blob/01a3a99db55c6f4e38a29528fee4c6c7619d719a/postgres-appliance/scripts/configure_spilo.py#L999

https://github.com/zalando/spilo/blob/01a3a99db55c6f4e38a29528fee4c6c7619d719a/postgres-appliance/scripts/configure_spilo.py#L551

https://github.com/zalando/spilo/blob/01a3a99db55c6f4e38a29528fee4c6c7619d719a/postgres-appliance/scripts/configure_spilo.py#L42

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/postgresql_type.go
+++ b/pkg/apis/acid.zalan.do/v1/postgresql_type.go
@@ -74,6 +74,7 @@ type PostgresSpec struct {
 	ServiceAnnotations    map[string]string           `json:"serviceAnnotations,omitempty"`
 	TLS                   *TLSDescription             `json:"tls,omitempty"`
 	AdditionalVolumes     []AdditionalVolume          `json:"additionalVolumes,omitempty"`
+	EnableWALArchiving    bool                        `json:"enableWALArchiving,omitempty" defaults:"true"`
 
 	// deprecated json tags
 	InitContainersOld       []v1.Container `json:"init_containers,omitempty"`
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -657,7 +657,7 @@ func (c *Cluster) generatePodTemplate(
 }
 
 // generatePodEnvVars generates environment variables for the Spilo Pod
-func (c *Cluster) generateSpiloPodEnvVars(uid types.UID, spiloConfiguration string, cloneDescription *acidv1.CloneDescription, standbyDescription *acidv1.StandbyDescription, customPodEnvVarsList []v1.EnvVar) []v1.EnvVar {
+func (c *Cluster) generateSpiloPodEnvVars(uid types.UID, spiloConfiguration string, cloneDescription *acidv1.CloneDescription, standbyDescription *acidv1.StandbyDescription, enableWALArchiving bool, customPodEnvVarsList []v1.EnvVar) []v1.EnvVar {
 	envVars := []v1.EnvVar{
 		{
 			Name:  "SCOPE",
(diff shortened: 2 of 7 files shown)


spohner:
Resolves #907

spohner:
Did a rebase to align with current master. Any comments on this PR by the maintainers?

FxKu:
@spohner thanks for you contribution. We have this fashion to name our flags "Enable..." and I would like to stick to that. The default should then be `true` which requires the *bool type to distinguish between `false` and `nil` setting. Could you change the CRD validation in the helm chart, too?

spohner:
Did the changes that you requested, @FxKu. Let me know if there is anymore I can do to get this merged.

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -69,6 +69,7 @@ type PostgresSpec struct {
 	ServiceAnnotations    map[string]string           `json:"serviceAnnotations"`
 	TLS                   *TLSDescription             `json:"tls"`
 	AdditionalVolumes     []AdditionalVolume          `json:"additionalVolumes,omitempty"`
+	EnableWALArchiving    *bool                       `json:"enableWALArchiving,omitempty"`
```
FxKu, line 72:
```suggestion
	EnableWALArchiving    *bool                       `json:"enableWALArchiving,omitempty" defaults:"true"``
```

FxKu:
Thanks. You have to update the generated code. Simply run `./hack/update-codegen.sh`. It should update [this](https://github.com/zalando/postgres-operator/blob/master/pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go) file.

MaxFedotov:
Hello @FxKu, @spohner. Are there any plans to continue with this PR? This looks like a very useful thing.

derekjanni:
@spohner @FxKu I second that, very useful feature here, would be cool to make it official

spohner:
Revisited this one and improved it. The previous attempt did not work when editing enableWALArchiving after the cluster was created. Now this should also work. @FxKu

fourstepper:
Would it be considerably difficult to make it possible to configure backup paths/locations per cluster?

Menzorg:
We really need it. Wish it be merged soon.

pkg/cluster/k8sres.go
```diff
@@ -769,7 +769,7 @@ func (c *Cluster) generateSpiloPodEnvVars(uid types.UID, spiloConfiguration stri
 		envVars = append(envVars, customPodEnvVarsList...)
 	}
 
-	if c.OpConfig.WALES3Bucket != "" {
+	if c.OpConfig.WALES3Bucket != "" && enableWALArchiving {
```
machine424, line 772: Unless I'm mistaken, not setting `WAL_S3_BUCKET` will not enable physical backups as well (https://github.com/zalando/spilo/blob/2.0-p7/postgres-appliance/scripts/configure_spilo.py#L862) so I think `enableWALArchiving` should be more generic.

machine424: And shouldn't we do the same for `c.OpConfig.WALGSBucket` as well?

machine424:
How can I help to get this merged? Thanks.

pawanku2:
Hi , Is this PR merged?

ibotty:
What's the status on that PR?  It looks very useful.

modzilla99:
Any news?

FxKu:
With #1794 merged we can close this PR. With the next release you will be able to disable archiving by overwriting WAl location to an empty string.

FxKu:
Thanks again @spohner for bringing it up. It's def. useful if you can turn off archiving for single clusters, but not all. Hope the new release will help you then.

spohner:
Great news! Thanks.

Menzorg:
Thanks!

mug3n451:
Hi, 
i'm trying to disable the WAL in the way described this thread, but even if i set, as a env in the postgresql's cluster manifest, WAL_S3_BUCKET to empty string, the wal at the first installtion is enabled and work. 
May i do something wrong? 
thanks

PR #916, opened 2020-04-14
Add option to exclude tables from logical backup

### LINKED ISSUE #917: Need option to exclude tables from logical backup
We have some large cache tables in our DB that we want to exclude from the logical backup.

mseiwald:
Fixes #917


mseiwald:
I'm not sure why the e2e test fails..

PR #918, opened 2020-04-15
Tls using additionalVolumes of pr736

ReSearchITEng:
make tls(pr798 of @zimbatm ) use additionalVolumes capability from pr736 (of @seuf @Thierry )
(as suggested by @frittentheke and @FxKu )
in 
https://github.com/zalando/postgres-operator/pull/736#discussion_r406226704
It also adds capability to hold ca.crt in a different secret.


FxKu:
Thanks @ReSearchITEng for following up on my suggestion from #736 :smiley: 
Can you rebase and also run `./hack/update-codgen.sh`?

docs/reference/cluster_manifest.md
```diff
@@ -435,5 +435,12 @@ Those parameters are grouped under the `tls` top-level key.
   client connects with `sslmode=verify-ca` or `sslmode=verify-full`.
   Default is empty.
 
+* **caSecretName**
+  By setting the `caSecretName` value, the ca certificate file defined by the
+  `caFile` will be fetched from this secret instead of `secretName` above.
+  This secret has to hold a file with that name in its root.
```
zimbatm, line 441: This is not strictly necessary as the user can provide an absolute path to `caFile` and point into the additional volume directly. The `caFile` only gets expanded with the `/tls` prefix if the path is relative.

ReSearchITEng: right, it could be done separately using the new additionalVolumeMounts capability and point to it (which is actually used in background).
Still, following the patterns for tls in similar projects, (therefore align), having a ca secret is might be more intuitive.
FYI, your nice ensurePath function has been reused, this time on top of /tlsca path.
btw, if the ca secret is not provided, the rest of functionality remains as is.

zimbatm: Right, the extra code complexity might be worth it if it makes things easier for the user. I am not in a position to judge that.

ReSearchITEng:
superseded by pr #920

Thierry:
Hello folks.

Please note that not all "Thierry" people in the world are using account @Thierry (mine) :)
I have nothing to do with this repository.

Thanks

PR #920, opened 2020-04-15
make tls pr798 use additionalVolumes capability from pr736

ReSearchITEng:
make tls(pr798 of @zimbatm ) use additionalVolumes capability from pr736 (of @seuf @Thierry )
(as suggested by @frittentheke and @FxKu )
in
#736 (comment)
It also adds capability to hold ca.crt in a different secret.
and codegen.sh


FxKu:
:+1:

sdudoladov:
:+1:

PR #922, opened 2020-04-16
Prevent empty syncs of connection pool

erthalion:
There is a possibility to pass nil as one of the specs and an empty spec
into syncConnectionPooler. In this case it will perfom a syncronization
because nil != empty struct. Avoid such cases and make it testable by
returning list of syncronization reasons on top together with the final
error.


Jan-M:
:+1:

erthalion:
:+1:

PR #928, opened 2020-04-23
Allow nodePort value for the postgres-operator-ui service

marcusportmann:
Added support for specifying a nodePort value for the postgres-operator-ui service when the service type is NodePort.

### DIFF
--- a/charts/postgres-operator-ui/templates/service.yaml
+++ b/charts/postgres-operator-ui/templates/service.yaml
@@ -11,7 +11,7 @@ spec:
   ports:
     - port: {{ .Values.service.port }}
       targetPort: 8081
-      {{- if .Values.service.nodePort }}
+      {{- if and (eq .Values.service.type "NodePort") .Values.service.nodePort }}
       nodePort: {{ .Values.service.nodePort }}
       {{- end }}
       protocol: TCP
--- a/charts/postgres-operator-ui/templates/service.yaml
+++ b/charts/postgres-operator-ui/templates/service.yaml
@@ -11,6 +11,9 @@ spec:
   ports:
     - port: {{ .Values.service.port }}
       targetPort: 8081
+      {{- if .Values.service.nodePort }}
+      nodePort: {{ .Values.service.nodePort }}
+      {{- end }}
       protocol: TCP
   selector:
     app.kubernetes.io/instance: {{ .Release.Name }}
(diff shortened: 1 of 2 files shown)


FxKu:
I think I would rather add the `nodePort` field to the values file and check if the service type is set to "NodePort" instead of checking if `nodePort` field is set.

marcusportmann:
Ok, do you have a preferred value for the node port? I will rework the change to what you suggested.

FxKu:
Could you still add the `nodePart` field in the values.yaml (and values-crd.yaml) without a value? Would it then receive a random port?

marcusportmann:
I have added a nodePort field to the values.yaml file with no value. This is for the postgres-operator-ui chart so there is no values-crd.yaml. If no value is specified for this field a random port is selected. If a value is specified for this field it is used as the NodePort port.

FxKu:
Will the `nodePort` field ignored when the service is of type `ClusterIP`

marcusportmann:
Adding the nodePort field to the values.yaml file causes the Travis CI build to fail. I attempted to remove the conditional addition of the field in the Helm chart to remedy this but the build unfortunately still fails. Both scenarios work for me when I test the operator locally. I will need to dig into why this is causing the build to fail. I am considering reverting my changes in the interim since it feels wrong to add a field that is conditional to the values.yaml file when it is not applicable i.e. the type for the service is not NodeType. Perhaps a commented out value would be a better option to indicate the field does exist.

FxKu:
The build error has nothing to do with this PR. Besides, the UI isn't really covered in our tests.
It's fine for me to have the blank nodePort field in the manifest. If it's ignored for ClusterIP service than we can save the conditional test in the template. I like that.

marcusportmann:
Are you happy with having the nodePort field commented out in the values.yaml file with a comment describing its usage?

FxKu:
Why not add `if type != clusterIP` to the service template, similar to what you had before?

marcusportmann:
Because the field is only valid if the type is NodePort. Removing the check in the service.yaml demonstrated this as the deployment failed. I believe there is a difference between an optional field, for which an empty value is valid, and a conditional field that is only valid as a result of the value of some other field. Having it commented out with a comment explaining its usage makes it clear that just setting the value is not sufficient and that the type also needs to be set to NodePort.

FxKu:
:+1:

sdudoladov:
:+1:

sdudoladov:
@marcusportmann thank you for the contribution :)

Issue #913, opened 2020-04-08
Allow propagating annotations to the StatefulSet


hjacobs:
We use the `deployment-time` annotation in Zalando to denote the "last update" time.
This annotation is automatically injected to all resources, incl. the Postgres CRD. The problem: the annotation is not passed to the StatefulSet which is therefore not considered to be "updated".
This creates an issue in combination with [kube-downscaler](https://github.com/hjacobs/kube-downscaler/).

My request: allow configuring certain annotations to be "propagated" to the StatefulSet.
Example annotations I would want to propagate:

* `deployment-time`
* `downscaler/exclude`
* `downscaler/exclude-until`
* `downscaler/uptime`
* `downscaler/downtime`

My implementation proposal would be:

* add a configuration option `statefulset_propagate_annotations` (for the ConfigMap) to specify a list of patterns for annotations to propagate (e.g. `deployment-time,downscaler/.*`)
* change the operator to take all annotations (filtered by the pattern list above) from the CRD and apply them on the StatefulSet

This allows users to configure the underlying StatefulSet via the CRD.

hjacobs:
As far as I can see, this line needs to be changed and dynamically build annotations (instead of hardcoded map with one entry): https://github.com/zalando/postgres-operator/blob/4dee8918bd45e4356e10acf15b9f71bf0a43f9bd/pkg/cluster/k8sres.go#L1138

FxKu:
Only for stateful set? For services and pods we already provide annotation options (local per manifest: `podAnnotations`, `service_annotations`, global per config: `custom_pod_annotations`, `custom_service_annotations`). It's not hard to add yet another option :smile:

hjacobs:
@FxKu I would like to avoid a custom config (like `podAnnotations`) as the annotation `deployment-time` is automatically injected by the CI/CD pipeline. I don't want to add some custom logic just for the Postgres CRD. Regarding the `Service` resource: yes, I think the annotation should also be propagated to the `Service`. I would see the Pods a bit different as they are managed by the StatefulSet itself, so they don't map 1-1 to the Postgres CRD and also don't need the `deployment-time` or `downscaler/*` annotations.

Jan-M:
Seems like something we can add, as much as I am not immediately a fan of scaling up and down Postgres stateful sets. But it works and works also well for single pod clusters e.g. in test environments outside of office hours.

[...]

PR #932, opened 2020-04-24
Propagate annotations to the StatefulSet

RafiaSabih:
To help with [this](https://github.com/zalando/postgres-operator/issues/913) reported issue.


sdudoladov:
@RafiaSabih we will need to test if adding such an annotation to the stafeful set causes a rolling update (it should not)

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -72,6 +72,7 @@ type KubernetesMetaConfiguration struct {
 	EnablePodAntiAffinity      bool                `json:"enable_pod_antiaffinity,omitempty"`
 	PodAntiAffinityTopologyKey string              `json:"pod_antiaffinity_topology_key,omitempty"`
 	PodManagementPolicy        string              `json:"pod_management_policy,omitempty"`
+	StatefulsetPropAnnotations []string            `json:"statefulset_propagate_annotations,omitempty"`
```
Jan-M, line 75: Prop / Propagate? Is it worth keeping it short?

Jan-M:
2. is out of scope for now.

pkg/cluster/sync.go
```diff
@@ -330,6 +330,20 @@ func (c *Cluster) syncStatefulSet() error {
...
 			}
 		}
+		ToPropagateAnnotations := c.OpConfig.StatefulsetPropAnnotations
+		PgCRDAnnotations := c.Postgresql.ObjectMeta.GetAnnotations()
+		annotations := make(map[string]string)
+
+		if ToPropagateAnnotations != nil && PgCRDAnnotations != nil {
+			for _, anno := range ToPropagateAnnotations {
+				for k, v := range PgCRDAnnotations {
+					if k == anno {
```
Jan-M, line 340: keep in mind this needs to be prefix pattern I dont see it right now.

pkg/cluster/k8sres.go
```diff
@@ -1148,12 +1148,14 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
 
+	annotations = make(map[string]string)
+
 	statefulSet := &appsv1.StatefulSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        c.statefulSetName(),
 			Namespace:   c.Namespace,
 			Labels:      c.labelsSet(true),
-			Annotations: map[string]string{rollingUpdateStatefulsetAnnotationKey: "false"},
+			Annotations: annotations,
```
Jan-M, line 1158: rollingUpdate key seems to have gotten lost

pkg/cluster/sync.go
```diff
@@ -354,6 +357,26 @@ func (c *Cluster) syncStatefulSet() error {
...
+// based on the annotations in postgres CRD
+func (c *Cluster) PropagateAnnotationsToStatefulsets(annotations map[string]string) map[string]string {
+	ToPropagateAnnotations := c.OpConfig.StatefulsetPropagateAnnotations
+	PgCRDAnnotations := c.Postgresql.ObjectMeta.GetAnnotations()
+
+	if ToPropagateAnnotations != nil && PgCRDAnnotations != nil {
+		for _, anno := range ToPropagateAnnotations {
+			for k, v := range PgCRDAnnotations {
+				matched, err := regexp.MatchString(anno, k)
+				if err == nil && matched {
```
Jan-M, line 370: Making it a bit more complex: Since regex can yield an error operator should validate regex on startup or log an error during comparison. Silent error here may hide problem too long.

e2e/tests/test_e2e.py
```diff
@@ -694,13 +728,23 @@ def get_service_type(self, svc_labels, namespace='default'):
...
+            if len(svc.metadata.annotations) < len(annotations):
                 return False
             for key in svc.metadata.annotations:
                 if svc.metadata.annotations[key] != annotations[key]:
                     return False
         return True
 
+    def check_statefulset_annotations(self, sset_labels, annotations, namespace='default'):
+        ssets = self.api.apps_v1.list_namespaced_stateful_set(namespace, label_selector=sset_labels, limit=1).items
+        for sset in ssets:
```
sdudoladov, line 740: Why loop if only one stateful set is fetched ?
It is effectively `sset = self.api.apps_v1.list_namespaced_stateful_set(namespace, label_selector=sset_labels, limit=1).items[0]`

e2e/tests/test_e2e.py
```diff
@@ -694,13 +728,23 @@ def get_service_type(self, svc_labels, namespace='default'):
...
+            if len(sset.metadata.annotations) < len(annotations):
```
sdudoladov, line 741: theses lines can be imo written slightly clearer:
```python
for key, value in annotations:
  if key not in sset.metadata.annotations or sset.metadata.annotations[key] != value:
    return False
```
the first condition avoids the `KeyError` if annotations indeed has not been propagated to the stateful set

e2e/tests/test_e2e.py
```diff
@@ -489,6 +489,40 @@ def test_service_annotations(self):
...
+        k8s.api.custom_objects_api.patch_namespaced_custom_object(
+            "acid.zalan.do", "v1", "default", "postgresqls", "acid-minimal-cluster", pg_crd_annotations)
+
+        k8s.delete_operator_pod()
+
+        annotations = {
+            "deployment-time": "2020-04-30 12:00:00",
+            "downscaler/downtime_replicas": "0",
+        }
+        self.assertTrue(k8s.check_statefulset_annotations(cluster_label, annotations))
```
sdudoladov, line 524: It would be really helpful to log here the error message with expected and actual annotations

pkg/cluster/resources.go
```diff
@@ -853,3 +853,24 @@ func (c *Cluster) updateConnectionPoolerDeployment(oldDeploymentSpec, newDeploym
...
+	}
+	result, err := c.KubeClient.Deployments(c.ConnectionPooler.Deployment.Namespace).Patch(
+		context.TODO(),
+		c.ConnectionPooler.Deployment.Name,
+		types.MergePatchType,
+		[]byte(patchData),
+		metav1.PatchOptions{},
+		"")
+	if err != nil {
+		return nil, fmt.Errorf("could not patch conenction pooler annotations %q: %v", patchData, err)
```
sdudoladov, line 872: typo: conenction -> connection

manifests/configmap.yaml
```diff
@@ -94,6 +94,7 @@ data:
   # sidecar_docker_images: ""
   # set_memory_request_to_limit: "false"
   spilo_privileged: "false"
+  # statefulset_propagate_annotation: "deployment-time,downscaler/*"
```
sdudoladov, line 97: typo: `statefulset_propagate_annotation` ->. `statefulset_propagate_annotationS`

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: postgresql-operator-default-configuration
 configuration:
   docker_image: registry.opensource.zalan.do/acid/spilo-cdp-12:1.6-p115
+  # statefulset_propagate_annotation:
```
sdudoladov, line 7: typo: `statefulset_propagate_annotationS`

FxKu:
:+1:

RafiaSabih:
:+1:

PR #933, opened 2020-04-24
fix typo in additionalVolume struct

siku4:
related to PR #736:

It is not possible to specify a `volumeSource` based on the given cluster manifests examples and the documentation: The name of the json field within the `AdditionalVolume` struct is `volume`. The name specified in the CRD is `volumeSource`: https://github.com/zalando/postgres-operator/blob/3c91bdeffadb5ec736a63548b5ffa08517c59de8/manifests/postgresql.crd.yaml#L45-L48

### DIFF
--- a/manifests/complete-postgres-manifest.yaml
+++ b/manifests/complete-postgres-manifest.yaml
@@ -31,28 +31,28 @@ spec:
     size: 1Gi
 #    storageClass: my-sc
   additionalVolumes:
-    - name: data
-      mountPath: /home/postgres/pgdata/partitions
-      targetContainers:
-        - postgres
-      volumeSource:
-        PersistentVolumeClaim:
-          claimName: pvc-postgresql-data-partitions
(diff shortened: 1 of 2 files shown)


FxKu:
Ah, good catch. Unfortunately, I can see that the multi_namespace tests is failing, which uses the complete manifest. Probably because the mentioned configmap and PVC does not exists. Can you comment out the first [two examples](https://github.com/zalando/postgres-operator/blob/3c91bdeffadb5ec736a63548b5ffa08517c59de8/manifests/complete-postgres-manifest.yaml#L34) for additionalVolumes? And then maybe move the emptyDir volume to the top. I think that one should work.

FxKu:
:+1:

sdudoladov:
:+1:

sdudoladov:
@siku4 welcome to our community and thanks for the contribution !

Issue #498, opened 2019-02-22
Add ownerReference to managed entities


aermakov-zalando:
If I delete a database, resources created by it (services, statefulsets, etc.) should be deleted automatically. This can be easily implemented by adding the Postgres resource as an `ownerReference` (with `controller: true`). Kubernetes will then automatically delete the resources once the master database resource is gone.

sdudoladov:
The operator normally deletes [k8s resources it creates](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L590) . Can you please provide more details on the problems with deletion you encountered ?

aermakov-zalando:
The linked function doesn't delete services, for one. Additionally, the operator can always be killed in the middle of `Delete()`, leaving dangling resources. It should also be easier to maintain since you won't have to write the cleanup code or test it, Kubernetes will always clean up the resources for you no matter what.

sdudoladov:
>  The linked function doesn't delete services, for one.

[service are deleted](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L623), but the point overall is valid

aermakov-zalando:
Isn't there also a `-config` service that should be deleted as well?

Jan-M:
Yeah, those come from the headless services via Patroni.

Guess we need some thinking here, who cleans up what and do we have any expectation on what deletes what and how one thing prevents mistakes. Let's think about this cascade or how to make it safe and clean.

dmayle:
The reason why I created the link #886 was mostly because of external tooling that we have for cluster monitoring and maintenance.  I know that the operator will clean up after itself, but having the link allows us to better manage the state of our clusters.

bchrobot:
> ... because of external tooling that we have for cluster monitoring and maintenance. ... having the reference allows us to better manage the state of our clusters.

This is true for us as well

dave08:
Apart from internal management, this would be very important for those using projects like ArgoCD for GitOps, since that's the way they know which CRD caused resources to be deployed... otherwise in the UI it looks pretty ugly to track "orphaned" resources (that ArgoCD thinks were deployed manually, but are actually managed in the Git repo with the custom CRD...).

frittentheke:
[...]

Issue #450, opened 2019-01-11
use finalizers avoid losing deletions


redbaron:
When Postgres CR is deleted it triggers deletion routine, which might not end completely due to crash, misconfiguration or intermittent Kubernetes errors. To be able to finish cleanup on restart, Postgres CR should remain present up until deletion is completed.

Kubernetes solutions for that is to use [finalizers](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers) where resources is still present and queryable until all finalizers are completed. This would allow operator to continue cleanup routine on restart

FxKu:
related to #387

frittentheke:
@redbaron  thanks for opening this ... I was just about to suggest this myself ;-)

Another use case for finalizers would a final basebackup or logical backup before the cluster is annihilated by the operator for good -  Just like AWS does with their final snapshot upon deletion of and RDS instance: https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_DeleteInstance.html

Issue #387, opened 2018-09-17
Prevent deletion of resources created on behalf of the operator. 


zerg-junior:
We've had cases when users manually deleted pods/statefulsets/etc of a PG cluster only to found out them to be re-created later. This leads to users' confusion and extra debugging overhead for maintenance teams. We should experiment with [k8s finalizers](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers) available since k8s v1.11 to prevent such deletions. Things to watch out are properly setting the deletion timestamp and effectively informing users about why their deletions of operator managed resources do not succeed.

A more sophisticated option is to use k8s admission webhooks.

FxKu:
Closing this tickets since finalizers support has been added.

PR #941, opened 2020-04-28
Use finalizers avoid losing deletions [starts on #450]

frittentheke:
Currently the Operator fully depends on the event received when running. While it has a sync loop iterating over clusters again and again in order to sync or repair them, if a delete event it lost the CR is deleted, but the resources might not, leaving the resources such as StatefulSets or Secrets dangling. Those resources a user usually does not have to interact with, but then it's his responsibility to clean those up.
To protect a CR from being deleted before proper cleanup has happend Finalizers are used (https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers) and best-practice.

This PR works towards issue https://github.com/zalando/postgres-operator/issues/450 as it adds a finalizer to the CR when a cluster is first created and will only remove it when it's resources are deleted, then causing the CR to be garbage collected by Kubernetes.

A deleted postgres resources will then only be "marked" with a deletion timestamp, but remains (with deletion timestamp set). To reconcile cases where the deletion event was "missed" (operator not running) I added a check for a deletion timestamp to exist when doing a sync loop (or initial sync of all clusters) if so the operator will then trigger the deletion.

I took the liberty to give the Delete function a returnable error, but I only added one condition (finalizer removal failed) as failing / error condition. To fully tackle https://github.com/zalando/postgres-operator/issues/450 (delete event missed) but even more importantly https://github.com/zalando/postgres-operator/issues/498 (resource cleanup incomplete/failing) full error handling should be added to the various resource deletions. If the operator cannot cleanup all the resources it created, it should report this (logs, events to the cr) but NOT remove the finalizer / CR until it can.
This will prevent a new CR (as the old one will not be deleted with the Finalizer on) from being created with the same (cluster) name causing conflicts with the not cleaned up resources.

I don't know if this relates at all to https://github.com/zalando/postgres-operator/issues/387, but @sdudoladov  mentioned Finalizers there - maybe this PR also helps providing initial scaffolding.


frittentheke:
@FxKu  on a local build testing this we ran into issues with null pointers on delete functions.
This is why I added commit https://github.com/zalando/postgres-operator/pull/941/commits/f36bba38bb4dcf1284a565cbcb95e86645e9b8c2 to make all those map accessing functions work simliar and to not crash on map entries for Services being "nil".

frittentheke:
I added some more commits to close the loop around the Finalizer:

1. https://github.com/zalando/postgres-operator/pull/941/commits/42ffefa3a7df6c12ee90f11f6b7d07276d087b77 makes the `deleteXYZresource()` not treat ResourceNotFound as an actual error

1. https://github.com/zalando/postgres-operator/pull/941/commits/9690c90adb678f4c66a81d48ddae9ca7bc374c8f sends out events of (actual) errors to the postgres resource so the user knows there are issues removing things

1. https://github.com/zalando/postgres-operator/pull/941/commits/795764e60799f3e6b27a129634fc94fd0376dd28 causes the Finalizer to not be removed until all resources could successfully be deleted (or don't exist anyways).


As for the `anyErrors` boolean, this is the smallest footprint in my opinion. Certainly straightening up the error handling and propagation a lot more is possible, as in collecting all the errors and then handling them jointly in the calling function. But making changes to the whole error logic is always possible, I wanted this PR to be about improving the way the operator carefully ensures that all created resources are deleted again and to use the Finalizer to a) allow for retries (next sync run)  and b) ensure no delete event is ever missed, even when the operator is not running.

frittentheke:
Two more additions:

1. 8e7c4ed6528b3bff1688ff2990ac19f847b4e631 - When doing a delete of a whole cluster and some resources are just not there (anymore), we should not treat this as an error. Log it and go on on our mission to delete the other things that belong to this cluster to reach a clean state.

1. f8a06fed80974d661dc53811e82f3a07d91c32ec - Maybe a bit late to the game of implementing the Finalizer logic, but when a Finalizer is on a resource there is no delete event received from the Kubernetes API by the operators informer, but rather an update to the resource with the deleteTimestamp set. This commit implements this logic to enable the operator handling of such "deletes" just like a regular deletion.

frittentheke:
@FxKu  are you still considering this PR fpr 1.6? If so, should I rebase it then?

frittentheke:
@FxKu could you kindly give me an indication if I should rebase this PR at some point or if you don't like it enough to even consider merging it?

frittentheke:
@FxKu @CyberDem0n is this PR still in any way in focus to you?

See i.e. https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/ on why finalizers are quite sensible to have.

frittentheke:
@FxKu @CyberDem0n @Jan-M is this PR still in any way in focus to you?

See i.e. https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/ on why finalizers are quite sensible to have.

frittentheke:
@FxKu @CyberDem0n  @Jan-M  may I bug you once more and ask if are you still interested in this change?

FxKu:
@frittentheke sorry to keep you hanging here for so long, because we [fear](https://github.com/zalando/postgres-operator/issues/498#issuecomment-607829097) it could break things. But since we have delete protection on the operator side (and also in our internal admission controller) we should be safe to allow cleanups when there was a delete event. Hope I will find time next week to look at it again.

frittentheke:
Maybe something to look into for 1.9 (https://github.com/zalando/postgres-operator/milestone/12) @FxKu  @Jan-M  ?

bwrobc:
I've rebased this patchset against current master [here](https://github.com/bwrobc/postgres-operator/tree/PR941) for anyone interested.

frittentheke:
> I've rebased this patchset against current master [here](https://github.com/bwrobc/postgres-operator/tree/PR941) for anyone interested.


I did not check out your branch yet ... but could I reset my PR to your kindly done rebase to keep this PR current then?

frittentheke:
> I've rebased this patchset against current master [here](https://github.com/bwrobc/postgres-operator/tree/PR941) for anyone interested.

@bwrobc 
I did not check out your branch yet ... but could I reset my PR to your kindly done rebase to keep this PR current then?

frittentheke:
> > I've rebased this patchset against current master [here](https://github.com/bwrobc/postgres-operator/tree/PR941) for anyone interested.
> 
> @bwrobc I did not check out your branch yet ... but could I reset my PR to your kindly done rebase to keep this PR current then?

@FxKu are you still interested in this addition / PR?
I'd then spend the time to rebase it again ... likely using @bwrobc's  work.

bwrobc:
Yes, please do.

-- 
Privileged and/or confidential information may be contained in this 
message. If you are not the addressee indicated in this message, or 
responsible for delivery of the message to such person, you may not copy or 
distribute this message or any part of its contents, or take any action in 
reliance on it. If you are not the intended recipient, you must destroy 
this message and notify the sender by reply e-mail immediately. Opinions, 
conclusions, and other information in this message that do not relate to 
the official business of Brandwatch shall be understood as neither given 
nor endorsed by Brandwatch.

bwrobc:
> > I've rebased this patchset against current master [here](https://github.com/bwrobc/postgres-operator/tree/PR941) for anyone interested.
> 
> @bwrobc I did not check out your branch yet ... but could I reset my PR to your kindly done rebase to keep this PR current then?

Yes, please feel free

FxKu:
Hey there :smiley: . I have not forgotten you. Before I give this PR another review I have one request. Can we make this feature optional we a global `enable_finalizers` option? Like this everybody can start testing and then using it at their own will. We would feel for comfortable if finalizers are optional and then we can finally merge this PR.

KunJakob:
Just wanted to add my interest in seeing this merged as a quiet spectator for a few months now. This would be such a joy to have in contexts like in https://tilt.dev/ local environment setups. Being able to rapidly tear down and set up a cluster without manual deletion intervention would be a massive boon for us!

frittentheke:
@FxKu @bwrobc @KunJakob  sorry for the delay. I pushed a rebased PR now and added a config option to enable finalizers. I'll have to wait for CI so see if / where I messed up somewhere ... but PTAL.

pkg/cluster/cluster.go
```diff
@@ -260,6 +262,12 @@ func (c *Cluster) Create() (err error) {
 	}()
 
 	c.KubeClient.SetPostgresCRDStatus(c.clusterName(), acidv1.ClusterStatusCreating)
+	if c.OpConfig.EnableFinalizers {
```
FxKu, line 265:
```suggestion
	if c.OpConfig.EnableFinalizers != nil && *c.OpConfig.EnableFinalizers {
```

FxKu:
:+1:

idanovinda:
👍

FxKu:
So finally, we have merged your PR @frittentheke . Thanks for this contribution and all the patience over the last years (!). There are still a few minor things I would change, but I will take it from here.

FxKu:
After adding way more work into it than anticipated (refactoring, adding sync support, writing unit tests and docs) I have to say, I don't see the appeal of using finalizers. At least not with our operator and the way we listen to events. With finalizers there can be a huge delay until the cluster is finally gone. And then it will emit the DELETE event to the queue when the job is already done - so logs look a little confusing now.

Anyway, all of this might soon be history if we switch to K8s controller runtime.

PR #943, opened 2020-04-28
Add support for pooler load balancer

### LINKED ISSUE #940: Add support for pooler LoadBalancer
There are currently options to enable load balancers for master and replica nodes, but that would be cool to support balancing for pooler service too.

Prototik:
Closes #940

### DIFF
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -163,8 +163,8 @@ configLoadBalancer:
   enable_master_load_balancer: false
   # toggles service type load balancer pointing to the replica pod of the cluster
   enable_replica_load_balancer: false
-  # toggles service type load balancer pointing to the pooler pod of the cluster
-  enable_pooler_load_balancer: false
+  # toggles service type load balancer pointing to the master pooler pod of the cluster
+  enable_master_pooler_load_balancer: false
   # defines the DNS name string template for the master load balancer cluster
   master_dns_name_format: "{cluster}.{team}.{hostedzone}"
   # defines the DNS name string template for the replica load balancer cluster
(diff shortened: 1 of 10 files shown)


charts/postgres-operator/values-crd.yaml
```diff
@@ -163,6 +163,8 @@ configLoadBalancer:
   enable_master_load_balancer: false
   # toggles service type load balancer pointing to the replica pod of the cluster
   enable_replica_load_balancer: false
+  # toggles service type load balancer pointing to the pooler pod of the cluster
+  enable_pooler_load_balancer: false
```
Jan-M, line 167: We did not consider the master/replica in the pooler yet, but probably good idea to make this `enable_master_pooler_load_balancer`

Prototik: Wait, pooler works only with master nodes? Shouldn't it delegate write requests to master and read request to all nodes?

Jan-M: This is not possible. This needs to be a conscious decision in your application. Supporting the pooler in front of the replicas will follow.

PGBouncer cannot decide this. Imagine a transaction spanning multiple commands and starting only with reads and ending with a write, where would you point it to.

Prototik: Oh, that makes sense. Renamed.

sdudoladov:
hello @Prototik 

1. Please add example to [the complete Postgres manifest](https://github.com/zalando/postgres-operator/blob/master/manifests/complete-postgres-manifest.yaml#L114)
2. This setting should be grouped with other [connection pooler settings](https://github.com/zalando/postgres-operator/blob/f32c615a531426804d031b16967d69a4ec364c16/pkg/apis/acid.zalan.do/v1/postgresql_type.go#L193) both in he example and the code

Jan-M:
Looks good to me if we get the tiny requested changes in,

charts/postgres-operator/values-crd.yaml
ntcong: pgpool do support routing read-only request to slave.

michaelmcmillan:
Did this never get merged in?

Issue #858, opened 2020-03-09
Documentation no clear on how to clone from S3 backup


walec51:
The documentation suggest that we can create a clone from a S3 bucket but in does not state explicitly in what format should be backup be there:

https://postgres-operator.readthedocs.io/en/latest/user/#how-to-clone-an-existing-postgresql-cluster

one could assume that this is the format used by logical backups and as their documentations states `pg_dumpall`:

https://postgres-operator.readthedocs.io/en/latest/administrator/#logical-backups

but I think that is not the case as in the further documentation I can see that this is a basebackup which would suggest the usage of `pg_basebackup`:

https://postgres-operator.readthedocs.io/en/latest/reference/cluster_manifest/#parameters-defining-how-to-clone-the-cluster-from-another-one

the two facts above would suggest that there logical backups feature is incompatible with the clone feature and I have to exec in to the cluster and do a `pg_basebackup` to S3 my self by hand.

But if that is the case then I don't understand this part of the documentation:

> The uid field is also mandatory. The operator will use it to find a correct key inside an S3 bucket

this would suggest that there is some mechanism or convention is which the S3 bucket should be created but I cannot find any documentation for this.

I would be grateful if someone would explain to me how this cloning functionality should work in detail and help to improve the documentation.

CyberDem0n:
The S3 bucket must contain a backup and WALs made by wal-e or wal-g.

walec51:
what do you mean by "backup"? files produced by a [logical backup](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#logical-backups) or something else?

how to I setup wal-e or wal-g using zalando operator?

FxKu:
@walec51 have a look at the [enviroment](https://github.com/zalando/spilo/blob/master/ENVIRONMENT.rst) variables of Spilo. You can overwrite them with an additional configmap which you can reference in the operator [configuration](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#custom-pod-environment-variables). Some things like S3 bucket or IAM role can already be set in the operator [configuration](https://github.com/zalando/postgres-operator/blob/650b8daf77f35d03b0fbd989908c3dea4a78108f/manifests/postgresql-operator-default-configuration.yaml#L79).

frittentheke:
This issue or discussion seems to be related to https://github.com/zalando/postgres-operator/issues/630 ...

[...]

Issue #597, opened 2019-06-25
Provide way to read config environment vars from secrets instead of configmap


mkjoerg:
If you want to use the backup functionality of WAL-E / WAL-G and you are not running on AWS, currently it is not possible to provide the secrets to the backup bucket in another way then defining a configmap that contains all secrets in plain text for the central Postgres Operator.

Delivering this information in the form of a configmap blocks us from using a GitOps centered approach, where we have tools to persist secrets in Git. Also as far as I noted, as soon as you enter the option in the operator config, that there is a separate configmap, each cluster has to have at least the empty configmap.

I know that there is the possibility to mount secrets for a cloud provider as file like aws account credentials or google ones and reference them in the configmaps for each cluster.

The problem with this approach is if you need direct access to secrets as ENV variables and you need these secrets just for certain environments. 
We, for example, run on OpenStack with additional S3 storage and so we can't make real use of the file-based approach, we just need to set up buckets for the relevant clusters and backup each one individually with own environment variables, where others don't require anything else.

Maybe it would be a good idea to add the Kubernetes feature that reads secrets and places all values as env vars to the Pod ( envFrom )

```yaml
envFrom:
 - secretRef:
    name: <secret-name>
```

This could be provided in a per cluster spec as:

secret_env:

or similar.

This would allow decoupling the configuration from the secrets.

Am I overlooking something or would this be a viable approach?

FxKu:
Closing as #946 got merged. Thanks for bringing it up.

Issue #480, opened 2019-02-04
allow populating pod environment from Secrets


kupson:
It would be nice to populate env variables from secrets e.g. `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` so they won't be exposed in ConfigMaps.

Jan-M:
Is there a reason you cannot use iam/pod iam role for this?

kupson:
I would like to use minio instead of Amazon S3.

kupson:
> Is there a reason you cannot use iam/pod iam role for this?

That would be `kube_iam_role` option?
I would like to use [minio](https://github.com/minio/minio) instead of Amazon S3 and I'm not sure that `kube2iam` could handle it.

erthalion:
Are there any other alternatives, maybe just mount these credentials and let the client library to discover it? If I'm not mistaken, this mentioned in the minio [documentation](https://github.com/minio/minio/blob/bd25f31100f44646cffae99dfec23179366593e2/docs/gateway/s3.md#using-binary-in-ec2)

kupson:
It's the wal-g that needs those environment variables. It would need the volume with credentials mounted so the file [`~postgres/.aws/credentials`](https://github.com/wal-g/wal-g#configuration) will be available.

But all the Pods created by postgres-operator would need the volume and I see no way to create it in the [postgres-operator configuration](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md) or the [postgresql object](https://github.com/zalando/postgres-operator/blob/master/manifests/complete-postgres-manifest.yaml) right now.

Is there a way to add custom volume to add spilo containers and I'm missing it?

erthalion:
Not yet, but probably we can leverage #535 for that purpose?

kupson:
I find #535  suboptimal as I still would need to run some script to turn secrets from Minio format (name: value) into `.aws/credentials`:
```
[default]
aws_access_key_id=EXAMPLE
aws_secret_access_key=EXAMPLEKEY
```
or maybe keep two secrets in sync with same values but different format.
I see environment variables as more flexible solution.

FxKu:
Closing this issue as now supported with #946 being merged.

PR #946, opened 2020-04-28
Allow pod environment variables to also be sourced from a secret

frittentheke:
Currently the Operator allows to configure a ConfigMap which key-value pairs are then added to the pod environent variables (https://postgres-operator.readthedocs.io/en/latest/administrator/#custom-pod-environment-variables).
Depending on your clusters RBAC roles and settings, ConfigMaps are usually accessible quite freely (compared to secrets, which can be restricted independently) and having the value exposed in the pod spec adds to this issue of exposing credentials to access all your valuable data.

This PR adds the same functionality, but using a Secret as a source and references to the keys in the secret instead of copying the key-value pairs - like suggested in the two issues https://github.com/zalando/postgres-operator/issues/480 and https://github.com/zalando/postgres-operator/issues/597

But to allow for a bit more flexibility I changed the order in which the `WAL_` and `LOG_` envVars are applied. This then allows for those variables to be overruled by those with the same key from the referenced Secret (or ConfigMap). This change applies only to this short "whilelist" of variables. There is still protection to not have someone fiddle with the variables required for the operator to configure the core functionality.

With Spilo taking all its configuration and switches from environment variables, this PR intends to allow for more flexibility, such as to set custom S3 buckets. Also just setting custom (AWS) credentials is now possible (see my issues with this in https://github.com/zalando/postgres-operator/issues/858#issuecomment-600541965)

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
+++ b/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
@@ -70,6 +70,7 @@ type KubernetesMetaConfiguration struct {
 	// TODO: use a proper toleration structure?
 	PodToleration              map[string]string   `json:"toleration,omitempty"`
 	PodEnvironmentConfigMap    spec.NamespacedName `json:"pod_environment_configmap,omitempty"`
+	PodEnvironmentSecret       string              `json:"pod_environment_secret,omitempty"`
 	PodPriorityClassName       string              `json:"pod_priority_class_name,omitempty"`
 	MasterPodMoveTimeout       Duration            `json:"master_pod_move_timeout,omitempty"`
 	EnablePodAntiAffinity      bool                `json:"enable_pod_antiaffinity,omitempty"`
--- a/pkg/util/config/config.go
+++ b/pkg/util/config/config.go
@@ -45,6 +45,7 @@ type Resources struct {
 	MinCPULimit             string              `name:"min_cpu_limit" default:"250m"`
 	MinMemoryLimit          string              `name:"min_memory_limit" default:"250Mi"`
 	PodEnvironmentConfigMap spec.NamespacedName `name:"pod_environment_configmap"`
+	PodEnvironmentSecret    string              `name:"pod_environment_secret"`
 	NodeReadinessLabel      map[string]string   `name:"node_readiness_label" default:""`
 	MaxInstances            int32               `name:"max_instances" default:"-1"`
 	MinInstances            int32               `name:"min_instances" default:"-1"`
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -149,6 +149,8 @@ spec:
                   type: string
                 pod_environment_configmap:
                   type: string
+                pod_environment_secret:
+                  type: string
                 pod_management_policy:
                   type: string
                   enum:
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -104,6 +104,8 @@ configKubernetes:
   pod_antiaffinity_topology_key: "kubernetes.io/hostname"
   # namespaced name of the ConfigMap with environment variables to populate on every pod
   # pod_environment_configmap: "default/my-custom-config"
+  # name of the Secret (in cluster namespace) with environment variables to populate on every pod
+  # pod_environment_secret: "my-custom-secret"
 
   # specify the pod management policy of stateful sets of Postgres clusters
   pod_management_policy: "ordered_ready"
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -95,6 +95,8 @@ configKubernetes:
   pod_antiaffinity_topology_key: "kubernetes.io/hostname"
   # namespaced name of the ConfigMap with environment variables to populate on every pod
   # pod_environment_configmap: "default/my-custom-config"
+  # name of the Secret (in cluster namespace) with environment variables to populate on every pod
+  # pod_environment_secret: "my-custom-secret"
 
   # specify the pod management policy of stateful sets of Postgres clusters
   pod_management_policy: "ordered_ready"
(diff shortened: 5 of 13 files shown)


FxKu:
reminds me #481, except this also allows for listing keys. Have you seen it? Maybe things like unit tests could be taken from their addressing the original owner @kupson

frittentheke:
@FxKu  lol ... I event replied to that other PR and toally forgot about it.
I absolutely did not want to make this a race for the "best" PR regarding this feature.

But if you agree that sourcing a secret for environment variables is a sensible addition to the operator (noting again, that Spilo just loves env variables  - just like @kupson said: https://github.com/zalando/postgres-operator/pull/481#issuecomment-507601272)

Also the ability to override the WAL_ and LOG_ parameters are an important addition to be able to let folks configure their own backup targets (we like to call it "bring your own bucket").

I gladly add some unit tests or copy some of @kupson if he does not mind and this PR might get merged.

pkg/util/config/config.go
```diff
@@ -45,6 +45,7 @@ type Resources struct {
 	MinCPULimit             string              `name:"min_cpu_limit" default:"250m"`
 	MinMemoryLimit          string              `name:"min_memory_limit" default:"250Mi"`
 	PodEnvironmentConfigMap spec.NamespacedName `name:"pod_environment_configmap"`
+	PodEnvironmentSecret    string              `name:"pod_environment_secret"`
```
FxKu, line 48: why isn't this of type `spec.NamespacedName` too?

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -69,6 +69,7 @@ type KubernetesMetaConfiguration struct {
 	// TODO: use a proper toleration structure?
 	PodToleration              map[string]string   `json:"toleration,omitempty"`
 	PodEnvironmentConfigMap    spec.NamespacedName `json:"pod_environment_configmap,omitempty"`
+	PodEnvironmentSecret       string              `json:"pod_environment_secret,omitempty"`
```
FxKu, line 72: Not `spec.NamespacedName`?

FxKu:
I would be fine with these changes. Unit test we've talked about are still missing, though ;)

pkg/util/config/config.go
frittentheke: To avoid exposing the contents of the secrets, main reason for this PR, they are not copied (as with the existing PodEnvironmentConfigMap ) but referenced. This means only a secret in the pods namespace can be used here (see https://kubernetes.io/docs/concepts/configuration/secret/#restrictions)

FxKu: I see, it's also documented. Fine for me then.

frittentheke:
> I would be fine with these changes. Unit test we've talked about are still missing, though ;)

Yeah, really sorry for the delay. I'll get those done soon.

frittentheke:
@FxKu  I did some cleaning up and added the unit tests greatly inspired by @kupson and his PR #481 

I refrained from adding the hashing of the secrets contents to recognize changes which @kupson did implement. This is a more general problem (https://github.com/kubernetes/kubernetes/issues/22368) in Kubernetes and applies to the ConfigMap which was a valid source for env variables just as well. Until Kubernetes "fixes" this upstream, there are tools like https://github.com/stakater/Reloader which could be added as a sidecar to achieve this functionality without it being a somewhat special and unexpected behavior among Kubernetes uses.

abdennour:
Guys! please let's move on and merge this PR.
We are waiting this feature for cloning clusters

FxKu:
@frittentheke thanks for the update. Can you merge with the master so that we get travis fixed.

3 very minor things I've overlooked so far. Can you add lines also in the configmap and default-config CRD manifests + one entry in the operatorconfigurations.yaml CRD in the helm chart. It's too many places to update, I have to admit.

frittentheke:
@FxKu I added the entries to the example operator config CRD and configmap.
The entries were already present in the values.yaml / values-crd.yaml and are rendered into the corresponding templates as well.

FxKu:
Thanks for fixing 2/3 :smiley: I should have used [links](https://github.com/zalando/postgres-operator/blob/b34db3077a36b0d8608f627376ff527a07afd92b/charts/postgres-operator/crds/operatorconfigurations.yaml#L152).

frittentheke:
@FxKu urgh, sorry for missing that one ... pretty much the most important one ;-)
It's in now, PTAL

FxKu:
:+1:

Jan-M:
:+1:

FxKu:
Thanks @frittentheke for your work (and patience). Also special thanks to @kupson for bringing up the idea and writing tests which were obtained here.

vijay-dcrust:
> @frittentheke thanks for the update. Can you rebase so that we get travis fixed.
> 
> 3 very minor things I've overlooked so far. Can you add lines also in the configmap and default-config CRD manifests + one entry in the operatorconfigurations.yaml CRD in the helm chart. It's too many places to update, I have to admit.

 @frittentheke @FxKu  guys, May I know in which version this feature (defining secret var in configmap) will be released.

haslersn:
This requires the secret to be in the same namespace as the Postgres cluster, right? I'm using Postgres clusters in many different namespaces. It would be nice if there could be a single secret in the postgres-operator namespace. Any ideas on how to implement this securely? Maybe, for each Postgres cluster, the postgres-operator could copy the secret from the postgres-operator namespace to the Postgres cluster namespace.

kupson:
This might be a little out-of-scope for postgres-operator I think. Sharing/cloning secrets across namespaces is a wider configuration issue with it's own solutions like:
- https://github.com/zakkg3/ClusterSecret operator
- https://github.com/IBM-Cloud/kube-samples/tree/master/secret-sync-operator

frittentheke:
@vijay-dcrust  it's released in 1.6 now https://github.com/zalando/postgres-operator/releases/tag/v1.6.0
@haslersn yes and for good reason. If you share something across all clusters anyways there is not much secrecy. So why not just use a ConfigMap configured via the `pod_environment_configmap` setting? This can then be global to the operator and to all clusters created.

Issue #950, opened 2020-04-29
type error in values-crd.yaml


hzliangbin:
I followed the docs quickstart to install the operator. when I run command :

`helm install postgres-operator ./charts/postgres-operator -f ./charts/postgres-operator/values-crd.yaml`

it returns as followd:

```
coalesce.go:195: warning: destination for cluster_labels is a table. Ignoring non-table value application:spilo
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
manifest_sorter.go:192: info: skipping unknown hook: "crd-install"
Error: OperatorConfiguration.acid.zalan.do "postgres-operator" is invalid: [configuration.connection_pooler.connection_pooler_max_db_connections: Invalid value: "string": configuration.connection_pooler.connection_pooler_max_db_connections in body must be of type integer: "string", configuration.connection_pooler.connection_pooler_number_of_instances: Invalid value: "string": configuration.connection_pooler.connection_pooler_number_of_instances in body must be of type integer: "string"]
```
**connection_pooler_max_db_connections**   and **connection_pooler_number_of_instances** should be **integer.**

PR #951, opened 2020-04-29
fix type error in values-crd.yaml

hzliangbin:
**connection_pooler_max_db_connections**  and **connection_pooler_number_of_instances** should be integer. as the issue describe [#950 ](https://github.com/zalando/postgres-operator/issues/950)

### DIFF
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -262,11 +262,11 @@ configConnectionPooler:
   # docker image
   connection_pooler_image: "registry.opensource.zalan.do/acid/pgbouncer"
   # max db connections the pooler should hold
-  connection_pooler_max_db_connections: "60"
+  connection_pooler_max_db_connections: 60
   # default pooling mode
   connection_pooler_mode: "transaction"
   # number of pooler instances
-  connection_pooler_number_of_instances: "2"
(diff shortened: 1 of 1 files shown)


FxKu:
You are using Helm 3, right? Thinking about the PR #901 which introduced the quotes @ReSearchITEng

hzliangbin:
@FxKu Thanks for reminding. I did't catch the details at my first reading the docs. So 
I should the command :
`helm install postgres-operator ./charts/postgres-operator`
instead in Helm3.

FxKu:
@hzliangbin this command use the values.yaml file which is the basis for ConfigMap-based configuration. This one definitely requires quotes. But maybe the crd configuration doesn't and in fact produces errors. Can you check again?

hzliangbin:
@FxKu I checked it again and the result is as followed.

FxKu:
Found time to test and its failing for me, too. Please reopen and we merge.

ReSearchITEng:
so in values.yaml has to be with quotes, while in values-crd without quotes, right?

FxKu:
yes

FxKu:
@hzliangbin @ReSearchITEng we will fix it with #932

hzliangbin:
@FxKu I had been deleted my fork ...

PR #952, opened 2020-04-29
set event broadcasting to Infof and update rbac

FxKu:
to make events more visible
plus, as recommended by our K8s team, "client-go needs to be able to see existing events and patch/update them"


sdudoladov:
:+1:

FxKu:
:+1:

PR #955, opened 2020-05-04
define more default values for opConfig CRD

### LINKED ISSUE #905: Cpu limit value cannot be set
After the latest update for the following input in the cluster manifest:
requests:
      cpu: "1"
limits:
      cpu: "4"
The bellow error occurred:
"level=error msg="could not create cluster: could not enforce minimum resource limits: could not compare defined CPU limit 4 with configured minimum value : could not parse limit  : quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'"
The same for like "500m" format also.

### LINKED ISSUE #944: Why the secret_name_template is different between helm and helm crd?
Hi, I installed the helm without crd, using the `chart/values.yaml` where there is this line:
`secret_name_template: '{username}.{cluster}.credentials'`

But I was expecting to have these credentials: `"{username}.{cluster}.credentials.{tprkind}.{tprgroup}"`
That's because I installed the helm chart using the crd previously.

Now I figured out that in the `values-crd.yaml` has this line:
`secret_name_template: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"`

Then my question is, why this value is different between these 2 ways of install the same thing?

Thank you.

FxKu:
Setting internal defaults for the CRD config requires writing coalesce functions that check for empty values. This PR defines defaults for more string fields and adds a coalesce check for `*bool` (fields that are true by default). Other data types are not so urgent to fix and maybe we better plan to refactor the configuration anyway.

Two more things:
- Secret template in `values.yaml` is aligned with internal default and `values.crd` file
- Internal default of `LogicalBackupS3SSE` is set to "" to allow omitting it in the backup script. I removed some comments in the ConfigMap regarding logical backup to make sure sse is set to `AES256` there, in case people work with this ConfigMap example and expect this setting

fixes #905 
closes #944


sdudoladov:
:+1:

FxKu:
:+1:

PR #960, opened 2020-05-05
Fix pooler delete

### LINKED ISSUE #961: Deletion of connection pooler incorrectly handles ResourceNotFound
With 1.5 the new connection pooler was added. In the function to delete the corresponding resources there is a check if the error returned from the K8S API is `ResourceNotFound` (good, as we wanted to delete it anyways), see https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/resources.go#L194

But it seems to me the condition wrongly checks for the opposite case - so if it NOT this error.
I prepared a PR to adjust this: https://github.com/zalando/postgres-operator/pull/960

frittentheke:
fixes #961

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1223,7 +1223,7 @@ func (c *Cluster) deleteClusterObject(
 				objType, namespacedName)
 
 			if err = del(name); err != nil {
-				return fmt.Errorf("could not Patroni delete cluster object %q with name %q: %v",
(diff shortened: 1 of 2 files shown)


FxKu:
Can you provide at least one sentence in the PR body. zappr doesn't like empty descriptions :)
LGTM, though.

FxKu:
:+1:

erthalion:
Thanks for noticing @frittentheke !

erthalion:
:+1:

Issue #173, opened 2017-12-08
Do not delete "postgres" credentials for cluster on delete


Jan-M:
Make it configurable that the postgres credentials secret is not deleted when cluster is deleted.

This is purely a workaround until the flow of delete and restore is working nicely. Unfortunately until then deleting the postgres secret breaks the usability of a restored cluster as the postgres password is restored from backup.

alexeyklyukin:
Merged.

PR #974, opened 2020-05-14
delete secrets of system users too

FxKu:
As we can clone or restore clusters, new secrets will get created and passwords get reset. Issue from #173 is obsolete by now.


Jan-M:
:+1:

FxKu:
:+1:

drivebyer:
hi @FxKu it's nice feture that we do need to re-configure secret any more. But maybe application side need re-configure the password in this way.

PR #975, opened 2020-05-15
Adding nodeaffinity support

spohner:
This PR adds support for the nodeaffinity spec alongside the node_readiness_label. 

An attempt to add the deprecated NodeSelector was done here https://github.com/zalando/postgres-operator/pull/689, but nodeaffinity is the more flexible and modern approach.

Will add this to docs if the general approach is accepted.


FxKu:
Thanks @spohner for your contribution. We went with a fully speced node affintiy now, see #1166

Issue #979, opened 2020-05-18
Order of env variables in sidecontainers changed


spohner:
We use https://github.com/wrouesnel/postgres_exporter as a global sidecar to export metrics. In the manifest file we define environment variables based on the injected environment variables like this: 

    env:
    - name: DATA_SOURCE_URI
      value: localhost/postgres?sslmode=disable
    - name: DATA_SOURCE_USER
      value: $(POSTGRES_USER)
    - name: DATA_SOURCE_PASS
      value: $(POSTGRES_PASSWORD)

With the merge of https://github.com/zalando/postgres-operator/pull/890 the order of which environment variables are added changed. The order of the env variables is actually a part of the API spec according to https://github.com/kubernetes/kubernetes/issues/40373 (could not find it in the spec docs directly).

By adding the common fields last in the env array a pattern like the one above is no longer supported and this could break some apps.

Should be an easy fix by swapping the order in the append method here https://github.com/zalando/postgres-operator/blob/d52296c3235956ea4cb34f78f6e4cf0aa374d09d/pkg/cluster/k8sres.go#L534. 

I will create a simple PR for this ASAP.

spohner:
PR: https://github.com/zalando/postgres-operator/pull/980

FxKu:
@fischerman can you also comment on this, as you are using are probably using this feature in prod?

fischerman:
Makes sense to have the operator environment variables first. Didn't think about that. PR from @spohner looks good!

frittentheke:
I sent in the PR https://github.com/zalando/postgres-operator/pull/946 which sources env variables from a secret. There I particularly allow to "override" certain variables related to backups and buckets - but in general those used by the operator for core functionality should come first if there is not use-case  and this would rather create unintended effects.

PR #980, opened 2020-05-18
Changed order of sidecar env vars

spohner:
Related to https://github.com/zalando/postgres-operator/issues/979

Changing the order of which env variables are added to sidecars containers to support referencing common fields that are injected in other env vars.

### DIFF
--- a/pkg/cluster/k8sres_test.go
+++ b/pkg/cluster/k8sres_test.go
@@ -1394,7 +1394,7 @@ func TestSidecars(t *testing.T) {
 
 	// replaced sidecar
 	// the order in env is important
-	scalyrEnv := append(env, []v1.EnvVar{v1.EnvVar{Name: "SCALYR_API_KEY", Value: "abc"}, v1.EnvVar{Name: "SCALYR_SERVER_HOST", Value: ""}}...)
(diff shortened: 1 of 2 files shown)


pkg/cluster/k8sres_test.go
```diff
@@ -1394,7 +1394,7 @@ func TestSidecars(t *testing.T) {
 
 	// replaced sidecar
 	// the order in env is important
-	scalyrEnv := append([]v1.EnvVar{v1.EnvVar{Name: "SCALYR_API_KEY", Value: "abc"}, v1.EnvVar{Name: "SCALYR_SERVER_HOST", Value: ""}}, env...)
```
fischerman, line 1397: Minor comment: The two added variables don't have to be added to an array just to applied as variadic arguments.

spohner: Thanks. An oversight. Should be resolved with latest commit.

FxKu:
:+1:

CyberDem0n:
:+1:

FxKu:
Thanks @spohner for your contribution and @fischerman for the review :+1:

PR #990, opened 2020-05-21
Added the ability to pass a alternative scheduler

mcoletti:
This PR will enable using the CRD to pass in the schedulerName to be used.  We use Portworx which leverages the stork scheduler.

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go
+++ b/pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go
@@ -646,6 +646,11 @@ func (in *PostgresSpec) DeepCopyInto(out *PostgresSpec) {
 			(*in)[i].DeepCopyInto(&(*out)[i])
 		}
 	}
+	if in.SchedulerName != nil {
+		in, out := &in.SchedulerName, &out.SchedulerName
+		*out = new(string)
+		**out = **in
+	}
 	return
 }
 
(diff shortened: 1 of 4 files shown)


Jan-M:
FYI https://portworx.com/stork-storage-orchestration-kubernetes/

pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go
```diff
@@ -646,6 +646,11 @@ func (in *PostgresSpec) DeepCopyInto(out *PostgresSpec) {
 			(*in)[i].DeepCopyInto(&(*out)[i])
 		}
 	}
+	if in.SchedulerName != nil {
```
jbagot, line 649: Hi, I compiled your PR because I'm interested in using it because I have to use stork as well. When I did `make docker` I'm receiving these errors:
`pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go:649:22: invalid operation: in.SchedulerName != nil (mismatched types string and nil)`
`pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go:651:8: cannot use new(string) (type *string) as type string in assignment`
`pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go:652:3: invalid indirect of *out (type string)`
`pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go:652:11: invalid indirect of *in (type string)`

FxKu: run `./hack/update-codegen.sh` and maybe `make deps` in case you get compiler errors

bbsbb:
What is the status on this? Is there intent to get it merged or is there a block I can help with solving around it?

FxKu:
Just noticed that we merged it with #1226 which looks nearly identical to this one. Thanks again @bbsbb

PR #992, opened 2020-05-22
new toggle to use internal ELB by default

### LINKED ISSUE #956: Support creating internal elastic load balancer on AWS
Support internal load balancer schema, via toggle in the manifest and maybe a default in the config map.

Defaults *first*

Config map:
```
  allow_public_load_balancers: false|true
  default_load_balancer_schema: internal|public
```

Postgres manifest:
```
  load_balancer_schema: public|internal
```

FxKu:
closes #956 

plus, changes internal default for enabling master LBs to false


Jan-M:
:+1:

PR #1001, opened 2020-06-02
Make selector match labels defined in the deployment

Kamilczak020:
Currently, the deployment manifest specifies two labels: `name` and `team`.
This fixes the service not matching the deployed pods by chosing a correct selector.


FxKu:
:+1:

PR #1003, opened 2020-06-04
Add ability to define additional volume claim templates

db-adrian:
I have added the ability to define additional persistent volume claim templates in the CRD.


bwrobc:
For anyone interested, I've rebased and forward-ported this patchset to current master [here](https://github.com/bwrobc/postgres-operator/tree/PR1003).

Things that I think would need fixing before merging:
- Unit tests need to be written
- Kubernetes doesn't allow many VolumeClaimTemplate options to be changed on-the-fly. If you try to do so the operator gets confused by the resulting errors and basically stops managing the cluster until it's deleted/recreated. It should do something sensible with the errors instead.
- Likewise if you try to change the number of VolumeClaimTemplates dynamically.
- Kubernetes does let you change the VolumeClaimTemplate size dynamically, but the operator should detect if an attempt fails (i.e. the underlying storage is not resizable) and act appropriately. (I'm guessing this counts for the base `pgroot` template too)
- "additionalVolumeClaimTemplates" is too long and unwieldy, but I don't know what it should be called instead

@FxKu or any other project owners, let me know if you'll consider merging this and comment on the above if so. It's working well enough for my purposes, but I'm prepared to take some time to clean it up if it will be useful.

Huge thanks to @adrianNEMO for writing this patchset in the first place.

Issue #988, opened 2020-05-20
Permission issue for tls cert after upgrading to 1.5


nmlc:
After upgrading `1.4 -> 1.5` my cluster couldn't init.

Logs from cluster pods:

```
...
selecting default max_connections ... 100
selecting default shared_buffers ... 128MB
selecting default time zone ... Etc/UTC
creating configuration files ... ok
running bootstrap script ... ok
performing post-bootstrap initialization ... ok
syncing data to disk ... ok

Success. You can now start the database server using:

    /usr/lib/postgresql/12/bin/pg_ctl -D /home/postgres/pgdata/pgroot/data -l logfile start

2020-05-20 14:40:25 UTC [301]: [1-1] 5ec54159.12d 0     FATAL:  could not load server certificate file "/tls/tls.crt": Permission denied
2020-05-20 14:40:25 UTC [301]: [2-1] 5ec54159.12d 0     LOG:  database system is shut down
2020-05-20 14:40:25,108 INFO: postmaster pid=301
/var/run/postgresql:5432 - no response
2020-05-20 14:40:25,121 INFO: removing initialize key after failed attempt to bootstrap the cluster
2020-05-20 14:40:25,137 INFO: renaming data directory to /home/postgres/pgdata/pgroot/data_2020-05-20-14-40-25
2020-05-20 14:40:25,587 INFO: Lock owner: None; I am mbox-grafana-cluster-0
Traceback (most recent call last):
  File "/usr/local/bin/patroni", line 11, in <module>
    load_entry_point('patroni==1.6.5', 'console_scripts', 'patroni')()
  File "/usr/local/lib/python3.6/dist-packages/patroni/__init__.py", line 235, in main
    return patroni_main()
  File "/usr/local/lib/python3.6/dist-packages/patroni/__init__.py", line 199, in patroni_main
    patroni.run()
  File "/usr/local/lib/python3.6/dist-packages/patroni/__init__.py", line 135, in run
    logger.info(self.ha.run_cycle())
  File "/usr/local/lib/python3.6/dist-packages/patroni/ha.py", line 1370, in run_cycle
    info = self._run_cycle()
  File "/usr/local/lib/python3.6/dist-packages/patroni/ha.py", line 1277, in _run_cycle
    return self.post_bootstrap()
  File "/usr/local/lib/python3.6/dist-packages/patroni/ha.py", line 1173, in post_bootstrap
    self.cancel_initialization()
[...]

PR #1007, opened 2020-06-05
trigger rolling update when securityContext of PodTemplate changes

FxKu:
addresses #988

we still need to check, if TLS and spiloFSGroup can be added at the same time, with this diff present


Jan-M:
:+1:

FxKu:
:+1:

PR #1013, opened 2020-06-11
skip creation later to improve visibility of errors

FxKu:
if the cluster name does not start with the `teamId` we are currently skipping the add event, which makes it hard to track the error from the UI. It is emitted neither to worker logs nor to K8s events. The `PostgresClusterStatus`, although set during marshaling, remains empty.

The PR does not skip the addEvent but will process it until the cluster would get created. This allows to show the error in logs accessible from UI and to record a K8s event with a working `ObjectReference`.

### DIFF
--- a/pkg/controller/postgresql.go
+++ b/pkg/controller/postgresql.go
@@ -14,9 +14,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/reference"
 
 	acidv1 "github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1"
 	"github.com/zalando/postgres-operator/pkg/cluster"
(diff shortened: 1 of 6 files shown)


pkg/controller/postgresql.go
```diff
@@ -429,10 +420,21 @@ func (c *Controller) queueClusterEvent(informerOldSpec, informerNewSpec *acidv1.
...
 
-	if clusterError != "" && eventType != EventDelete && eventType != EventAdd {
-		c.logger.
-			WithField("cluster-name", clusterName).
-			Debugf("skipping %q event for the invalid cluster: %s", eventType, clusterError)
+	if clusterError != "" && eventType != EventDelete {
+		c.logger.WithField("cluster-name", clusterName).Debugf("skipping %q event for the invalid cluster: %s", eventType, clusterError)
+
+		switch eventType {
+		case EventAdd:
```
Jan-M, line 427: is this Go format style no indentation for case?

pkg/util/k8sutil/k8sutil.go
```diff
@@ -156,6 +159,34 @@ func NewFromConfig(cfg *rest.Config) (KubernetesClient, error) {
 	return kubeClient, nil
 }
 
+// SetStatus of Postgres cluster
+// TODO: eventually switch to updateStatus() for kubernetes 1.11 and above
```
Jan-M, line 163: Whats with this comment?

pkg/controller/postgresql.go
FxKu: [yes](https://gobyexample.com/switch) :smiley:

pkg/util/k8sutil/k8sutil.go
FxKu: traveled with the code. I'll remove it. `UpdateStatus` requires a Postgresql struct with updated status. Seems easier as of now to just patch the CRD.

FxKu: I wonder if SetStatus should be called differently now that it lives inside k8sutil. Maybe `SetPgStatus` or `SetCRDStatus`

Jan-M:
:+1:

FxKu:
:+1:

PR #1016, opened 2020-06-12
include volume in list of required fields

FxKu:
If volume section is left out in manifest, cluster setup fails. So CRD validation should mark this field as required.


Jan-M:
:+1:

FxKu:
:+1:

PR #1020, opened 2020-06-16
change Clone attribute of PostgresSpec to *CloneDescription

hlihhovac:
When creating `postgresqls.acid.zalan.do` CRD using [client-go](https://github.com/kubernetes/client-go) and `type PostgresSpec`, process fails in `func (c *postgresqls) Create(ctx context.Context, postgresql *v1.Postgresql, opts metav1.CreateOptions) (result *v1.Postgresql, err error)`due to the missing `PostgresSpec.Clone` attribute.
This was resolved changing `Clone` attribute to pointer `*CloneDescription` in `PostgresSpec` struct.
Other changes were required to adopt for above change.


pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -53,7 +53,7 @@ type PostgresSpec struct {
 	NumberOfInstances     int32                       `json:"numberOfInstances"`
 	Users                 map[string]UserFlags        `json:"users"`
 	MaintenanceWindows    []MaintenanceWindow         `json:"maintenanceWindows,omitempty"`
-	Clone                 CloneDescription            `json:"clone"`
+	Clone                 *CloneDescription            `json:"clone"`
```
FxKu, line 56:
```suggestion
	Clone                 *CloneDescription           `json:"clone"`
```

FxKu:
:+1:

pkg/cluster/sync_test.go
```diff
@@ -63,44 +63,47 @@ func noEmptySync(cluster *Cluster, err error, reason SyncReason) error {
...
-					SuperUsername:       superUserName,
-					ReplicationUsername: replicationUserName,
-				},
-				ConnectionPooler: config.ConnectionPooler{
-					ConnectionPoolerDefaultCPURequest:    "100m",
-					ConnectionPoolerDefaultCPULimit:      "100m",
-					ConnectionPoolerDefaultMemoryRequest: "100Mi",
-					ConnectionPoolerDefaultMemoryLimit:   "100Mi",
-					NumberOfInstances:                    int32ToPointer(1),
+	newCluster := func() *Cluster {
```
erthalion, line 66: Why these changes are necessary for the `CloneDescription`?

Jan-M: Ping, should this be removed for the scope of having a smaller PR=

pashagolub: @Jan-M do you want to remove test completely?

pashagolub: @erthalion these changes fix error during manual creation of `postgresqls.acid.zalan.do` CRD.

erthalion: > @erthalion these changes fix error during manual creation of postgresqls.acid.zalan.do CRD.

That's the goal of the whole PR, isn't it? I'm just trying to figure out if there is any mysterious value in changing unit test that seems to be completely unrelated (connection pooler test).

pashagolub: That unit test was broken with new functionality:
`--- FAIL: TestConnectionPoolerSynchronization (0.00s)`

Objects for testing before changes were static objects, after change introduced they became pointers... So before changes, the test just copies static objects and takes the address of it. And now copying pointer will still point to the same object! So tests always ruined the same underlying object with mock copies.

That's why all new mocked objects now created using `New()`, but not copying.

I hope it's clear enough :)

erthalion: But only `CloneDescription` became a point, right? And it's not involved in this test, so still sounds strange for me. But ok, at least it's not included by mistake.

erthalion:
:+1:

FxKu:
Thanks @hlihhovac and @pashagolub for your contribution!

pkg/cluster/sync_test.go
pashagolub: The thing is the test itself was written for static properties. As soon as we've added pointers the simple copying became a problem. That's why we do not copy mock objects but create them separately so they do not interfere

pashagolub:
Thanks for cooperation!

hlihhovac:
Thx everybody!
It is great honour to be part of this community!

PR #1035, opened 2020-06-24
Standby cluster promotion by changing manifest

ghost:
Added support for promoting the standby cluster by removing the standby section from the manifest.
This avoids the need to edit the patroni config manually inside the postgres container.

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -393,6 +393,11 @@ func (c *Cluster) syncStatefulSet() error {
 		return fmt.Errorf("could not set cluster-wide PostgreSQL configuration options: %v", err)
 	}
 
+	// Apply Standby Cluster parameters
+	if err := c.checkAndSetStandbyClusterConfiguration(); err != nil {
+		return fmt.Errorf("could not set StandbyCluster configuration options: %v", err)
+	}
+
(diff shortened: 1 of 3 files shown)


Jan-M:
The idea is good, but e.g. needs to look into updating postgres and standby credentials. These are right now per se out of sync since they are generated by the operator and do not respect the "source" databases credentials. Needs a bit of a discussion or maybe this minimal solution is a start.

@CyberDem0n open for comments

pkg/cluster/sync.go
```diff
@@ -393,6 +393,11 @@ func (c *Cluster) syncStatefulSet() error {
 		return fmt.Errorf("could not set cluster-wide PostgreSQL configuration options: %v", err)
 	}
 
+	// Apply Standby Cluster parameters
+	if err := c.checkAndSetStandbyClusterConfiguration(); err != nil {
+		return fmt.Errorf("could not set StandbyCluster configuration options: %v", err)
+	}
```
FxKu, line 399: Adding or removing the standby section must happen after the pod rotation. Otherwise you promote the standby and then immediately do a rolling update.

FxKu:
I created a follow-up PR with #2472

PR #1037, opened 2020-06-29
helm-chart: allow configmaps instead of endpoints for leader elections…

Gottox:
This fixes leader election in 1.5 on Openshift.

### DIFF
--- a/charts/postgres-operator/templates/clusterrole-postgres-pod.yaml
+++ b/charts/postgres-operator/templates/clusterrole-postgres-pod.yaml
@@ -10,14 +10,11 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
 rules:
 # Patroni needs to watch and manage endpoints
+{{- if toString .Values.configGeneral.kubernetes_use_configmaps | eq "true" }}
 - apiGroups:
   - ""
   resources:
-{{- if toString .Values.configGeneral.kubernetes_use_configmaps | eq "true" }}
   - configmaps
-{{- else }}
-  - endpoints
-{{- end }}
   verbs:
   - create
   - delete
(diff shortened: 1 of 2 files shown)


charts/postgres-operator/templates/clusterrole-postgres-pod.yaml
```diff
@@ -13,7 +13,11 @@ rules:
 - apiGroups:
   - ""
   resources:
+{{- if toString .Values.configGeneral.kubernetes_use_configmaps | eq "true" }}
+  - configmaps
+{{- else }}
   - endpoints
+{{- end }}
```
CyberDem0n, line 20: I would recommend to at least allow `get endpoints` even in the case when configmaps are used.

Gottox: Can you explain why this would be helpful? - As far as I understand endpoints aren't used at all if `configGeneral.kubernetes_use_configmaps` is set.

Gottox:
Any idea what's wrong with the tests? I can't see how my change has triggered the issue.

charts/postgres-operator/templates/clusterrole-postgres-pod.yaml
CyberDem0n: I have some ideas to bypass `kubernetes` service when talking with K8s API servers, that would require having access to `get endpoints`. It would actually good to check it is ever allowed on openshift, but I don't have access to any OCP setups.

Gottox: Changed the patch according to your suggestions.

ReSearchITEng:
Can we get the same conditional also for operator roles?
https://github.com/zalando/postgres-operator/blob/88735a798a8c9ef98897ac94efa09a14527de83b/charts/postgres-operator/templates/clusterrole.yaml#L57

Gottox:
@ReSearchITEng applied!

FxKu:
@Gottox can you change the operator cluster role in the same way? Because the operator can also create a binding between the `postgres-pod` clusterrole and pod service account, but IIRC requires to have at least the same privileges for this. So if the pod role has a section for endpoints and the operator role has not this would fail I think.

Gottox:
I'm currently on vacations, will do it on monday!  :smile:

FxKu:
@Gottox any chance you can have a look at this PR again and update the cluster role? Thanks!

Jan-M:
:+1:

FxKu:
:+1:

FxKu:
Thanks @Gottox for your contribution.

PR #1064, opened 2020-07-16
[WIP] Extend infrastructure roles handling

erthalion:
Postgres Operator uses infrastructure roles to provide access to a
database for external users e.g. for monitoring purposes. Such
infrastructure roles are expected to be present in the form of k8s
secrets with the following content:

    inrole1: some_encrypted_role
    password1: some_encrypted_password
    user1: some_entrypted_name

    inrole2: some_encrypted_role
    password2: some_encrypted_password
    user2: some_entrypted_name

The format of this content is implied implicitely and not flexible
enough. In case if we do not have possibility to change the format of a
secret we want to use in the Operator, we need to recreate it in this
format.

To address this lets make the format of secret content explicitely. The
idea is to introduce a new configuration option for the Operator.

    infrastructure_roles_secrets:
    - secret: k8s_secret_name
      name: some_encrypted_name
      password: some_encrypted_password
      role: some_encrypted_role

    - secret: k8s_secret_name
      name: some_encrypted_name
      password: some_encrypted_password
      role: some_encrypted_role

This would allow Operator to use any avalable secrets to prepare
infrastructure roles. To make it backward compatible simulate the old
behaviour if the new option is not present.

The new configuration option is intended be used mainly from CRD, but
it's also available via Operator ConfigMap in a limited fashion. For
ConfigMap one can put there only a string with one secret definition in
the following format (as a string):

    infrastructure_roles_secret_name: |
        secret: k8s_secret_name,
        name: some_encrypted_name,
        password: some_encrypted_password,
        role: some_encrypted_role


pkg/util/config/config.go
```diff
@@ -51,16 +51,42 @@ type Resources struct {
 	ShmVolume               *bool               `name:"enable_shm_volume" default:"true"`
 }
 
+type InfrastructureRole struct {
+	// Name of a secret which describes the role, and optionally name of a
+	// configmap with an extra information
+	Secret spec.NamespacedName
```
FxKu, line 57: I wonder if this should also be called `SecretName` to be clear it's just a reference. What is the common way in K8s to name reference to other resources?

pkg/controller/util.go
```diff
@@ -109,8 +110,158 @@ func readDecodedRole(s string) (*spec.PgUser, error) {
...
+	var errors []error
+	var noRolesProvided = true
+
+	roles := []spec.PgUser{}
+	uniqRoles := map[string]spec.PgUser{}
+
+	// To be compatible with the legacy implementation we need to return nil if
+	// the provided secret name is empty. The equivalent situation in the
+	// current implementation is an empty rolesSecrets slice or all its items
+	// are empty.
```
FxKu, line 202: Maybe I'm overlooking something, but doesn't `getInfrastructureRoleDefinitions` set the secret name when the old format is used? Can it be empty then?

pkg/controller/util.go
```diff
@@ -119,52 +270,84 @@ func (c *Controller) getInfrastructureRoles(rolesSecret *spec.NamespacedName) (m
...
+					}
 				}
+				// XXX: This is a part of the original implementation, which is
+				// rather obscure. Why do we delete this key? Wouldn't it be
+				// used later in comparison for configmap?
+				delete(secretData, key)
 			}
-			delete(secretData, key)
+
+			roles = append(roles, t)
```
FxKu, line 315: this part here is gone?

```
if t.Name != "" {
	if t.Password == "" {
		c.logger.Warningf("infrastructure role %q has no password defined and is ignored", t.Name)
		continue
	}
	result[t.Name] = t
}
```

pkg/controller/util.go
erthalion: Yes, it sets if from the old field, which apparently could be empty (judging from the previous tests implementation).

pkg/util/k8sutil/k8sutil.go
```diff
@@ -271,31 +271,73 @@ func SameLogicalBackupJob(cur, new *batchv1beta1.CronJob) (match bool, reason st
...
+	oldFormatConfigmap.Name = "testcluster"
+	oldFormatConfigmap.Data = map[string]string{
 		"foobar": "{}",
 	}
-	return configmap, nil
+
+	newFormatConfigmap := &v1.ConfigMap{}
+	newFormatConfigmap.Name = "testcluster"
+	newFormatConfigmap.Data = map[string]string{
+		"new-foobar": "{}",
```
FxKu, line 328: shouldn't we also test with fields like `user_flags` to test if a proper pgUser struct is marshalled?

pkg/util/config/config.go
```diff
@@ -51,16 +51,42 @@ type Resources struct {
 	ShmVolume               *bool               `name:"enable_shm_volume" default:"true"`
 }
 
+type InfrastructureRole struct {
+	// Name of a secret which describes the role, and optionally name of a
+	// configmap with an extra information
+	SecretName spec.NamespacedName
+
+	Name     string
```
FxKu, line 59: Should this field better be called `User`? I got confused while configuring the `infrastructure_roles_secret` field in the config map (where I used `user:_<username>_` first, and then `name:user` and finally `name:_<username>_`).

pkg/controller/util.go
```diff
@@ -109,8 +110,158 @@ func readDecodedRole(s string) (*spec.PgUser, error) {
...
+
+		for _, property := range properties {
+			values := strings.Split(property, valueSep)
+			if len(values) < 2 {
+				continue
+			}
+			name := strings.TrimSpace(values[0])
+			value := strings.TrimSpace(values[1])
+
+			switch name {
```
FxKu, line 155: if you would also extract the `secretname` key here, we could mark the `infrastructure_roles_secret_name` as deprecated and one could solely rely on `infrastructure_roles_secrets`. The surrounding conditions need to be adapted then:

```
if c.opConfig.InfrastructureRolesSecretName == emptyName {
  if c.opConfig.InfrastructureRolesDefs == "" {
    return rolesDefs
  } else {
    do extraction from configmap

} else {
  older format
} 
```

Now, it's a bit confusing that you have to use both config options together when dealing with configmap. But if you do the same for CRD only the old format is used and the new format is ignored (see `if c.opConfig.InfrastructureRolesSecretName == emptyName`)

docs/reference/operator_parameters.md
```diff
@@ -252,8 +252,15 @@ configuration they are grouped under the `kubernetes` key.
   teams API. The default is `postgresql-operator`.
 
 * **infrastructure_roles_secret_name**
-  namespaced name of the secret containing infrastructure roles names and
-  passwords.
+  *deprecated*: namespaced name of the secret containing infrastructure roles
+  names and passwords.
```
RafiaSabih, line 256: might be a minor thing, but should be like roles, names, and passwords, or is it role names and passwords?

docs/reference/operator_parameters.md
```diff
@@ -252,8 +252,15 @@ configuration they are grouped under the `kubernetes` key.
...
+
+* **infrastructure_roles_secrets**
+  array of infrastructure role definitions which reference existing secrets
+  and specify the keys for extract credentials from them such as user name,
+  password and role membership. For the configmap this has to be a string
+  which allows referencing only one infrastructure roles secret.
+  The default is empty.
```
RafiaSabih, line 263: I am finding first part of the sentence hard to follow,  array of infrastructure role definitions that references existing secrets and specify the keys to extract credentials from them such as user name, password, and role membership, does this phrasing says what you meant.
Also, when you mention for the configmap, can it be specified somewhere else also...?
Also, is it role's secret...?

docs/user.md
```diff
@@ -150,23 +150,62 @@ user. There are two ways to define them:
 
 #### Infrastructure roles secret
 
-The infrastructure roles secret is specified by the `infrastructure_roles_secret_name`
-parameter. The role definition looks like this (values are base64 encoded):
+Infrastructure roles can be specified by the `infrastructure_roles_secrets`
+parameter where you can reference multiple existing secrets. Prior to `v1.6.0`
+the operator could only reference one secret with the
+`infrastructure_roles_secret_name` option. However, this secret could contain
+multiple roles using the same set of keys with incremented indexes.
```
RafiaSabih, line 157: Indexes? Does it refers to array indexes? Does it need to be mentioned like this...?

docs/user.md
```diff
@@ -150,23 +150,62 @@ user. There are two ways to define them:
...
-'secret' that is a member of the 'operator' role. For the following definitions
-one must increase the index, i.e. the next role will be defined as 'user2' and
-so on. The resulting role will automatically be a login role.
+'secret' that is a member of the 'operator' role. The resulting role will
+automatically be a login role.
+
+With the new option the user can configure the names of secret keys that
+contain the user name, password etc. The secret itself is referenced by the
+`secretname` key. If the secret uses a template for multiple roles as described
+above, list them as separately.
```
RafiaSabih, line 178: list them separately.

docs/user.md
FxKu: slightly rewritten, hopefully clearer now ... I thought the example shows it nicely what I'm referring to

docs/reference/operator_parameters.md
FxKu: it's roles in this case. Did some rephrasing.

erthalion:
:+1:

FxKu:
:+1:

PR #1065, opened 2020-07-16
set search_path for default roles

### LINKED ISSUE #1014: Unable to create relations with schema other than postgres
Hi Team,

We are creating the schema:  a3sschema with user- **a3s** and database - **a3sdb**.
Below is the configuration used to  deploy the cluster ,
```
spec:
  numberOfInstances: 1
  patroni:
    pg_hba:
    - host    all all 0.0.0.0/0 md5
  postgresql:
    version: "12"
  preparedDatabases:
    a3sdb:
      defaultRoles: false
      defaultUsers: false
      schemas:
        a3sschema:
          defaultRoles: false
          defaultUsers: false
  resources:
    limits:
      cpu: 500m
      memory: 500Mi
    requests:
      cpu: 10m
      memory: 100Mi
  teamId: postgres
  tls:
    secretName: ""
  users:
    a3suser:
    - createdb
  volume:
    size: 1Gi
```

Upon connecting to the database with a3suser user and a3sdb database we create a table.
However we see the table is created with the public schema and not the a3sschema which was created during deployment.

```
Please see logs below:

a3sdb=> CREATE TABLE student (
a3sdb(>     name            varchar(80),
a3sdb(>     location        point
a3sdb(> );
CREATE TABLE
a3sdb=>
a3sdb=> \dt
         List of relations
 Schema |  Name   | Type  |  Owner
--------+---------+-------+---------
 public | student | table | a3suser
(1 row)
```

Also we see the numerous roles created even after setting the   defaultRoles: false
```
a3sdb=> \du
                                                                                List of roles
       Role name        |                         Attributes                         |                                       Member of
------------------------+------------------------------------------------------------+---------------------------------------------------------------------------------------
 a3sdb_a3sschema_owner  | Cannot login                                               | {a3sdb_a3sschema_reader,a3sdb_a3sschema_writer}
 a3sdb_a3sschema_reader | Cannot login                                               | {}
 a3sdb_a3sschema_writer | Cannot login                                               | {a3sdb_a3sschema_reader}
 a3sdb_owner            | Cannot login                                               | {a3sdb_a3sschema_owner,a3sdb_reader,a3sdb_writer,a3sdb_reader_user,a3sdb_writer_user}
 a3sdb_owner_user       |                                                            | {a3sdb_owner}
 a3sdb_reader           | Cannot login                                               | {}
 a3sdb_reader_user      |                                                            | {a3sdb_reader}
 a3sdb_writer           | Cannot login                                               | {a3sdb_reader}
 a3sdb_writer_user      |                                                            | {a3sdb_writer}
 a3suser                | Create DB                                                  | {}
 admin                  | Create DB, Cannot login                                    | {a3sdb_owner,a3suser,a3sdb_owner_user}
 postgres               | Superuser, Create role, Create DB, Replication, Bypass RLS | {}
 robot_zmon             | Cannot login                                               | {}
 standby                | Replication                                                | {}
 zalandos               | Create DB, Cannot login                                    | {}
```

We need you support in order to understand  how do we create the relations in our database with our created schema and not the public schema.
Also how do we eliminate the extra roles despite setting the values to "false"

Thanks,
Regards
Neelam Sharma

FxKu:
fixes #1014

When creating databases through the `preparedDatabases` feature the `search_path` is not configured hence all created roles need to put schema qualifiers in their queries. But, the existing `PgUser` user type already has a field `Parameters` to allow for setting `search_path` individually. Even code for altering the role has already been there, although it wasn't used probably.

This PR defines a default `search_path: "$user"` for each role (note `public` is not included) and then appends all schemas, if it's a `<dbname>_[owner|reader|writer]` role or only one schema when it's a  `<dbname>_<schema>_[owner|reader|writer]` role.

Adding databases with two schemas will have the following settings:

```
  preparedDatabases:
    bar:
      schemas:
        prod: {}
        audit: {}

SELECT r.rolname, d.datname, rs.setconfig
  FROM pg_db_role_setting rs
  LEFT JOIN pg_roles r ON r.oid = rs.setrole
  LEFT JOIN pg_database d ON d.oid = rs.setdatabase
  ORDER BY rolname;

     rolname      | datname |               setconfig                
------------------+---------+----------------------------------------
 bar_audit_owner  |         | {"search_path=\"$user\", audit"}
 bar_audit_reader |         | {"search_path=\"$user\", audit"}
 bar_audit_writer |         | {"search_path=\"$user\", audit"}
 bar_owner        |         | {"search_path=\"$user\", audit, prod"}
 bar_prod_owner   |         | {"search_path=\"$user\", prod"}
 bar_prod_reader  |         | {"search_path=\"$user\", prod"}
 bar_prod_writer  |         | {"search_path=\"$user\", prod"}
 bar_reader       |         | {"search_path=\"$user\", audit, prod"}
 bar_writer       |         | {"search_path=\"$user\", audit, prod"}
```

**Open question 1:**
If the database already contains roles that follow the naming pattern and existing databases and schemas are moved under the `preparedDatabases` key, existing `search_path` settings will be overwritten. Other example: The minimal cluster defines no schemas which would create a default schema `data`. At one point somebody defines new `schemas` under `preparedDatabases` and only these schemas are put into the search_path for the `<dbname>_[owner|reader|writer]` roles. `data` is lost. Solution: Get the current search_path and merge / concat it with the new one.

**Open question 2:**
Should the ALTER ROLE command be enriched with the IN DATABASE term? Probably, it's already fine because of default access privileges.

**Open question 3:**
Should there be an option to add `public` to the default `search_path` even if it's not that safe.

**Open question 4:**
In case the user is called like the schema, the default "$user" is good enough. Should we still append the schema name in case the role is renamed at some point in the future.

### DIFF
--- a/manifests/postgres-operator.yaml
+++ b/manifests/postgres-operator.yaml
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: postgres-operator
       containers:
       - name: postgres-operator
-        image: registry.opensource.zalan.do/acid/postgres-operator:v1.5.0-19-g37596342-dirty
+        image: registry.opensource.zalan.do/acid/postgres-operator:v1.5.0
         imagePullPolicy: IfNotPresent
         resources:
           requests:
--- a/manifests/postgres-operator.yaml
+++ b/manifests/postgres-operator.yaml
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: postgres-operator
       containers:
       - name: postgres-operator
-        image: registry.opensource.zalan.do/acid/postgres-operator:v1.5.0
+        image: registry.opensource.zalan.do/acid/postgres-operator:v1.5.0-19-g37596342-dirty
         imagePullPolicy: IfNotPresent
         resources:
           requests:
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -955,32 +955,42 @@ func (c *Cluster) initPreparedDatabaseRoles() error {
 	}
 
 	for preparedDbName, preparedDB := range c.Spec.PreparedDatabases {
+		// get list of prepared schemas to set in search_path
+		preparedSchemas := preparedDB.PreparedSchemas
+		if len(preparedDB.PreparedSchemas) == 0 {
+			preparedSchemas = map[string]acidv1.PreparedSchema{"data": {DefaultRoles: util.True()}}
+		}
+
+		var searchPath strings.Builder
+		searchPath.WriteString(constants.DefaultSearchPath)
+		for preparedSchemaName := range preparedSchemas {
+			searchPath.WriteString(", " + preparedSchemaName)
+		}
+
 		// default roles per database
-		if err := c.initDefaultRoles(defaultRoles, "admin", preparedDbName); err != nil {
+		if err := c.initDefaultRoles(defaultRoles, "admin", preparedDbName, searchPath.String()); err != nil {
 			return fmt.Errorf("could not initialize default roles for database %s: %v", preparedDbName, err)
 		}
 		if preparedDB.DefaultUsers {
-			if err := c.initDefaultRoles(defaultUsers, "admin", preparedDbName); err != nil {
+			if err := c.initDefaultRoles(defaultUsers, "admin", preparedDbName, searchPath.String()); err != nil {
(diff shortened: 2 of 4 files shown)


Jan-M:
:+1:

FxKu:
:+1:

PR #1066, opened 2020-07-18
Dockerfile based build plus create postgres-operator images

ReSearchITEng:
1. Dockerfile based build of code (using existing Makefile) on alpine (instead of setting full developer env for simple/trivial tests)
plus 
2. create postgres-operator-secure image as well as optional postgres-operator-alpine which holds full alpine os (less secure).

All in same Dockerfile.

This solution is ideal when someone wants to test/share an image with their PR directly from docker.io.
Locating this Dockerfile under root of the project, allows docker hub (docker.io) simple and automatic build of images for those forking the project.

Dockerfile builds two versions of the image: one with full alpine os, and a more secure one, from scratch (which is the last one, and the default).

BUILD.md publishes this information plus points to the docs/developer.md .

solves https://github.com/zalando/postgres-operator/issues/1067


FxKu:
Should this replace the existing [Dockerfile](https://github.com/zalando/postgres-operator/tree/master/docker)? In any case I would put new docker related files into this folder.

ReSearchITEng:
> Should this replace the existing [Dockerfile](https://github.com/zalando/postgres-operator/tree/master/docker)? In any case I 

As the one in docker/ directory is used by Travis (and possibly by other tools also), I have renamed it.

Jan-M:
We will look into this, I like not having to deal with the setup in any way to build and test it. Once we are through with improving kind and e2e this should get some attention.

Would also need to understand the Docker hub situation, this may be useful indeed to make everyone's life easier.

ReSearchITEng:
Hi @Jan-M 
Yes, the idea is to have a quick way for us (non core devs) to easily build the operator, usually only for adding more debug info or some small experimental changes, before we do PRs or open issues.
We want do build using dockerhub. By default, when we connect github repo with dockerhub, it will look for a Dockerfile in the root of the project (hence my request to move this new docker/WithBuildDockerfile to the root of the project and name it Dockerfile).
This Dockerfile first builds the binary and then adds it in the final clean image.

So, it's cleaning the path to create the pgo and its image with zero setup and zero prerequisites.

ReSearchITEng:
Hi @Jan-M @FxKu ,
is there anything we can do to have this merged?
E.g. Should I simply add only the root Dockerfile for the rest of us, without touching at all the existing folder with dockerfiles.
It would help us greatly, as we don't have full blown pgo build machines and dependencies. The proposed Dockerfile is only for us, while existing systems will remain in place as they are, without any change, and without being affected at all.

docker/Dockerfile
```diff
@@ -9,6 +9,7 @@ COPY build/* /
 RUN addgroup -g 1000 pgo
 RUN adduser -D -u 1000 -G pgo -g 'Postgres Operator' pgo
 
-USER 1000:1000
+USER pgo:pgo
```
sylvainOL, line 12: Hello, having "named" user prevents kubernetes to check for non root. I think it would be best to keep as it's now.

PR #1068, opened 2020-07-21
InitContainers should include AdditionalVolumes as well

yanchenko-igor:
Hi, I believe that init containers should have additional volumes as well. this patch fixes the issue.


yanchenko-igor:
Hi, any opinion on this PR?

FxKu:
Hm, with initContainers we already support the full `v1.Container` spec. Can one not list additional volumes there?
What if somebody doesn't want the volume mounts in the initContainers? Why do you need it? :smiley:

gyrter:
Hello. Init containers can help us with different maintenance tasks. You have mechanism in CRD. It would be great if this feature would be  unified.

szuecs:
@FxKu you define volumes in pod spec and reference them in initContainers, so if the podSpec does not contain "my-custom-vol", the container can not reference it.

PR #1069, opened 2020-07-21
allow delete only if annotations meet configured criteria

FxKu:
Right now a Postgres cluster can be easily removed with `kubectl delete pg acid-minimal-cluster`. If clusters are called very similar it might happen that one accidently removes the wrong resource. While it could be recovered from S3 pretty quickly (depending on the size, of course) a few more steps might be necessary to get back to normal operating mode.

To avoid all the stress, this PRs introduces a simple delete protection mechanism, that checks for existing annotations in the manifest before allowing the delete process to continue. Actually, we would assume this is something to be set in the K8s infrastructure, which blocks a delete request already at the ApiServer level. With this approach here, the CR still gets removed but all managed child resources remain. Either if the cluster was deleted accidentally or the delete annotations were forgotten, one has to recreate the cluster again to continue.

The two delete protection checks are:
- cluster name must match
- data must be the current data

The name for the two annotations keys can be configured. Each check can be disabled individually if the option is not set (empty). This is also the default.

This PR also fixes three more things:
- re-enables all e2e tests
- some typos on K8s words

### DIFF
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -67,6 +67,12 @@ configKubernetes:
   #   keya: valuea
   #   keyb: valueb
 
+  # key name for annotation that compares manifest value with current date
+  # delete_annotation_date_key: "delete-date"
+
+  # key name for annotation that compares manifest value with cluster name
+  # delete_annotation_name_key: "delete-clustername"
+
   # list of annotations propagated from cluster manifest to statefulset and deployment
   # downscaler_annotations:
   #   - deployment-time
--- a/pkg/controller/postgresql.go
+++ b/pkg/controller/postgresql.go
@@ -429,9 +429,9 @@ func (c *Controller) queueClusterEvent(informerOldSpec, informerNewSpec *acidv1.
 			c.logger.WithField("cluster-name", clusterName).Warnf(
 				"please, recreate Postgresql resource %q and set annotations to delete properly", clusterName)
 			if currentManifest, marshalErr := json.Marshal(informerOldSpec); marshalErr != nil {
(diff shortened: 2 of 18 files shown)


charts/postgres-operator/values-crd.yaml
```diff
@@ -67,6 +67,12 @@ configKubernetes:
   #   keya: valuea
   #   keyb: valueb
 
+  # key name for annotation that compares manifest value with current date
+  # delete_annotation_date_key: "acid.zalan.do/deletedate"
```
Jan-M, line 71: `/delete-date`?

charts/postgres-operator/values-crd.yaml
```diff
@@ -67,6 +67,12 @@ configKubernetes:
...
+
+  # key name for annotation that compares manifest value with cluster name
+  # delete_annotation_name_key: "acid.zalan.do/deleteclustername"
```
Jan-M, line 74: `delete-clustername`?

pkg/controller/postgresql.go
```diff
@@ -420,6 +420,14 @@ func (c *Controller) queueClusterEvent(informerOldSpec, informerNewSpec *acidv1.
 		clusterError = informerNewSpec.Error
 	}
 
+	// only allow deletion if delete annotations are set and conditions are met
+	if eventType == EventDelete {
+		if err := c.meetsClusterDeleteAnnotations(informerOldSpec); err != nil {
+			c.logger.WithField("cluster-name", clusterName).Warnf("skipping %q event as delete criteria not fulfilled: %s", eventType, err)
```
Jan-M, line 426: `skipping %q event as delete criteria not fulfilled: %s`
To
`Ignoring delete event for cluster %q - manifest does not fulfill delete requirements (annotations): %s`

FxKu:
:+1:

Jan-M:
:+1:

PR #1084, opened 2020-07-30
fix index in TestGenerateSpiloPodEnvVars

FxKu:
PR #946 changed the order of Pod env variables to allow overwriting by additional variables coming from ConfigMap or Secret. Thus the `DCS_ENABLE_KUBERNETES_API` or `ETCD_HOST` variables now come before the additional variables


Jan-M:
:+1:

FxKu:
:+1:

PR #1085, opened 2020-07-30
fix random order for pod environment tests

FxKu:
make sure the order of env variables returned from `getPodEnvironmentConfigMapVariables` and `getPodEnvironmentSecretVariables` is the same for the slices defined in the tests


FxKu:
:+1:

Jan-M:
:+1:

PR #1090, opened 2020-08-07
allow using both infrastructure_roles_options

FxKu:
Current master only allows to set one of the two infrastructure roles options. If `infrastructure_roles_secrets` is set, `infrastructure_roles_secret_name` is ignored. However, as the new option doesn't allow "templating", so configmap users might still want to use old option but also being able to reference another secret with the new flexibility of choosing the keys.

This PR also fixes a small "bug": When using the new configmap option one had to specify a `rolekey`. But it could be that this info isn't available. CRD validation also doesn't enforce it.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -138,6 +138,7 @@ spec:
                     type: object
                     required:
                       - secretname
+                      - userkey
                       - passwordkey
                     properties:
                       secretname:
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -138,7 +138,6 @@ spec:
                     type: object
                     required:
                       - secretname
-                      - userkey
                       - passwordkey
                     properties:
                       secretname:
@@ -148,6 +147,10 @@ spec:
                       passwordkey:
                         type: string
                       rolekey:
+                        type: string
+                      defaultuservalue:
+                        type: string
+                      defaultrolevalue:
                         type: string
                       details:
                         type: string
--- a/manifests/infrastructure-roles-new.yaml
+++ b/manifests/infrastructure-roles-new.yaml
@@ -4,7 +4,7 @@ data:
   # robot_zmon_acid_monitoring_new
   user: cm9ib3Rfem1vbl9hY2lkX21vbml0b3JpbmdfbmV3
   # robot_zmon
-  inrole: cm9ib3Rfem1vbg==
+  role: cm9ib3Rfem1vbg==
   # foobar_new
   password: Zm9vYmFyX25ldw==
 kind: Secret
--- a/manifests/infrastructure-roles-new.yaml
+++ b/manifests/infrastructure-roles-new.yaml
@@ -4,7 +4,7 @@ data:
   # robot_zmon_acid_monitoring_new
   user: cm9ib3Rfem1vbl9hY2lkX21vbml0b3JpbmdfbmV3
   # robot_zmon
-  role: cm9ib3Rfem1vbg==
+  memberof: cm9ib3Rfem1vbg==
   # foobar_new
   password: Zm9vYmFyX25ldw==
 kind: Secret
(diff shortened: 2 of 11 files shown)


charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -138,7 +138,6 @@ spec:
                     type: object
                     required:
                       - secretname
-                      - userkey
```
erthalion, line 141: Why does it not required?

manifests/infrastructure-roles-new.yaml
```diff
@@ -3,8 +3,8 @@ data:
   # infrastructure role definition in the new format
   # robot_zmon_acid_monitoring_new
   user: cm9ib3Rfem1vbl9hY2lkX21vbml0b3JpbmdfbmV3
-  # robot_zmon_new
-  role: cm9ib3Rfem1vbl9uZXc=
+  # robot_zmon
```
erthalion, line 6: To test default values it makes sense to omit this key.

pkg/cluster/sync.go
```diff
@@ -500,6 +500,7 @@ func (c *Cluster) syncSecrets() error {
 				c.logger.Warningf("secret %q does not contain the role %q", secretSpec.Name, secretUsername)
 				continue
 			}
+			c.Secrets[secret.UID] = secret
```
erthalion, line 503: How is this related to the infrastructure roles?

pkg/controller/util.go
```diff
@@ -118,16 +119,12 @@ var emptyName = (spec.NamespacedName{})
...
-		// All the other possibilities require secret name to be present, so if
-		// it is not, then nothing else to be done here.
-		return rolesDefs
-	}
+	// take from CRD configuration
+	rolesDefs := c.opConfig.InfrastructureRoles
 
 	// check if we can extract something from the configmap config option
-	if c.opConfig.InfrastructureRolesDefs != "" {
+	if len(rolesDefs) == 0 && c.opConfig.InfrastructureRolesDefs != "" {
```
erthalion, line 127: No one prevents having both new roles definitions and old format, AFAICS that was the idea.

pkg/controller/util.go
```diff
@@ -163,27 +160,33 @@ func (c *Controller) getInfrastructureRoleDefinitions() []*config.Infrastructure
...
+				roleDef.DefaultUserValue = value
+			case "defaultrolevalue":
+				roleDef.DefaultRoleValue = value
 			default:
 				c.logger.Warningf("Role description is not known: %s", properties)
 			}
 		}
-	} else {
+
+		if roleDef.SecretName != emptyName &&
```
erthalion, line 172: It's probably better to have "validate" method.

pkg/controller/util_test.go
```diff
@@ -301,14 +301,51 @@ func TestInfrastructureRoleDefinitions(t *testing.T) {
...
+			"secretname: infrastructureroles-new-test, userkey: test-user, passwordkey: test-password, defaultrolevalue: test-role",
+			[]*config.InfrastructureRole{
+				&config.InfrastructureRole{
+					SecretName: spec.NamespacedName{
+						Namespace: v1.NamespaceDefault,
+						Name:      testInfrastructureRolesNewSecretName,
+					},
+					UserKey:          "test-user",
+					PasswordKey:      "test-password",
+					RoleKey:          "",
```
erthalion, line 342: It's probably better to test with no role value present in the secret.

charts/postgres-operator/crds/operatorconfigurations.yaml
FxKu: when we can have a default user either that or the userkey has to be set. That's why I excluded it here

manifests/infrastructure-roles-new.yaml
FxKu: yes, thought about it too. Will update

pkg/cluster/sync.go
FxKu: moved to #1091

charts/postgres-operator/crds/operatorconfigurations.yaml
erthalion: Userkey is mandatory, default value should kicks in when a userkey value is not found within a secret, no?

FxKu: true, mixed it up in my head

erthalion:
:+1:

FxKu:
:+1:

PR #1091, opened 2020-08-10
Remove all secrets on delete incl. pooler

FxKu:
This PR sets secrets on SYNC, so that they can be deleted.

On connection pooler deletion the corresponding secret is now removed via API call. Because of the way roles and secrets are synced with the `needCoonectionPooler` condition, the pooler user and secret are no longer set after deletion. But resources would still be there. For the database user it's fine - operator does not delete those in general - but the secret should go.

### DIFF
--- a/pkg/cluster/resources.go
+++ b/pkg/cluster/resources.go
@@ -771,9 +771,7 @@ func (c *Cluster) deleteSecret(uid types.UID, secret v1.Secret) error {
 	c.setProcessName("deleting secret")
 	c.logger.Debugf("deleting secret %q", util.NameFromMeta(secret.ObjectMeta))
 	err := c.KubeClient.Secrets(secret.Namespace).Delete(context.TODO(), secret.Name, c.deleteOptions)
-	if k8sutil.ResourceNotFound(err) {
-		c.logger.Debugf("Connection pooler secret was already deleted")
-	} else if err != nil {
+	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/cluster/resources.go
```diff
@@ -221,6 +219,23 @@ func (c *Cluster) deleteConnectionPooler() (err error) {
...
+	secret, err := c.KubeClient.
+		Secrets(c.Namespace).
+		Get(context.TODO(), secretName, metav1.GetOptions{})
+
+	if err != nil {
+		c.logger.Debugf("could not get connection pooler secret %q: %v", secretName, err)
+	} else {
+		uid := secret.UID
+		if err = c.deleteSecret(uid, *secret); err != nil {
+			return fmt.Errorf("could not delete pooler secret: %v", err)
```
Jan-M, line 234: does this prevent deletion of e.g deployment? disable pooler should cleanup as good as it gets.

Jan-M:
:+1:

Jan-M:
:+1:

pkg/cluster/resources.go
FxKu: nope. 1. Deployment, 2. Service, 3. Secret

It's true, however, that if 1. or 2. return an error the rest is not deleted.

FxKu:
:+1:

PR #1096, opened 2020-08-11
UI Service port forwarding internal port is updated to 80 from 8081.

### LINKED ISSUE #1093: Operator UI Port Forwarding Documentation Mistake
When following documentation, UI  port forwarding command is as follows: 

> `kubectl port-forward svc/postgres-operator-ui 8081:8081`

But working on Windows, I got:

> kubectl port-forward svc/postgres-operator-ui 8081:8081
>  error: Service postgres-operator-ui does not have a service port 8081

Then, when run 
Capture

So I thought the internal 8081 in fact should be 80 as:

>  kubectl port-forward svc/postgres-operator-ui 8081:80  

And it worked. Either there is an OS dependent discrepancy or the default port has changed without documentation update.

sonaysevik:
Fixes #1093

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -160,7 +160,7 @@ You can now access the web interface by port forwarding the UI pod (mind the
 label selector) and enter `localhost:8081` in your browser:
 
 ```bash
-kubectl port-forward svc/postgres-operator-ui 8081:8081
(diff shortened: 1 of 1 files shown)


FxKu:
Thanks for the quick fix @sonaysevik

FxKu:
:+1:

Jan-M:
:+1:

PR #1103, opened 2020-08-12
restart instances via rest api instead of recreating pods, fixes bug with being unable to decrease some values, like max_connections

yanchenko-igor:
This PR is intended to solve an issue with values like max_connections being decreased, at the moment after the pod restarting the new decreased value doesn't apply unless instance is restarted by api. When we compare statefulsets we ignore bootstrap.dcs section so pod will not restart after changing parameters which matters only on cluster bootstrap.


yanchenko-igor:
Can this PR be reviewed? if there other way to fix the bug let's discuss it.

sdudoladov:
hello @yanchenko-igor 

I will have a look today

pkg/cluster/k8sres.go
```diff
@@ -407,13 +407,33 @@ func tolerations(tolerationsSpec *[]v1.Toleration, podToleration map[string]stri
...
 // See http://patroni.readthedocs.io/en/latest/dynamic_configuration.html.
 func isBootstrapOnlyParameter(param string) bool {
-	return param == "max_connections" ||
-		param == "max_locks_per_transaction" ||
-		param == "max_worker_processes" ||
-		param == "max_prepared_transactions" ||
-		param == "wal_level" ||
-		param == "wal_log_hints" ||
-		param == "track_commit_timestamp"
+	params := map[string]bool{
```
Jan-M, line 410: i wonder if this list should not be in the config map, this could be the hard coded default. Spilo image may change and not sure this always warrants a new operator build

Jan-M:
Thanks for chipping in on this important issue.

For such a change to rolling upgrades maybe worth having an end 2 end test added or confirm there is one checking respective changes.

pkg/cluster/k8sres.go
yanchenko-igor: I am not sure I understand what you mean, this list is what @CyberDem0n gave me when we discussed how it should work.

sdudoladov (commented):
Please also comment in the [admin docs](https://github.com/zalando/postgres-operator/blob/3ddc56e5b9236de656cd5793b10029a567ff38c1/docs/administrator.md#understanding-rolling-update-of-spilo-pods) on when restarts should happen instead of pod re-creation

e2e/tests/test_e2e.py
```diff
@@ -680,6 +680,42 @@ def test_x_cluster_deletion(self):
...
+            "acid.zalan.do", "v1", "default", "postgresqls", "acid-minimal-cluster", pg_patch_max_connections)
+        # wait for service recreation
+        time.sleep(60)
+
+        pods = k8s.api.core_v1.list_namespaced_pod(
+            'default', label_selector=labels).items
+        self.assert_master_is_unique()
+        masterPod = pods[0]
+        get_max_connections_cmd = '''psql -At -U postgres -c "SELECT setting FROM pg_settings WHERE name = 'max_connections';"'''
+        result = k8s.exec_with_kubectl(masterPpod.metadata.name, get_max_connections_cmd)
```
sdudoladov, line 713: there is a typo here: `masterPpod` should be `masterPod`. Tests fail because of that
```
ERROR: test_decrease_max_connections (test_e2e.EndToEndTestCase)
  File "/usr/local/lib/python3.6/dist-packages/timeout_decorator/timeout_decorator.py", line 81, in new_function
  File "/test_e2e.py", line 713, in test_decrease_max_connections
    result = k8s.exec_with_kubectl(masterPpod.metadata.name, get_max_connections_cmd)
NameError: name 'masterPpod' is not defined
```

pkg/cluster/sync.go
```diff
@@ -410,6 +421,52 @@ func (c *Cluster) syncStatefulSet() error {
 	return nil
 }
 
+func (c *Cluster) restartInstances() error {
+	c.setProcessName("starting to recreate pods")
```
sdudoladov, line 425: I assume here and elsewhere within this function what is meant here are Postgres. Please fix relevant log lines

pkg/cluster/k8sres.go
yanchenko-igor: @sdudoladov it was private discussion in Slack.

pkg/cluster/cluster_test.go
```diff
@@ -845,3 +846,156 @@ func TestPreparedDatabases(t *testing.T) {
...
+		{
+			`{"postgresql":{"bin_dir":"/usr/lib/postgresql/12/bin","parameters":{"autovacuum_analyze_scale_factor":"0.1"},"pg_hba":["hostssl all all 0.0.0.0/0 md5","host all all 0.0.0.0/0 md5"]},"bootstrap":{"initdb":[{"auth-host":"md5"},{"auth-local":"trust"},"data-checksums",{"encoding":"UTF8"},{"locale":"en_US.UTF-8"}],"users":{"test":{"password":"","options":["CREATEDB","NOLOGIN"]}},"dcs":{"ttl":30,"loop_wait":10,"retry_timeout":10,"maximum_lag_on_failover":33554432,"postgresql":{"parameters":{"max_connections":"100","max_locks_per_transaction":"64","max_worker_processes":"4"}}}}}`,
+			true,
+		},
+		{
+			`{"postgresql":{"bin_dir":"/usr/lib/postgresql/12/bin","parameters":{"autovacuum_analyze_scale_factor":"0.1"},"pg_hba":["hostssl all all 0.0.0.0/0 md5","host all all 0.0.0.0/0 md5"]},"bootstrap":{"initdb":[{"auth-host":"md5"},{"auth-local":"trust"},"data-checksums",{"encoding":"UTF8"},{"locale":"en_US.UTF-8"}],"users":{"test":{"password":"","options":["CREATEDB","NOLOGIN"]}},"dcs":{"ttl":30,"loop_wait":10,"retry_timeout":10,"maximum_lag_on_failover":33554432,"postgresql":{"parameters":{"max_connections":"200","max_locks_per_transaction":"64","max_worker_processes":"4"}}}}}`,
```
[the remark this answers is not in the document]
yanchenko-igor: Yes, we don't want to restart pod if there only bootstrap.dcs is different.

pkg/cluster/cluster.go
```diff
@@ -555,6 +556,56 @@ func compareResourcesAssumeFirstNotNil(a *v1.ResourceRequirements, b *v1.Resourc
...
+				if !equal {
+					return equal
+				}
+			}
+		}
+		if !hasmatch {
+			return false
+		}
+	}
+	return equal
```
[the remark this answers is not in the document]
yanchenko-igor: Yes, changed to true for readability.

pkg/cluster/sync.go
```diff
@@ -389,10 +389,21 @@ func (c *Cluster) syncStatefulSet() error {
...
 	// it is important to do it after the statefulset pods are there, but before the rolling update
 	// since those parameters require PostgreSQL restart.
-	if err := c.checkAndSetGlobalPostgreSQLConfiguration(); err != nil {
+	instancesRestartRequired, err = c.checkAndSetGlobalPostgreSQLConfiguration()
+	if err != nil {
 		return fmt.Errorf("could not set cluster-wide PostgreSQL configuration options: %v", err)
 	}
 
+	if instancesRestartRequired {
+		c.logger.Debugln("performing instance restart")
```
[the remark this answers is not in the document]
Jan-M: Which begs to question what is the proper variable name and what to use consistently within the operator?

pkg/cluster/sync.go
```diff
@@ -469,11 +523,12 @@ func (c *Cluster) checkAndSetGlobalPostgreSQLConfiguration() error {
 		c.logger.Debugf("calling Patroni API on a pod %s to set the following Postgres options: %v",
 			podName, optionsToSet)
 		if err = c.patroni.SetPostgresParameters(&pod, optionsToSet); err == nil {
-			return nil
+			restartRequired = true
```
CyberDem0n, line 526: Let's say that we changed `work_mem` from '4MB' to '16MB'.
Patroni will update `postgresql.conf` and `do pg_ctl reload`.
That doesn't require a restart.

yanchenko-igor: How we can distinguish parameters which require restart from those which don't?

CyberDem0n: You have to wait for a while (ttl seconds should be enough), and call Patroni API on every pod:
```
$ curl -s http://localhost:8008/patroni | jq .
{
  "pending_restart": true,
```

If there is `pending_restart` flag is set - postgres in this pod must be restarted.

Also, the `/restart` [endpoint](https://patroni.readthedocs.io/en/latest/rest_api.html#restart-endpoint) expects to get the json object, where you can specify `{"restart_pending":true}`, so Patroni will restart postgres only if it is required.

Waiting for `ttl` seconds is mandatory.

yanchenko-igor: Added `{"restart_pending":true}` and waiting after restart.

yanchenko-igor:
Hello, I need help with e2e tests, I don't understand why they fail, can someone support me? They seem to be unrelated to my changes.

Jan-M:
Hi! Right now you should not spend time on the e2e test, we are aware of this not being as reliable as we want it to be and also the feedback time seems really slow.

We are looking into this in a separate PR, lets continue this once we have this sorted out.

yanchenko-igor:
Thanks for the feedback, please let me know if there is anything left except e2e tests, I think I've done everything that was requested additionally.

yanchenko-igor:
All tests passed! :tada:

kustodian:
I'm not 100% sure from the code, but will this PR also fix restarting instances when only a reload is needed?

yanchenko-igor:
> I'm not 100% sure from the code, but will this PR also fix restarting instances when only a reload is needed? Or to ask in a different way, will the PG instances only be reloaded when a reload is enough?

@kustodian In some cases when updating value with patronictl is enough, it will not restart instances, I am not sure, you can check patroni code to check if reload is happening in such cases, but it doesn't restart nodes if patronictl does the job.

yanchenko-igor:
FYI I removed timeout from e2e tests as I noticed that even master branch doesn't pass them at the moment, let me know if I should rollback this change.

yanchenko-igor:
I would like to remind about this PR

FxKu:
:+1:

FxKu:
:+1:

FxKu:
@yanchenko-igor with `pending_restart` an actual restart will not happen on each sync, but still seeing the messages in the logs and events makes it quite confusing:
```
Events:
  Type    Reason  Age                  From               Message
  ----    ------  ----                 ----               -------
  Normal  Update  119s (x79 over 44h)  postgres-operator  restarting Postgres server within pods
  Normal  Update  87s (x79 over 44h)   postgres-operator  Postgres server restart done - all instances have been restarted
```

I think, it's better if we read the config from patroni and compare it with what's in the manifest under `postgresql.parameters`.

FxKu:
@yanchenko-igor Before you start working on it again, maybe we will merge the PR now as is, because I have to build up on it anyway and could then refactor that part.

FxKu:
:+1:

FxKu:
:+1:

sdudoladov:
:+1:

FxKu:
Thanks again @yanchenko-igor  for this important contribution and sorry it took so long to merge. Thanks for your patience and responsiveness.

PR #1111, opened 2020-08-18
Improve e2e tests

### LINKED ISSUE #1101: Improve e2e tests
We need to convert the e2e test image to a test runner without any tests inside.
That would enable us to avoid frequent image rebuilds.

For that we need to get rid of the COPY commands in favor of mounting directories with tests.

[example of running the tests from ZMON](https://github.com/zalando-zmon/zmon-worker#running-unit-tests)

sdudoladov:
Speed up tests:

- [x] increase VM size for builds
- [x] cache build results

Fix the issue with the test runner sometimes failing to fetch the `registry.opensource.zalan.do/acid/postgres-operator-e2e-tests:latest`. 
Description  of the fix :
- [x] There is no such image in the registry. In fact the image was never supposed to be there but  was rather supposed to  be rebuilt on every test run locally. So I removed the pull command for it
- [x] The pull command was only executed if the image with the `${operator_tag}` was  not found locally. 
1. The e2e image is built first with the clean working dir so it lacks the `-dirty` suffix added y git in `VERSION          ?= $(shell git describe --tags --always --dirty)`
2. Makefile installs `kind` via `go get`. That from time to time modifies  `go.mod` and `go.sum`and in that way pollutes the working dir.
3. The `run.sh` computes the image tag again, but this time the tag gets the `-dirty` suffix. So we actually have `image-name` locally, but try to fetch `image-name-dirty`, which obviuosly fails and the pull command from above is triggerd, and that fails as well.   Installing `kind` manually removes the issue.


Jan-M:
:+1:

FxKu:
:-1:

Jan-M:
:+1:

FxKu:
:+1:

PR #871, opened 2020-03-19
Update Kind to version 0.8.1 (Kubernetes 1.18.0) 

lalbers:
It would be nice if the e2e tests would run on a more recent kind version. Since most of the k8s versions are newer, it makes sense to test the operator with a more recent api. 

- [x] Adjust _kubeconfig_ management according to https://github.com/kubernetes-sigs/kind/pull/1029.
- [x] Update [kind](https://github.com/kubernetes-sigs/kind) from 0.5.1 -> 0.70

PR #1121, opened 2020-08-26
update kind and use with old storage class

FxKu:
to test if [this idea](https://github.com/zalando/postgres-operator/pull/871#issuecomment-679872320) from #871 works


sdudoladov:
:+1:

FxKu:
:+1:

PR #1127, opened 2020-08-28
Connection pooler for replica

RafiaSabih:
Add a new service connection pooler service for replica.


pkg/cluster/k8sres.go
```diff
@@ -2304,24 +2357,26 @@ func (c *Cluster) generateConnectionPoolerService(spec *acidv1.PostgresSpec) *v1
...
 			},
 		},
 		Type: v1.ServiceTypeClusterIP,
-		Selector: map[string]string{
-			"connection-pooler": c.connectionPoolerName(),
-		},
 	}
-
+	if role == Replica {
+		serviceSpec.Selector = c.roleLabelsSet(false, Replica)
```
Jan-M, line 2373: We probably need to rethink this here at the current stage. Does not look right. Essentially we need to have two non overlapping selectors.

Ideally also including our "application/version" labels

charts/postgres-operator/crds/postgresqls.yaml
```diff
@@ -185,8 +185,10 @@ spec:
               # Note: usernames specified here as database owners must be declared in the users key of the spec key.
             dockerImage:
               type: string
-            enableConnectionPooler:
+            enableMasterConnectionPooler:
```
Jan-M, line 188: unfortunately I am not convinced that we can rename right now, so we should add the new master field and then it needs to be 'pooler || masterPooler' for backwards compatibility

pkg/cluster/k8sres.go
```diff
@@ -2304,26 +2311,26 @@ func (c *Cluster) generateConnectionPoolerService(spec *acidv1.PostgresSpec) *v1
 	if spec.ConnectionPooler == nil {
 		spec.ConnectionPooler = &acidv1.ConnectionPooler{}
 	}
+	name := c.connectionPoolerName(role)
```
Jan-M, line 2314: same here

pkg/cluster/k8sres.go
```diff
@@ -2304,26 +2311,26 @@ func (c *Cluster) generateConnectionPoolerService(spec *acidv1.PostgresSpec) *v1
...
 
 	serviceSpec := v1.ServiceSpec{
 		Ports: []v1.ServicePort{
 			{
-				Name:       c.connectionPoolerName(),
+				Name:       name,
```
Jan-M, line 2319: interesting this is named dynamically. could just be "postgres"

pkg/cluster/k8sres.go
```diff
@@ -2304,26 +2311,26 @@ func (c *Cluster) generateConnectionPoolerService(spec *acidv1.PostgresSpec) *v1
...
 				Port:       pgPort,
-				TargetPort: intstr.IntOrString{StrVal: c.servicePort(Master)},
+				TargetPort: intstr.IntOrString{StrVal: c.servicePort(role)},
 			},
 		},
 		Type: v1.ServiceTypeClusterIP,
-		Selector: map[string]string{
```
Jan-M, line 2317: why remove this? Service needs a selector?

pkg/cluster/sync.go
```diff
@@ -895,8 +895,10 @@ func (c *Cluster) syncConnectionPooler(oldSpec,
 
 	if oldNeedConnectionPooler && !newNeedConnectionPooler {
 		// delete and cleanup resources
-		if err = c.deleteConnectionPooler(); err != nil {
-			c.logger.Warningf("could not remove connection pooler: %v", err)
+		for _, role := range [2]PostgresRole{Master, Replica} {
```
Jan-M, line 905: This is a bit too ambitious in the sync. What if master is disable and replica is enabled?

pkg/cluster/resources.go
```diff
@@ -126,44 +126,82 @@ func (c *Cluster) createConnectionPooler(lookup InstallFunction) (*ConnectionPoo
...
+
+		replServiceSpec := c.generateConnectionPoolerService(&c.Spec, Replica)
+		replService, err := c.KubeClient.
+			Services(replServiceSpec.Namespace).
+			Create(context.TODO(), replServiceSpec, metav1.CreateOptions{})
+
+		if err != nil {
+			return nil, err
+		}
+		c.ConnectionPooler = &ConnectionPoolerObjects{
```
Jan-M, line 192: does this work, if e.g. master connection pooler is enabled? does this not overwrite?

pkg/cluster/resources.go
```diff
@@ -176,9 +214,14 @@ func (c *Cluster) deleteConnectionPooler() (err error) {
 
 	// Clean up the deployment object. If deployment resource we've remembered
 	// is somehow empty, try to delete based on what would we generate
-	deploymentName := c.connectionPoolerName()
-	deployment := c.ConnectionPooler.Deployment
+	deploymentName := c.connectionPoolerName(role)
```
Jan-M, line 217: is there any reason to compute this, we have a pointer to the deployment.

docs/reference/cluster_manifest.md
```diff
@@ -141,10 +141,15 @@ These parameters are grouped directly under  the `spec` key in the manifest.
   configured (so you can override the operator configuration). Optional.
 
 * **enableConnectionPooler**
-  Tells the operator to create a connection pooler with a database. If this
-  field is true, a connection pooler deployment will be created even if
+  Tells the operator to create a connection pooler with a database for the master
+  service. If this field is true, a connection pooler deployment will be created even if
   `connectionPooler` section is empty. Optional, not set by default.
 
+* **enableReplicaConnectionPooler**
```
erthalion, line 148: Someone down the line in `connectionPooler` section needs to be mentioned, that this section alone will enable only a connection pooler for master service.

pkg/cluster/cluster.go
```diff
@@ -346,7 +348,7 @@ func (c *Cluster) Create() error {
 	//
 	// Do not consider connection pooler as a strict requirement, and if
 	// something fails, report warning
-	if c.needConnectionPooler() {
+	if c.needMasterConnectionPooler() || c.needReplicaConnectionPooler() {
```
erthalion, line 351: I have suspicions that without distinction `needMasterConnectionPooler`/`needReplicaConnectionPooler` the code would be more clean. What do you thing about having only one function here for connection pool in general, and one helper function that returns the list of for what roles we need a pooler to be deployed? Similar to what is already implemented in some other place below in this pr.

pkg/cluster/k8sres.go
```diff
@@ -75,8 +75,12 @@ func (c *Cluster) statefulSetName() string {
 	return c.Name
 }
 
-func (c *Cluster) connectionPoolerName() string {
-	return c.Name + "-pooler"
+func (c *Cluster) connectionPoolerName(role PostgresRole) string {
+	name := c.Name + "-pooler"
+	if role == Replica {
+		name = name + "-repl"
```
erthalion, line 81: Since `c.Name` is a cluster name, I guess there should be both `pooler` and `repl` parts.

pkg/cluster/k8sres_test.go
```diff
@@ -1073,7 +1073,8 @@ func TestConnectionPoolerPodSpec(t *testing.T) {
 		{
 			subTest: "labels for service",
 			spec: &acidv1.PostgresSpec{
-				ConnectionPooler: &acidv1.ConnectionPooler{},
+				ConnectionPooler:              &acidv1.ConnectionPooler{},
+				EnableReplicaConnectionPooler: boolToPointer(true),
```
erthalion, line 1077: There is a small blind spot for testing with `EnableReplicaConnectionPooler: false`.

pkg/cluster/k8sres_test.go
```diff
@@ -1166,7 +1170,8 @@ func TestConnectionPoolerDeploymentSpec(t *testing.T) {
 		{
 			subTest: "default configuration",
 			spec: &acidv1.PostgresSpec{
-				ConnectionPooler: &acidv1.ConnectionPooler{},
+				ConnectionPooler:              &acidv1.ConnectionPooler{},
+				EnableReplicaConnectionPooler: boolToPointer(true),
```
erthalion, line 1174: The same as above, why only with replica pooler?

pkg/cluster/resources_test.go
```diff
@@ -62,10 +62,35 @@ func TestConnectionPoolerCreationAndDeletion(t *testing.T) {
 		t.Errorf("%s: Connection pooler service is empty", testName)
 	}
 
-	err = cluster.deleteConnectionPooler()
+	err = cluster.deleteConnectionPooler(Master)
 	if err != nil {
 		t.Errorf("%s: Cannot delete connection pooler, %s", testName, err)
 	}
+
+	//Check if Replica connection pooler can be create and deleted successfully
```
erthalion, line 70: Maybe it makes sense to move out replica pooler test into a separate test function.

pkg/cluster/cluster.go
```diff
@@ -55,8 +55,10 @@ type Config struct {
 
 // K8S objects that are belongs to a connection pooler
 type ConnectionPoolerObjects struct {
-	Deployment *appsv1.Deployment
-	Service    *v1.Service
+	Deployment     *appsv1.Deployment
```
erthalion, line 58: Can we define this as a map with role as a key? This will allow to avoid constant if/else.

pkg/cluster/sync_test.go
```diff
@@ -177,6 +216,41 @@ func TestConnectionPoolerSynchronization(t *testing.T) {
 			defaultInstances: 1,
 			check:            objectsAreDeleted,
 		},
+		{
+			subTest: "delete only master if not needed",
```
erthalion, line 220: Judging form the specification, a master connection pooler here was not created, which somehow contradicts with the test description. What is the point of this test?

pkg/cluster/sync.go
```diff
@@ -973,11 +1005,10 @@ func (c *Cluster) syncConnectionPoolerWorker(oldSpec, newSpec *acidv1.Postgresql
 		specSync, specReason := c.needSyncConnectionPoolerSpecs(oldConnectionPooler, newConnectionPooler)
 		defaultsSync, defaultsReason := c.needSyncConnectionPoolerDefaults(newConnectionPooler, deployment)
 		reason := append(specReason, defaultsReason...)
+
 		if specSync || defaultsSync {
-			c.logger.Infof("Update connection pooler deployment %s, reason: %+v",
```
erthalion, line 977: Why was the log removed?

pkg/cluster/sync_test.go
```diff
@@ -177,6 +216,41 @@ func TestConnectionPoolerSynchronization(t *testing.T) {
...
+					EnableReplicaConnectionPooler: boolToPointer(true),
+				},
+			},
+			cluster:          clusterMock,
+			defaultImage:     "pooler:1.0",
+			defaultInstances: 1,
+			check:            OnlyMasterDeleted,
+		},
+		{
+			subTest: "delete only replica if not needed",
```
erthalion, line 237: It's probably necessary to test other transitions, e.g. from only master pooler to both, from both to one etc.

pkg/cluster/util.go
```diff
@@ -519,16 +519,21 @@ func (c *Cluster) patroniKubernetesUseConfigMaps() bool {
...
 
-func (c *Cluster) needConnectionPoolerWorker(spec *acidv1.PostgresSpec) bool {
-	if spec.EnableConnectionPooler == nil {
-		return spec.ConnectionPooler != nil
-	} else {
-		return *spec.EnableConnectionPooler
-	}
+// isConnectionPoolerEnabled
+func (c *Cluster) needMasterConnectionPoolerWorker(spec *acidv1.PostgresSpec) bool {
+	return (nil != spec.EnableConnectionPooler && *spec.EnableConnectionPooler) || (spec.ConnectionPooler != nil && spec.EnableConnectionPooler == nil)
```
erthalion, line 524: This order of arguments seems strange. What is the reason of changing the original implementation anyway, if they're equivalent?

erthalion:
Are there any e2e tests coming with this feature?

pkg/cluster/k8sres.go
RafiaSabih: I don't understand this, currently we are appending repl to **name** which is c.Name + "-pooler". Do you expect some thing else...?

pkg/cluster/cluster.go
erthalion: Seems like this is still a thing. I wanted to suggest removing `need{Master/Replica}ConnectionPooler` completely as they used mostly together anyway.

pkg/cluster/resources_test.go
erthalion: This is also still actual. To clarify, I've meant different initial condition for the test case framed into a separate function.

pkg/cluster/sync.go
```diff
@@ -845,69 +845,100 @@ func (c *Cluster) syncConnectionPooler(oldSpec,
...
-		}
-	}
-
-	if oldNeedConnectionPooler && !newNeedConnectionPooler {
-		// delete and cleanup resources
-		if err = c.deleteConnectionPooler(); err != nil {
-			c.logger.Warningf("could not remove connection pooler: %v", err)
+		if oldNeedConnectionPooler && !newNeedConnectionPooler {
+			// delete and cleanup resources
+			otherRole := role
```
erthalion, line 907: If I understand this part correctly, this seems to be too detailed and unnecessary for this function. What do you think about introducing a new function defined for `ConnectionPoolerObject` to reset it's value if needed? And then inside one can all present keys.

pkg/cluster/cluster.go
RafiaSabih: Well, no it doesn't exist in the code anymore, it is also marked outdated. Maybe you accidentally had a look at old code.

Jan-M:
Since Dmitrii hints at the blind spots in the test, lets double check that we have the meaningful cases covered. Especially one of the "on/off" cases. With unit tests ideally complete case.

pkg/cluster/k8sres.go
```diff
@@ -2237,7 +2241,7 @@ func (c *Cluster) ownerReferences() []metav1.OwnerReference {
 	}
 }
 
-func (c *Cluster) generateConnectionPoolerDeployment(spec *acidv1.PostgresSpec) (
+func (c *Cluster) generateConnectionPoolerDeployment(spec *acidv1.PostgresSpec, role PostgresRole) (
```
Jan-M, line 2244: pod template for pooler should also inherit labels, according to statefulset pod label inheriatance

pkg/cluster/resources.go
```diff
@@ -176,8 +180,9 @@ func (c *Cluster) deleteConnectionPooler() (err error) {
 
 	// Clean up the deployment object. If deployment resource we've remembered
 	// is somehow empty, try to delete based on what would we generate
-	deploymentName := c.connectionPoolerName()
-	deployment := c.ConnectionPooler.Deployment
+	var deploymentName string
+	var deployment *appsv1.Deployment
+	deployment = c.ConnectionPooler.Deployment[role]
```
Jan-M, line 185: ` c.ConnectionPooler.Deployment[role]` -> `c.connectionPooler[role].deployment`

pkg/cluster/sync.go
```diff
@@ -845,69 +845,105 @@ func (c *Cluster) syncConnectionPooler(oldSpec,
...
+					otherRole = Master
+				}
+			}
+			if c.ConnectionPooler != nil &&
+				(c.ConnectionPooler.Deployment[role] != nil ||
+					c.ConnectionPooler.Service[role] != nil) {
 
-			if err = c.deleteConnectionPooler(); err != nil {
-				c.logger.Warningf("could not remove connection pooler: %v", err)
+				if err = c.deleteConnectionPooler(role); err != nil {
```
Jan-M, line 942: `c.connectionPooler[role].delete()`

pkg/cluster/k8sres.go
erthalion: To clarify, that's just my misunderstanding, the question can be ignored.

docs/reference/cluster_manifest.md
erthalion: In any way the difference when for a leader one can enable connection pool via flag or config section, but for replica only via flag, is somewhat mind boggling. The easiest way to make it more consistent is to prevent config section from enabling connection pooler for master and require flag for that as well.

e2e/tests/test_e2e.py
```diff
@@ -417,7 +423,7 @@ def test_lazy_spilo_upgrade(self):
         }
         k8s.update_config(patch_lazy_spilo_upgrade, step="Init baseline image version")
 
-        self.eventuallyEqual(lambda: k8s.get_statefulset_image(), SPILO_CURRENT, "Stagefulset not updated initially")        
+        self.eventuallyEqual(lambda: k8s.get_statefulset_image(), SPILO_CURRENT, "Stagefulset not updated initially")
```
Jan-M, line 426: lets also fix "stagefulset" in message

docs/reference/cluster_manifest.md
Jan-M: I totally agree. Both options only toggle able via "flag". Question this leaves is there anyone out there relying on the old logic where we remove the "pool" with the operator upgrade.

pkg/cluster/cluster.go
```diff
@@ -347,19 +331,7 @@ func (c *Cluster) Create() error {
...
-		}
-		connectionPooler, err := c.createConnectionPooler(c.installLookupFunction)
-		if err != nil {
-			c.logger.Warningf("could not create connection pooler: %v", err)
-			return nil
-		}
-		c.logger.Infof("connection pooler %q has been successfully created",
-			util.NameFromMeta(connectionPooler.Deployment.ObjectMeta))
-	}
+	c.createConnectionPooler(c.installLookupFunction)
```
Jan-M, line 334: Parameter is not needed or? both `c` are the cluster

pkg/cluster/database.go
```diff
@@ -571,9 +571,6 @@ func (c *Cluster) installLookupFunction(poolerSchema, poolerUser string) error {
 		c.logger.Infof("pooler lookup function installed into %s", dbname)
 	}
 
-	if len(failedDatabases) == 0 {
-		c.ConnectionPooler.LookupFunction = true
-	}
-
+	c.ConnectionPooler[role].LookupFunction = true
```
Jan-M, line 574: Please double check this function with the master, this looks out of sync.

Jan-M:
:+1:

RafiaSabih:
:+1:

PR #1128, opened 2020-08-30
Fix omitempty of attributes of PostgresSpec in postgresql_type.go

hlihhovac:
PostgresSpec struc contains attributes of pointer type which, when used from client-go, need to have "omitempty" json directive - otherwise error is reported when API is used to grammatically create PostgreSQL CRD.


Jan-M:
:+1:

erthalion:
:+1:

Issue #1134, opened 2020-09-09
FAILED: <nil> on a successful switchover


mwpeterson:
https://github.com/zalando/postgres-operator/blob/dfd0dd90ed07365dba7da8f080d266201685fa31/pkg/cluster/cluster.go#L1273

When a successful switchover occurs, `FAILED: <nil>` is emitted as an event. Seeing FAILED after a successful switchover is very unexpected. 
`c.eventRecorder.Eventf()` should be wrapped in `if err != nil {}` so that the FAILED event is only emitted if there is an actual failure

mwpeterson:
Updating with ISSUE_TEMPLATE details:

- Which image of the operator are you using? registry.opensource.zalan.do/acid/postgres-operator:v1.5.0-27-gf3ddce8-dirty
- Where do you run it - cloud or metal? Kubernetes or OpenShift? AWS EC2 k8s via kops
- Are you running Postgres Operator in production? yes
- Type of issue? bug

StepanKuksenko:
also faced with it 
```
8m43s       Normal    Update                   postgresql/acid-pgcluster    Performing rolling update
8m21s       Normal    Switchover               postgresql/acid-pgcluster    Switching over from "acid-pgcluster-0" to "database/acid-pgcluster-1"
8m20s       Normal    Switchover               postgresql/acid-pgcluster    Successfully switched over from "acid-pgcluster-0" to "database/acid-pgcluster-1"
8m20s       Normal    Switchover               postgresql/acid-pgcluster    Switchover from "acid-pgcluster-0" to "database/acid-pgcluster-1" FAILED: <nil>
7m49s       Normal    Update                   postgresql/acid-pgcluster    Rolling update done - pods have been recreated
38m         Normal    Switchover               postgresql/acid-pgcluster    Switching over from "acid-pgcluster-1" to "database/acid-pgcluster-0"
38m         Normal    Switchover               postgresql/acid-pgcluster    Successfully switched over from "acid-pgcluster-1" to "database/acid-pgcluster-0"
38m         Normal    Switchover               postgresql/acid-pgcluster    Switchover from "acid-pgcluster-1" to "database/acid-pgcluster-0" FAILED: <nil>
```

FxKu:
I think we fixed this message with one of the last releases.

PR #1135, opened 2020-09-09
Don't emit FAILED k8s event on successful switchover

mwpeterson:
Fixes issue #1134


Jan-M:
Thanks for the PR.

Would you mind to actually move this emit into the two different error cases above, and emit different messages.

Reason:
1) Switch failed because patroni call failed , which may also still be successfull
2) Switch call was succesfull, but promote is slow (which means user just needs to wait)

Jan-M:
And 3. emit success if this is not happening already somewhere else.

Jan-M:
@mwpeterson any chance you will make the changes?

mwpeterson:
Yes. This weekend.

mwpeterson:
Closing until I _actually_ can find time to rework this. My apologies.

PR #1136, opened 2020-09-10
Allow to overwrite default ExternalTrafficPolicy for the service

yanchenko-igor:
We would like to use Local ExternalTrafficPolicy for the services, so we can properly see the source IP of the clients connecting to the database, for this reason we would like to overwrite detault ExternalTraffixPolicy for the services created by postgres-operator.

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -1619,6 +1619,13 @@ func (c *Cluster) generateService(role PostgresRole, spec *acidv1.PostgresSpec) 
 		}
 
 		c.logger.Debugf("final load balancer source ranges as seen in a service spec (not necessarily applied): %q", serviceSpec.LoadBalancerSourceRanges)
+		// spec.externalTrafficPolicy can define externalTrafficPolicy for the service
+		// if spec.externalTrafficPolicy is not defined default value is "Cluster"
+		if spec.ExternalTrafficPolicy != nil {
+			serviceSpec.ExternalTrafficPolicy = *spec.ExternalTrafficPolicy
+		} else {
+			serviceSpec.ExternalTrafficPolicy = v1.ServiceExternalTrafficPolicyTypeCluster
+		}
 		serviceSpec.Type = v1.ServiceTypeLoadBalancer
 	} else if role == Replica {
 		// before PR #258, the replica service was only created if allocated a LB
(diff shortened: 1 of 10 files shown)


Jan-M:
Since you want to overwrite, should this not be in config map / crd config vs in the Postgres manifest?

Jan-M:
It could just come with "cluster" from config map as default and allow overwrite in manifest like we do with other properties and then this sounds good.

pkg/cluster/k8sres.go
```diff
@@ -1619,12 +1619,10 @@ func (c *Cluster) generateService(role PostgresRole, spec *acidv1.PostgresSpec)
 		}
 
 		c.logger.Debugf("final load balancer source ranges as seen in a service spec (not necessarily applied): %q", serviceSpec.LoadBalancerSourceRanges)
-		// spec.externalTrafficPolicy can define externalTrafficPolicy for the service
-		// if spec.externalTrafficPolicy is not defined default value is "Cluster"
-		if spec.ExternalTrafficPolicy != nil {
-			serviceSpec.ExternalTrafficPolicy = *spec.ExternalTrafficPolicy
+		if c.OpConfig.ExternalTrafficPolicy == "" {
```
Jan-M, line 1622: With the absence of any kind of error handling just dealing with the empty case does not make sense or? it is already set to default to "Cluster" if not set. so just line 1625 is good enough.

yanchenko-igor: I am not sure if I am doing anything wrong, but I don't get default values in the test.

Jan-M: Yes but in no case is c.opconfig.externaltrafficpolicy "", which is fine. I would just argue this whole if statement can just be

```
serviceSpec.ExternalTrafficPolicy = v1.ServiceExternalTrafficPolicyType(c.OpConfig.ExternalTrafficPolicy)
```

You have the default properly set, so it should be cluster for everyone.

Everyone else switches it to Local.

And everyone else using something invalid gets an error.

So why deal/cover `"""` case extra?

yanchenko-igor: I mean, if I run tests and I don't define c.opConfig.ExternalTrafficPolicy, then I get "" here. Not the default "Cluster" that I expect.

yanchenko-igor: I guess it's just because the way how the config is defined in tests.

Jan-M:
Looks good to me, thanks for the fixes.

Jan-M:
:+1:

sdudoladov:
:+1:

sdudoladov:
@yanchenko-igor thanks for a yet another contribution !

PR #1144, opened 2020-09-22
Compare Postgres pod priority on Sync

sdudoladov:
Priority classes are immutable. One option to change a priority class for existing pods is to delete it and then submit the newly configured class under the same name. We deem this approach as too dangerous because in the meantime no Postgres pod will be able to start.

So we take the alternative approach  and rely on the existing Sync machinery to do the rolling update of existing pods.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -459,18 +459,13 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 		}
 	}
 
+	// we assume any change in priority happens by rolling out a new priority class
+	// changing the priority value in an existing class is not supproted
 	if c.Statefulset.Spec.Template.Spec.PriorityClassName != statefulSet.Spec.Template.Spec.PriorityClassName {
 		match = false
 		needsReplace = true
 		needsRollUpdate = true
 		reasons = append(reasons, "new statefulset's pod priority class in spec doesn't match the current one")
-	}
-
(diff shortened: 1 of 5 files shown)


pkg/cluster/cluster.go
```diff
@@ -459,6 +459,20 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
...
 	}
 
+	if c.Statefulset.Spec.Template.Spec.PriorityClassName != statefulSet.Spec.Template.Spec.PriorityClassName {
+		match = false
+		needsReplace = true
+		needsRollUpdate = true
+		reasons = append(reasons, "new statefulset's pod priority class in spec doesn't match the current one")
+	}
+
+	if c.Statefulset.Spec.Template.Spec.Priority != statefulSet.Spec.Template.Spec.Priority {
```
Jan-M, line 469: Where does this priority come from?

sdudoladov: From the new priority class.

the update procedure works as follows:
1. submit the new priority class
2. edit the configmap and restart the operator to  use the new class
3. wait until the operator completes the rolling upgrade

Jan-M: Slightly confusing, can we limit this just to the change of the class name.

the priority we know cannot be changed and is not in scope of the operator/feature now.

Jan-M:
:+1:

sdudoladov:
:+1:

FxKu:
:+1:

PR #1149, opened 2020-09-28
Changed -w to -o wide

vfrans:
-w now means watch for changes not wide output


FxKu:
-w is intended here :smiley:

Watching how first the master comes up and then the replica.

PR #1154, opened 2020-10-01
Support Pgversion13

SanjeevChoubey:
Support Postgresql 13


FxKu:
Thanks @SanjeevChoubey. When we have a 13 Spilo release image ready, we can merge this PR :smiley: Should be soonish.

FxKu:
Was now merged with #1212. Still, thanks for bringing this up!

PR #1159, opened 2020-10-05
Refactor connection pooler

RafiaSabih:
Refactor connection pooler to be a map of PostgresRole, as  now we require connection pooler for each service.


FxKu:
@RafiaSabih can you rebase your PR, so that it does not list unrelated commits?

RafiaSabih:
Current issues to discuss ,

- [ ] some of the functions in connection_pooler package still require cluster package for things like Spec (to get the values of flag EnableConnectionPooler, etc.), OpConfig( to get ConnectionPooler for schema, user, etc. and also for some not so related stuff like PodServiceAccountName). 
 For connection pooler related stuff, maybe we can have separate config for it and thus would not require cluster package for them and also they are not used anywhere else. For PodServiceAccountName, I am not very sure how to handle that.

- [ ] Other issue is in accessing c.KubeClient, which is also not very clear to me on how to handle that dependency

One way could be to let the functions like createConnectionPooler, deleteConnectionPooler, updateConenctionPooler which are using Kubeclient in the cluster package itself, but then it kind of defeats the purpose of refactoring.
Another way could be to have a third package which does this stuff for connection pooler and maybe other resources and both cluster and connection pooler could use import it.

pkg/cluster/cluster.go
```diff
@@ -102,9 +77,8 @@ type Cluster struct {
 	currentProcess   Process
 	processMu        sync.RWMutex // protects the current operation for reporting, no need to hold the master mutex
 	specMu           sync.RWMutex // protects the spec for reporting, no need to hold the master mutex
-
+	ConnectionPooler map[PostgresRole]*connection_pooler.ConnectionPoolerObjects
```
sdudoladov, line 80: So this whole PR is not solely a refactoring but rather a code change that enables separate connection poolers for the master and replicas ?

pkg/cluster/cluster.go
```diff
@@ -346,21 +320,27 @@ func (c *Cluster) Create() error {
...
-	roles := c.RolesConnectionPooler()
-	for _, r := range roles {
-		if c.ConnectionPooler != nil {
-			c.logger.Warning("Connection pooler already exists in the cluster")
-			return nil
+	if c.needConnectionPooler() {
+
+		roles := c.RolesConnectionPooler()
+		for _, r := range roles {
+			c.logger.Warningf("found roles are %v", r)
```
sdudoladov, line 327: why `Warningf` and not `Infof` ? at this point there is technically nothing bad happening

pkg/cluster/cluster.go
```diff
@@ -346,21 +320,27 @@ func (c *Cluster) Create() error {
...
 		}
-		connectionPooler, err := c.createConnectionPooler(c.installLookupFunction)
-		if err != nil {
-			c.logger.Warningf("could not create connection pooler: %v", err)
-			return nil
+
+		for _, r := range c.RolesConnectionPooler() {
+			if c.ConnectionPooler[r] != nil {
+				c.logger.Warning("Connection pooler already exists in the cluster")
```
sdudoladov, line 332: It would be more descriptive here to log also the role for which the pooler exists, and the name of this pooler similar to how errors are processed below in this code

pkg/cluster/cluster.go
```diff
@@ -782,15 +762,31 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
...
+			c.installLookupFunction); err != nil {
+			c.logger.Errorf("could not sync connection pooler: %v", err)
+			updateFailed = true
+		}
 	}
 
 	return nil
 }
 
+func syncResources(a, b *v1.ResourceRequirements) bool {
```
sdudoladov, line 776: I understand this function was named like that in the original code, but judging from what it does, `shouldSyncResources` or `needToSyncResouces` is a better name for it from a readability perspective.

pkg/pooler_interface/pooler_interface.go
```diff
@@ -0,0 +1,18 @@
+package pooler_interface
+
+//functions of cluster package used in connection_pooler package
+type pooler interface {
```
sdudoladov, line 4: Is this interface meant to help resolving  cyclic dependencies ?

pkg/connection_pooler/connection_pooler.go
```diff
@@ -0,0 +1,938 @@
...
+	"github.com/zalando/postgres-operator/pkg/util/constants"
+	"github.com/zalando/postgres-operator/pkg/util/k8sutil"
+)
+
+const (
+	connectionPoolerContainer = "connection-pooler"
+	pgPort                    = 5432
+)
+
+// K8S objects that are belongs to a connection pooler
```
sdudoladov, line 29: typo: `are belongs` -> `belong`

pkg/connection_pooler/connection_pooler.go
```diff
@@ -0,0 +1,938 @@
...
+
+// no sync happened, empty value
+var NoSync SyncReason = []string{}
+
+// PostgresRole describes role of the node
+type PostgresRole string
+
+const (
+	// Master role
+	Master PostgresRole = "master"
```
sdudoladov, line 56: Role constants are already defined in [pkg/cluster/types.go](https://github.com/zalando/postgres-operator/blob/6a689cdc1c15bef7498cd29c0ad3e79a5f840996/pkg/cluster/types.go#L16). Why are they re-defined here ?

sdudoladov:
> Current issues open for discussion,

I feel they are all caused by the fundamental circular dependency between `pkg/cluster` and `pkg/connection_pooler`

We can try to resolve the dependency by making the `pkg/connection_pooler` depend on an additional `ClusterInterface` instead of directly importing objects from `pkg/cluster`. The interface would
1. live in the `connection_pooler` package
2. contain all vars/methods required by a connection pooler

One can have the effective implementation of the interface in the `cluster` package. [When the operator creates a PG cluster](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L349), it will just provide an implementation of the `ClusterInterface` to a connection pooler.

sdudoladov:
@RafiaSabih could you update the PR description with  
1. the scope of this PR. It does not seem to me to be limited to refactoring only as it also manipulates with the objects belonging to the replica connection pooler
2. the relationship between this PR and [#1127](https://github.com/zalando/postgres-operator/pull/1127) . Do you still work on the latter ? there are commits there from the last week

RafiaSabih:
:+1:

Jan-M:
:+1:

pkg/cluster/connection_pooler.go
```diff
@@ -0,0 +1,872 @@
...
+		},
+		Spec: serviceSpec,
+	}
+
+	return service
+}
+
+//delete connection pooler
+func (c *Cluster) deleteConnectionPooler(role PostgresRole) (err error) {
+	//c.setProcessName("deleting connection pooler")
```
erthalion, line 409: Should it be commented out?

erthalion:
:+1:

PR #1160, opened 2020-10-07
Remove operator checks that prevent PG major version upgrade

sdudoladov:
**UPD 2** I re-tested the PR with the most recent version of the Spilo image that supports  major version upgrade. A few minor issues are gone ; the PR is now ready for review.

**UPD** I will introduce the support of the major version upgrade in a series of smaller PRs

As the first step, this PR:
1. removes the checks in the operator that  prevent major version upgrade and
2. documents how to manually carry out the upgrade. 

We rely on Spilo to start the right PG version. For example, if a manifest - and hence the sts template - has version 12 but the actual cluster was not manually upgraded and still runs version 10, Spilo will still start version 10 despite having
```
SPILO_CONFIGURATION={"postgresql":{"bin_dir":"/usr/lib/postgresql/12/bin"}
```
in pod environment.

To do the in-place major upgrade with this pull request:
0. Use / build the Spilo from [this PR](https://github.com/zalando/spilo/pull/488)
```bash
IMAGE=test-image
docker build --build-arg PGVERSION=13 -t $IMAGE .
kind load docker-image test-image:latest
```
1. Submit a PG manifest, and after the cluster is created, increase the version in the manifest
2. Wait until the `UPDATE` event is completely processed. In particular, all pods must successfully start Postgres after the rolling upgrade.
3. Execute in the master pod
```bash
python3  /scripts/inplace_upgrade.py 2
```
where `2` is the number of pods.


pkg/cluster/sync.go
```diff
@@ -332,18 +332,6 @@ func (c *Cluster) syncStatefulSet() error {
 		// statefulset is already there, make sure we use its definition in order to compare with the spec.
 		c.Statefulset = sset
 
-		// check if there is no Postgres version mismatch
-		for _, container := range c.Statefulset.Spec.Template.Spec.Containers {
```
Jan-M, line 336: Is this code obsolete?

sdudoladov: This was initially done [to prevent breaking the cluster by altering the PG version](https://github.com/zalando/postgres-operator/pull/860).  With this PR, increasing a PG version becomes a legitimate operation; decreasing a PG version is still caught with this PR.

Jan-M:
:+1:

sdudoladov:
:+1:

PR #1162, opened 2020-10-09
fixed case where, no ready label is defined, but node is unscheduable

arminfelder:
in case there is no ready label defined, an unschedulable should make a node not ready

### DIFF
--- a/pkg/controller/node_test.go
+++ b/pkg/controller/node_test.go
@@ -15,7 +15,6 @@ const (
 
 func newNodeTestController() *Controller {
 	var controller = NewController(&spec.ControllerConfig{}, "node-test")
-	controller.opConfig.NodeReadinessLabel = map[string]string{readyLabel: readyValue}
 	return controller
 }
 
@@ -38,6 +37,7 @@ func TestNodeIsReady(t *testing.T) {
 	var testTable = []struct {
 		in  *v1.Node
 		out bool
+		readinessLabel map[string]string
 	}{
 		{
 			in:  makeNode(map[string]string{"foo": "bar"}, true),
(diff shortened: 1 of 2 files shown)


pkg/controller/node_test.go
```diff
@@ -63,3 +63,36 @@ func TestNodeIsReady(t *testing.T) {
 		}
 	}
 }
+
+func TestNodeIsReadyWithoutReadyLabel(t *testing.T) {
```
FxKu, line 67: Can we make a more general `TestNodeIsReady` that also covers the `Unscheduable` case? Still, thanks a lot for this fix and provided unit test :smiley:

arminfelder: @FxKu I just changed the TestNodeReady Test to be more generic, please have a look

pkg/controller/node_test.go
```diff
@@ -55,8 +55,38 @@ func TestNodeIsReady(t *testing.T) {
...
+			out:            true,
+			readinessLabel: map[string]string{},
+		},
 	}
 	for _, tt := range testTable {
+		if tt.readinessLabel != nil {
+			nodeTestController.opConfig.NodeReadinessLabel = tt.readinessLabel
+		} else {
+			nodeTestController.opConfig.NodeReadinessLabel = map[string]string{readyLabel: readyValue}
+		}
```
FxKu, line 88:
```suggestion
		nodeTestController.opConfig.NodeReadinessLabel = tt.readinessLabel
```
I would prefer to define readinessLabel for each test to keep this part here short. Maybe we don't need all 8 combinations, e.g. when the node is scheduable

arminfelder: done

FxKu:
:+1:

sdudoladov:
:+1:

sdudoladov:
@arminfelder thanks for the contribution :)

PR #1165, opened 2020-10-12
PostgresTeam CRD for advanced team management

FxKu:
a new CRD for describing relationships between teams to allow for adding more human users to clusters and even map additional members to teams


docs/user.md
```diff
@@ -269,6 +269,67 @@ to choose superusers, group roles, [PAM configuration](https://github.com/CyberD
...
+```
+
+One `PostgresTeam` resource could contain mappings of multiple teams but you
+can choose to create separate CRDs, alternatively. On each CRD creation or
+update the operator will gather all mappings to create additional human users
+in databases the next time they are synced. Additional teams are resolved
+transitively, meaning you will also add users for their `additionalTeams`
+or (not and) `additionalSuperuserTeams`.
+
+For each additional team the Teams API would be queried. Additional members
```
sdudoladov, line 305: `Teams API` is strictly  internal I am afraid. There is an [old dummy implementation](https://github.com/mkabilov/fake-teams-api) from the original operator author though.

pkg/controller/util.go
```diff
@@ -394,6 +396,36 @@ func (c *Controller) getInfrastructureRole(
...
+	c.logger.Debugf("PostgreTeam %q added. Reloading postgres team CRDs and overwriting cached map", pgTeam.Name)
+	c.loadPostgresTeams()
+}
+
+func (c *Controller) postgresTeamUpdate(prev, obj interface{}) {
+	pgTeam, ok := obj.(*acidv1.PostgresTeam)
+	if !ok {
+		c.logger.Errorf("could not cast to PostgresTeam spec")
+	}
+	c.logger.Debugf("PostgreTeam %q updated. Reloading postgres team CRDs and overwriting cached map", pgTeam.Name)
```
sdudoladov, line 425: Is it possible to log contents of `pgTeam` here for debugging purposes ?

pkg/cluster/util.go
```diff
@@ -217,24 +217,37 @@ func (c *Cluster) getTeamMembers(teamID string) ([]string, error) {
...
+	members := []string{}
+	additionalMembers := c.PgTeamMap[c.Spec.TeamID].AdditionalMembers
+	for _, member := range additionalMembers {
+		members = append(members, member)
+	}
+
 	if !c.OpConfig.EnableTeamsAPI {
-		c.logger.Debugf("team API is disabled, returning empty list of members for team %q", teamID)
-		return []string{}, nil
+		c.logger.Debugf("team API is disabled, returning only additional members for team %q if set", teamID)
```
sdudoladov, line 228: that additional list might as well be empty. It is slightly better to log this new message only if there are actually additional users.

pkg/cluster/cluster.go
```diff
@@ -1129,18 +1130,34 @@ func (c *Cluster) initTeamMembers(teamID string, isPostgresSuperuserTeam bool) e
...
 			clusterIsOwnedBySuperuserTeam = true
 		}
 	}
 
+	additionalTeams := c.PgTeamMap.GetAdditionalTeams(c.Spec.TeamID, true)
+	for _, additionalTeam := range additionalTeams {
+		if !(util.SliceContains(superuserTeams, additionalTeam)) {
+			err := c.initTeamMembers(additionalTeam, false)
+			if err != nil {
+				return fmt.Errorf("Cannot create additional team %q for cluster owner by %q: %v", additionalTeam, c.Spec.TeamID, err)
```
sdudoladov, line 1156: a more accurate log message would be `Cannot initialize team members for an additional team`

pkg/util/util_test.go
```diff
@@ -43,6 +43,16 @@ var prettyDiffTest = []struct {
...
 }
 
+var isEqualIgnoreOrderTest = []struct {
+	inA      []string
+	inB      []string
+	outEqual bool
+}{
+	{[]string{"a", "b", "c"}, []string{"a", "b", "c"}, true},
+	{[]string{"a", "b", "c"}, []string{"a", "c", "b"}, true},
+	{[]string{"a", "b"}, []string{"a", "c", "b"}, false},
```
sdudoladov, line 53: another test case would be "equal length / unequal content"
```go
[]string{"a", "b", "c"}, []string{"a", "d", "c"}, true
```
The second slice has `d` instead of `b`

charts/postgres-operator/crds/postgresteams.yaml
```diff
@@ -0,0 +1,67 @@
...
+spec:
+  group: acid.zalan.do
+  names:
+    kind: PostgresTeam
+    listKind: PostgresTeamList
+    plural: postgresteams
+    singular: postgresteam
+    shortNames:
+    - pgteam
+  scope: Namespaced
```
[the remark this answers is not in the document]
Jan-M: No, the operator watches all namespaces for team crds and merges them into one big team configuration. Since namespaces do not necessarily correlate with teams or administrative domains that is fine. Aim is to have a convenient team setup.

charts/postgres-operator/crds/postgresteams.yaml
```diff
@@ -0,0 +1,67 @@
...
+        spec:
+          type: object
+          properties:
+            additionalSuperuserTeams:
+              type: object
+              description: "Map for teamId and associted additional superuser teams"
+              additionalProperties:
+                type: array
+                nullable: true
+                description: "List of teams to become Postgres superusers"
```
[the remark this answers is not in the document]
Jan-M: We will add 

`enable_additional_teams_support` to enable/disable this feature in total (default=True)
`enable_additional_superusers_support` to enable/disable the additional of superusers (default=False)

To make it easy. But a lot of this responsibility also comes from having proper K8s rights for users and admission control.

docs/user.md
```diff
@@ -269,6 +269,67 @@ to choose superusers, group roles, [PAM configuration](https://github.com/CyberD
...
+    - "postgres_superusers"
+  additionalTeams:
+    acid: []
+  additionalMembers:
+    acid:
+    - "elephant"
+```
+
+One `PostgresTeam` resource could contain mappings of multiple teams but you
+can choose to create separate CRDs, alternatively. On each CRD creation or
```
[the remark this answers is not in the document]
Jan-M: They do not belong to a cluster. This is a global team setup.

pkg/cluster/util.go
Jan-M: `returning ${len(members) additional membes`

pkg/teams/postgres_team_test.go
```diff
@@ -0,0 +1,180 @@
...
+					AdditionalMembers:        []string{"drno"},
+				},
+				"teamC": {
+					AdditionalSuperuserTeams: []string{"team24/7"},
+					AdditionalTeams:          []string{"teamA", "teamB", "acid"},
+					AdditionalMembers:        nil,
+				},
+				"team24/7": {
+					AdditionalSuperuserTeams: nil,
+					AdditionalTeams:          nil,
```
Jan-M, line 80: `[]` vs `nil`

pkg/teams/postgres_team_test.go
```diff
@@ -0,0 +1,180 @@
...
+			[]string{"teamC"},
+			"GetAdditionalTeams returns wrong list",
+		},
+		{
+			"Check that additional teams are returned incl. transitive teams",
+			"teamA",
+			true,
+			[]string{"teamC", "teamB", "acid"},
+			"GetAdditionalTeams returns wrong list",
+		},
```
Jan-M, line 132: `24x7-team` -> '[]`

charts/postgres-operator/crds/postgresteams.yaml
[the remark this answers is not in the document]
FxKu: https://github.com/zalando/postgres-operator/blob/61225998ff6227bc9d58f97e8469ccc5ad6f6739/pkg/controller/util.go#L403

docs/user.md
FxKu: The paragraph explaining Teams API (and the dummy version) comes right before this chapter. Should be fine.

sdudoladov:
:+1:

FxKu:
:+1:

Issue #493, opened 2019-02-20
Extend affinity support for PG pods 


sdudoladov:
We need the ability to put pods of a particular Postgres DB on particular nodes,  for example fur purposes of load testing. 

Ideally we also need to deprecate `nodeSelector` along the way, as the k8s eventually plans to remove this feature in favor of pod affinity.
 
See [Affinity/Anti-affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity)

related #489

digitalassembly:
In our case, we use AWS Spot and OnDemand nodes each one label accordingly eg: nodegroup=od and nodegroup=spot.  Currently we have no way to determine node-affinity per PG cluster.

Issue #420, opened 2018-11-20
Specifying a nodeSelector for Postgres Pods


insertjokehere:
Hi,

Its entirely possible that I missed something when I looked through the documentation; if so, sorry in advance.

I'm trying to setup postgres-operator so that it only schedules Postgres pods on specific nodes; our cluster has several different types of nodes, including a mix of CPU architectures. In a 'normal' StatefulSet I would do this by setting `nodeSelector` - is there an equivalent field I can set on the `postgresql` object to achieve the same effect?

Thanks

insertjokehere:
For those finding this in the future; currently (operator v1.1.0) it isn't possible to do this on a per-postgres-cluster level, but you can do it globally using the `node_readiness_label` config option.

* Label the nodes that you want to have Postgres pods on them: `kubectl label node <node name> should_have_pgdb_pods=true`
* Update the operator config map:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: postgres-operator
data:  
  node_readiness_label: 'should_have_pgdb_pods:true'
```

I had to restart the operator pod for it pick up the config change, but once I did it automatically updated the existing StatefulSets and moved the pods onto the labeled nodes

FxKu:
As pointed out, there is the `node_readiness_label` and you can use node taints (as described in #547). If we would implement a node affinity then we should better use the more flexible `NodeAffinity` instead of the rather simple `NodeSelector`.

Closing this for now as there are existing workarounds. Feel free to reopen if there are new questions regarding this topic.

hpurmann:
So this means if I'd want to have separate node pools for separate postgres clusters, I would have to deploy multiple postgres-operator instances? Or how do you guys do it?

I'm asking because we mostly likely want to configure different node types per environment (staging/production).

dekelmidlink:
https://github.com/zalando/postgres-operator/blob/master/manifests/complete-postgres-manifest.yaml
In the end of the file there is node affinity section that can be used

PR #1166, opened 2020-10-12
Add node affinity support

sagor999:
Hi,

this pull request adds node affinity support. 
Added documentation.
Added e2e test. 
All tests complete successfully locally. 

Please let me know if anything is missing, would like to get this to finish line, as I have seen numerous requests and PR that did not make it. 
Just going to reference them here so that once this is accepted, they all can be closed:
https://github.com/zalando/postgres-operator/pull/975
https://github.com/zalando/postgres-operator/pull/689
https://github.com/zalando/postgres-operator/issues/493
https://github.com/zalando/postgres-operator/issues/420
https://github.com/zalando/postgres-operator/pull/209


sagor999:
Travis-CI failing with this:
```
hack/../pkg is out of date. Please run hack/update-codegen.sh
The command "hack/verify-codegen.sh" exited with 1.
```
I ran locally and it works fine. And I don't see any local changes. 
Do I need to do anything to make CI work with this change?

pkg/cluster/k8sres.go
```diff
@@ -2228,6 +2241,9 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(spec *acidv1.PostgresSpec)
 			ServiceAccountName:            c.OpConfig.PodServiceAccountName,
 			TerminationGracePeriodSeconds: &gracePeriod,
 			Containers:                    []v1.Container{poolerContainer},
+			Affinity: &v1.Affinity{
+				NodeAffinity: &spec.ConnectionPooler.NodeAffinity,
+			},
```
FxKu, line 2246: Hm, should we also apply the configured `c.OpConfig.NodeReadinessLabel` here? Like we do in `generatePodTemplate` for db and backup pods.

e2e/tests/test_e2e.py
```diff
@@ -680,6 +680,100 @@ def test_taint_based_eviction(self):
...
+            "metadata": {
+                "labels": {
+                    "environment": "postgres"
+                }
+            }
+        }
+
+        try:
+            # patch current master node with the label
+            print('patching master node: %s'%current_master_node)
```
FxKu, line 705:
```suggestion
            print('patching master node: %s', current_master_node)
```
, instead of % ?

pkg/cluster/k8sres.go
sagor999: Makes sense. Fixed!

e2e/tests/test_e2e.py
sagor999: Fixed.

FxKu:
Can you run `./hack/update-codegen.sh`? Then travis should pass, I think.

sagor999:
@FxKu hm, I just did run that script, but it did not produce any commitable changes. 
```
./hack/update-codegen.sh 
Generating deepcopy funcs
Generating clientset for acid.zalan.do:v1 at github.com/zalando/postgres-operator/pkg/generated/clientset
Generating listers for acid.zalan.do:v1 at github.com/zalando/postgres-operator/pkg/generated/listers
Generating informers for acid.zalan.do:v1 at github.com/zalando/postgres-operator/pkg/generated/informers
```
Which is a bit odd. Do I need to do anything else?

FxKu:
Hm, not the first time, I hearing this. Also doesn't work for one of the committers. Anyway, you can take it from here: https://gist.github.com/FxKu/2eaae658c2c2fd2e4d11ccdeb28cb07f

sagor999:
@FxKu CI is now passing. :)

Jan-M:
@sagor999 thanks for the PR and that it does include tests.

For now I just want to comment on that in the future we should somehow start or engage a discussion early on on how the Postgres manifest should change. It is not my opinion that mapping through all these API objects is the right way to go.

E.g we could have aligned on a field for node affinity that just takes labels, making the "manifest" slim and not overblown. The type of affinity could have been some default.

`nodeAffinity: "node-type=i3,reserved-for=postgres"` would be something that comes to mind to specify affinity to node labels.

Or in your case simply: `nodeAffinity: "environment=pci"`

I want to hide some complexity of the K8s configuration from the user of our manifests.

sagor999:
@Jan-M That would have been `nodeSelector`, which does support what you are describing, for example: 
```
nodeSelector:
  node-type: i3
  reserved-for: postgres
```
Which internally is mapped into nodeAffinity by kubernetes.
nodeAffinity does give you more flexibility though. 
There were some discussion about deprecating `nodeSelector` from kubernetes API, but just checked and it seems like that idea was shut down. 

So I think we can still support `nodeAffinity` as an advanced feature, and maybe add another PR that would add support for simpler `nodeSelector` instead?

Jan-M:
Interesting discussion, from your point are you really looking for the full complexity of node affinity?

I think it is good that the operator creates "node affinity" question should just be what is define on the "user" side. Sometimes it can be helpful to limit the user to the useful options. If in the end we build "affinityComplexity" again we better just support it like you did.

In the end the topic of moving from nodeSelector to nodeAffinity also highlights that this could have been hidden from the user, what was originally nodeSelector in the Postgres manifest could just be translated to affinity.

ReSearchITEng:
Hi,
I could not find any nodeSelector parameter either in crd.
Not sure how to make a production deployments done if there is neither nodeSelector nor affinity. We see this is as very important feature as it stops reaching production.

Regarding the nodeSelector vs affinity, I was checking what others are doing:
ElasticSearch: https://www.elastic.co/guide/en/cloud-on-k8s/current/k8s-advanced-node-scheduling.html#k8s_node_affinity -> both nodeSelector and affinity (more exactly, it exposes the pod spec)
CouchBase: https://docs.couchbase.com/operator/current/reference-couchbasecluster.html#spec-servers-pod-spec-nodeselector - only nodeSelector, but note they have additional concepts like ServerGroups which brings some extra functionality on top.

As production deployments scenarios become complex relatively quickly, and probably that's one of the reasons the affinity was introduced: more complex logical conditions: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity

If I were to choose which would bring more production support, I would go with affinity.
(as for keeping pg simple: these are optional fields and won't clutter the pg definition when are not required).

FxKu:
@sagor999 can to you rebase with the master to resolve the conflict. We did some changes in the e2e tests.

sagor999:
@FxKu done. Do you have an idea as to when you will merge this PR?

e2e/tests/test_e2e.py
```diff
@@ -891,6 +891,102 @@ def test_zzz_taint_based_eviction(self):
...
+                namespace="default",
+                plural="postgresqls",
+                name="acid-minimal-cluster",
+                body=patch_node_affinity_config)
+
+            # bounce replica and check if it has migrated to proper node that has a label
+            k8s.api.core_v1.delete_namespaced_pod("acid-minimal-cluster-1", "default")
+
+            # wait for operator to apply node affinity changes
+            time.sleep(120)
```
FxKu, line 954: doesn't the previous pod trigger a failover? maybe you can use the `assert_failover` moethod then. Instead of using `sleep` we have recently switched to using `eventuallyEqual` and `eventuallyTrue` that retry and potentially finish way sooner than just waiting.

sagor999: Fixed. Removed hard coded sleep. and removed second sleep, that was not necessary there.

sagor999:
@FxKu updated e2e and removed hard coded sleep from it.

pkg/cluster/cluster.go
```diff
@@ -432,6 +432,12 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 		}
 	}
 
+	if !reflect.DeepEqual(c.Statefulset.Spec.Template.Spec.Affinity.NodeAffinity, statefulSet.Spec.Template.Spec.Affinity.NodeAffinity) {
+		needsReplace = true
+		needsRollUpdate = true
+		reasons = append(reasons, "new statefulset's NodeAffinity doesn't match the current one")
+	}
```
FxKu, line 439: This isn't already covered with a [previous](https://github.com/zalando/postgres-operator/blob/9dc5f2427d68150aa91dc5e9a26661d016551825/pkg/cluster/cluster.go#L375) check on affinity?

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -569,6 +654,91 @@ var PostgresCRDResourceValidation = apiextv1.CustomResourceValidation{
...
+														},
+													},
+												},
+											},
+											"weight": {
+												Type:   "integer",
+												Format: "int32",
+											},
+										},
+										Required: []string{"preference, weight"},
```
FxKu, line 698: can you also move it above `Properties`? I know, it's not alphabetical to `Required`, but improves reability a little.

e2e/tests/test_e2e.py
```diff
@@ -891,6 +891,98 @@ def test_zzz_taint_based_eviction(self):
...
+                version="v1",
+                namespace="default",
+                plural="postgresqls",
+                name="acid-minimal-cluster",
+                body=patch_node_affinity_config)
+
+            # bounce replica and check if it has migrated to proper node that has a label
+            k8s.api.core_v1.delete_namespaced_pod("acid-minimal-cluster-1", "default")
+            # wait for pod to be started
+            k8s.wait_for_pod_start('spilo-role=replica')
```
FxKu, line 953: changing the node affinity triggers a rolling update. Thus, we can check here if the failover happens (`asser_failover`) instead of deleting the replica pod.

e2e/tests/test_e2e.py
```diff
@@ -891,6 +891,98 @@ def test_zzz_taint_based_eviction(self):
...
+                    "nodeAffinity" : None
+                }
+            }
+            k8s.api.custom_objects_api.patch_namespaced_custom_object(
+                group="acid.zalan.do",
+                version="v1",
+                namespace="default",
+                plural="postgresqls",
+                name="acid-minimal-cluster",
+                body=patch_node_remove_affinity_config)
```
FxKu, line 980: Removing the affinity is another rolling update. Therefore, we should guarantee that everything is running when we leave the test.. Check other e2e tests like `test_zz_node_readiness_label` or `test_zzz_taint_based_eviction` how it is handled there.

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -247,6 +247,91 @@ var PostgresCRDResourceValidation = apiextv1.CustomResourceValidation{
...
+																},
+															},
+														},
+													},
+													"weight": {
+														Type:   "integer",
+														Format: "int32",
+													},
+												},
+												Required: []string{"preference, weight"},
```
FxKu, line 291: can you also move it above `Properties`? I know, it's not alphabetical order, but improves readability a little.

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -208,12 +209,13 @@ type PostgresStatus struct {
...
-	Mode              string `json:"mode,omitempty"`
-	DockerImage       string `json:"dockerImage,omitempty"`
-	MaxDBConnections  *int32 `json:"maxDBConnections,omitempty"`
+	NumberOfInstances *int32          `json:"numberOfInstances,omitempty"`
+	Schema            string          `json:"schema,omitempty"`
+	User              string          `json:"user,omitempty"`
+	Mode              string          `json:"mode,omitempty"`
+	DockerImage       string          `json:"dockerImage,omitempty"`
+	MaxDBConnections  *int32          `json:"maxDBConnections,omitempty"`
+	NodeAffinity      v1.NodeAffinity `json:"nodeAffinity,omitempty"`
```
FxKu, line 218: Do we need a `NodeAffinity` setting for the pooler, too? @erthalion @Jan-M 
Can we not reuse the affinity defined in the `Postgresql` spec?
If I see it correctly this field isn't even reflected on creation and syncing, e.g. it is not set in `generateConnectionPoolerPodTemplate`.

pkg/cluster/k8sres.go
```diff
@@ -1183,7 +1196,7 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
 		effectiveRunAsUser,
 		effectiveRunAsGroup,
 		effectiveFSGroup,
-		nodeAffinity(c.OpConfig.NodeReadinessLabel),
+		nodeAffinity(c.OpConfig.NodeReadinessLabel, &spec.NodeAffinity),
```
FxKu, line 1199: as mentioned above, this isn't done for the pooler pod template. Another question would be, if we should apply `c.OpConfig.NodeReadinessLabel`, too?

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -247,6 +247,91 @@ var PostgresCRDResourceValidation = apiextv1.CustomResourceValidation{
 							"user": {
 								Type: "string",
 							},
+							"nodeAffinity": {
```
FxKu, line 250: gosh, that's a lot of lines. Is this validation covering all affinity options? Maybe we just cut it short with:
```
"nodeAffinity": {
	Type:                   "object",
	XPreserveUnknownFields: util.True(),
},
```

FxKu: at least, makes it easier to copy it to the CRD yaml [manifest](https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql.crd.yaml) where this part is still missing, I think

FxKu:
@sagor999 did a proper review now. Sorry it took so long and thanks for your patience. Unfortunately, I've found a few more issues.

1. The `NodeAffinity` support for the connection pooler looks only half done. Maybe we better remove it from this PR to get this merged quicker. But, lets wait for opinions from @Jan-M and @erthalion. 
2. e2e tests could still be improved.
3. Do we diff the affinity twice on sync?
4. Extend CRD validation also in the yaml manifests (manifests folder and helm chart). You could also add an example in the [complete](https://github.com/zalando/postgres-operator/blob/master/manifests/complete-postgres-manifest.yaml) postgres manifest. With comments maybe.

Are you fine with these requests?

sagor999:
@FxKu 
1. I would agree on removing connection pooler support for node affinity for now just to make this PR go through. :) 
2. What type of improvements you are looking for?
3. Hm. We shouldn't. What makes you think so? 
4. Will do.

Give me a bit more info about 2 and 3 and I will try to work on those items soon.

FxKu:
It's all there in the inline comments ;)

sagor999:
@FxKu oh. haha. for some reason github did not send me notification email about those, so I totally missed them!

pkg/apis/acid.zalan.do/v1/crds.go
sagor999: @FxKu for some reason, it was just kept trimming extra fields, even with `XPreserveUnknownFields`. That is why I had to expand it almost fully. :(

pkg/cluster/cluster.go
sagor999: You are right, it should be covered by it. I will remove those lines.

pkg/apis/acid.zalan.do/v1/crds.go
sagor999: will do!

pkg/apis/acid.zalan.do/v1/crds.go
sagor999: will do!

e2e/tests/test_e2e.py
sagor999: Will redo that portion!

e2e/tests/test_e2e.py
sagor999: Will update!

pkg/cluster/k8sres.go
sagor999: Is that a question for me or for other members of postgres-operator team?

pkg/apis/acid.zalan.do/v1/crds.go
sagor999: I just tested it again, and yes, even with `XPreserveUnknownFields` it still trims `unknown` fields and completely breaks node affinity support. Btw, before Nov 17th, it used to work fine.

sagor999:
@FxKu I think I addressed all issues. But, I think something broke in github actions? It gives this error currently:
```
Error:  "2020-12-06T21:30:13.185Z"  'error from getOptions'
Error: Error: Command failed: git cat-file -p 135a13f18b0a9a96db03f327200b01d8e28cdb23
fatal: Not a valid object name 135a13f18b0a9a96db03f327200b01d8e28cdb23
```
e2e tests pass locally. 

Let me know also what do you want to do about pooler's node affinity.

sagor999:
@FxKu my e2e test completed successfully, but this one failed:
```
======================================================================
FAIL: test_zzzz_cluster_deletion (test_e2e.EndToEndTestCase)
Test deletion with configured protection
  File "/usr/local/lib/python3.8/dist-packages/timeout_decorator/timeout_decorator.py", line 81, in new_function
  File "/tests/test_e2e.py", line 1106, in test_zzzz_cluster_deletion
    self.eventuallyEqual(lambda: k8s.count_pdbs_with_label(cluster_label), 0, "Pod disruption budget not deleted")
  File "/tests/test_e2e.py", line 42, in eventuallyEqual
    self.assertEqual(y, x, m.format(y))
AssertionError: 1 != 0 : Pod disruption budget not deleted
```
Which doesn't seem related to any of my changes. Is it known that this test might be flaky?
Btw, it runs fine locally.

pkg/apis/acid.zalan.do/v1/crds.go
FxKu: ok, can you remove this part from the validation as it covers the anti_affinity of the pooler

pkg/apis/acid.zalan.do/v1/postgresql_type.go
FxKu: Can you remove the field `NodeAffintity` from pooler API?

FxKu:
@sagor999 can you remove NodeAffinity from pooler config and validation? See my new inline comments.

pkg/apis/acid.zalan.do/v1/crds.go
sagor999: sounds good! will remove

pkg/apis/acid.zalan.do/v1/postgresql_type.go
sagor999: Will do!

sagor999:
@FxKu I think I addressed all issues. Let me know if there is anything else.

sagor999:
@FxKu this test has failed: `test_zz_node_readiness_label`. But it ran fine locally on my machine. Is it known that this test might be flaky? Should I submit empty change to re-trigger tests again?

pkg/cluster/k8sres_test.go
```diff
@@ -864,6 +864,31 @@ func testEnvs(cluster *Cluster, podSpec *v1.PodTemplateSpec, role PostgresRole)
 	return nil
 }
 
+func testNodeAffinity(cluster *Cluster, podSpec *v1.PodTemplateSpec) error {
```
FxKu, line 867: can you add a unit test where this function is actually used :smiley:

FxKu:
@sagor999 we found that your addition to unit tests is only a function. It's not called from test case (t *testing.T).

pkg/cluster/k8sres_test.go
sagor999: Facepalm. :rofl: 
Fixed. 
btw, I found there are a bunch of functions like that in that file that are not used in actual tests.

sagor999:
@FxKu fix submitted.

pkg/cluster/k8sres_test.go
FxKu: Thanks. Yes, we found that, too :smiley:

Jan-M:
:+1:

Jan-M:
Thank you @sagor999 for the quick responses to our not so quick responses. This allows us to put it into the 1.6 we try to sum up now.

FxKu:
:+1:

PR #1171, opened 2020-10-13
Lookup function installation

### LINKED ISSUE #1060: Pooler issue after upgrade to 1.5. With "error: unexpected response from login query".
Hi All!
I ran into the following problem. I worked with operator version 1.4. I created 2 postgres clusters ("dev" and "staging") with identical configurations and databases names (in different namespace k8s), then I restored sql-backup of the database from "dev" to the "staging" database, while replication and backups on s3 continued to work correctly. After that, I updated the operator to version 1.5 and tried to configure the connection through pooler. For the base from the "dev" cluster, everything worked correctly. And when connecting to the pooler for the "staging" cluster, I encountered the error "error: unexpected response from login query". Then I noticed that pooler schemes were not created automatically for the "staging" database, I think this is due to the fact that I restored in the "staging" base from the sql-backup and the operator didn’t touch upon them 
when updating. This is perceived as a mistake. :(

Now I try understand what is missing for the base on "staging" to use pooler.

Any thoughts?

UPD. Yes. This is the problem of the lack of a "pooler" scheme for the required database. After adding the "pooler" scheme and assigning rights for the roles pooler (u) and postgres (uc), the connection through pgbouncer works without errors.

erthalion:
Apparently lookup function installation process is somewhat fragile right now.
It seems that it happens because a connection to an original db could be
somehow reused when we iterate over available databases, and the required
pooler appliance is installed into a wrong db (does it happen due to some
golang optimization reshuffling the code?). To prevent that and make the
implementation more robust, put more debug logging about installation process,
close connection before opening a new one and check to which database we
actually connected to. One more minor fix is to reset stmtBytes, as it was
accumulating stuff in between (which does not lead to any issues, as appliance
queries are idempotent, but still).

Fixes #1060

### DIFF
--- a/pkg/cluster/database.go
+++ b/pkg/cluster/database.go
@@ -101,14 +101,19 @@ func (c *Cluster) databaseAccessDisabled() bool {
 }
 
 func (c *Cluster) initDbConn() error {
-	return c.initDbConnWithName("")
-}
-
-func (c *Cluster) initDbConnWithName(dbname string) error {
-	c.setProcessName("initializing db connection")
 	if c.pgDb != nil {
 		return nil
 	}
+
+	return c.initDbConnWithName("")
+}
+
+// Worker function for connection initialization. This function does not check
+// if the connection is already open, if it is then it will be overwritten.
+// Callers need to make sure no connection is open, otherwise we could leak
+// connections
+func (c *Cluster) initDbConnWithName(dbname string) error {
+	c.setProcessName("initializing db connection")
 
 	var conn *sql.DB
 	connstring := c.pgConnectionString(dbname)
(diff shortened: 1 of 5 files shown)


pkg/cluster/database.go
```diff
@@ -101,14 +101,19 @@ func (c *Cluster) databaseAccessDisabled() bool {
...
 	return c.initDbConnWithName("")
 }
 
+// Worker function for connection initialization. This function does not check
+// if the connection is already open, if it is then it will be overwritten.
+// Callers need to make sure no connection is open, otherwise we could leak
+// connections
 func (c *Cluster) initDbConnWithName(dbname string) error {
 	c.setProcessName("initializing db connection")
-	if c.pgDb != nil {
```
Jan-M, line 109: We can for now assume we do not have any kind of parallel work per Postgres cluster.

Thus we should for now close the connection and set to nil, if we end up here and that is not the case already. We have now fallen once over this code, not doing what it is supposed to do and pgDb being shared.

This is also no real overhead, since this code now always establishes a new connection.

mboutet:
Would this PR address the problem described in this issue https://github.com/zalando/postgres-operator/issues/1060?

erthalion:
> Would this PR address the problem described in this issue #1060?

Yes, the goal is exactly that. Sorry for taking so long :)

erthalion:
:+1:

erthalion:
:+1:

Jan-M:
:+1:

PR #1172, opened 2020-10-14
Introduce PGVERSION

sdudoladov:
Spilo doc:
```
PGVERSION: Specifies the version of postgreSQL to reference in the bin_dir variable (/usr/lib/postgresql/PGVERSION/bin) if postgresql.bin_dir wasn't set in SPILO_CONFIGURATION
```

Using `PGVERSION` is now preferred to setting `postgresql.bin_dir` in `SPILO_CONFIGURATION`

This changes works just fine with any recent Spilo. Anyway as a precaution, we provide the toggle `pgversion_env_var`
When it is "true", the environment contains only `PGVERSION`
```bash
postgres@acid-minimal-cluster-0:~$ echo $SPILO_CONFIGURATION
{"postgresql":{},"bootstrap":{"initdb":[{"auth-host":"md5"},{"auth-local":"trust"}],"users":{"zalandos":{"password":"","options":["CREATEDB","NOLOGIN"]}},"dcs":{}}}
postgres@acid-minimal-cluster-0:~$ echo $PGVERSION
12
```

Otherwise `SPILO_CONFIGURATION` contains the `postgresql.bin_dir` subsection and `PGVERSION` is unset. The latter behavior is not strictly necessary because in Spilo the `bin_dir` takes precedence over `PGVERSION` when both are set. I deliberately made this case more explicit on the operator side and decided not to set `PGVERSION` when `bin_dir` is provided.

Flipping `pgversion_env_var` costs a rolling upgrade because pod environment changes.

per request from @CyberDem0n 

See [Spilo PR 222](https://github.com/zalando/spilo/commit/a7ebb51cb9eebf2fb57e090956d1af270ebdd5f7) for details.
blocked by #1160
closes #554


Jan-M:
So this PR will also remove bin dir from spilo configuration?

sdudoladov:
> So this PR will also remove bin dir from spilo configuration?

yes, it will.

Jan-M:
:+1:

sdudoladov:
:+1:

PR #873, opened 2020-03-23
Add ability to upload logical backup to gcs

db-adrian:
I have added gsutil to the logical backup container and two options to the backup job:
`logical_backup_provider` which allows to select between s3 or gcs for uploading the dump.

`logical_backup_provider` has s3 as default to not break current s3 backups.
`logical_backup_google_application_credentials` sets the path for the service account file which I added via `additional_secret_mount`.

PR #1173, opened 2020-10-14
Add ability to upload logical backup to gcs

sagor999:
This is based on: 
https://github.com/zalando/postgres-operator/pull/873

Resolved conflicts that original PR had. 
@FxKu Would appreciate if you will have time to take a look at this one and hopefully approve it for merging.


pkg/controller/operator_config.go
```diff
@@ -141,12 +141,14 @@ func (c *Controller) importConfigurationFromCRD(fromCRD *acidv1.OperatorConfigur
 	// logical backup config
 	result.LogicalBackupSchedule = util.Coalesce(fromCRD.LogicalBackup.Schedule, "30 00 * * *")
 	result.LogicalBackupDockerImage = util.Coalesce(fromCRD.LogicalBackup.DockerImage, "registry.opensource.zalan.do/acid/logical-backup")
+	result.LogicalBackupProvider = fromCRD.LogicalBackup.BackupProvider
```
FxKu, line 144:
```suggestion
	result.LogicalBackupProvider = util.Coalesce(fromCRD.LogicalBackup.BackupProvider, "s3")
```
We have to do this in order to set a default when using CRD config.

FxKu:
I'm fine with these changes. Just wonder if the AWS default should be called `s3` and not `aws`. Will poll my colleagues :smile:

pkg/controller/operator_config.go
sagor999: Good catch! Fixed!

sdudoladov:
:+1:

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -181,14 +181,16 @@ type ConnectionPoolerConfiguration struct {
...
+	Schedule                     string `json:"logical_backup_schedule,omitempty"`
+	DockerImage                  string `json:"logical_backup_docker_image,omitempty"`
+	BackupProvider               string `json:"logical_backup_provider,omitempty"`
+	S3Bucket                     string `json:"logical_backup_s3_bucket,omitempty"`
+	S3Region                     string `json:"logical_backup_s3_region,omitempty"`
+	S3Endpoint                   string `json:"logical_backup_s3_endpoint,omitempty"`
+	S3AccessKeyID                string `json:"logical_backup_s3_access_key_id,omitempty"`
+	S3SecretAccessKey            string `json:"logical_backup_s3_secret_access_key,omitempty"`
+	S3SSE                        string `json:"logical_backup_s3_sse,omitempty"`
+	GoogleApplicationCredentials string `json:"logical_backup_google_application_credentials,omitempty"`
```
Jan-M, line 193: Should this be `..Path` or `...File` ?

sagor999: @Jan-M Hm. Technically it should be `Path`, since it refers to a path where google credentials are. 
But, in `AWSGCPConfiguration` it is also simply `gcp_credentials`:
https://github.com/zalando/postgres-operator/blob/925491fa1eb429986bc7a6db06e9688b45ad2c02/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go#L123

So if we were to update it here, we should probably update it there too? But that will be a breaking change then. 
What do you think?

cambierr:
Even if this is not directly related, what about using a tool like rclone to export the backups ?

This would make it possible to support S3 / GCS / Openstack Swift / ... transparently by just injecting the required env vars ?

Jan-M:
:+1:

Jan-M:
@cambierr sure in theory we / I do consider it desireable that we have job/pod specs that others can leverage by providing customer docker images.

PR #1175, opened 2020-10-15
add maxLength constraint for CRD

FxKu:
Because otherwise cluster creation will silently fail when label length is exceeded (e.g. cluster name + random revision value)


FxKu:
:+1:

sdudoladov:
:+1:

PR #1176, opened 2020-10-15
Fix cloning from GCS

sagor999:
If using wal_gs_bucket, pass correct env variables for cloning to work.


docs/administrator.md
```diff
@@ -617,6 +617,8 @@ of the backup cron job.
 `cronjobs` resource from the `batch` API group for the operator service account.
 See [example RBAC](../manifests/operator-service-account-rbac.yaml)
 
+You cannot use logical backups when you are cloning cluster. For that you need to setup WAL writing into S3 or GCS.
```
FxKu, line 620: Can you remove this sentence? I understand the intention behind it, but I find it rather confusing then helping. We should add a paragraph here, that we advice to use WAL archiving as backup strategy as we do not support restoring from logical backups as of now. But then we should also point to a corresponding chapter which is still lacking, unfortunately. We should change that soon.

sagor999: @FxKu done!

FxKu:
:+1:

docs/administrator.md
```diff
@@ -685,6 +685,31 @@ aws_or_gcp:
 ...
 ```
 
+### Setup pod environment configmap
```
erthalion, line 688: I believe the documentation part needs to be formulated differently.

* Title of the section should include "GCS" to indicate it's not just general configmap.

*  The operator doesn't have any preferences or suggestions about which method to use, so "we want to use WAL-G" need to be omitted. Instead it could be formulated as "To make operator work with GCS do the following: ...".

FxKu: this paragraph is part of a bigger section about GCP so title is fine.

erthalion: Indeed, then the title part is fine. But I would still ask to avoid indirect suggestions about using WAL-G (it could be a note though). Other than that looks good.

sagor999: @erthalion I re-worded that section based on your feedback. I left a note as to why we are using WAL-G instead of WAL-E, so that end-user understands the reasoning, instead of just using "magical" configmap. Let me know if you agree with the changes.

erthalion: @sagor999 Yes, looks good, thanks. Do you have any references we can also add to make the claim look more respectable?

sagor999: @erthalion valid point. I looked it up, and it seems like it is possible to make WAL-E work with GCS. So I have re-worded that phrase.

erthalion:
:+1:

FxKu:
:+1:

PR #1177, opened 2020-10-15
Fixed PodEnvironmentSecret location namespace

preved911:
Hello. If I understand correct, secret must be located in the same namespace as the postgresql cluster manifest.


FxKu:
Looks indeed wrong. Thanks for fixing. @frittentheke fyi

FxKu:
:+1:

sdudoladov:
:+1:

PR #1178, opened 2020-10-15
Fix timestamp regex when cloning cluster

sagor999:
This fixes regex for timestamp in `clone` section. Postgres doesn't know how to parse timestamp that has 'Z' at the end, you need to explicitly provide timezone instead `+00:00`. 
It throws this error when cloning is complete and it attempts to start up:
`invalid value for parameter "recovery_target_time": "2020-10-15T14:47:00Z"`


FxKu:
ok, but this is about CRD validation. So if there is a manifest with a Zulu time timestamp this would then be rejected, tight? I see that we already have a `|` there. Can we make the regex also cover your case then?

sagor999:
@FxKu it doesn't seem like postgres itself understands Zulu timestamps. So this change will reject manifest if it has zulu timestamp in it. 
>Can we make the regex also cover your case then?

I am not sure I understand what you are asking. Existing regex allows to specify Zulu timestamp and my change removes it.

FxKu:
Ok, so maybe then the operator should make sure to set the env variable correctly. Simply forbidding things in CRD validation that were allowed before can break deployments. Will double check what timestamps we use. Maybe it's only with `+00:00` as `Z` produces the error you mentioned.

FxKu:
@sagor999 can you resolve the conflicts? we want to merge this one

Jan-M:
Seems good. Existing deployments do not matter, they already cloned and can be adjusted to be valid against new pattern.

This prevents user error early on.

Jan-M:
:+1:

FxKu:
:+1:

Jan-M:
:+1:

PR #1179, opened 2020-10-15
Improve gcs docs

sagor999:
This adds documentation on how to make cloning work when using GCS.
It relies on this PR to be approved first: 
https://github.com/zalando/postgres-operator/pull/1176


erthalion:
If it relies on #1176, why not move this change set into it?

sagor999:
@erthalion I prefer not to bundle several changes together into one PR. But if you prefer, I can easily do that. Let me know.

erthalion:
It's not a big deal I guess, but just in case I would ask you to pack the relevant changes together. It could be a separate commit to isolate stuff, but otherwise there is a chance to miss something. Does it sound good for you?

sagor999:
@erthalion not a problem! I merged changes into https://github.com/zalando/postgres-operator/pull/1176 and closing this one.

PR #1192, opened 2020-11-04
fix redundant appending of infrastructure roles

FxKu:
only append once after the roleDescr checks. Should solve the `Conflicting infrastructure roles` error which one could see in the debug logs on operator start up


erthalion:
:+1:

FxKu:
:+1:

PR #1194, opened 2020-11-04
check resize mode on update events

FxKu:
Risizing volumes by updating the PVC storage field has been introduced with #958. However the UPDATE event wasn't covered. Using `pvc` mode requires more privileges than what we currently list in the operator cluster role. In the helm chart we can arrange the RBAC nicely by checking the `storage_resize_mode` setting.

This PR also changes the `storage_resize_mode` in example manifests and charts to `pvc`. The internal default however remains `ebs`.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -140,7 +140,7 @@ func New(cfg Config, kubeClient k8sutil.KubernetesClient, pgSpec acidv1.Postgres
 			Secrets:   make(map[types.UID]*v1.Secret),
 			Services:  make(map[PostgresRole]*v1.Service),
 			Endpoints: make(map[PostgresRole]*v1.Endpoints)},
-		userSyncStrategy: users.DefaultUserSyncStrategy{password_encryption},
+		userSyncStrategy: users.DefaultUserSyncStrategy{PasswordEncryption: password_encryption},
 		deleteOptions:    metav1.DeleteOptions{PropagationPolicy: &deletePropagationPolicy},
 		podEventsQueue:   podEventsQueue,
 		KubeClient:       kubeClient,
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -672,7 +672,7 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
 	// Volume
 	if oldSpec.Spec.Size != newSpec.Spec.Size {
 		c.logVolumeChanges(oldSpec.Spec.Volume, newSpec.Spec.Volume)
-		c.logger.Debugf("syncing volumes using %q resize mode", c.OpConfig.StorageResizeMode)
+		c.logger.Debugf("syncing volumes using %q storage resize mode", c.OpConfig.StorageResizeMode)
 		if c.OpConfig.StorageResizeMode == "pvc" {
 			if err := c.syncVolumeClaims(); err != nil {
 				c.logger.Errorf("could not sync persistent volume claims: %v", err)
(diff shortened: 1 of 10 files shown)


pkg/cluster/cluster.go
```diff
@@ -671,12 +671,20 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
...
 	// Volume
 	if oldSpec.Spec.Size != newSpec.Spec.Size {
-		c.logger.Debugf("syncing persistent volumes")
 		c.logVolumeChanges(oldSpec.Spec.Volume, newSpec.Spec.Volume)
 
-		if err := c.syncVolumes(); err != nil {
-			c.logger.Errorf("could not sync persistent volumes: %v", err)
-			updateFailed = true
+		if c.OpConfig.StorageResizeMode == "pvc" {
+			c.logger.Debugf("syncing persistent volume claims")
```
Jan-M, line 677: `syncing persistent volume claims using 'pvc' resize mode`

pkg/cluster/cluster.go
```diff
@@ -671,12 +671,20 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
...
+			if err := c.syncVolumeClaims(); err != nil {
+				c.logger.Errorf("could not sync persistent volume claims: %v", err)
+				updateFailed = true
+			}
+		} else if c.OpConfig.StorageResizeMode == "ebs" {
+			c.logger.Debugf("syncing persistent volumes")
```
Jan-M, line 683: `syncing persistent volumes using 'ebs' resize mode`

pkg/cluster/cluster.go
```diff
@@ -140,7 +140,7 @@ func New(cfg Config, kubeClient k8sutil.KubernetesClient, pgSpec acidv1.Postgres
 			Secrets:   make(map[types.UID]*v1.Secret),
 			Services:  make(map[PostgresRole]*v1.Service),
 			Endpoints: make(map[PostgresRole]*v1.Endpoints)},
-		userSyncStrategy: users.DefaultUserSyncStrategy{password_encryption},
+		userSyncStrategy: users.DefaultUserSyncStrategy{PasswordEncryption: password_encryption},
```
RafiaSabih, line 143: Is this change related to this PR?

FxKu: no it isn't. That was my IDE showing an error there.

sdudoladov:
the test failure you see in this PR is addressed in #1202

pkg/cluster/cluster.go
sdudoladov: that looks not like an error but rather go best practice

pkg/cluster/sync.go
```diff
@@ -57,8 +57,8 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 		return err
 	}
 
+	c.logger.Debugf("syncing volumes using %q resize mode", c.OpConfig.StorageResizeMode)
```
sdudoladov, line 60: I believe that does not bring much to people unfamiliar with operator internals. "pvc / ebs resize" is also too generic to google. It would be better to log
```
syncing volumes using kubernetes-native PVC resize (%q mode)
```
or
```
syncing volumes by exec'ing into a pod and calling AWS API on its EBS volumes (%q mode)
```

FxKu: I've now added the word "storage" to make the link to the config option. This should be fine.

sdudoladov:
:+1:

FxKu:
:+1:

PR #1208, opened 2020-11-11
Truncate cronjob name at 52 characters

mseiwald:
K8s cronjobs must be no longer than 52 characters. This fixes the issue of backup cronjobs not being created for DBs with long names.

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -2079,7 +2079,7 @@ func (c *Cluster) generateLogicalBackupPodEnvVars() []v1.EnvVar {
 
 // getLogicalBackupJobName returns the name; the job itself may not exists
 func (c *Cluster) getLogicalBackupJobName() (jobName string) {
-	return c.OpConfig.LogicalBackupJobPrefix + c.clusterName().Name
+	return trimCronjobName(c.OpConfig.LogicalBackupJobPrefix + c.clusterName().Name)
 }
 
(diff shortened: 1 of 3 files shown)


FxKu:
Can you add a simple unit test for your new function? And maybe call it trim and not truncate?

mseiwald:
Done

Jan-M:
We see several objects where the cluster name length is a problem.

Internally we have discussed other options also to ensure uniqueness for child objects, e.g. the connection pooler deployment potentially leveraging the "uid" of the CRD object. But that also does not look nice in kubectl or other tools.

Just stripping it short may be a good start, but maybe not the best final solution.

mseiwald:
I see and understand that this is not the ideal solution. However, for us this is currently a bug which affects us and we cannot enable backups for one of our DBs without re-deploying it with a shorter name...

mseiwald:
Any news on this?

FxKu:
@mseiwald we've just merged an option to specify a shorter prefix for the logical backup jobs with #1287. I think, this should also solve your issue.

mseiwald:
That's a good workaround but there could still be situations where the name is too long.

In any case I updated my PR and fixed the merge conflicts. If you want you can still consider it. Otherwise I would suggest that we close it.

Jan-M:
:+1:

FxKu:
:+1:

FxKu:
Thanks @mseiwald for your contribution. You're right, it could happen that the name is still too long.

PR #1212, opened 2020-11-13
CRD: preserve unknown fields and add to all category

### LINKED ISSUE #1206: Postgres-operator doesn't let to create sidecars anymore

```
  sidecars:
  - env:
    - {}
    - {}
    - {}
    ports:
    - {}
    resources: {}
```

FxKu:
With #746 we switched the CRDs to `v1`. It seems that this api version does require to explicitly enable preserving unknown fields (e.g. whenever we allow full specs). Plus, the toggle moved from `crd.spec` to the schema.

This PR will also register the Postgres Operator CRDs in the `all` category, so that they are listed when `kubectl get all` is executed.

fixes #1206

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -15,6 +15,8 @@ spec:
     singular: operatorconfiguration
     shortNames:
     - opconfig
+    categories:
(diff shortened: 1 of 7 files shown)


Jan-M:
:+1:

FxKu:
:+1:

PR #1213, opened 2020-11-13
Fix connection pooler deployment selectors

Jan-M:
The currently generated deployment selectors are not correct, including too many labels.

The selector should include:

`application: db-connection-pooler`
`connection-pooler: <name><-repl>`

These two elements identify all the pods belonging to the deployment. Compare with the test manifest.

When the operator behaves properly the included e2e test should be green, and the deployment adjusted.

This PR should ideally also contain some documentation on the involved objects and their labels:
deployments and service come to mind, and the statefulset could also be mentioned, having similar selector.


pkg/cluster/connection_pooler.go
```diff
@@ -78,16 +78,25 @@ func needReplicaConnectionPoolerWorker(spec *acidv1.PostgresSpec) bool {
 // have e.g. different `application` label, so that recreatePod operation will
 // not interfere with it (it lists all the pods via labels, and if there would
 // be no difference, it will recreate also pooler pods).
-func (c *Cluster) connectionPoolerLabelsSelector(role PostgresRole) *metav1.LabelSelector {
+func (c *Cluster) connectionPoolerLabelsSelector(role PostgresRole, moreLabels bool) *metav1.LabelSelector {
```
Jan-M, line 81: Maybe we can rename this function now to not be called "Selector" - that would be more fitting

e2e/tests/test_e2e.py
```diff
@@ -165,6 +166,7 @@ def test_overwrite_pooler_deployment(self):
             }
         })
 
+        time.sleep(10)
```
Jan-M, line 169: no more sleeps, artificially slowing things down. lets talk about the issue here

pkg/cluster/connection_pooler.go
```diff
@@ -284,7 +291,7 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 
 	podTemplate := &v1.PodTemplateSpec{
 		ObjectMeta: metav1.ObjectMeta{
-			Labels:      c.connectionPoolerLabelsSelector(role).MatchLabels,
+			Labels:      c.connectionPoolerLabels(role, false).MatchLabels,
```
FxKu, line 294: question here is: Does it have to be the same list of what's in the selector? Or has the selector only be contained in the pod template label list (for watching)

manifests/minimal-fake-pooler-deployment.yaml
```diff
@@ -0,0 +1,34 @@
...
+  selector:
+    matchLabels:
+      application: db-connection-pooler
+      connection-pooler: acid-minimal-cluster-pooler
+      cluster-name: acid-minimal-cluster
+  template:
+    metadata:
+      labels:
+        application: db-connection-pooler
+        connection-pooler: acid-minimal-cluster-pooler
```
Jan-M, line 20:
```
 kubectl create -f minimal-fake-pooler-deployment.yaml
The Deployment "acid-minimal-cluster-pooler" is invalid: spec.template.metadata.labels: Invalid value: map[string]string{"application":"db-connection-pooler", "connection-pooler":"acid-minimal-cluster-pooler"}: `selector` does not match template `labels`
```

Jan-M:
:+1:

Jan-M:
:+1:

FxKu:
:+1:

PR #1022, opened 2020-06-16
Update postgresql_type.go

Ghostbaby:
when we use client-go create `PostgreSQL`,  get the below error message.

``
[...]

PR #1223, opened 2020-11-18
add to postgresql_type.go omitempty annotation

### LINKED ISSUE #1221: How to use Cluster manifest in golang?

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.5.0 - cloned to private registry and installed with helm chart
- **Type of issue?** question

Can someone explain to me how the minimal-postgres-manifest.yaml https://github.com/zalando/postgres-operator/blob/master/manifests/minimal-postgres-manifest.yaml works at all, when at the same time parameters like "clone" are actually required in the CRD: https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql.crd.yaml#L110

In the following, a little more profoundly my concrete problem, but think that this is the essential question.

---

I am currently writing an operator that manages a postgresqls.acid.zalan.do resource (and does lots of other stuff.), using the  CreateOrUpdate function https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L197

However, a `k8serrors.isInvalid(err)==true` (https://github.com/kubernetes/apimachinery/blob/master/pkg/api/errors/errors.go#L498) forced me to configure additional parameters when using "github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1" as a dependency

```
"error": "postgresql.acid.zalan.do \"***-sample-db\" is invalid: [spec.clone.cluster: Required value, spec.tls: Invalid va
lue: \"null\": spec.tls in body must be of type object: \"null\", spec.postgresql.parameters: Invalid value: \"null\": spec.postgresql.parameters in body must be of type object: \"null\", spec.standby: Invalid value: \"null\": spec.standby in body must be of type object: \"null\", spec.patroni.pg_hba: Invalid value: \"null\": spec.patroni.pg_hba in body m
ust be of type array: \"null\", spec.patroni.initdb: Invalid value: \"null\": spec.patroni.initdb in body must be of type object: \"null\", spec.serviceAnnotations: Invalid value: \"null\": spec.serviceAnnotations in body must be of type object: \"null\", spec.podAnnotations: Invalid value: \"null\": spec.podAnnotations in body must be of type object: \"n
ull\"]"}
```

...What can be created as a resource in k8s, but the object is misconfigured like the following snippet, because I have to fill in all possible additional fields.
```
r.Spec.Clone = zalando.CloneDescription{
		ClusterName: r.Name, // Self-Reference removes standby-cluster error message ?
	}
```

I tried https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/scheme.go#L346 but that doesn't work either.

Thunderbolt32:
I use the PostgreSQL type to configure a struct instance in golang to create the resource in the k8s cluster.

To create the https://github.com/zalando/postgres-operator/blob/master/manifests/minimal-postgres-manifest.yaml in golang it is necessary to add some more `omitempty` annotations to unset values.

Closes #1221

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/postgresql_type.go
+++ b/pkg/apis/acid.zalan.do/v1/postgresql_type.go
@@ -53,7 +53,7 @@ type PostgresSpec struct {
 	// load balancers' source ranges are the same for master and replica services
 	AllowedSourceRanges []string `json:"allowedSourceRanges"`
 
-	NumberOfInstances     int32                       `json:"numberOfInstances" defaults:"2"`
+	NumberOfInstances     int32                       `json:"numberOfInstances"`
 	Users                 map[string]UserFlags        `json:"users,omitempty"`
 	MaintenanceWindows    []MaintenanceWindow         `json:"maintenanceWindows,omitempty"`
 	Clone                 *CloneDescription           `json:"clone,omitempty"`
--- a/pkg/apis/acid.zalan.do/v1/postgresql_type.go
+++ b/pkg/apis/acid.zalan.do/v1/postgresql_type.go
@@ -127,7 +127,7 @@ type AdditionalVolume struct {
 // PostgresqlParam describes PostgreSQL version and pairs of configuration parameter name - values.
 type PostgresqlParam struct {
 	PgVersion  string            `json:"version"`
-	Parameters map[string]string `json:"parameters"`
+	Parameters map[string]string `json:"parameters,omitempty"`
 }
 
 // ResourceDescription describes CPU and memory resources defined for a cluster.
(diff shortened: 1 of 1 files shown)


FxKu:
Looks similar to #1022 
Can you check if there are more sections, that require `omitempty`?

Thunderbolt32:
In fact it is similar to #1022. As I was still working with v.1.5.0, I had a lot more bugs until I switched to the current master branch.

@FxKu Yes, I'll look at it again. It seems that more pointers would also be useful, because no structure values can be set on maps in Golang. However, I do not know if this would break compatibility. However, I will change it first.

Thunderbolt32:
One thing that would most likely require a code change would be to solve the "not addressable" problem by using struct Pointers (instead of structs) https://stackoverflow.com/a/32751792

However, it only seems to be seriously needed in one case
````
type PostgresSpec struct {
        ...
	PreparedDatabases     map[string]PreparedDatabase `json:"preparedDatabases,omitempty"`
        ...
}

type PreparedDatabase struct {
	PreparedSchemas map[string]PreparedSchema `json:"schemas,omitempty"`
	...
}
````

File: pkg/apis/acid.zalan.do/v1/postgresql_type.go [hunk 1]
```diff
@@ -53,8 +53,8 @@ type PostgresSpec struct {
 	// load balancers' source ranges are the same for master and replica services
 	AllowedSourceRanges []string `json:"allowedSourceRanges"`
 
-	NumberOfInstances     int32                       `json:"numberOfInstances"`
-	Users                 map[string]UserFlags        `json:"users"`
+	NumberOfInstances     int32                       `json:"numberOfInstances" defaults:"2"`
```
FxKu, line 56: would not add the default here. It's also a required field in CRD validation.

FxKu:
using pointers for the PreparedDatabase structs sound like a story for another PR. I'm also not sure how easy it is to make this backwards compatible.

Thunderbolt32:
I have removed the default value and resolved your review.

Thunderbolt32:
Is the E2E test Flaky? I do not quite understand why the test should fail after this change.

FxKu:
:+1:

Jan-M:
:+1:

PR #1228, opened 2020-11-23
preserving fields only when using k8s specs

### LINKED ISSUE #1206: Postgres-operator doesn't let to create sidecars anymore

FxKu:
The `Postgresql` CRD has been messed up by #746. This PR cleans it up. Plus for validation it replaces `additionalProperties` with `XPreserveUnknownFields: true` on the object level wherever we allow K8s specs in our custom resources.

Tries to fix #1206


e2e/tests/test_e2e.py
```diff
@@ -117,6 +117,8 @@ def setUpClass(cls):
             yaml.dump(configmap, f, Dumper=yaml.Dumper)
 
         for filename in ["operator-service-account-rbac.yaml",
+                         "postgresql.crd.yaml",
```
sdudoladov, line 120: why configure operator twice here ? once with configmap and once with CRD?

charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -589,4 +436,4 @@ spec:
           status:
             type: object
             additionalProperties:
-              type: string
+              type: string
```
Jan-M, line 439: new line

charts/postgres-operator/crds/postgresqls.yaml
```diff
@@ -609,4 +475,4 @@ spec:
           status:
             type: object
             additionalProperties:
-              type: string
\ No newline at end of file
+              type: string
```
Jan-M, line 478: new line

pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go
```diff
@@ -532,6 +532,11 @@ func (in *PostgresSpec) DeepCopyInto(out *PostgresSpec) {
 		*out = new(bool)
 		**out = **in
 	}
+	if in.EnableReplicaConnectionPooler != nil {
+		in, out := &in.EnableReplicaConnectionPooler, &out.EnableReplicaConnectionPooler
```
sdudoladov, line 536: Why was this code generated ? was there any change to the connection pooler ?

e2e/tests/test_e2e.py
FxKu: the operatorconfiguration is not used. This is only to test if the CRD can be created. Maybe I should add the sample manifest too.

pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go
FxKu: Yes, and seem the generated code wasn't updated.

e2e/tests/test_e2e.py
sdudoladov: In that case it might be a first small e2e test of its own. The whole purpose of this `setUp` class method is to prepare the infra for actual tests, not to test anything

pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go
sdudoladov: not very good, but let's commit it in this PR with a commit message acknowledging this fix

Jan-M:
:+1:

sdudoladov:
:+1:

debu99:
how to use this image?

PR #1232, opened 2020-11-25
Add Support for Custom TLS Certificates in Connection Pooler

### LINKED ISSUE #1230: PgBouncer does not use Custom TLS Certificate
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** GKE
- **Type of issue?** Bug/Feature

The Postgres Operator currently supports setting [custom TLS certificates](https://postgres-operator.readthedocs.io/en/latest/reference/cluster_manifest/#custom-tls-certificates). However, regardless of whether this option is set, the pooler (pgBouncer) generates its own self-signed TLS certificates. This is a problem when clients try to verify the presented TLS certificate.

In order to make this work, at least two things have to be done:

- The operator must mount the TLS secret into the pooler Pods
- The pooler image must use the TLS certificates that are mounted by the secret instead of generating its own certificates

As I'm not completely familiar with pgBouncer, it might also be necessary to periodically check for certificate changes and reload the Pods if something has changed. If pgBouncer implements hot reloads, this code should reside in the pooler image. Otherwise, the operator should watch the TLS secret and redeploy the pooler deployment to prevent all Pods from restarting simultaneously.

borchero:
Fixes #1230 
Dockerfile for new pgBouncer image: https://github.com/borchero/pgbouncer

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -150,6 +150,8 @@ spec:
                   numberOfInstances:
                     type: integer
                     minimum: 2
+                  deploymentAnnotations:
+                    type: object
                   resources:
                     type: object
                     required:
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -434,6 +434,9 @@ for both master and replica pooler services (if `enableReplicaConnectionPooler`
 * **resources**
   Resource configuration for connection pooler deployment.
 
+* **deploymentAnnotations**
+  Custom annotations to set on the connection pooler deployment.
+
 ## Custom TLS certificates
 
 Those parameters are grouped under the `tls` top-level key.
--- a/docs/user.md
+++ b/docs/user.md
@@ -933,3 +933,29 @@ Alternatively, it is also possible to use
 
 Certificate rotation is handled in the Spilo image which checks every 5
 minutes if the certificates have changed and reloads postgres accordingly.
+
+### Custom TLS Certificates with Connection Pooler
+
+If the postgres resource is configured with a custom TLS secret, the connection
+pooler uses that secret as well. Note that neither pgBouncer nor the operator
+handle the reload of TLS certificates as they are changed.
+
+As an alternative, the postgres resource allows setting custom annotations on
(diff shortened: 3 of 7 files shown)


FxKu:
Thank @borchero for your contribution. We have just [merged](https://github.com/zalando/postgres-operator/pull/1236) the inherited annotation feature to allow passing annotations from the postgres manifest down to child resources. You can also use the `downscaler_annotations`  option to only inherit annotations to the deployment and statefulset. I think, you can use this rather than adding yet another field. Could you strip back this PR then to only the TLS passing to pooler?

borchero:
Removing pooler-specific annotations is a hacky solution imo. The reason for these annotations is to allow external operators to reload the pooler Pods once the TLS secret changes. When using `downscaler_annotations`, my understanding is that these annotations are also added to the Postgres StatefulSet.

Unfortunately, this means that these "reloading" annotations also trigger a redeploy of the Postgres Pods although they handle hot reloading internally.

Jan-M:
@borchero I would like to see this added to the operator config, don#t you agree? this seems like a potential global config for all poolers?

borchero:
Yes, that sounds reasonable. So can I update the PR with pooler specific annotations in the operator config?

FxKu:
Sure, you can add it [here](https://github.com/zalando/postgres-operator/blob/636ba9b84624a3552182e5156b1c6d752b213c97/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go#L174).

There are a few more places where you have to reflect the change. See our short [docs](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters) on this topic.

Dayde:
Really interested in this too, in fact we consider it to be more a bugfix than a feature as we can’t use MTLS through pgBouncer because of this.
Looking at the two years old comments @FxKu I’m not sure what is still to be done.
@bchrobot have you already start to work on that ? We’re gonna need that soon and we’d be glad to lend a hand.

Dayde:
Should I open another PR ? I’ve rebased the branch but I can’t push it here.
A brief sum up of what’s left to do in your opinion would be really nice :pray:

borchero:
Closing this in favor of #2146.

PR #1235, opened 2020-11-27
raise log level for malformed secrets

sdudoladov:
Otherwise this message gets lost in the log output though when a secret is malformed.


Jan-M:
:+1:

sdudoladov:
:+1:

PR #1236, opened 2020-11-27
Support inherited annotations for all major objects

FxKu:
In preparation for an inherited annotations feature, similar to inherited labels, to estimate how much it would take.

### DIFF
--- a/pkg/cluster/connection_pooler.go
+++ b/pkg/cluster/connection_pooler.go
@@ -334,12 +334,15 @@ func (c *Cluster) generateConnectionPoolerDeployment(connectionPooler *Connectio
 		return nil, err
 	}
 
+	annotations := c.annotationsSet(nil)
+	annotations = c.AnnotationsToPropagate(annotations)
+
 	deployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:        connectionPooler.Name,
 			Namespace:   connectionPooler.Namespace,
 			Labels:      c.connectionPoolerLabels(connectionPooler.Role, true).MatchLabels,
-			Annotations: c.annotationsSet(map[string]string{}),
+			Annotations: annotations,
 			// make StatefulSet object its owner to represent the dependency.
 			// By itself StatefulSet is being deleted with "Orphaned"
 			// propagation policy, which means that it's deletion will not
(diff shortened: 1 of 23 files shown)


pkg/cluster/connection_pooler.go
```diff
@@ -334,12 +334,15 @@ func (c *Cluster) generateConnectionPoolerDeployment(connectionPooler *Connectio
 		return nil, err
 	}
 
+	annotations := c.annotationsSet(nil)
+	annotations = c.AnnotationsToPropagate(annotations)
+
```
RafiaSabih, line 339: Why not directly passing nil to AnnotationsToPropagate?

RafiaSabih, line 339: Why not, annotations := c.AnnotationsToPropagate(c.annotationsSet(nil))

pkg/cluster/connection_pooler.go
```diff
@@ -866,8 +869,9 @@ func (c *Cluster) syncConnectionPoolerWorker(oldSpec, newSpec *acidv1.Postgresql
 		}
 	}
 
-	newAnnotations := c.AnnotationsToPropagate(c.ConnectionPooler[role].Deployment.Annotations)
-	if newAnnotations != nil {
+	newAnnotations := c.annotationsSet(c.ConnectionPooler[role].Deployment.Annotations)
+	newAnnotations = c.AnnotationsToPropagate(newAnnotations)
```
RafiaSabih, line 873: Again, could be just one line like the previous comment.

pkg/cluster/util.go
```diff
@@ -271,6 +271,33 @@ func (c *Cluster) getTeamMembers(teamID string) ([]string, error) {
...
+		}
+	}
+
+	if len(annotations) > 0 {
+		return annotations
+	}
+
+	return nil
+}
+
```
RafiaSabih, line 300: Not sure if returning nil is used anywhere, can simply return annotations.

FxKu: it should return nil if there are no Annotations defined in the config and on the resource. For example `TestGeneratePodDisruptionBudget` fails for me if I return an empty map as this test does not define any Annotations at all. I have also added this behavior to the `AnnotationsToPropagate` function, too.

FxKu:
:+1:

RafiaSabih:
:+1:

PR #1242, opened 2020-12-07
Support EBS gp2 to gp3 migration on sync for below 1tb volumes

Jan-M:
Support migration of existing gp2 volumes below a certain threshold to be migrated to gp3.


Jan-M:
:+1:

pkg/cluster/cluster.go
```diff
@@ -89,7 +90,10 @@ type Cluster struct {
 	processMu        sync.RWMutex // protects the current operation for reporting, no need to hold the master mutex
 	specMu           sync.RWMutex // protects the spec for reporting, no need to hold the master mutex
 	ConnectionPooler map[PostgresRole]*ConnectionPoolerObjects
+	EBSVolumes       map[string]volumes.VolumeProperties
```
FxKu, line 93:
```suggestion
	EBSVolumes       map[string]*volumes.VolumeProperties
```
not sure, but since we had a similar case with PreparedDatabases recently, maybe this should be with pointer?

pkg/cluster/sync.go
```diff
@@ -55,7 +54,23 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
...
-	if c.OpConfig.StorageResizeMode == "pvc" {
+
+	if c.OpConfig.EnableEBSGp3Migration {
+		err = c.executeEBSMigration()
+		if nil != err {
+			return err
+		}
+	}
+
+	if c.OpConfig.StorageResizeMode == "mixed" {
```
FxKu, line 65: I fail to see the difference to `c.OpConfig.StorageResizeMode == "pvc"`. If there is another case for `StorageResizeMode`, it should be documented.

pkg/cluster/volumes.go
```diff
@@ -119,19 +118,26 @@ func (c *Cluster) listPersistentVolumes() ([]*v1.PersistentVolume, error) {
...
 
 // resizeVolumes resize persistent volumes compatible with the given resizer interface
-func (c *Cluster) resizeVolumes(newVolume acidv1.Volume, resizers []volumes.VolumeResizer) error {
-	c.setProcessName("resizing volumes")
+func (c *Cluster) resizeVolumes() error {
+	if c.VolumeResizer == nil {
+		return fmt.Errorf("no volume resizer set for EBS volume handling")
+	}
+
+	c.setProcessName("resizing EBS volumes")
```
FxKu, line 126: too AWS-specific, no? Might be irritating to see this process name on GCP, Azure etc.

pkg/cluster/cluster.go
Jan-M: Nope, Cluster object is never generated anywhere. No pointer needed here.

The pointer non pointer is for K8s API/Client objects.

pkg/cluster/sync.go
Jan-M: Future work, we need "mixed" to do call to EBS for volume changes and rely on PVC to do file system extension without privileged pods.

pkg/cluster/volumes.go
Jan-M: Yes, exactly.

pkg/cluster/volumes_test.go
```diff
@@ -169,3 +175,132 @@ func TestQuantityToGigabyte(t *testing.T) {
 		}
 	}
 }
+
+func CreatePVCs(namespace string, clusterName string, labels labels.Set, n int, size string) v1.PersistentVolumeClaimList {
```
FxKu, line 179: we could reuse this for the already existing `TestResizeVolumeClaim` test

FxKu:
I will add the new config fields to CRD validation and helm chart and update `volume_test.go` to reuse your `CreatePVC` method in other tests.

Jan-M:
:+1:

FxKu:
:+1:

PR #1248, opened 2020-12-10
replace AdditionalProperties with XPreserveUnknownFields FIXES  #1206

### LINKED ISSUE #1206: Postgres-operator doesn't let to create sidecars anymore

yanchenko-igor:
This patch fixes the bug I reported before #1206


yanchenko-igor:
Can you help me with the test, I don't understand why it failed.

FxKu:
Ah, I've only changed the validation for OperatorConfiguration CRD, but not Postgres CRD. Thanks. Not sure, what coveralls is complaining. We merge it anyway.

FxKu:
:+1:

RafiaSabih:
:+1:

PR #1256, opened 2020-12-15
Sync sts at pgversion upgrade

RafiaSabih:
Like the other resources, services, etc. sync statefulset when postgres version is upgraded.


pkg/cluster/cluster.go
```diff
@@ -623,6 +623,10 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
 	} else if oldSpec.Spec.PostgresqlParam.PgVersion < newSpec.Spec.PostgresqlParam.PgVersion {
 		c.logger.Infof("postgresql version increased (%q -> %q), major version upgrade can be done manually after StatefulSet Sync",
 			oldSpec.Spec.PostgresqlParam.PgVersion, newSpec.Spec.PostgresqlParam.PgVersion)
+		if err := c.syncStatefulSet(); err != nil {
+			c.logger.Errorf("could not sync statefulSet: %v", err)
+			updateFailed = true
+		}
```
FxKu, line 629: c.syncStatefulSet() might be called again when comparing annotations further down. Maybe set a flag here to sync it only once?

FxKu:
:+1:

RafiaSabih:
:+1:

PR #1260, opened 2020-12-16
[UI] reflect new backup paths and cluster status

FxKu:
* reflect new backup path that got introduced with PG_VERSION env variable
* display cluster status to easier spot a failed setup
* do not print K8s managed fields in manifest
* update cluster costs with min number of pods


Jan-M:
:+1:

FxKu:
:+1:

PR #1263, opened 2020-12-17
use pointer type for nodeAffinity

FxKu:
if NodeAffinity is not specified the current code will put an empty affinity on the statefulset which can cause a rolling update if there was no `node_readiness_label` configured so far. With the pointer type the affinity will be omitted if not specified.

Plus, two minor fixes where go compiler complained (exported function with wrong comment and variable name with underscore)


FxKu:
:+1:

Jan-M:
:+1:

PR #1266, opened 2020-12-18
fix ui service port and update quickstart

FxKu:
following UI setup in quickstart fails at port-forward when using helm chart, as this one uses port 8081 for the service. It was changed for a [reason](https://github.com/zalando/postgres-operator/pull/1096), but not reflected in the chart. Use a simple addition to quickstart to remove OLM section before people will start asking for this again.


FxKu:
:+1:

Jan-M:
:+1:

PR #1268, opened 2020-12-21
Logicalbackup switcher

SanjeevChoubey:
This PR will handle when the user wanted to run the logical backup on the master. By default, it will run as it is working currently. If LOGICAL_BACKUP_ON_MASTER will be true then only it will run on master.


FxKu:
We want to keep the current behavior.

PR #1274, opened 2020-12-28
Avoid syncing when possible

RafiaSabih:
Do not sync connection pooler when there are no changes in pooler configurations.

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -275,7 +275,7 @@ class EndToEndTestCase(unittest.TestCase):
         self.eventuallyEqual(lambda: k8s.count_secrets_with_label('application=db-connection-pooler,cluster-name=acid-minimal-cluster'),
                              1, "Secret not created")
 
-        scale up connection pooler deployment
+        # scale up connection pooler deployment
         k8s.api.custom_objects_api.patch_namespaced_custom_object(
             'acid.zalan.do', 'v1', 'default',
             'postgresqls', 'acid-minimal-cluster',
@@ -292,7 +292,7 @@ class EndToEndTestCase(unittest.TestCase):
         self.eventuallyEqual(lambda: k8s.count_running_pods("connection-pooler=acid-minimal-cluster-pooler"),
                              3, "Scale up of pooler pods does not work")
 
-        turn it off, keeping config should be overwritten by false
+        # turn it off, keeping config should be overwritten by false
         k8s.api.custom_objects_api.patch_namespaced_custom_object(
             'acid.zalan.do', 'v1', 'default',
             'postgresqls', 'acid-minimal-cluster',
@@ -310,9 +310,9 @@ class EndToEndTestCase(unittest.TestCase):
         self.eventuallyEqual(lambda: k8s.count_secrets_with_label('application=spilo,cluster-name=acid-minimal-cluster'),
                              4, "Secrets not deleted")
 
-        Verify that all the databases have pooler schema installed.
-        Do this via psql, since otherwise we need to deal with
-        credentials.
+        # Verify that all the databases have pooler schema installed.
+        # Do this via psql, since otherwise we need to deal with
+        # credentials.
         dbList = []
 
         leader = k8s.get_cluster_leader_pod('acid-minimal-cluster')
@@ -362,7 +362,7 @@ class EndToEndTestCase(unittest.TestCase):
         else:
             print('Could not find leader pod')
 
-        remove config section to make test work next time
+        # remove config section to make test work next time
         k8s.api.custom_objects_api.patch_namespaced_custom_object(
             'acid.zalan.do', 'v1', 'default',
             'postgresqls', 'acid-minimal-cluster',
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -275,44 +275,44 @@ class EndToEndTestCase(unittest.TestCase):
         self.eventuallyEqual(lambda: k8s.count_secrets_with_label('application=db-connection-pooler,cluster-name=acid-minimal-cluster'),
                              1, "Secret not created")
 
-        # scale up connection pooler deployment
-        # k8s.api.custom_objects_api.patch_namespaced_custom_object(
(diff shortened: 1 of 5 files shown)


e2e/tests/test_e2e.py
```diff
@@ -221,6 +221,8 @@ def test_enable_disable_connection_pooler(self):
         self.eventuallyEqual(lambda: k8s.count_services_with_label(
                             'application=db-connection-pooler,cluster-name=acid-minimal-cluster'),
                             2, "No pooler service found")
+        self.eventuallyEqual(lambda: k8s.count_secrets_with_label('application=spilo,cluster-name=acid-minimal-cluster'),
+                             5, "Secret not created")
```
FxKu, line 225: Can you maybe include the `db-connection-pooler` label to only get the secret of interest? Just in case the number of the other secrets changes we don't have to care about the code here.

pkg/cluster/connection_pooler.go
```diff
@@ -681,13 +681,31 @@ func logPoolerEssentials(log *logrus.Entry, oldSpec, newSpec *acidv1.Postgresql)
...
+			},
+		}
+	}
+
+	needSync, _ := needSyncConnectionPoolerSpecs(oldSpec.Spec.ConnectionPooler, newSpec.Spec.ConnectionPooler, c.logger)
+	masterChanges, _ := diff.Diff(oldSpec.Spec.EnableConnectionPooler, newSpec.Spec.EnableConnectionPooler)
+	replicaChanges, _ := diff.Diff(oldSpec.Spec.EnableReplicaConnectionPooler, newSpec.Spec.EnableReplicaConnectionPooler)
+
+	if !needSync && len(masterChanges) <= 0 && len(replicaChanges) <= 0 {
+		c.logger.Warningln("no need for pooler sync")
```
FxKu, line 703: Debug level should be fine, I think.

pkg/cluster/connection_pooler_test.go
```diff
@@ -366,9 +366,7 @@ func TestConnectionPoolerSynchronization(t *testing.T) {
 		{
 			subTest: "create if doesn't exist",
 			oldSpec: &acidv1.Postgresql{
-				Spec: acidv1.PostgresSpec{
-					ConnectionPooler: &acidv1.ConnectionPooler{},
-				},
```
FxKu, line 371: This change I don't get. Now it's the same like the first unit test. Shouldn't the operator try to create the connection pooler when it's not there - even if the old spec says so? I think this was the original idea.

pkg/cluster/connection_pooler.go
```diff
@@ -539,13 +539,13 @@ func updateConnectionPoolerAnnotations(KubeClient k8sutil.KubernetesClient, depl
...
 // sync if there is any difference.
-func needSyncConnectionPoolerSpecs(oldSpec, newSpec *acidv1.ConnectionPooler) (sync bool, reasons []string) {
+func needSyncConnectionPoolerSpecs(oldSpec, newSpec *acidv1.ConnectionPooler, logger *logrus.Entry) (sync bool, reasons []string) {
 	reasons = []string{}
 	sync = false
 
 	changelog, err := diff.Diff(oldSpec, newSpec)
 	if err != nil {
-		//c.logger.Infof("Cannot get diff, do not do anything, %+v", err)
+		logger.Infof("Cannot get diff, do not do anything, %+v", err)
```
Jan-M, line 548: `cannot`

pkg/cluster/connection_pooler.go
```diff
@@ -539,13 +539,13 @@ func updateConnectionPoolerAnnotations(KubeClient k8sutil.KubernetesClient, depl
...
 // sync if there is any difference.
-func needSyncConnectionPoolerSpecs(oldSpec, newSpec *acidv1.ConnectionPooler) (sync bool, reasons []string) {
+func needSyncConnectionPoolerSpecs(oldSpec, newSpec *acidv1.ConnectionPooler, logger *logrus.Entry) (sync bool, reasons []string) {
 	reasons = []string{}
 	sync = false
 
 	changelog, err := diff.Diff(oldSpec, newSpec)
 	if err != nil {
-		//c.logger.Infof("Cannot get diff, do not do anything, %+v", err)
+		//logger.Infof("Cannot get diff, do not do anything, %+v", err)
```
sdudoladov, line 548: First, it is dead code. Second, it makes the `logger` parameter in this function useless.

pkg/cluster/connection_pooler.go
```diff
@@ -681,13 +681,39 @@ func logPoolerEssentials(log *logrus.Entry, oldSpec, newSpec *acidv1.Postgresql)
...
+	if oldSpec == nil {
+		oldSpec = &acidv1.Postgresql{
+			Spec: acidv1.PostgresSpec{
+				ConnectionPooler: &acidv1.ConnectionPooler{},
+			},
+		}
+	}
+
+	needSync, _ := needSyncConnectionPoolerSpecs(oldSpec.Spec.ConnectionPooler, newSpec.Spec.ConnectionPooler, c.logger)
+	masterChanges, _ := diff.Diff(oldSpec.Spec.EnableConnectionPooler, newSpec.Spec.EnableConnectionPooler)
```
sdudoladov, line 699: Please do not ignore errors returned by `diff.Diff`:
```golang
// Diff returns a changelog of all mutated values from both
func Diff(a, b interface{}) (Changelog, error) {
```

pkg/cluster/k8sres.go
```diff
@@ -1561,11 +1561,17 @@ func (c *Cluster) generateSingleUserSecret(namespace string, pgUser spec.PgUser)
 	}
 
 	username := pgUser.Name
+	lbls := c.labelsSet(true)
+
+	if username == "pooler" {
```
sdudoladov, line 1566: Please use the `ConnectionPoolerUserName` constant from `pkg/util/constants/pooler.go` and not the hard-coded value.

sdudoladov (commented):
> Do not sync connection pooler when there are no changes in pooler configurations.

Why does this extra pooler Sync present a problem ?

sdudoladov:
:+1:

RafiaSabih:
:+1:

PR #1278, opened 2020-12-30
Add missing configuration

possani:
_logical_backup_provider_ configuration was added in [PR](https://github.com/zalando/postgres-operator/pull/873) to support GCS. However, it missed the configmap which is still an option when deploying the operator. 

Without it, the users will get the following error:
```
+ case $LOGICAL_BACKUP_PROVIDER in
/dump.sh: line 56: LOGICAL_BACKUP_PROVIDER: unbound variable 
```

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -62,6 +62,7 @@ data:
   # inherited_labels: application,environment
   # kube_iam_role: ""
   # log_s3_bucket: ""
+  logical_backup_provider: "s3"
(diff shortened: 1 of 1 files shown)


FxKu:
Thanks @possani for raising this. This new option as well as the GCS credentials path are indeed missing. I've created a separate PR (#1282) to cover everything. However, this is only about examples. Internally the operator default to `s3` so I would expect it to set the `LOGICAL_BACKUP_PROVIDER` environment variable. Are you maybe using the new logical backup image with an older operator version?

PR #1282, opened 2021-01-04
add logical backup provider and gcs examples

FxKu:
Two new logical backup options were introduced with #1173 but the manifest examples as well as the helm chart values files do not reflect it. However, these are just example files. Functionality and validation is there and should work.

closes #1278


Jan-M:
:+1:

FxKu:
:+1:

Issue #1277, opened 2020-12-29
[1.6.0] cannot convert int64 to float64 


vitobotta:
- **Which image of the operator are you using?** e.g. registry.opensource.zalan.do/acid/postgres-operator:v1.6.0
1.6.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** [AWS K8s | GCP ... | Bare Metal K8s]
Self managed Rancher cluster
- **Are you running Postgres Operator in production?** [yes | no]
Yes, but I am testing the upgrade from 1.5.0 to 1.6.0 in a throwaway cluster
- **Type of issue?** [Bug report, question, feature request, etc.]
I think bug

Hi,

I am testing the upgrade of the operator from 1.5.0 to 1.6.0. I actually use Pulumi to install the Helm chart but it's like installing manually basically - I have been using this for a while with no problems.

When I try to install 1.6.0 the upgrade fails because of "convert int64 to float64 " errors when updating the CRDs. I can't figure out which fields are causing this. I have commented out values like workers / min_instances / max_instances so to use the defaults according to an old, similar issue, but it doesn't make any difference. How can I find which fields are causing the problem?

These are the values I am using for the chart:

```
  const customValues = {
      configTarget: "OperatorConfigurationCRD",
      configKubernetes: {
        enable_pod_antiaffinity: true,
        pod_environment_configmap: "postgres-pod-config",
        watched_namespace: "*",
        enable_init_containers: true,
        enable_pod_disruption_budget: true,
        enable_sidecars: true,
        spilo_privileged: false
      },
      configAwsOrGcp: {
        aws_region: s3Region,
        aws_endpoint: s3Endpoint,
        wal_s3_bucket: s3Bucket
      },
      configLoadBalancer: {
        enable_master_load_balancer: false,
        enable_replica_load_balancer: false
      },
      configDebug: {
        debug_logging: true,
        enable_database_access: true
      },
      configLogicalBackup: {
        logical_backup_docker_image: logicalBackupsImage,
        logical_backup_s3_access_key_id: s3AccessKeyId,
        logical_backup_s3_bucket: s3Bucket,
        logical_backup_s3_region: s3Region,
        logical_backup_s3_endpoint: s3Endpoint,
        logical_backup_s3_secret_access_key: s3SecretAccessKey,
        logical_backup_s3_sse: logicalBackupS3SSE,
        logical_backup_schedule: logicalBackupDefaultSchedule
      },
      configGeneral: {
[...]

PR #1283, opened 2021-01-04
add default values to operatorconfiguration crd

FxKu:
since v1 and K8s 1.17 allows defining defaults in the CRD schema validation. With this we could at some point get gid of all the coalescing in [operator_config.go](https://github.com/zalando/postgres-operator/blob/master/pkg/controller/operator_config.go).

For compatibility with older K8s versions a coalesce function for `int64` is added. The [go code](https://github.com/zalando/postgres-operator/blob/master/pkg/apis/acid.zalan.do/v1/crds.go) for the CRD validation does not use defaults at the moment because values have to be defined as `*JSON` (which looks like too much work right now - and maybe we drop this code at some point anyway to only rely on manifests).

This PR incl. a few more minor changes:
- adding `logical_backup_job_prefix` to CRD validation and all examples
- adding `v1.6.0` tag to every mention of the logical backup image
- adding `master-12` tag to default pooler image
- setting minimum of `connection_pooler_number_of_instances` to 1

addresses #1277


FxKu:
:+1:

sdudoladov:
:+1:

PR #1083, opened 2020-07-30
Set user and group in security context

ricoberger:
Since version 1.6-p3 of the Spilo image it is possible to run the processes inside the container without root privileges. This PR introduces two new options, so a user can run the container without root and without the need to build a custom image.

The two new options for the security context, can be set via ConfigMap or the CR:

- `spiloRunAsUser` / `spilo_runasuser`: Set the user ID in the security context
- `spiloRunAsGroup` / `spilo_runasGroup`: Set the group ID in the security context

This PR addresses #1052.

Issue #1052, opened 2020-07-08
Set Security Context in postgresql manifest


connorearl:
I don't see any way to set a securitycontext in the postgresql manifest so I can run the pods not as root. I know Spilo has support from rootless containers now, but without making a custom spilo image I don't have a way setting this. I also would like to implement Pod Security Policies and will need the pods to conform to them.

ReSearchITEng:
Hi @connorearl ,
Well, this should be ok already now.
E.g. OpenShift  allocates the users and groups dynamically (based on scc), and their range is different in every namespace.
And this operator, using the latest spilo (developed exactly for this) it's working without any issues.
Do not have to set any parameter like spilo_fsgroup or like spiloFSGroup in either operator or cluster request.
While you may not be looking exactly for openshift, by searching  in the docs & discussions on this project for openshift should clarify few things.

marcoslarsen:
Hello! We're using v1.5.0 of the operator, and we require to run the pods as non root user. The mentioned PR seems to fix the issue, but it's not available on 1.5.0. We cannot upgrade to 1.6 yet because of kubernetes version (we're using 1.15 and we cannot update it yet). Is there a way to apply this to 1.5 without having to to build a custom image? Thanks!

ckotzbauer:
Hi,
are there any news on this? The Spilo-Image already runs as a non-root user, but the flag `securityContext.runAsNonRoot` cannot be set to `true`.
Thanks!

kndoni:
Hello is anyone still working on this topic?

PR #1286, opened 2021-01-05
Patch v1.5.0 to set user and group in security context

marcoslarsen:
I've created this draft PR to address the comment I left on https://github.com/zalando/postgres-operator/issues/1052 .
This is basically a cherry pick of  #1083 applied to version 1.5.0. It would be great to have this patch version.


FxKu:
We cannot update existing images. Is it maybe only the CRD jump to v1 which doesn't work on K8s 1.15?
I've went through our build history to a place before we merged this one, but after the PR you mentioned was merged. Maybe try with this image:
```
registry.opensource.zalan.do/acid/postgres-operator:v1.5.0-55-gd9f5d1c-dirty
```

marcoslarsen:
Yes, the issue is with the crd version bump to v1. I was thinking not to update existing images but create a new one with version 1.5.1. But if the one you mentioned already supports this I'll give it a try! Thanks! :)

marcoslarsen:
One more note, I've opened this as a draft PR because i haven't found an existing branch that made sense to do this.... Maybe a new branch for version 1.5.1 if it made sense. My intention wasn't to update the existing release branch

FxKu:
I think this is obsolete by now.

PR #1287, opened 2021-01-05
add a prefix for the name of a logical backup job

sdudoladov:
Operator currently prepends the `logical-backup-` prefix to the name of  a CronJob doing the backup.
The prefix proved to be too long as it counts towards the name length limit imposed by k8s.
This patch makes the prefix configurable.

Automatic job renaming after prefix change is out of scope. Operator will automatically create a new job with the updated  prefix. The old job will stay around and has to be deleted manually. Logical backup jobs for `$CLUSTERNAME` are visible with
```bash
kubectl get cronjobs -l application=spilo,cluster-name=$CLUSTERNAME
``` 
Operator currently does not assign any specific labels to the jobs.

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -571,6 +571,9 @@ grouped under the `logical_backup` key.
   S3 bucket to store backup results. The bucket has to be present and
   accessible by Postgres pods. Default: empty.
 
+* **logical_backup_job_prefix**
+  The prefix to be prepended to the name of a k8s CronJob running the backups. Beware the prefix counts towards the name length restrictions imposed by k8s. Empty string is a legitimate value. Operator does not do the actual renaming: If you change the prefix, you will have to recreate respective CronJobs manually. Default: "logical-backup-".
+
 * **logical_backup_s3_region**
   Specifies the region of the bucket which is required with some non-AWS S3 storage services. The default is empty.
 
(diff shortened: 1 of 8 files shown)


manifests/operatorconfiguration.crd.yaml
```diff
@@ -334,6 +334,8 @@ spec:
                   logical_backup_schedule:
                     type: string
                     pattern: '^(\d+|\*)(/\d+)?(\s+(\d+|\*)(/\d+)?){4}$'
+                  logical_backup_job_prefix:
+                    type: string
```
FxKu, line 338: can you stick to alphabetical sorting here and in the sample configuration file? Maybe docs, too. Makes things easier to find.

sdudoladov: We did not follow alphabetical order in the logical backup settings at all. I restored the order in a separate commit.

docs/reference/operator_parameters.md
```diff
@@ -571,6 +571,9 @@ grouped under the `logical_backup` key.
   S3 bucket to store backup results. The bucket has to be present and
   accessible by Postgres pods. Default: empty.
 
+* **logical_backup_job_prefix**
+  The prefix to be prepended to the name of a k8s CronJob running the backups. Beware the prefix counts towards the name length restrictions imposed by k8s. Empty string is a legitimate value. Operator does not do the actual renaming: If you change the prefix, you will have to recreate respective CronJobs manually. Default: "logical-backup-".
```
Jan-M, line 575: I think wording is off. The operator creates a new cron job with the proper name, but does not delete the old named one. So users would need to delete old job or end up with two backups.

Correct?

sdudoladov: yes. I already updated the Pr description and working on the docs right now

Jan-M:
:+1:

sdudoladov:
:+1:

PR #1290, opened 2021-01-06
raise interval for node affinity test

FxKu:
currently e2e tests in Gh actions fail mostly on this test, although they are passing in our internal pipeline.


sdudoladov:
:+1:

PR #1293, opened 2021-01-07
wrap getting Patroni state into retry

sdudoladov:
e2e tests reveal operator is sometimes too fast to fetch  Patroni state. A few retires can help to alleviate the issue.


sdudoladov:
:+1:

Jan-M:
:+1:

PR #1297, opened 2021-01-08
[UI] on read_pods filter only spilo pods

FxKu:
Pods from enabled connection pooler also receive the `spilo-role=master` label. Therefore, only fetching pods by the cluster name would return too many labeled masters hence showing in the UI that the master isn't ready.


FxKu:
:+1:

Jan-M:
:+1:

PR #1301, opened 2021-01-11
Use fake client  for connection pooler

RafiaSabih:
Test for creation and deletion of connection pooler, using k8s fake client API.


FxKu:
was their a discussion to put it in an extra file an not add it to `connection_pooler_test.go`?

pkg/cluster/connection_pooler_test.go
```diff
@@ -210,133 +251,178 @@ func TestNeedConnectionPooler(t *testing.T) {
...
-					},
+	var cluster = New(
+		Config{
+			OpConfig: config.Config{
+				ConnectionPooler: config.ConnectionPooler{
+					ConnectionPoolerDefaultCPURequest:    "100m",
+					ConnectionPoolerDefaultCPULimit:      "100m",
+					ConnectionPoolerDefaultMemoryRequest: "100Mi",
+					ConnectionPoolerDefaultMemoryLimit:   "100Mi",
+					NumberOfInstances:                    int32ToPointer(1),
```
sdudoladov, line 398: Why `int32ToPointer(1)` instead of simply `1` ?

RafiaSabih: because NumberOfInstances is of type *int32...!

sdudoladov:
:+1:

RafiaSabih:
:+1:

PR #1308, opened 2021-01-12
Added imagePullSecrets option for spilo pod

dervoeti:
I wanted to deploy my own spilo image in a cluster, the image is hosted in a private registry. Therefore i added an option to specify imagePullSecrets for the pods running spilo. It can be specified in the operator configuration as well as in the cluster manifest (in CRDs and/or ConfigMap).
I have built the operator image and ran it in my cluster, it works fine there (tested with and without imagePullSecrets). The change should be backwards compatible. A bit of documentation and examples are added as well.

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -191,6 +191,17 @@ spec:
                 # Note: usernames specified here as database owners must be declared in the users key of the spec key.
               dockerImage:
                 type: string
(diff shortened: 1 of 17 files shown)


HusseinKabbout:
Since this PR is almost 3 years old, I don't think it can be reviewed / merged as a lot has probably changed since then. Are the maintainers even open for such a change? I would be open to help but I don't want to invest time in a feature that the maintainers are not interested in.

PR #1314, opened 2021-01-14
allow additional members from other teams

FxKu:
`getTeamMembers` should also work for other teams associated to a cluster, e.g. superusers, additional teams etc.


Jan-M:
:+1:

FxKu:
:+1:

PR #1319, opened 2021-01-19
add host info to connection docs

### LINKED ISSUE #1316: psql: error: could not connect to server: No such file or directory

- **Which image of the operator are you using?** n/a
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** K3s via K3D
- **Type of issue?** Documentation issue

I followed the quick start and, since I'm not using Minikube, dropped into the "Connect to PostgreSQL" section of the "User Guide" which suggests to simply connect to Postgres as follows once `PGPASSWORD` and `PGSSLMODE` are set:

```sh
psql -U postgres -p 6432
```

Resulting in the following error:

```term
psql: error: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/tmp/.s.PGSQL.6432"?
```

My port-forward was running so I did a:

```
kubectl exec -it acid-minimal-cluster-0 -- bash
```

And confirmed I was able to access the database that way. Then I tried connecting again from outside the Pod using SQLPro for Postgres and it also couldn't connect. Then I tried adding the host like to connect from command line outside the Pod:

```sh
psql -U postgres -p 6432 -h localhost
```

And receive the same error mentioned above:

```term
psql: error: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/tmp/.s.PGSQL.6432"?
```

At which point I stopped and restarted my port forward and was only then able to connect. This was a bit confusing as quickstarts should generally be, well, quick. So my suggestions:

- [ ] Add the `-h` flag to the `pgsql` connect command for both the User Guide and README
- [ ] Explain to the user they are supposed to run the connect command outside the pod (it's not obvious)
- [ ] Note that running `psql` from outside the pod requires `psql` installed (somewhat obvious but not completely)

As to why the port forwarder acts flaky when attempting to connect without `-h` I cannot say. But what I can say as that if users cannot easily finish the quickstart they're likely to bail on using the operator.

FxKu:
closes #1316 

the quickstart is not changed as the host variable (PGHOST) is set there - so `-h` is not necessary

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -71,26 +71,26 @@ kubectl describe postgresql acid-minimal-cluster
 ## Connect to PostgreSQL
 
 With a `port-forward` on one of the database pods (e.g. the master) you can
-connect to the PostgreSQL database. Use labels to filter for the master pod of
-our test cluster.
+connect to the PostgreSQL database from your machine. Use labels to filter for
+the master pod of our test cluster.
 
 ```bash
 # get name of master pod of acid-minimal-cluster
-export PGMASTER=$(kubectl get pods -o jsonpath={.items..metadata.name} -l application=spilo,cluster-name=acid-minimal-cluster,spilo-role=master)
+export PGMASTER=$(kubectl get pods -o jsonpath={.items..metadata.name} -l application=spilo,cluster-name=acid-minimal-cluster,spilo-role=master -n default)
 
 # set up port forward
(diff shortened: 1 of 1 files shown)


FxKu:
:+1:

Jan-M:
:+1:

PR #1326, opened 2021-01-22
set AllowPrivilegeEscalation on container securityContext

FxKu:
in the process of switching all pods to non-privileged pods the `use` privilege for a `privileged` PodSecurityPolicy is obsolete. However, Pods holding this privilege before might still escalate to such a PSP and will then face issues when the psp privilege is missing in the RBAC. Setting `AllowPrivilegeEscalation: false` on the container's securityContext should prevent this. The value will be based on what's configured for `spilo_privileged`.


Jan-M:
:+1:

FxKu:
:+1:

PR #1328, opened 2021-01-25
set allowPrivilegeEscalation for deployment templates

FxKu:
#1326 has added `allowPrivilegeEscalation` field to postgres containers, but operator and pooler containers might need it, too, since the service account they have used so far holds or has hold the `use` privilege for a `privileged` PSP.

This PR also introduces configuring the operator container's `securityContext` in the helm chart.

### DIFF
--- a/pkg/cluster/connection_pooler.go
+++ b/pkg/cluster/connection_pooler.go
@@ -280,6 +280,9 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 				},
 			},
 		},
+		SecurityContext: &v1.SecurityContext{
+			AllowPrivilegeEscalation: util.False(),
+		},
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/cluster/connection_pooler.go
```diff
@@ -289,7 +292,6 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 			Annotations: c.annotationsSet(c.generatePodAnnotations(spec)),
 		},
 		Spec: v1.PodSpec{
-			ServiceAccountName:            c.OpConfig.PodServiceAccountName,
```
sdudoladov, line 292: Which service account will connection pooler pods use ? Are there any special privileges required for the pods ?

sdudoladov:
:+1:

FxKu:
:+1:

PR #1341, opened 2021-02-01
Improve rolling upgrades and rolling upgrade continue

FxKu:
### Problem

We have seen scenarios where a cluster ends up in a loop of failovers on every sync event. This can happen if the recreated replica doesn't receive it's label after a wait timeout. We've added code to not recreate the pods when Patroni says they are still rebuilding. But still, this does not prevent the operator from terminating that pod again if the rolling update did not complete during one sync event.

The source of the problem lies upon managing the rolling update based on a flag (annotation) on the statefulset. If you not all pods could be rotated during between the next sync, it remains `true`, although some of the pods might already been updated.

### Idea

Have the rolling update annotation flag on the pod level rather than on the statefulset, that is patched to `true` when a template diff reveals changes we want to consider for a rolling update (`needsRollUpdate`). A recreated pod will have the flag set to false again.

### Final solution

As there will be not annotation on the statefulset, recreated Pod's will not have the flag - hence there's no code where the flag is set to false. The `recreatePods` method was changed to allow for passing a list of pods to roll. The function will not list pods anymore. The sync code should already now about all pods to roll and possible replication targets from cached/unfinished updates.

Tested cases:
- [x] flag on both pods = normal rolling update
- [x] flag added only on master = artificial cashed unfinished rolling update
- [x] flag only on one pod because recreation did not happen soon enough and SYNC failed


pkg/cluster/k8sres.go
```diff
@@ -1305,6 +1306,7 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
 
 func (c *Cluster) generatePodAnnotations(spec *acidv1.PostgresSpec) map[string]string {
 	annotations := make(map[string]string)
+	// TODO annotations[rollingUpdatePodAnnotationKey] = strconv.FormatBool(false)
```
Jan-M, line 1309: I am likely not a fan of this approach.

I suggest we have a good point, where we explicitly "patch" the pods.

pkg/cluster/pod.go
```diff
@@ -384,6 +384,7 @@ func (c *Cluster) recreatePods() error {
 
 	if masterPod != nil {
 		// failover if we have not observed a master pod when re-creating former replicas.
+		// TODO if masterPod has no rolling update label anymore skip switchover too
```
Jan-M, line 387: Can we write down the idea of this change? how will the new strategy look like?

pkg/cluster/pod.go
```diff
@@ -45,6 +47,93 @@ func (c *Cluster) getRolePods(role PostgresRole) ([]v1.Pod, error) {
 	return pods.Items, nil
 }
 
+// enableRollingUpdateFlagForPod sets the indicator for the rolling update requirement
+// in the Pod annotation.
+func (c *Cluster) enableRollingUpdateFlagForPod(pod v1.Pod, msg string) error {
```
Jan-M, line 52: setRollingUpdateFlag(pod, msg, value) {}

can potentially include value so it could be set to false despite not being used.

Jan-M: `markForRollingUpgrade`

pkg/cluster/pod.go
```diff
@@ -45,6 +47,93 @@ func (c *Cluster) getRolePods(role PostgresRole) ([]v1.Pod, error) {
...
+	c.logger.Debugf("enable rolling update annotation for %s: reason %s", pod.Name, msg)
+	flag := make(map[string]string)
+	flag[rollingUpdatePodAnnotationKey] = strconv.FormatBool(true)
+
+	patchData, err := metaAnnotationsPatch(flag)
+	if err != nil {
+		return fmt.Errorf("could not form patch for pod's rolling update flag: %v", err)
+	}
+
+	_, err = c.KubeClient.Pods(pod.Namespace).Patch(
```
Jan-M, line 62: could have 1-3 retries

pkg/cluster/pod.go
```diff
@@ -45,6 +47,93 @@ func (c *Cluster) getRolePods(role PostgresRole) ([]v1.Pod, error) {
...
+	}
+
+	return nil
+}
+
+// enableRollingUpdateFlagForPods sets the indicator for the rolling update requirement
+// on pods that do not have it yet
+func (c *Cluster) enableRollingUpdateFlagForPods(pods []v1.Pod, msg string) error {
+	for _, pod := range pods {
+		if c.getRollingUpdateFlagFromPod(&pod, true) {
```
Jan-M, line 80: should be `false`, true would not mark all pods.

pkg/cluster/sync.go
```diff
@@ -304,60 +306,86 @@ func (c *Cluster) syncStatefulSet() error {
...
 			//  (b) some of the pods were not restarted when the lazy update was still in place
-			podsRollingUpdateRequired, err = c.mustUpdatePodsAfterLazyUpdate(desiredSS)
-			if err != nil {
-				return fmt.Errorf("could not list pods of the statefulset: %v", err)
+			for _, pod := range pods {
+				effectivePodImage := pod.Spec.Containers[0].Image
+				stsImage := desiredSts.Spec.Template.Spec.Containers[0].Image
+
+				if stsImage != effectivePodImage {
+					if err = c.markRollingUpdateFlagForPod(&pod, "pod not yet restarted due to lazy update"); err != nil {
```
sdudoladov, line 380: what does `markRollingUpdateFlagForPod` return if a pod already has the desired annotation ?

pkg/cluster/sync.go
```diff
@@ -304,60 +306,86 @@ func (c *Cluster) syncStatefulSet() error {
...
-			if cmp.rollingUpdate && !podsRollingUpdateRequired {
-				podsRollingUpdateRequired = true
-				c.setRollingUpdateFlagForStatefulSet(desiredSS, podsRollingUpdateRequired, "statefulset changes")
+			if cmp.rollingUpdate {
+				podsToRecreate = make([]v1.Pod, 0)
+				for _, pod := range pods {
+					if err = c.markRollingUpdateFlagForPod(&pod, "pod changes"); err != nil {
+						return fmt.Errorf("updating rolling update flag for pod failed: %v", err)
+					}
+					podsToRecreate = append(podsToRecreate, pod)
```
sdudoladov, line 349: this update is all-or-nothing: either all  of `pods` make it to `podsToRecreate` or none, and we re-try later, It is better to ack the behaviour explicitly by copying `pods` to `podsToRecreate` after   `markRollingUpdateFlagForPod` succeeds for all of them

e2e/tests/test_e2e.py
```diff
@@ -784,6 +782,145 @@ def test_multi_namespace_support(self):
...
+                "annotations": {
+                    "zalando-postgres-operator-rolling-update-required": "true",
+                }
+            }
+        }
+
+        try:
+            podsList = k8s.api.core_v1.list_namespaced_pod('default', label_selector=cluster_label)
+            for pod in podsList.items:
+                # add flag only to the master to make it appear to the operator as a leftover from a rolling update
```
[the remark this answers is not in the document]
FxKu: Would be too long. The current docstring is good enough imo

pkg/cluster/sync.go
```diff
@@ -304,60 +306,86 @@ func (c *Cluster) syncStatefulSet() error {
 			return fmt.Errorf("cluster is not ready: %v", err)
 		}
 
-		podsRollingUpdateRequired = (len(pods) > 0)
-		if podsRollingUpdateRequired {
-			c.logger.Warningf("found pods from the previous statefulset: trigger rolling update")
-			if err := c.applyRollingUpdateFlagforStatefulSet(podsRollingUpdateRequired); err != nil {
-				return fmt.Errorf("could not set rolling update flag for the statefulset: %v", err)
+		if len(pods) > 0 {
```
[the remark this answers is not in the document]
FxKu: it was like this before and I removed `podsRollingUpdateRequired`. I think with the log message in the if block it's pretty clear what's the expected situation here

pkg/cluster/sync.go
```diff
@@ -304,60 +306,86 @@ func (c *Cluster) syncStatefulSet() error {
...
 		c.logger.Infof("created missing statefulset %q", util.NameFromMeta(sset.ObjectMeta))
 
 	} else {
-		podsRollingUpdateRequired = c.mergeRollingUpdateFlagUsingCache(sset)
+		// check if there are still pods with a rolling update flag
+		for _, pod := range pods {
+			if c.getRollingUpdateFlagFromPod(&pod) {
+				podsToRecreate = append(podsToRecreate, pod)
+			} else {
+				switchoverCandidates = append(switchoverCandidates, util.NameFromMeta(pod.ObjectMeta))
```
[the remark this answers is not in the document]
FxKu: good catch! Fixed it

pkg/cluster/sync.go
FxKu: nil. I've added a check inside the function to not patch the pod when the flag is present

pkg/cluster/pod.go
```diff
@@ -45,6 +47,59 @@ func (c *Cluster) getRolePods(role PostgresRole) ([]v1.Pod, error) {
 	return pods.Items, nil
 }
 
+// markRollingUpdateFlagForPod sets the indicator for the rolling update requirement
+// in the Pod annotation.
+func (c *Cluster) markRollingUpdateFlagForPod(pod *v1.Pod, msg string) error {
```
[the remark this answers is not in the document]
FxKu: I would assume `set` implies I can set it either `true` or `false` (otoh, in that case it could also be called `enableRollingUpdateFlagForPod` then). It was a suggestion by @Jan-M

e2e/tests/test_e2e.py
```diff
@@ -784,6 +782,145 @@ def test_multi_namespace_support(self):
...
+
+            # operator should now recreate the master pod and do a switchover before
+            k8s.wait_for_pod_failover(replica_nodes, 'spilo-role=master,' + cluster_label)
+
+            # check if the former replica is now the new master
+            leader = k8s.get_cluster_leader_pod()
+            self.eventuallyEqual(lambda: leader.metadata.name, switchover_target, "Rolling update flag did not trigger switchover")
+
+            # check that the old master has been recreated
+            k8s.wait_for_pod_start('spilo-role=replica,' + cluster_label)
```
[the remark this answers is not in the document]
FxKu: yeah ... another time :)

e2e/tests/test_e2e.py
```diff
@@ -784,6 +782,145 @@ def test_multi_namespace_support(self):
...
+
+            # check if the former replica is now the new master
+            leader = k8s.get_cluster_leader_pod()
+            self.eventuallyEqual(lambda: leader.metadata.name, switchover_target, "Rolling update flag did not trigger switchover")
+
+            # wait for the old master to get restarted
+            k8s.wait_for_pod_start('spilo-role=replica,' + cluster_label)
+
+            # status should again be "SyncFailed" but turn into "Running" on the next sync
+            time.sleep(10)
```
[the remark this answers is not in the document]
FxKu: my thought here is, if I know that the status will be `SyncFailed` for one sync cycle (taking 20 seconds in this e2e test) I don't need to "hammer" eventuallyEqual the whole time. And eventuallyEqual would "only" try for 30s. So if the sync (to turn the cluster into `Running` state) is somewhat delayed there's a higher chance the test could fail. But maybe it's better that it fails then?

pkg/cluster/sync.go
FxKu: Note that `podsToRecreate` is reset every time we pass this code. If it fails for one pod the statefulset will not be updated and we end up here again on the next sync. I don't think we need an extra pods loop to append all pods to `podsToRecreate`.

Instead, I found another issue. The `switchoverCandidates` should be reset, too. Otherwise the current code will consider pods where `markRollingUpdateFlagForPod` failed the first time as already updated :grimacing:

sdudoladov:
:+1:

FxKu:
:+1:

PR #1342, opened 2021-02-01
Match formatting in both chart values files

apeschel:
There are currently two values files:

* values.yaml
* values-crd.yaml

There are a large number of insignificant formatting changes between the
two, which obscures the actual changes between the two files.

This commit standardizes the formatting between the two files.

### DIFF
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -15,6 +15,9 @@ podLabels: {}
 
 configTarget: "OperatorConfigurationCRD"
 
+# JSON logging format
+enableJsonLogging: false
+
 # general top-level configuration parameters
 configGeneral:
   # choose if deployment creates/updates CRDs with OpenAPIV3Validation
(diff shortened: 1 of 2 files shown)


apeschel:
Hello, while looking at values.yaml and values-crd.yaml to see what the difference was between the two files, I noticed that there are a lot of formatting difference, which have no impact but obscure the actual differences between the two files. This PR just cleans up the formatting to make it consistent between the two files. There is no logical difference between the two files.

apeschel:
Looks like there's a problem with CI at the moment. This changeset has no impact on the CI pipeline at all.

apeschel:
CI is fixed in https://github.com/zalando/postgres-operator/pull/1345

apeschel:
It looks like more work is required here, since the validator validates each file with different types for some reason.

FxKu:
Will close this PR since we merged the two `values` files. Could be that there is is still some inconsistency with our manifest examples, but that's stuff for another PR.

PR #1349, opened 2021-02-05
making pgTeamMap a pointer

FxKu:
when a PostgresTeam map is updated with more teams/users, it's immediately reflected in the controller's pgTeamMap (can be seen from logs). However, `initUsers` function do not seem to pick them up. Only after the operator is restarted, it seems to work.

Essential change 1: Make cluster's PgTeamMap a pointer:
```
--- PgTeamMap                    pgteams.PostgresTeamMap
+++ PgTeamMap                    *pgteams.PostgresTeamMap
```

Essential change 2: Additional teams and members are only fetched if the feature is enabled and the PgTeamMap is not `nil`

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1143,7 +1143,7 @@ func (c *Cluster) initHumanUsers() error {
 	var clusterIsOwnedBySuperuserTeam bool
 	superuserTeams := []string{}
 
-	if c.OpConfig.EnablePostgresTeamCRDSuperusers {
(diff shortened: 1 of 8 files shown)


pkg/cluster/cluster.go
```diff
@@ -49,7 +49,7 @@ var (
 type Config struct {
 	OpConfig                     config.Config
 	RestConfig                   *rest.Config
-	PgTeamMap                    pgteams.PostgresTeamMap
+	PgTeamMap                    *pgteams.PostgresTeamMap
```
Jan-M, line 52: did we not say that we could just not use this var at all and always just use

cluster: c.opconfig.pgteams?

FxKu:
:+1:

Jan-M:
:+1:

PR #1350, opened 2021-02-07
code optimization for generateCapabilities

zvier:
* pre-allocate cap for slice structure
* if clause is no needed because access slice with range, and kubelet also use range
  method to get each capability so there is no side-effect

Signed-off-by: Jeff Zvier <zvier20@gmail.com>


FxKu:
:+1:

Jan-M:
:+1:

PR #1356, opened 2021-02-10
WIP: some code refactoring about controller.go

zvier:
* no logic change, only for clean code
* reduce if-else nested for it looks more clean

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -321,16 +321,13 @@ func getLocalAndBoostrapPostgreSQLParameters(parameters map[string]string) (loca
 }
 
 func generateCapabilities(capabilities []string) v1.Capabilities {
-	if len(capabilities) > 0 {
-		additionalCapabilities := []v1.Capability{}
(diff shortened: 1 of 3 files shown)


zvier:
Hello, I'm new and interest for this project, start with some small optimizations for it. Thanks for review~

sdudoladov:
:+1:

sdudoladov:
@zvier thanks for the contribution !

FxKu:
:+1:

zvier:
> I think you did something wrong in your last push. Your changes are gone. Note, that when we approve a PR you should not push other commits because then we have to approve again. In the end all commits will be squashed anyway on commit.

Hello,  I really do this for try to found the problem whether it is introduced by this PR because the e2e tests ci failed. But even I do nothing even with an empty commit(only delete a blank line) it also failed. So, is the e2e tests any problem ?

FxKu:
Will close this now. Yes the nodeAffinity e2e test fails on GH regularly. Something we should fix at some point.

PR #1361, opened 2021-02-11
Fix operator configs runasuser and runasgroup

### LINKED ISSUE #1359: Operator runasuser/runasgroup configs not working
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.6.0
- **Where do you run it - cloud or metal?** Kubernetes 1.19

If the following configs are set in the operator as defaults they seem to be ignored
```
  spilo_runasuser: xxxx
  spilo_runasgroup: xxxx
```
while if they get set in the postgresql resource they work as expected, setting the properties on the postgres pods
```
  spiloRunAsUser: xxxx
  spiloRunAsGroup: xxxx
```

I think the issue is a typo in the annotation of the two configs [here](https://github.com/zalando/postgres-operator/blob/master/pkg/util/config/config.go#L31) and [here](https://github.com/zalando/postgres-operator/blob/master/pkg/util/config/config.go#L32) which should be `name` rather than `json`

tommasopozzetti:
This commit fixes a typo in the operator's configuration that would lead to the configs spilo_runasuser and spilo_runasgroup to be ignored.

fixes #1359

### DIFF
--- a/pkg/util/config/config.go
+++ b/pkg/util/config/config.go
@@ -28,8 +28,8 @@ type Resources struct {
 	PodLabelWaitTimeout       time.Duration       `name:"pod_label_wait_timeout" default:"10m"`
 	PodDeletionWaitTimeout    time.Duration       `name:"pod_deletion_wait_timeout" default:"10m"`
 	PodTerminateGracePeriod   time.Duration       `name:"pod_terminate_grace_period" default:"5m"`
-	SpiloRunAsUser            *int64              `json:"spilo_runasuser,omitempty"`
-	SpiloRunAsGroup           *int64              `json:"spilo_runasgroup,omitempty"`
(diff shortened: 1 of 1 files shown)


FxKu:
:+1:

Jan-M:
:+1:

FxKu:
Thanks @tommasopozzetti for your contribution :+1:

caiconkhicon:
@FxKu : sorry I know this commit had been merged and added to v1.6.1, but has anyone tested it yet? I tried with v1.6.1 and runasuser and runasgroup were still ignored.

cortopy:
Same here with 1.6.2

FxKu:
Thanks @caiconkhicon and @cortopy for bringing it up again. The `omitempty` is indeed wrong there and leads to the error. I'll fix it in #1503

PR #1372, opened 2021-02-18
Support major version upgrade via manifest and global upgrades via min version

Jan-M:
This PR introduces support to upgrade major version via manifest for a healthy cluster.

This PR also introduces the ability to configure a minimum and desired Postgres major version for all clusters. Clusters with smaller versions will automatically be upgraded to the respective target version, which is 13 for the time being.

### DIFF
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -1,8 +1,6 @@
 package cluster
 
 import (
-	"fmt"
-
 	"github.com/zalando/postgres-operator/pkg/spec"
 	v1 "k8s.io/api/core/v1"
 )
@@ -80,8 +78,7 @@ func (c *Cluster) majorVersionUpgrade() error {
 		if c.currentMajorVersion < desiredVersion {
 			podName := &spec.NamespacedName{Namespace: masterPod.Namespace, Name: masterPod.Name}
 			c.logger.Infof("triggering major version upgrade on pod %s of %d pods", masterPod.Name, numberOfPods)
-			command := fmt.Sprintf("/bin/su postgres -c \"/usr/bin/python3 /scripts/inplace_upgrade.py %d 2>&1 | tee last_upgrade.log\"", numberOfPods)
-			_, err := c.ExecCommand(podName, command)
+			_, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", "\"/usr/bin/python3 /scripts/inplace_upgrade.py %d 2>&1 | tee last_upgrade.log\"")
 			if err != nil {
 				return err
 			}
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -48,6 +48,13 @@ func (c *Cluster) majorVersionUpgrade() error {
 		return nil
 	}
 
+	desiredVersion := c.GetDesiredMajorVersionAsInt()
+
+	if c.currentMajorVersion >= desiredVersion {
+		c.logger.Infof("cluster version up to date. current: %d desired: %d", c.currentMajorVersion, desiredVersion)
+		return nil
+	}
+
 	pods, _ := c.listPods()
 	allRunning := true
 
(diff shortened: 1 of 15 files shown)


pkg/util/patroni/patroni_test.go
```diff
@@ -72,3 +79,32 @@ func TestApiURL(t *testing.T) {
 		}
 	}
 }
+
+func TestPatroniAPI(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	json := `{"state": "running", "postmaster_start_time": "2021-02-19 14:31:50.053 CET", "role": "master", "server_version": 90621, "cluster_unlocked": false, "xlog": {"location": 55978296057856}, "timeline": 6, "database_system_identifier": "6462555844314089962", "pending_restart": true, "patroni": {"version": "2.0.1", "scope": "acid-rest92-standby"}}`
```
sdudoladov, line 87: For the sake of readability,  this line should split into multiple ones.

pkg/controller/util_test.go
```diff
@@ -480,3 +480,45 @@ func TestInfrastructureRoleDefinitions(t *testing.T) {
...
+	ssc.sub.teammap["teapot"] = "Mikkel"
+
+	if len(teams) != 3 {
+		t.Errorf("Team Map does not contain 3 elements")
+	}
+
+	teams = make(map[string]string)
+
+	if len(ssc.sub.teammap) != 3 {
+		t.Errorf("Team Map does not contain 0 elements")
```
sdudoladov, line 518: Wrong error message or wrong error condition ?  The condition check the length is not equal to 3 but the error talks about 0 elements.

pkg/cluster/sync.go
```diff
@@ -118,6 +118,10 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 		return fmt.Errorf("could not sync connection pooler: %v", err)
 	}
 
+	if err := c.majorVersionUpgrade(); err != nil {
```
sdudoladov, line 121: Is there an implicit requirement that `c.majorVersionUpgrade()` must be called after all other `Sync` actions have completed successfully ?

pkg/cluster/majorversionupgrade.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+
+	desiredVersion := c.GetDesiredMajorVersionAsInt()
+
+	if c.currentMajorVersion >= desiredVersion {
+		c.logger.Infof("cluster version up to date. current: %d desired: %d", c.currentMajorVersion, desiredVersion)
+		return nil
+	}
+
+	pods, _ := c.listPods()
```
sdudoladov, line 58: Please don't swallow the errors.

pkg/cluster/majorversionupgrade.go
```diff
@@ -0,0 +1,92 @@
...
+	allRunning := true
+
+	var masterPod *v1.Pod
+
+	for _, pod := range pods {
+		ps, _ := c.patroni.GetMemberData(&pod)
```
sdudoladov, line 64: Shall one retry if that fails on the first try ?

pkg/cluster/majorversionupgrade.go
```diff
@@ -0,0 +1,92 @@
...
+
+		if ps.State != "running" {
+			allRunning = false
+			c.logger.Infof("identified non running pod, potentially skipping major version upgrade")
```
sdudoladov, line 68: Judging from the [condition below](https://github.com/zalando/postgres-operator/pull/1372/files#diff-46d182d6fa2f77000bdf0906dc6d1ea032a074500f2eb07b3ce730384702e3afR78) the upgrade is always skipped if there is at least one non-running pod, which totally makes sense.  The word "potentially" is misleading in this case.

sdudoladov (commented):
Please add a paragraph describing the feature to the admin docs.

Jan-M:
:+1:

sdudoladov:
:+1:

Jan-M:
:+1:

sdudoladov:
:+1:

PR #1381, opened 2021-02-22
Add logic to look for container by name instead of defaulting to to index 0

guicholeo:
## Problem description

The postgres pods would restart every 30 minutes due to Postgres-Operator applying a rolling update because the container image name was defaulting to the index 0 and the actual name of the container in index 0 was a different one. 

In our case, we have Istio injecting the sidecar manually, and when injects the sidecar, it injects it to the top of the container images. When the operator is checking if there has been a change, it detects the image name `istio-proxy` different than `postgres` which causes to roll an update every 30 minutes.

## Checklist

- [ ] Your go code is [formatted](https://blog.golang.org/gofmt). Your IDE should do it automatically for you.
- [ ] You have updated [generated code](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#code-generation) when introducing new fields to the `acid.zalan.do` api package.
- [ ] New [configuration options](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters) are reflected in CRD validation, helm charts and sample manifests.
- [ ] New functionality is covered by [unit](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#unit-tests) and/or [e2e](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#end-to-end-tests) tests.
- [ ] You have checked existing open PRs for possible overlay and referenced them.

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -263,8 +263,21 @@ func (c *Cluster) mustUpdatePodsAfterLazyUpdate(desiredSset *appsv1.StatefulSet)
 
 	for _, pod := range pods {
 
-		effectivePodImage := pod.Spec.Containers[0].Image
-		ssImage := desiredSset.Spec.Template.Spec.Containers[0].Image
+		var (
+			effectivePodImage, ssImage string
+		)
+
+		for i := range pod.Spec.Containers {
+			if pod.Spec.Containers[i].Name == "postgres" {
+				effectivePodImage = pod.Spec.Containers[i].Image
+			}
+		}
+
+		for j := range desiredSset.Spec.Template.Spec.Containers {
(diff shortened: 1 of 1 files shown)


FxKu:
@guicholeo this is a good first step in the right direction. I've checked if we have this hard-coded assumptions in other places and we do:
```
grep -irn  "\[0\]" pkg/
pkg/cluster/k8sres.go:1456
pkg/cluster/cluster.go:459
pkg/cluster/connection_pooler_test.go
pkg/cluster/k8sres_test.go
pkg/cluster/sync.go
```

So this would likely be a problem for your setup, too, I think. You're eager to update more Go code? :smiley:

FxKu:
picked it up with #1504 and #1505. Closing this PR then. Thanks again for raising.

Issue #1131, opened 2020-09-04
Clearer way to map PVCs to PVs for Static Volumes


curionim:
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.5.0
- **Where do you run it - cloud or metal? Bare Metal K8s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** feature request

When running on cluster without dynamic volumes, i.e. PVs are statically created, the operator will try to create a PVC for each replica and bind it to available PVs based on storage class name and size.
We would like to be able to explicitly select which PVs to use, for instance by allowing for static PVCs, or by using selectors.
Is it possible already?

Thanks,
Michele

Jan-M:
If you can elaborate more what is needed for this we might at least give some feedback.

I would wonder if this is not something K8s can already solve for you?

curionim:
Hi Jan,
what is needed is a way in postgres-manifest to be able to specify which PV the various replicas of Postgres will bind to.
In our system we have about 10 PVs created statically, each using the same storage class.
3 of those PVs are meant for Postgres.
When the Operator creates the Postgres cluster with 3 instances, it will request 3 available PVs from Kubernetes based on the storage class specified in postgres-manifest. Without any specific selector, Kubernetes will offer any 3 of the above 10 PVs.
What we want is to be able to target the 3 PVs that were created specifically for Postgres, and not pick any others.
I don't see how that's possible with the current parameters available in postgres-manifest.

curionim:
Hi Jan,
what is needed is a way in postgres-manifest to be able to specify which PV the various replicas of Postgres will bind to.
In our system we have about 10 PVs created statically, each using the same storage class.
3 of those PVs are meant for Postgres.
When the Operator creates the Postgres cluster with 3 instances, it will request 3 available PVs from Kubernetes based on the storage class specified in postgres-manifest. Without any specific selector, Kubernetes will offer any 3 of the above 10 PVs.
What we want is to be able to target the 3 PVs that were created specifically for Postgres, and not pick any others.
I don't see how that's possible with the current parameters available in postgres-manifest.

curionim:
Hi Jan,
what is needed is a way in postgres-manifest to be able to specify which PV the various replicas of Postgres will bind to.
In our system we have about 10 PVs created statically, each using the same storage class.
3 of those PVs are meant for Postgres.
[...]

PR #1385, opened 2021-02-25
Add volume selector

jcrood:
As asked/discussed in https://github.com/zalando/postgres-operator/issues/1131 this PR adds a volume selector for PVs as an alternative to using storage classes.


FxKu:
Looks good. Default is empty so behavior will stay the same, right.
Did you update the generated code running `./hack/update-codegen.sh` since you've introduced a new field to the API?

jcrood:
Yup, the selector is optional and can be left out or when desired, be combined with a storageClass. I did run `update-codegen.sh` but this did not yield any changes. I came across another PR mentioning that this occasionally happens, but I have no idea how to remedy this.

FxKu:
Hm ok, I've check out your PR, run the update and got a change. You can copy it from here: https://gist.github.bus.zalan.do/fekunde/f27db5c22c43a9360b24f55d104647cd

jcrood:
Thanks! But I appear not to have access to your private github ;)

FxKu:
Oh! Here you go: https://gist.github.com/FxKu/1a0fae553303e90b5f39cfd9d0a2e3d6

jcrood:
Much appreciated; I've updated the deepcopy src.

jcrood:
As it turns out, `update-codegen.sh` outputs to `${GOPATH}/github.com/zalando/postgres-operator`, hence it not producing apparent changes for  me. With this fixed, I noticed your earlier copy also sneaked in a deepcopy for MajorVersionUpgradeConfiguration which comes from a different branch, so I've removed this.

FxKu:
Thanks for the update. I think the last thing to do is add a test for it. I think a unit test would be fine. Have a look on how we [mock volume claims.](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/volumes_test.go) Should not be too hard to add.

jcrood:
I've added a test to check whether the selector ends up in the StatefulSet's volumeClaimTemplate, but beyond that I don't really know what to test. Mocking PVCs seems to be a bit out of scope as it's up to k8s to create the PVC based on the template.

FxKu:
:+1:

Issue #1371, opened 2021-02-18
Support for hostNetwork / hostPort in statefulset pods/containers


JJJJJones:
Please, answer some short questions which should help us to understand your problem / question better?

- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare-metal Kubernetes
- **Type of issue?** Feature request

Add support for the pods postgres-operator generates to use hostNetwork (statefulset.spec.template.spec.hostNetwork or for the containers to utilize hostPort (statefulset.spec.tempplate.spec.containers.spec).

I'll make a PR for this issue.

PR #1388, opened 2021-02-26
Support for hostNetwork in statefulset pods

JJJJJones:
Implementation for part of #1371 

This adds support for statefulset pods to be run with hostNetwork enabled. It also adds support for readOnlyRootFilesystem containers. Documentation is added for both new flags.

These are desired features for our own environment that we are currently supporting via a patch. This MR was made with assistance/contributions from @kwilliams057 and @tommasopozzetti.

I am still working on implementing hostPort support for containers still, the ConfigMap configuration is being resistant.


JJJJJones:
Postgres-operator has difficulties managing pods running in host mode, closing this PR

PR #1405, opened 2021-03-14
Added IPv6 support for K8s API Service.

Cellebyte:
## Issue

When you run the backup script inside a K8s which is DualStack and prefers IPv6 the api requests are failing.

## Solution

Check if the IP address is v4 or v6 and add brackets accordingly.


Cellebyte:
What is the status of this pull request?

L

sdudoladov:
:+1:

FxKu:
:+1:

PR #1407, opened 2021-03-15
Set runAsUser/runAsGroup/FSGroup on logical-backup job podtemplates

### LINKED ISSUE #1306: Run logical_backup as non-root

- **Which image of the operator are you using?** e.g. registry.opensource.zalan.do/acid/postgres-operator:v1.5.0

Our security policy requires that our pods don't run as root. This support was added to the operator on v1.6.0 (and we're using a patched v1.5.0 with it), but the logical_backup image still runs as root. Would it be possible to also add the capability to run as non-root to this image? Thanks!

timricese:
This PR makes logical-backup CronJobs be created with a securityContext containing `runAsUser`, `runAsGroup` and `FSGroup` based on the SpiloRunAsUser, SpiloRunAsGroup and SpiloFSGroup operator configuration options, allowing the logical-backup jobs to run as non-root.

resolves #1306

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ We introduce the major version into the backup path to smoothen the [major versi
 The new operator configuration can set a compatibility flag *enable_spilo_wal_path_compat* to make Spilo look for wal segments in the current path but also old format paths.
 This comes at potential performance costs and should be disabled after a few days.
 
-The newest Spilo image is: `registry.opensource.zalan.do/acid/spilo-14:2.1-p6`
+The newest Spilo image is: `registry.opensource.zalan.do/acid/spilo-14:2.1-p7`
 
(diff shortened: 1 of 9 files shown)


joschi36:
Just ran in the non-root problem for logical backup container. Is there a plan to merge this change, or is there another way?

FxKu:
Thanks @timricese. Let's do some more refactoring here. Can you move [this code block](https://github.com/zalando/postgres-operator/blob/c6f2c685884a98a4121641256e4ee36b3f745ad6/pkg/cluster/k8sres.go#L1164) (lines 1164 - 1178) inside generatePodTemplate function? We can then remove arguments `spiloRunAsUser *int64, spiloRunAsGroup *int64, spiloFSGroup *int64` from the function spec and it will be used for all pods - spilo and logical backup.

Issue #1403, opened 2021-03-12
Operator 1.6 : Remove "AllowPrivilegeEscalation" or decouple "AllowPrivilegeEscalation" from "Privileged"


neelasha-09:
Hi Team,

While using OPR v1.6.1, we get the error  `seteuid: Operation not permitted` due to cron binary unable to run.

https://github.com/zalando/spilo/issues/562

We compared the old (1.5.0)and new operator (1.6.1) and observed the error is due to the `allowPrivilegeEscalation: false` under `securityContext`.

We see required  either to remove the  `allowPrivilegeEscalation` ( as in 1.5.0 ) or to decouple the `allowPrivilegeEscalation` from `Privileged` by having 2 different parameters:

Reference links below for the code:

**Old OPR: 1.5.0 - Working**
https://github.com/zalando/postgres-operator/blob/a9b677c957be09de5e02f3b0eec4221c551eb2ca/pkg/cluster/k8sres.go#L456

**New OPR: 1.6.1 - cron error**
https://github.com/zalando/postgres-operator/blob/ff8143770c9ca5b2e7b36d1d29d8c914e424b9c9/pkg/cluster/k8sres.go#L469

In conclusion, at least for OpenShift  we need the parameter  `allowPrivilegeEscalation` to be set as `true` and  `Privileged` to be set as `false`

neelasha-09:
@FxKu : Could you please review the PR  #1412

PR #1412, opened 2021-03-19
Fix for AllowPrivilegeEscalation : issue-1403

neelasha-09:
Fix for #1403

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -374,6 +374,11 @@ configuration they are grouped under the `kubernetes` key.
   used for AWS volume resizing and not required if you don't need that
   capability. The default is `false`.
 
+* **spilo_allow_privilege_escalation**
+  Controls whether a process can gain more privileges than its parent
+  process. Required by cron which needs setuid. Without this parameter, certification
+  rotation & backups will not be done. The default is `true`.
+
 * **additional_pod_capabilities**
   list of additional capabilities to be added to the postgres container's
   SecurityContext (e.g. SYS_NICE etc.). Please, make sure first that the
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -376,8 +376,8 @@ configuration they are grouped under the `kubernetes` key.
 
 * **spilo_allow_privilege_escalation**
   Controls whether a process can gain more privileges than its parent
-  process. Required by cron which needs setuid. Without this parameter, certification
-  rotation & backups will not be done. The default is `true`.
+  process. Required by cron which needs setuid. Without this parameter,
+  certification rotation & backups will not be done. The default is `true`.
 
 * **additional_pod_capabilities**
   list of additional capabilities to be added to the postgres container's
--- a/pkg/controller/operator_config.go
+++ b/pkg/controller/operator_config.go
@@ -68,6 +68,7 @@ func (c *Controller) importConfigurationFromCRD(fromCRD *acidv1.OperatorConfigur
 	result.PodEnvironmentSecret = fromCRD.Kubernetes.PodEnvironmentSecret
 	result.PodTerminateGracePeriod = util.CoalesceDuration(time.Duration(fromCRD.Kubernetes.PodTerminateGracePeriod), "5m")
 	result.SpiloPrivileged = fromCRD.Kubernetes.SpiloPrivileged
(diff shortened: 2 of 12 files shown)


neelasha-09:
@FxKu @sdudoladov  - This is the fix for the bug introduced in OPR 1.6 that does not allow `cron` to work any longer.
A quick review would be highly appreciated. We tested and everything is working fine again with no errors. 
Spilo Issue: https://github.com/zalando/spilo/issues/562 which we reviewed with @CyberDem0n

FxKu:
Thanks for your contribution @neelasha-09 . Looks good. The name for the parameter is fine. Two minor things.
1. In yaml manifests I'd like to keep alphabetical sorting so move `spilo_allow_privilege_escalation` further up. 
2. I'm thinking if the default should be `false`, too, like with privileged mode. OTOH looking at the use cases you've mentioned (cron, cert rotation) it seems to be more user-friendly if it's set to true straight away. Other opinions?

neelasha-09:
> Thanks for your contribution @neelasha-09 . Looks good. The name for the parameter is fine. Two minor things.
> 
> 1. In yaml manifests I'd like to keep alphabetical sorting so move `spilo_allow_privilege_escalation` further up.
We thought of keeping the `spilo_allow_privilege_escalation` and `spilo_privileged`  next to each other as they are closely related.
> 2. I'm thinking if the default should be `false`, too, like with privileged mode. OTOH looking at the use cases you've mentioned (cron, cert rotation) it seems to be more user-friendly if it's set to true straight away. Other opinions?
The default value to `false` would cause BWC with OPR  version1.5.0. 
Without this parameter, functionality would be decreased. In our opinion, it should not be postponed  till next milestone & rather consider as fix within same release.

FxKu:
1. Makes sense, but would also apply to many other existing options. Alphabetical sorting makes it easier in the end to find stuff. And with the `spilo_` prefix the options are not that far away from each other anyway.
2. Good point keeping it consistent with 1.5.0.

neelasha-09:
> 1. Makes sense, but would also apply to many other existing options. Alphabetical sorting makes it easier in the end to find stuff. And with the `spilo_` prefix the options are not that far away from each other anyway.

                 Sorted the parameters in alphabetical order as suggested. 
> 2. Good point keeping it consistent with 1.5.0.

                 Also, could you please consider of not postponing the fix until 1.7.0  and fix is within the same release

pkg/controller/operator_config.go
```diff
@@ -68,6 +68,7 @@ func (c *Controller) importConfigurationFromCRD(fromCRD *acidv1.OperatorConfigur
 	result.PodEnvironmentSecret = fromCRD.Kubernetes.PodEnvironmentSecret
 	result.PodTerminateGracePeriod = util.CoalesceDuration(time.Duration(fromCRD.Kubernetes.PodTerminateGracePeriod), "5m")
 	result.SpiloPrivileged = fromCRD.Kubernetes.SpiloPrivileged
+	result.SpiloAllowPrivilegeEscalation =fromCRD.Kubernetes.SpiloAllowPrivilegeEscalation
```
FxKu, line 71:
```suggestion
	result.SpiloAllowPrivilegeEscalation = fromCRD.Kubernetes.SpiloAllowPrivilegeEscalation
```

docs/reference/operator_parameters.md
```diff
@@ -374,6 +374,11 @@ configuration they are grouped under the `kubernetes` key.
   used for AWS volume resizing and not required if you don't need that
   capability. The default is `false`.
 
+* **spilo_allow_privilege_escalation**
+  Controls whether a process can gain more privileges than its parent
+  process. Required by cron which needs setuid. Without this parameter, certification
+  rotation & backups will not be done. The default is `true`.
```
FxKu, line 380:
```suggestion
  Controls whether a process can gain more privileges than its parent
  process. Required by cron which needs setuid. Without this parameter,
  certification rotation & backups will not be done. The default is `true`.
```

FxKu:
:+1:

sdudoladov:
:+1:

FxKu:
Thanks @neelasha-09 . We are making it a 1.6.2 release now. We cannot replace previous images/releases.

neelasha-09:
Thanks @FxKu for the PR merge 
However, I've raised another PR with the correction in the default values for the parameter  `SpiloAllowPrivilegeEscalation` under `pkg/util/config/config.go` -  https://github.com/zalando/postgres-operator/pull/1434

Issue #1414, opened 2021-03-21
Building dependencies requires go 1.15+


ssinger:
docs/developing.md says that GO 1.7+ is supported but 1.15+ is actually required.

Building with golang 1.11  results in

~~~~
 make deps
GO111MODULE=on go get k8s.io/client-go@kubernetes-1.20.2
go: finding k8s.io/client-go kubernetes-1.20.2
GO111MODULE=on go get github.com/golang/mock/mockgen@v1.4.4
go: finding github.com/golang/mock/mockgen v1.4.4
# golang.org/x/tools/go/packages
../../gopath/pkg/mod/golang.org/x/tools@v0.0.0-20201207204333-a835c872fcea/go/packages/golist_overlay.go:547:7: undefined: strings.ReplaceAll
note: module requires Go 1.12make: *** [Makefile:89: tools] Error 2

~~~~

Building with 1.12 gets a bit further but the make step results in

~~~
.
/tmp/gopath/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/errors/errors.go:99:10: undefined: errors.Is
note: module requires Go 1.15
k8s.io/apimachinery/pkg/util/net
# k8s.io/apimachinery/pkg/api/resource
/tmp/gopath/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/api/resource/quantity.go:470:34: invalid operation: int64(1) << (exponent * 10) (shift count type int, must be unsigned integer)
note: module requires Go 1.15
# k8s.io/utils/trace
/tmp/gopath/pkg/mod/k8s.io/utils@v0.0.0-20201110183641-67b214c5f920/trace/trace.go:200:119: totalTime.Milliseconds undefined (type time.Duration has no field or method Milliseconds)
# k8s.io/apimachinery/pkg/util/net
/tmp/gopath/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/net/http.go:67:5: undefined: errors.As
/tmp/gopath/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/net/http.go:84:5: undefined: errors.As
/tmp/gopath/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/net/util.go:43:5: undefined: errors.As
/tmp/gopath/pkg/mod/k8s.io/apimachinery@v0.20.2/pkg/util/net/util.go:52:5: undefined: errors.As
note: module requires Go 1.15
~~~

[...]

PR #1415, opened 2021-03-21
Update developer docs to say that GO 1.15 or newer is required.

ssinger:
Fixes issue #1414


FxKu:
:+1:

Jan-M:
:+1:

FxKu:
Thanks @ssinger for finding it. Wasn't so clear to me that when we raise the Go version it would not compile on older versions anymore. But 1.7 is def too old :smiley:

Issue #1420, opened 2021-03-25
Default privileges are not set in public schema


machine424:
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.6.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Kube on Cloud
- **Are you running Postgres Operator in production?** no
- **Type of issue?** Bug report?

When I deploy a postgresl manifest with:

```
  preparedDatabases:
    foo:
      defaultUsers: true
      schemas:
        public:
          defaultUsers: false
          defaultRoles: false
      extensions:
        pg_stat_statements: public
        btree_gist: public
```

The Global Default privileges are not created in the `public` schema, even though PG says that it applies the globalDefaultPrivilegesSQL statements in [here](https://github.com/zalando/postgres-operator/blob/v1.6.1/pkg/cluster/database.go#L42), we cannot fallback to schemaDefaultPrivilegesSQL because `public` schema  gets removed [here](https://github.com/zalando/postgres-operator/blob/v1.6.1/pkg/cluster/sync.go#L679) as already present. 

Of course, when I manually re-execute the globalDefaultPrivilegesSQL statements (that I get from PG logs), the default privileges get created in `public` schema.

When I try with another schema name, the operator creates the default privileges as expected.

Am I doing something wrong?

machine424:
I think I get it, the operator creates the default privileges in the wrong database (always in `postgres`), I'll prepare a PR for this.

machine424:
here: https://github.com/zalando/postgres-operator/pull/1421

---

I still don't see why we exclude already created schemas (`public` for example) [here](https://github.com/zalando/postgres-operator/blob/v1.6.1/pkg/cluster/sync.go#L679), `executeCreateDatabaseSchema` seems to be idempotent, so why not running it on already existing schemas that we add under `preparedDatabases`?

machine424:
> I still don't see why we exclude already created schemas (`public` for example) [here](https://github.com/zalando/postgres-operator/blob/v1.6.1/pkg/cluster/sync.go#L679), `executeCreateDatabaseSchema` seems to be idempotent, so why not running it on already existing schemas that we add under `preparedDatabases`?
> 
> EDIT: this can also help to sync again if the first call fails

Any thoughts?

PR #1421, opened 2021-03-25
create global default privileges in the appropriate prepared databases

machine424:
fixes https://github.com/zalando/postgres-operator/issues/1420
I can see the default privileges now.
I don't see how to easily test this, maybe the preparedDatabases feature needs some e2e tests?

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -661,6 +661,9 @@ func (c *Cluster) syncDatabases() error {
 
 	// set default privileges for prepared database
 	for _, preparedDatabase := range preparedDatabases {
+		if err := c.initDbConnWithName(preparedDatabase); err != nil {
+			return fmt.Errorf("could not init database connection to %s", preparedDatabase)
+		}
 		if err = c.execAlterGlobalDefaultPrivileges(preparedDatabase+constants.OwnerRoleNameSuffix, preparedDatabase); err != nil {
 			return err
 		}


pkg/cluster/sync.go
```diff
@@ -612,6 +612,9 @@ func (c *Cluster) syncDatabases() error {
 	if err := c.initDbConn(); err != nil {
 		return fmt.Errorf("could not init database connection")
 	}
+
+	// only try to close the last connection that will be initialized, as c.initDbConn/c.initDbConnWithName
+	// try to close the existing connection in c.pgDb before initializing a new one
```
Jan-M, line 617: Can we remove this, this does not do anything and imho the comment is also not correct, the defer part comes post init.

machine424: `the defer part comes post init` I didn't get it?
I was trying to say that, in case we enter in the loop in line 666, the first `initDbConnWithName` in line 667 will try to close the connection we open in line 612 and that this defer will close the last connection that we open in line 667 so we don't need to add another close.
I'll remove it.

FxKu: maybe remove it if it confuses Jan ;)
I think from the code it should be clear that the last connection opened in this method will be closed in the defer part.

machine424: yep, and to underline the fact that initDbConnWithName (tries to) does close the conn before creating another one so we don't risk a conn risk.
Done.

FxKu:
:+1:

Jan-M:
:+1:

Jan-M:
Thank you @machine424

PR #1424, opened 2021-03-25
Wait for statefulset pods after sts modification

ynnt:
Sometimes operators tries to recreate pods that are in Pending state

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -371,6 +371,9 @@ func (c *Cluster) syncStatefulSet() error {
 					return fmt.Errorf("could not replace statefulset: %v", err)
 				}
 			}
+			if err = c.waitStatefulsetPodsReady(); err != nil {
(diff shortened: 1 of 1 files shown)


Jan-M:
My quick response is, we need to hold out on this.

Imagine scenarios where the earlier statefulset config is broken or produces broken pods or unschedulable pods. Then the operator and user cannot fix situation anymore, as it waits for ready pods. Need to double check if this is the case here.

ynnt:
Thanks for the quick response!

> Imagine scenarios where the earlier statefulset config is broken or produces broken pods or unschedulable pods
But I am applying the check _after_ the statefulset was modified / replaced. Why would there be any hanging pods? Are you reffering statefulsets with `OnDelete` update strategy?

FxKu:
With #1767 merged the operator will replace only healthy pods. This also means that the rolling update label remains on the pod and the operator will retry to recreate the pod on subsequent sync cycles. Therefore we do not really need another wait function.

PR #1428, opened 2021-03-30
Add documentation for using Azure storage accounts for WAL

### LINKED ISSUE #1429: Administrator Guide is missing a section on sending WAL to Azure storage accounts
The Administrator Guide has a section on how to configure WAL to be sent to AWS S3 buckets and Google GCP buckets but not for azure storage accounts.

ssinger:
Add documentation for sending WAL to azure storage accounts

Fixes #1429


docs/administrator.md
```diff
@@ -804,6 +804,52 @@ pod_environment_configmap: "postgres-operator-system/pod-env-overrides"
 ...
 ```
 
+### Azure Storage Account Setup
+
+To configure phyiscal backups and WAL to be sent to an azure storage account perform the
+following steps.
+
+1. Generate the K8s secret resource that will contain your service account's
+credentials. It's highly recommended to use a service account and limit its
```
apeschel, line 813: Can you actually use a service principal for this? It seems like you can only use an access token or SAS token. I don't see any WAL-G configuration to provide access via a SP.

docs/administrator.md
```diff
@@ -804,6 +804,52 @@ pod_environment_configmap: "postgres-operator-system/pod-env-overrides"
...
+scope to just the WAL-e bucket.
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: psql-wale-creds
+  namespace: postgres-operator-system
+type: Opaque
+stringData:
+  AZURE_STORAGE_ACCOUNT: 'mystorageaccount'
```
apeschel, line 823: This should be  base64 encoded.

docs/administrator.md
ssinger: I've removed the reference to service-account.

PR #1433, opened 2021-03-30
Bump v1.6.2

FxKu:
- bump operator version
- bump spilo to 13-2.0-p6
- bump pooler to master-16
- switch data type of `SpiloAllowPrivilegeEscalation` to `*bool` as it's true by default
- fixes #1431 
- update helm charts
- do not log query on error when creating or altering users
- update version diff message: "desired" -> "min desired"

PR #1434, opened 2021-03-31
Corrected default value for SpiloAllowPrivilegeEscalation

neelasha-09:
Corrected default value for `SpiloAllowPrivilegeEscalation` from `false` to `true`
This is to align it with the rest of the files, including documentation, manifests, etc


FxKu:
Thanks @neelasha-09. I've included it in #1433. There were also a few other issues. Changed the type to *bool, because the option is true by default. And it was without quotes in the helm chart's values file.

PR #1437, opened 2021-04-01
Unit tests for sync and config modules

ajaybhat:


FxKu:
Thanks @ajaybhat for providing unit test on your own initiative :smiley: 

However, this still works with the old mocking client. For new unit tests we have switched to a new mock client. Have a look at the new [sync_test.go](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/sync_test.go) file to understand how it works. It can be a bit tricky with what needs to be configured but the new client is closer to the actual workflows of the operator. Hope you're still up for working on this PR.

ajaybhat:
Thanks for your comments @FxKu 

I'll take a look.

PR #1448, opened 2021-04-09
fix pooler sync with empty cluster name

FxKu:
Pooler deployment is updated on each sync:
```
time="2021-04-08T16:55:56Z" level=info msg="Update connection pooler deployment acid-minimal-cluster-pooler, reason: [pooler user is different (having pooler.acid-minimal-cluster.credentials.postgresql.acid.zalan.do, required pooler)]" cluster-name=default/acid-minimal-cluster pkg=cluster
```

The message is a bit confusing because it's the secrets which are compared. The current code uses `deployment.ClusterName` which is (or can be?) empty, hence producing a reason to update the deployment on every sync. The solution suggested in this PR is to use the `Name` field of the cluster. Thus, making `needSyncConnectionPoolerDefaults` a cluster method.


sdudoladov:
:+1:

FxKu:
:+1:

PR #1457, opened 2021-04-14
Rename roles that are removed from PostgresTeam CRD

FxKu:
The operator never deletes users (and databases) which where specified once in the cluster manifest or in the PostgresTeams CRD. With the latter it's very easy to add roles to a lot of clusters, but reversing such action manually is cumbersome. We could give some power to the operator to at least rename such "outdated" roles, so that logins with the old username will fail. Deleting the role and corresponding secret might be too dangerous due to depending database objects or deployments.

~In this first approach of this PR the `readPgUsersFromDatabase` method is changed to not only retrieve given DB roles, but all roles that can login (or not but have a PW set like roles for preparedDatabases). In `ProduceSyncRequests`, DB roles that do not exists in the `pgUser` map would then trigger a rename request where a suffix is added.~

Safer: Create a cache for pgUsers and add team roles to name list to pass to `readPgUsersFromDatabase` that do not exist in pgUsers anymore. Lowers the risk of renaming existing DB roles that are not (yet) in the manifest maybe due to cloning/migration etc. However, if the operator pod would crash the users cache is gone and we don't know if roles can be renamed.

A new field is added to pgUser struct to mark it for deprecation. The flow is as follows:

### New Role
1. New role will end up in pgUser map during initUsers phase
2. Create user name array with pgUsers **and** role names with deprecation suffix (we do this every time just in case).
3. New role is not found in DB and therefore not in dbUsers map.
4. ProduceSyncRequests will trigger a new PGSyncUserAdd request because pgUsers != dbUsers.

### Role is removed
1. Removed role will not be added to pgUsers, but pgUserCache still knows it
2. Create map with pgUsers and name+suffix (the removed role is still not covered here)
3. Because there is a cached role that is not in the pgUser map it is also added to the DB query.
4. The removed role is found and added to dbUsers. It doesn't have the suffix yet so the `Deprecated` field is still left to `false`
5. In, ProduceSyncRequests an extra role in dbUsers map triggers a rename request
6. The removed role is renamed and gets the suffix from the strategy struct

### Role is added again
1. New role will end up in pgUser map during initUsers phase
2. Create map with pgUsers and name+suffix. The re-added role is not in the cache, but this is not checked.
3. The new (old) role is not found and hence not in the dbUsers map
4. The role with the deprecation suffix however is found. It's added to dbUsers with `Deprecated` field set to `true`.
5. Check the counterpart (without the suffix) in the pgUsers map and set the Deprecated field to `true` as well. This is fine because this pgUser is not part of dbUsers. It will have no affect in the renaming later.
6. In ProduceSyncRequests the user creation for the new (old) role is omitted because `Deprecated` is set to `true`. 
7. The extra dbUser triggers a rename request. But, because this user was found in the DB with the deprecation suffix the role will be renamed to the original role name.


pkg/util/users/users.go
```diff
@@ -124,6 +138,30 @@ func (strategy DefaultUserSyncStrategy) alterPgUserSet(user spec.PgUser, db *sql
...
 
+func (strategy DefaultUserSyncStrategy) alterPgUserRename(user spec.PgUser, db *sql.DB) error {
+	var query string
+
+	renamedBack := false
+
+	nameSuffixDiff := len(user.Name) - len(constants.RoleRenameSuffix)
+
+	if nameSuffixDiff > 0 {
+		if user.Name[nameSuffixDiff:] == constants.RoleRenameSuffix {
```
sdudoladov, line 149: imo the usage of standard functions makes the code clearer:
```golang
strings.HasSuffix(user.Name, constants.RoleRenameSuffix)
```
and
```golang
newName := strings.TrimSuffix(user.Name, constants.RoleRenameSuffix)
```
so no need for `nameSuffixDiff`

sdudoladov (commented):
That is a visible change, so it has to be mentioned at least in the admin docs.

sdudoladov:
another interesting bur minor  question is what happens if someone creates a role that is named exactly like `RoleRenameSuffix`, e.g, `_delete_me`

pkg/util/users/users.go
FxKu: ah, good to know :smiley:

pkg/util/users/users.go
```diff
@@ -124,6 +153,22 @@ func (strategy DefaultUserSyncStrategy) alterPgUserSet(user spec.PgUser, db *sql
 	return nil
 }
 
+func (strategy DefaultUserSyncStrategy) alterPgUserRename(user spec.PgUser, db *sql.DB) error {
+	var query string
+
+	if user.Deleted {
```
sdudoladov, line 159: I find the naming here misleading:
if the user is deleted why do we actually trim the `RoleDeletionSuffix` from the `newName` ?
That reads non-obvious to me

pkg/cluster/sync.go
```diff
@@ -569,6 +588,16 @@ func (c *Cluster) syncRoles() (err error) {
 		return fmt.Errorf("error getting users from the database: %v", err)
 	}
 
+	// update pgUsers where a deleted role was found
+	// so that they are skipped in ProduceSyncRequests
+	for _, dbUser := range dbUsers {
+		if originalUser, exists := deletedUsers[dbUser.Name]; exists {
+			recreatedUser := c.pgUsers[originalUser]
+			recreatedUser.Deleted = true
+			c.pgUsers[originalUser] = recreatedUser
```
sdudoladov, line 597: what about sth like
```golang
c.pgUsers[originalUser].Deleted = true
```
? Just to elimiante the extra variable "recreatedUser"

pkg/util/users/users.go
```diff
@@ -70,6 +75,25 @@ func (strategy DefaultUserSyncStrategy) ProduceSyncRequests(dbUsers spec.PgUserM
 		}
 	}
 
+	// No existing roles are deleted or stripped of role membership/flags
+	// but team roles will be renamed and denied from LOGIN
```
sdudoladov, line 79: What would happen to roles coming from the Teams API if the API becomes unavailable ?

pkg/cluster/sync.go
FxKu: `cannot assign to struct field c.pgUsers[originalUser].Deleted in map`

pkg/util/users/users.go
FxKu: Good point. I thought, there will be an error and the sync will stop, but no - it continues only with additional members. I return the errors now.

Jan-M:
:+1:

FxKu:
I will add an option to enable/disable the feature.

FxKu:
:+1:

sdudoladov:
:+1:

PR #1463, opened 2021-04-19
docs: add note about archive timeline reset on upgrade

bchrobot:
For example, a Postgres 12 cluster with backups saved for 7 days may have timelines 5-8. Upon major version upgrade, the timeline is reset to 1. After a few days there may now be timelines 1-4 (PG 13) and 5-8 (PG 12) in cloud storage.  A recovery attempt without a specific timeline set via `recovery_target_timeline` will fail as the PG 12 timeline 5 does not follow the PG 13 timeline 4.

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -48,6 +48,11 @@ upgrade procedure, refer to the [corresponding PR in Spilo](https://github.com/z
 When `major_version_upgrade_mode` is set to `manual` the operator will run
 the upgrade script for you after the manifest is updated and pods are rotated.
 
+A major version upgrade will reset the archive timeline to 1. If you are archiving
+to cloud storage you will need to manually remove basebackups and WAL segments from
+pre-upgrade timelines so Postgres does not try to restore from these seemingly-later
+timelines.
+
 ## CRD Validation
 
 [CustomResourceDefinitions](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#customresourcedefinitions)


CyberDem0n:
> Upon major version upgrade, the timeline is reset to 1

Yes, this is absolutely standard behavior. Major upgrade with pg_upgrade involves initializing the new PGDATA with `initdb`, which has timeline=1

> A recovery attempt without a specific timeline set via recovery_target_timeline will fail as the PG 12 timeline 5 does not follow the PG 13 timeline 4.

Sorry, but this is not true. Backups for different major versions are written to different places in the bucket:
https://github.com/zalando/spilo/blob/c91248e26e2ea910304d04a3acbeda1e965e2e42/postgres-appliance/scripts/configure_spilo.py#L763
```python
bucket_path = '/spilo/{WAL_BUCKET_SCOPE_PREFIX}{SCOPE}{WAL_BUCKET_SCOPE_SUFFIX}/wal/{PGVERSION}'.format(**wale)
```

I.e., for version 12 it would be `/spilo/very-long-uid/my-cluster-name/wal/12` and for version 13 it would be `/spilo/very-long-uid/my-cluster-name/wal/13`.

When restoring from the backup you either have to specify the exact location where to restore from, or the backup-restore script will try to restore from all possible locations until it finds something:
```
/spilo/very-long-uid/my-cluster-name/wal/13
/spilo/very-long-uid/my-cluster-name/wal/12
/spilo/very-long-uid/my-cluster-name/wal/11
/spilo/very-long-uid/my-cluster-name/wal/10
/spilo/very-long-uid/my-cluster-name/wal/9.6
/spilo/very-long-uid/my-cluster-name/wal/9.5
/spilo/very-long-uid/my-cluster-name/wal
```

And once the suitable backup was found it will stick to this location.

If it started restoring the backup from version 13 there is no way it could jump back to 12.

PR #1471, opened 2021-04-26
Support for readOnlyRootFilesystem and not setting requests/limits

JJJJJones:
#1371 fell through but I still had some improvements to contribute that i've rebuilt: readOnlyRootFilesystem support (best security practice where doable) and the ability to not set resource requests/limits of containers/sidecars.

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -353,6 +353,10 @@ configuration they are grouped under the `kubernetes` key.
   that should be assigned to the Postgres pods. The priority class itself must
   be defined in advance. Default is empty (use the default priority class).
 
+  * **read_only_root_filesystem**
+  boolean parameter controlling whether containers should use [readOnlyRootFilesystem](https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems).
+  `kubectl explain statefulset.spec.template.spec.securityContext.readOnlyRootFilesystem`
+
 * **spilo_runasuser**
   sets the user ID which should be used in the container to run the process.
   This must be set to run the container without root. By default the container
--- a/manifests/postgresql.crd.yaml
+++ b/manifests/postgresql.crd.yaml
@@ -299,6 +299,8 @@ spec:
                             type: boolean
                           defaultRoles:
                             type: boolean
+              readOnlyRootFilesystem:
+                type: boolean
               replicaLoadBalancer:  # deprecated
                 type: boolean
               resources:
--- a/pkg/apis/acid.zalan.do/v1/postgresql_type.go
+++ b/pkg/apis/acid.zalan.do/v1/postgresql_type.go
@@ -53,27 +53,28 @@ type PostgresSpec struct {
 	// load balancers' source ranges are the same for master and replica services
 	AllowedSourceRanges []string `json:"allowedSourceRanges"`
 
-	NumberOfInstances     int32                       `json:"numberOfInstances"`
-	Users                 map[string]UserFlags        `json:"users,omitempty"`
-	MaintenanceWindows    []MaintenanceWindow         `json:"maintenanceWindows,omitempty"`
-	Clone                 *CloneDescription           `json:"clone,omitempty"`
-	ClusterName           string                      `json:"-"`
-	Databases             map[string]string           `json:"databases,omitempty"`
-	PreparedDatabases     map[string]PreparedDatabase `json:"preparedDatabases,omitempty"`
-	SchedulerName         *string                     `json:"schedulerName,omitempty"`
-	NodeAffinity          *v1.NodeAffinity            `json:"nodeAffinity,omitempty"`
-	Tolerations           []v1.Toleration             `json:"tolerations,omitempty"`
-	Sidecars              []Sidecar                   `json:"sidecars,omitempty"`
-	InitContainers        []v1.Container              `json:"initContainers,omitempty"`
-	PodPriorityClassName  string                      `json:"podPriorityClassName,omitempty"`
-	ShmVolume             *bool                       `json:"enableShmVolume,omitempty"`
-	EnableLogicalBackup   bool                        `json:"enableLogicalBackup,omitempty"`
-	LogicalBackupSchedule string                      `json:"logicalBackupSchedule,omitempty"`
-	StandbyCluster        *StandbyDescription         `json:"standby,omitempty"`
-	PodAnnotations        map[string]string           `json:"podAnnotations,omitempty"`
-	ServiceAnnotations    map[string]string           `json:"serviceAnnotations,omitempty"`
-	TLS                   *TLSDescription             `json:"tls,omitempty"`
-	AdditionalVolumes     []AdditionalVolume          `json:"additionalVolumes,omitempty"`
+	NumberOfInstances      int32                       `json:"numberOfInstances"`
+	Users                  map[string]UserFlags        `json:"users,omitempty"`
+	MaintenanceWindows     []MaintenanceWindow         `json:"maintenanceWindows,omitempty"`
+	Clone                  *CloneDescription           `json:"clone,omitempty"`
+	ClusterName            string                      `json:"-"`
+	Databases              map[string]string           `json:"databases,omitempty"`
+	PreparedDatabases      map[string]PreparedDatabase `json:"preparedDatabases,omitempty"`
+	SchedulerName          *string                     `json:"schedulerName,omitempty"`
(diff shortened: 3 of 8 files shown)


ReSearchITEng:
Hi,
this functionality has been introduced recently and it's already available in latest release 1.6.2

JJJJJones:
Hello @ReSearchITEng, in our experience yes. We've been testing readOnlyRootFilesystem for months now without issue.

FxKu:
Thank you @JJJJJones. You are introducing the option also to the `postgresql` CRD? Would somebody want to define such option per cluster and not only globally. Maybe you've mixed it up with the `OperatorConfiguration` CRD, because there it is currently missing.

JJJJJones:
@FxKu i'll correct that.

JJJJJones:
@FxKu 
OperatorConfiguration CRD: I added support for readOnlyRootFilesystem in the CRD and operator_config.go. I've tested the regex for the OperatorConfiguration CRD and it supports null values correctly. Since the OperatorConfiguration CRD is imported into a *config.Config the backend should support it correctly. Documentation has been added to reflect the changes.

docs/reference/operator_parameters.md
```diff
@@ -419,7 +427,7 @@ configuration they are grouped under the `kubernetes` key.
 
 This group allows you to configure resource requests for the Postgres pods.
 Those parameters are grouped under the `postgres_pod_resources` key in a
-CRD-based configuration.
+CRD-based configuration. `0` is the null value for all resource requests. Any limit that's set requires its request counterpart to be set for scheduling.
```
FxKu, line 430:
```suggestion
CRD-based configuration. `0` is the null value for all resource requests. 
Any limit that's set requires its request counterpart to be set for scheduling.
```

pkg/util/util.go
```diff
@@ -338,5 +338,9 @@ func IsSmallerQuantity(requestStr, limitStr string) (bool, error) {
 		return false, fmt.Errorf("could not parse limit %v : %v", limitStr, err2)
 	}
 
+	if limitStr == "0" { // Zero disables setting a limit
+		return false, nil
+	}
```
FxKu, line 343: not sure about this. Maybe we should better lift the `required` setting in the CRD schema in another PR.

FxKu:
@JJJJJones can you resolve the conflicts? I want to merge this PR for the next release.

pkg/util/util.go
JJJJJones: To the best of my memory, the reason for this conditional was to prevent the code after from executing because `return request.Cmp(limit) == -1, nil` would always, given a non-zero request and a zero (null) limit, return false even though setting a request but not a limit is perfectly valid, legitimate, and potentially desirable behavior.

As for the required setting I assume you're referring to [this](https://github.com/zalando/postgres-operator/blob/32743e2058f0e7e7895cb1a36f7f873d965d82bd/manifests/postgresql.crd.yaml#L411)? I'm not sure how that's related though.

JJJJJones:
@FxKu i've cleared up the conflicts.

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -68,6 +68,7 @@ type PostgresSpec struct {
 	Tolerations           []v1.Toleration             `json:"tolerations,omitempty"`
 	Sidecars              []Sidecar                   `json:"sidecars,omitempty"`
 	InitContainers        []v1.Container              `json:"initContainers,omitempty"`
+	ReadOnlyRootFilesystem *bool                       `json:"readOnlyRootFilesystem,omitempty"`
```
FxKu, line 71: looks like go.fmt did not run here. All other lines should normally adapt to the new option. Remove also the two spaces introduced above.

manifests/postgresql.crd.yaml
```diff
@@ -395,6 +395,8 @@ spec:
                             type: boolean
                     secretNamespace:
                       type: string
+              readOnlyRootFilesystem:
+                type: boolean
```
FxKu, line 399: remove, please. Should only be a global option for now

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -242,6 +242,7 @@ type OperatorConfigurationData struct {
 	MajorVersionUpgrade        MajorVersionUpgradeConfiguration   `json:"major_version_upgrade"`
 	Kubernetes                 KubernetesMetaConfiguration        `json:"kubernetes"`
 	PostgresPodResources       PostgresPodResourcesDefaults       `json:"postgres_pod_resources"`
+	ReadOnlyRootFilesystem     *bool                              `json:"read_only_root_filesystem,omitempty"`
```
FxKu, line 245: please move it into the KubernetesMetaConfiguration struct next to spiloPrivilege option

pkg/controller/operator_config.go
```diff
@@ -122,6 +122,8 @@ func (c *Controller) importConfigurationFromCRD(fromCRD *acidv1.OperatorConfigur
 	result.PodAntiAffinityTopologyKey = util.Coalesce(fromCRD.Kubernetes.PodAntiAffinityTopologyKey, "kubernetes.io/hostname")
 	result.PodToleration = fromCRD.Kubernetes.PodToleration
 
+	result.ReadOnlyRootFilesystem = fromCRD.ReadOnlyRootFilesystem
```
FxKu, line 125:
```suggestion
	result.ReadOnlyRootFilesystem = fromCRD.Kubernetes.ReadOnlyRootFilesystem
```

docs/reference/operator_parameters.md
```diff
@@ -164,6 +164,10 @@ Those are top-level keys, containing both leaf keys and groups.
   [operator deployment manually](https://github.com/zalando/postgres-operator/blob/master/manifests/postgres-operator.yaml#L20).
   The default is `false`.
 
+* **read_only_root_filesystem**
+  This sets readOnlyRootFilesystem for the Postgres pods.
+  The default is `false`.
```
FxKu, line 169: remove. You already have it under Kubernetes options

docs/reference/operator_parameters.md
```diff
@@ -398,6 +402,10 @@ configuration they are grouped under the `kubernetes` key.
   that should be assigned to the Postgres pods. The priority class itself must
   be defined in advance. Default is empty (use the default priority class).
 
+  * **read_only_root_filesystem**
```
FxKu, line 405: fix indentation of this part, please

manifests/operatorconfiguration.crd.yaml
```diff
@@ -317,6 +317,8 @@ spec:
                       type: string
                   watched_namespace:
                     type: string
+              read_only_root_filesystem:
+                type: boolean
```
FxKu, line 321: Please move under `kubernetes` section and follow alphabetical sorting. Can you also add it [here](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/crds/operatorconfigurations.yaml) and [here](https://github.com/zalando/postgres-operator/blob/695ad44caffa7c03e19a1b59796d93942d8cb70c/pkg/apis/acid.zalan.do/v1/crds.go#L1175).

pkg/cluster/k8sres.go
```diff
@@ -159,32 +159,38 @@ func generateResourceRequirements(resources acidv1.Resources, defaultResources a
...
 	} else {
-		requests[v1.ResourceMemory], err = resource.ParseQuantity(defaults.Memory)
+		resources[v1.ResourceMemory], err = resource.ParseQuantity(defaults.Memory)
 		if err != nil {
 			return nil, fmt.Errorf("could not parse default memory quantity: %v", err)
 		}
 	}
 
-	return requests, nil
+	return resources, nil
```
FxKu, line 193: Can also these changes maybe go up in a seperate PR? Let's focus solely on the `readOnlyRootFilesystem` option here.

FxKu:
@JJJJJones did a proper review now and found a few more issues:
1. Remove `readOnlyRootFilesystem` as a manifest option. You've implemented obtaining the value only from the config, anyway.
2. Move the option under `Kubernetes` struct of `OperatorConfiguration` CRD
3. Add options to examples in manifest and values.yaml, like it's documented [here](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters).
4. Put changes related to request/limits in a second PR so that we only discuss readOnlyRootFilesystem here

PR #1475, opened 2021-04-28
Mount additional volumes to 'postgres' container when 'targetContains` is an empty list

### LINKED ISSUE #1459: No additionalVolumes mounted when targetContainers is an empty list
- **Which image of the operator are you using?** `registry.opensource.zalan.do/acid/postgres-operator:v1.6.2`
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K8s (`kubeadm`)
- **Type of issue?** [Bug report, question, feature request, etc.] Documentation/actual behavior mismatch

For `additionalVolumes` in the postgres manifest, [the docs](https://postgres-operator.readthedocs.io/en/latest/reference/cluster_manifest/) state:

> If `targetContainers` is empty, additional volumes will be mounted only in the `postgres` container.

This currently is correct only if you set `null` for _empty_. Assigning an empty list (`[]`) (as I did initially) causes the operator not to mount the volume to the `postgres` container (but still adds it to `volumes`).

So, either the code does not show the desired behavior or the docs need an update. IMHO, taking an empty list as _empty_ besides `null` makes sense.

cucxabong:
Fix #1459

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -1455,7 +1455,7 @@ func (c *Cluster) addAdditionalVolumes(podSpec *v1.PodSpec,
 			continue
 		}
 
-		if v.TargetContainers == nil {
+		if v.TargetContainers == nil || len(v.TargetContainers) == 0 {
 			spiloContainer := podSpec.Containers[0]
 			additionalVolumes[i].TargetContainers = []string{spiloContainer.Name}
 		}
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -1455,7 +1455,7 @@ func (c *Cluster) addAdditionalVolumes(podSpec *v1.PodSpec,
 			continue
 		}
 
-		if v.TargetContainers == nil || len(v.TargetContainers) == 0 {
+		if len(v.TargetContainers) == 0 {
 			spiloContainer := podSpec.Containers[0]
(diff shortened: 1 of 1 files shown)


pkg/cluster/k8sres.go
```diff
@@ -1455,7 +1455,7 @@ func (c *Cluster) addAdditionalVolumes(podSpec *v1.PodSpec,
 			continue
 		}
 
-		if v.TargetContainers == nil {
+		if v.TargetContainers == nil || len(v.TargetContainers) == 0 {
```
FxKu, line 1458:
```suggestion
		if len(v.TargetContainers) == 0 {
```

FxKu, line 1458:
```suggestion
		if len(v.TargetContainers) == 0 {
```

FxKu:
:+1:

FxKu:
@cucxabong could you add test cases [here](https://github.com/zalando/postgres-operator/blob/f0f7f25d309590713fe549e0c0664b46fdece11b/pkg/cluster/k8sres_test.go#L1179) for targetContainers being `nil` and empty (`[]string{}`). I tried with adding the cases to `additionalVolumeMount` variable and raised `numMounts` by it's length. But maybe you got a better idea.

FxKu:
:+1:

RafiaSabih:
:+1:

sdudoladov:
:+1:

PR #1318, opened 2021-01-18
fix helm install

Gottox:
When installing the postgres operator version 1.6.0 through helm I get the following error message:

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(OperatorConfiguration.configuration.aws_or_gcp.enable_ebs_gp3_migration): invalid type for do.zalan.acid.v1.OperatorConfiguration.configuration.aws_or_gcp.enable_ebs_gp3_migration: got "string", expected "boolean", ValidationError(OperatorConfiguration.configuration.enable_pgversion_env_var): invalid type for do.zalan.acid.v1.OperatorConfiguration.configuration.enable_pgversion_env_var: got "string", expected "boolean", ValidationError(OperatorConfiguration.configuration.enable_spilo_wal_path_compat): invalid type for do.zalan.acid.v1.OperatorConfiguration.configuration.enable_spilo_wal_path_compat: got "string", expected "boolean", ValidationError(OperatorConfiguration.configuration.teams_api.enable_postgres_team_crd): invalid type for do.zalan.acid.v1.OperatorConfiguration.configuration.teams_api.enable_postgres_team_crd: got "string", expected "boolean"]
```

The patch below fixes this behavior.

PR #1480, opened 2021-05-03
fix helm's postgres-operator/values.yaml

flypenguin:
the data types in values.yaml are "wrong". if you just perform an "helm install" then the installation fails with a LOT of error messages about wrong data types. 

for example configGeneral.enable_crd_validation is '"true"' (i.e. a string value) in the values.yaml file, but should be just 'true' (a yaml boolean value).

weirdly, configPostgresPodResources.default_cpu_limit is expected to be string although a numerical value (wrong as well?), just like configConnectionPooler.connection_pooler_default_cpu_limit.

this commit fixes it. AFTER applying you can just do this:

    helm upgrade -i pg-operator zalando/pg-operator

TESTED WITH 1.6.2.

see also:
  - https://git.io/J3z4S

### DIFF
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -21,15 +21,15 @@ enableJsonLogging: false
 # general configuration parameters
 configGeneral:
   # choose if deployment creates/updates CRDs with OpenAPIV3Validation
-  enable_crd_validation: "true"
+  enable_crd_validation: true
   # update only the statefulsets without immediately doing the rolling update
-  enable_lazy_spilo_upgrade: "false"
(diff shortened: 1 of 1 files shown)


FxKu:
:-1: see https://github.com/zalando/postgres-operator/pull/1318, but maybe things have change for helm in the meantime.

flypenguin:
oh look two identical PRs. maybe it's a flawed situation breaking expectations of all people using helm 😆

thanks for the quick reply though, that's indeed rare.

FxKu:
Closing, since we merged #1224 .

PR #1482, opened 2021-05-04
Introduce "ignored_service_annotations" parameter

### LINKED ISSUE #1478: Ignore service annotations

registry.opensource.zalan.do/acid/postgres-operator:v1.6.2

AWS EC2 with Rancher 2.5.7

No but planned.

- **Type of issue?** [Bug report, question, feature request, etc.]

Question / Feature request

- **Details**

I am running a test setup with 1 main and 1 replica. The overall setup worked fine, and the cluster is up and running. Occasionally, however, the cluster will re-sync and cuts connections to do so. I first thought it was caused by #1447 but then realized that Rancher caused the resync by adding a service annotation `field.cattle.io/publicEndpoints`. That is, I see messages like:

```
time="2021-05-03T11:28:20Z" level=info msg="reason: new service's annotations does not match the current one: Removed 'field.cattle.io/publicEndpoints'." cluster-name=default/magic-minimal-swarm64da-tpch-zwn
65w pkg=cluster
```

Since this is dynamically added, I cannot simply add it to `custom_service_annotations` (resp. I don't know whether it is possible to make this dynamic). So, is there a way currently to ignore certain annotations? If not, I'd be happy to make a PR given I get a pointer to the correct location.

sdressler:
This PR attempts to fix #1478 by introducing a parameter to ignore specific service annotations.

Rationale: on, e.g., Rancher, additional annotations are added to the LB services. They seem to cause a connection drop on service syncs. These annotations are added again after removal by the operator, enforcing recurring but unnecessary syncs. A solution to prevent this is to ignore these if they are known.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -344,6 +344,10 @@ spec:
                     type: object
                     additionalProperties:
                       type: string
+                  ignored_service_annotations:
+                    type: array
+                    items:
+                      type: string
                   db_hosted_zone:
                     type: string
                     default: "db.example.com"
--- a/charts/postgres-operator/values-crd.yaml
+++ b/charts/postgres-operator/values-crd.yaml
@@ -202,6 +202,11 @@ configLoadBalancer:
   #   keyx: valuez
   #   keya: valuea
 
+  # annotations to be ignored during sync to not cause restarts
+  # ignored_service_annotations:
+  #   - foo
+  #   - bar
+
   # toggles service type load balancer pointing to the master pod of the cluster
   enable_master_load_balancer: false
   # toggles service type load balancer pointing to the replica pod of the cluster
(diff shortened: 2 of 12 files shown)


hjacobs:
Thanks for the PR! (coming here from https://twitter.com/theDressler/status/1389482796872347648)

Aisuko:
We have the same issue here what you said some like random cut connections and sync service due to custom service annotations and thanks for the PR. It's looking good and clean, we will try and do feedback if it is necessary, and help this PR to merge.

sdressler:
Hi @Aisuko, thanks for the feedback; very much appreciated. It seems #1485 makes it into 1.7. Hence this PR will then be obsolete unless you see value in having an explicit distinction between ignored service annotations and ignored "global" annotations.

Aisuko:
Hi @sdressler , thanks for mention this.

FxKu:
Finally, we merged this feature with #1823. Thanks again @sdressler for your first approach.

PR #1490, opened 2021-05-14
Create cross namespace secrets

RafiaSabih:
If namespace is provided with the username in manifest,
then create the secret for that user in that namespace.
Also, the name of the secret will contain the namespace.


pkg/cluster/cluster.go
```diff
@@ -1089,6 +1089,16 @@ func (c *Cluster) initRobotUsers() error {
 		if c.shouldAvoidProtectedOrSystemRole(username, "manifest robot role") {
 			continue
 		}
+		name := username
+		namespace := "default"
```
FxKu, line 1093: default should be the namespace of the cluster, so `c.Namespace`

pkg/cluster/cluster.go
```diff
@@ -1089,6 +1089,16 @@ func (c *Cluster) initRobotUsers() error {
...
+
+		if strings.Contains(username, ".") {
+			splits := strings.Split(username, ".")
+			name = splits[1]
+			namespace = splits[0]
```
FxKu, line 1098: what if the username is `.user`? Then namespace is `""` and should become the default namespace

RafiaSabih: Right now, such a username is considered invalid username and I think this should continue. I don't see a reason why this should be allowed, if no namespace is provided then username only should be provided.

pkg/cluster/cluster.go
```diff
@@ -1089,6 +1093,18 @@ func (c *Cluster) initRobotUsers() error {
 		if c.shouldAvoidProtectedOrSystemRole(username, "manifest robot role") {
 			continue
 		}
+		name := username
+		namespace := c.Namespace
+
+		if strings.Contains(username, ".") {
```
Jan-M, line 1099: Needs some attention.

Also include update to regex in manifest schema.

e2e/tests/test_e2e.py
```diff
@@ -587,6 +587,27 @@ def verify_role():
...
             raise
 
+    @timeout_decorator.timeout(TEST_TIMEOUT_SEC)
+    def test_cross_namespace_secrets(self):
+        '''
+            Test secrets in different namespace
+        '''
+        app_namespace = "appspace"
+        k8s = self.k8s
+        k8s.api.core_v1.create_namespace(app_namespace)
```
sdudoladov, line 597: I am unsure if `kind` always creates a new namespace fast enough. It is better to have some type of timeout / waiting here to avoid spurious test failures.

e2e/tests/test_e2e.py
```diff
@@ -587,6 +587,27 @@ def verify_role():
...
+            'postgresqls', 'acid-minimal-cluster',
+            {
+                'spec': {
+                    'users':{
+                        'appspace.db_user': [],
+                    }
+                }
+            })
+        self.eventuallyEqual(lambda: k8s.count_secrets_in_namespace(app_namespace),
+                             1, "Secret not created in user namespace")
```
sdudoladov, line 609: To  make it more informative, please log the name of the secret and the namespace

pkg/cluster/cluster_test.go
```diff
@@ -845,3 +847,84 @@ func TestPreparedDatabases(t *testing.T) {
...
+		}, client, pg, logger, eventRecorder)
+
+	userNamespaceMap := map[string]string{
+		cluster.Namespace: "db_user",
+		"appspace":        "appspace.db_user",
+	}
+
+	err := cluster.initRobotUsers()
+	if err != nil {
+		t.Errorf("%s Could not create namespaced users with error: %s", testName, err)
```
sdudoladov, line 910: 1. Does `namespaced user` here really mean `namespaced secret` ?
2. I doubt there is a need to log a test name, the golang test framework likely reports it automatically

pkg/cluster/cluster_test.go
```diff
@@ -845,3 +847,84 @@ func TestPreparedDatabases(t *testing.T) {
...
+		if u.Name != userNamespaceMap[u.Namespace] {
+			t.Errorf("%s: Could not create namespaced user in its correct namespaces for user %s in namespace %s", testName, u.Name, u.Namespace)
+		}
+	}
+}
+
+func TestValidUsernames(t *testing.T) {
+	testName := "test username validity"
+
+	invalidUsernames := []string{"_", ".", ".user", "appspace.", "appspace.user.extra", "user_", "_user", "-user", "user-", ",", ",user", "user,", "namespace,user"}
```
sdudoladov, line 923: Is the name `-` disallowed by other means ?

pkg/cluster/sync.go
```diff
@@ -556,20 +556,13 @@ func (c *Cluster) syncRoles() (err error) {
...
-			userNames = append(userNames, u.Name+c.OpConfig.RoleDeletionSuffix)
-		}
-	}
-
-	// add team members that exist only in cache
-	// to trigger a rename of the role in ProduceSyncRequests
-	for _, cachedUser := range c.pgUsersCache {
-		if _, exists := c.pgUsers[cachedUser.Name]; !exists {
-			userNames = append(userNames, cachedUser.Name)
+		pg_role := u.Name
```
sdudoladov, line 559: `pg_role` should be `pgRole` because of the golang [naming conventions](https://golang.org/doc/effective_go#mixed-caps).

pkg/cluster/sync.go
```diff
@@ -556,20 +556,13 @@ func (c *Cluster) syncRoles() (err error) {
...
-			deletedUsers[u.Name+c.OpConfig.RoleDeletionSuffix] = u.Name
-			userNames = append(userNames, u.Name+c.OpConfig.RoleDeletionSuffix)
-		}
-	}
-
-	// add team members that exist only in cache
-	// to trigger a rename of the role in ProduceSyncRequests
-	for _, cachedUser := range c.pgUsersCache {
-		if _, exists := c.pgUsers[cachedUser.Name]; !exists {
-			userNames = append(userNames, cachedUser.Name)
```
sdudoladov, line 571: What happened to these lines ? they are still needed for renaming roles removed fro PostgresTeam CRD

pkg/cluster/sync.go
```diff
@@ -556,20 +556,13 @@ func (c *Cluster) syncRoles() (err error) {
...
+		pg_role := u.Name
+		if u.Namespace != c.Namespace {
+			// to avoid the conflict of having multiple users of same name
+			// but each in different namespace.
+			pg_role = fmt.Sprintf("%s.%s", u.Name, u.Namespace)
```
sdudoladov, line 563: so the template for the secret name is `namespace.username` but (in case of this name conflict) the role also gets a namespace as part of its name ? Will the resultant secret contain the namespace twice ?

charts/postgres-operator/values-crd.yaml
```diff
@@ -145,7 +145,9 @@ configKubernetes:
 
   # Postgres pods are terminated forcefully after this timeout
   pod_terminate_grace_period: 5m
-  # template for database user secrets generated by the operator
+  # template for database user secrets generated by the operator,
+  # here username contains the namespace in the format namespace.username
+  # if the user is in different namespace than cluster
```
sdudoladov, line 150: I think it is the secret in the different namespace not the user.

e2e/tests/test_e2e.py
```diff
@@ -587,6 +586,33 @@ def verify_role():
...
+    @timeout_decorator.timeout(TEST_TIMEOUT_SEC)
+    def test_cross_namespace_secrets(self):
+        '''
+            Test secrets in different namespace
+        '''
+        app_namespace = "appspace"
+        k8s = self.k8s
+        v1_appnamespace = client.V1Namespace(metadata=client.V1ObjectMeta(name=app_namespace))
+        try:
+            k8s.api.core_v1.create_namespace(v1_appnamespace)
```
sdudoladov, line 598: It is unlikely the actual namespace creation will timeout.  What I meant in my previous review was the following:

1. The API call to create namespace may take some time, especially on `kind` where we know from experience even basic k8s operations sometimes take minutes
2. the patch at [line 603](https://github.com/zalando/postgres-operator/pull/1490/files#diff-0b56893556c9ee8b6a3f37f0e8984f343da0fd397154b370938a433fa4aa475dR603) may run into problems because of that
3. Since we run e2e tests in different environments (locally, GitHub Action, CDP etc) that may lead to unwanted spurious test failures

Earlier I used simple timeouts in such cases,  but now we have function like `self.k8s.wait_for_pod_start` and so on. You may just add a new one named `self.k8s.wait_for_namespace_creation`

e2e/tests/test_e2e.py
```diff
@@ -587,6 +586,33 @@ def verify_role():
...
+            'postgresqls', 'acid-minimal-cluster',
+            {
+                'spec': {
+                    'users':{
+                        'appspace.db_user': [],
+                    }
+                }
+            })
+        self.eventuallyEqual(lambda: k8s.count_secrets_with_label("cluster_name=acid-minimal-cluster,application=spilo", app_namespace),
+                             1, "Secret not created for user in namespace", app_namespace)
```
sdudoladov, line 614: To which [parameter of `eventuallyEqual`](https://github.com/zalando/postgres-operator/blob/b2171beb67d78972eb2565d693d73f390376b134/e2e/tests/test_e2e.py#L38) does `app_namespace` correspond ?

pkg/cluster/sync.go
```diff
@@ -556,7 +556,13 @@ func (c *Cluster) syncRoles() (err error) {
 
 	// create list of database roles to query
 	for _, u := range c.pgUsers {
-		userNames = append(userNames, u.Name)
+		pgRole := u.Name
+		if u.Namespace != c.Namespace {
+			// to avoid the conflict of having multiple users of same name
+			// but each in different namespace.
+			pgRole = fmt.Sprintf("%s.%s", u.Name, u.Namespace)
```
sdudoladov, line 563: I feel that behavior needs to be logged and mentioned in the docs. Just because it may happen an end user unaware of the name conflict expects to get `username` but gets instead `namespace.username`

e2e/tests/test_e2e.py
```diff
@@ -314,15 +314,14 @@ def test_overwrite_pooler_deployment(self):
         'postgresqls', 'acid-minimal-cluster',
         {
             'spec': {
-                'enableConnectionPooler': False
+                c
```
FxKu, line 317:
```suggestion
                'enableConnectionPooler': False
```

pkg/cluster/cluster.go
```diff
@@ -1105,6 +1109,17 @@ func (c *Cluster) initRobotUsers() error {
...
 			continue
 		}
+		namespace := c.Namespace
+
+		//if namespaced secrets are allowed
+		if c.Postgresql.Spec.EnableNamespacedSecret != nil &&
+			*c.Postgresql.Spec.EnableNamespacedSecret {
+			if strings.Contains(username, ".") {
+				splits := strings.Split(username, ".")
+				namespace = splits[0]
```
FxKu, line 1119: don't we have to update the username, too?

```suggestion
				namespace = splits[0]
				username = splits[1]
```

pkg/cluster/sync.go
```diff
@@ -556,7 +556,13 @@ func (c *Cluster) syncRoles() (err error) {
 
 	// create list of database roles to query
 	for _, u := range c.pgUsers {
-		userNames = append(userNames, u.Name)
+		pgRole := u.Name
+		if u.Namespace != c.Namespace {
```
FxKu, line 560: make sure `u.Namespace` is always set. Currently it's missing in `initTeamMembers` and `readPgUsersFromDatabase` methods which leads to the wrong name appended here. This makes the second e2e test fail atm (it wants to create a user that already exists).

pkg/cluster/cluster.go
Jan-M: We use namespace in username to prevent conflicts of usernames.

Jan-M:
:+1:

RafiaSabih:
:+1:

FxKu:
:+1:

PR #1504, opened 2021-05-27
get postgres container by name, not index

FxKu:
In some code parts we assume that containers[0] is the postgres container. #1381 has shown that this might not always be the case. This PR tries to eliminate every code with that assumption.


sdudoladov:
:+1:

FxKu:
:+1:

PR #1505, opened 2021-05-28
pick first container if postgres is not found

FxKu:
with #1504 we stopped assuming the first container being the postgres container. Hower, to avoid breaking things, bring back the old behavior as a fallback. Plus, some more refactoring.

### DIFF
--- a/pkg/cluster/util.go
+++ b/pkg/cluster/util.go
@@ -227,12 +227,18 @@ func (c *Cluster) logServiceChanges(role PostgresRole, old, new *v1.Service, isU
 	}
 }
 
-func (c *Cluster) getPostgresContainer(podSpec *v1.PodSpec) v1.Container {
+func getPostgresContainer(podSpec *v1.PodSpec) v1.Container {
 	var pgContainer v1.Container
+
 	for _, container := range podSpec.Containers {
 		if container.Name == constants.PostgresContainerName {
 			pgContainer = container
 		}
+	}
+
+	// if no postgres container was found, take the first one in the podSpec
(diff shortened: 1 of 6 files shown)


FxKu:
:+1:

pkg/cluster/util.go
```diff
@@ -227,13 +227,19 @@ func (c *Cluster) logServiceChanges(role PostgresRole, old, new *v1.Service, isU
 	}
 }
 
-func (c *Cluster) getPostgresContainer(podSpec *v1.PodSpec) v1.Container {
+func getPostgresContainer(podSpec *v1.PodSpec) v1.Container {
```
sdudoladov, line 230: you can make `pgContainer` a [named return value](https://tour.golang.org/basics/7)

pkg/cluster/util.go
```diff
@@ -227,13 +227,19 @@ func (c *Cluster) logServiceChanges(role PostgresRole, old, new *v1.Service, isU
...
 	var pgContainer v1.Container
+
 	for _, container := range podSpec.Containers {
 		if container.Name == constants.PostgresContainerName {
 			pgContainer = container
 		}
 	}
+
+	// if no postgres container was found, take the first one in the podSpec
```
sdudoladov, line 239: that sounds very suspicious. Either it is simply an error or a PG operator deployment with non-standard container names (for example a very old one from a couple of years ago). Silently choosing the very first container is not the very best option in both cases.
That is still likely to work because of how we used container names before, but at least that case need to be acknowledged as a warning.

sdudoladov:
:+1:

FxKu:
:+1:

PR #1512, opened 2021-06-02
define default access privileges for default users too

### LINKED ISSUE #1170: Default privileges not setup up for the <db>_owner_user role
Hi,

I'm using the default roles created by postgres-operator (reader/writer/owner) and I'm logging in to the DB with `<db>_owner_user` (which has the LOGIN privilege). However if we now create tables with this user (`<db>_owner_user` ), the default privileges for the other users (_reader/_writer) are not applied. I assume this is because the default privilegs are only applied for the owner of a schema which is `<db>_owner`. Is there a way to work around this or what is the recommended approach? (i.e. always running `SET ROLE <db>_owner` before creating tables?)

Thank you

- **Where do you run it - cloud or metal?** Bare Metal K8s
- **Type of issue?** Feature Request/Guidance/Bug?

FxKu:
Using preparedDatabases allows you to create a set of pre-defined owner, reader and writer NOLOGIN roles. With `defaultUsers` enabled LOGIN roles can be created to which inherit all privileges from their NOLOGIN counterparts except for defalt access privileges. This means, when using tools such as flyway one has to make sure that the NOLOGIN owner role is set before creating new tables. If the migration is executed with the `<dbName>[_<schemaName>]_owner_user` role, reader and writer roles will have no access.

Because such behavior is not very obvious to users, this PR suggests to define default access privileges on the LOGIN roles, too. 

Closes #1170


sdudoladov:
:+1:

sdudoladov:
:+1:

FxKu:
:+1:

Issue #1199, opened 2020-11-06
postgres-operator-ui v1.5.0 cannot be installed using helm 3


dennisjac:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which repo tag are you using?** v.1.50
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** GKE v1.16.13-gke.401
- **Are you running Postgres Operator in production?** not yet
- **Type of issue?** Bug report

I'm trying to follow the quickstart guide but ran into an issue with installing the postgres-operator-ui via helm 3:
```
helm install postgres-operator-ui ./charts/postgres-operator-ui

Error: Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false},{"na|..., bigger context ...|uster-name"},{"name":"RESOURCES_VISIBLE","value":false},{"name":"TARGET_NAMESPACE","value":"default"|...
```
This seems to be caused by a lack of quotes in the deployment template. When I change this:
```
            - name: "RESOURCES_VISIBLE"
              value: {{ .Values.envs.resourcesVisible }}
```
into this:
```
            - name: "RESOURCES_VISIBLE"
              value: "{{ .Values.envs.resourcesVisible }}"
```
then the installation succeeds.

FxKu:
This should already be [fixed](https://github.com/zalando/postgres-operator/blob/b379db20edc05fa946c6cc50ec13161b80a89c5d/charts/postgres-operator-ui/templates/deployment.yaml#L47) #966.

dennisjac:
I will try that, thanks. I'm a bit surprised though that the new `v1.5.0-dirty` image was released back in May but no `v1.5.1` release was made since given the severity of this and the crash-loop bug in #1200 . Is there a reason for this?

Jan-M:
@dennisjac fair point, I guess it comes down to too little visibility of the GKE impact and what people need.

Expect 1.6 sometime end of November, potentially we can indeed squeze in a 1.5.1. and remember such a release next time something like this happens.

Thunderbolt32:
[...]

PR #1514, opened 2021-06-03
using quote in UI deployment template

FxKu:
as suggested here: https://github.com/zalando/postgres-operator/issues/1199#issuecomment-729660273


FxKu:
:+1:

sdudoladov:
:+1:

PR #1058, opened 2020-07-13
standby-cluster-gs

anshudutta:
Currently, there is a bug which doesn't allow standby cluster to be created from WAL archives from GCP Bucket

Error:

```
Error: postgresql.acid.zalan.do "alchemy-database" is invalid: []: Invalid value: map[string]interface {}{"apiVersion":"acid.zalan.do/v1", "kind":"postgresql", "metadata":map[string]interface {}{"annotations":map[string]interface {}{"meta.helm.sh/release-name":"alchemy", "meta.helm.sh/release-namespace":"alchemy"}, "creationTimestamp":"2020-07-13T06:19:13Z", "generation":1, "labels":map[string]interface {}{"app.kubernetes.io/managed-by":"Helm"}, "name":"alchemy-database", "namespace":"postgres-operator", "uid":"80eca2c8-8959-4c8e-a84e-bc8ff64a965a"}, "spec":map[string]interface {}{"numberOfInstances":1, "patroni":map[string]interface {}{"pg_hba":[]interface {}{"host    all all 0.0.0.0/0 md5"}}, "postgresql":map[string]interface {}{"version":"11"}, "standby":map[string]interface {}{"gs_wal_path":"gs://pgbucket-rl-af7b0aec/spilo/alchemy-database"}, "teamId":"alchemy", "volume":map[string]interface {}{"size":"10Gi", "storageClass":"ssd-alchemy"}}}: validation failure list:
 spec.standby.s3_wal_path in body is required
```
This is because the `s3_wal_path` property is mandatory even when using GCP.

This PR is aimed at fixing that bug

Changes:
1. Spin up a standby cluster using WAL archives from GCP bucket
2. Either s3_wal_path or gs_wal_path are mandatory
3. Setting gs_wal_path will set set environment variable STANDBY_WALE_GS_PREFIX which will then be used by wal-g to use gcp sdk

PR #1515, opened 2021-06-04
Allow configuration of gs_wall_path in for standby clusters

d0x2f:
Hello,

This PR is essentially the same as #1058, but rebased on current master.
I've been using a version of the operator with these modifications for some time and have had no issues.
These changes allow the gs_wal_path value to  be passed along to wale, which is already able to make use of it.

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -374,10 +374,10 @@ spec:
                 type: integer
               standby:
                 type: object
-                required:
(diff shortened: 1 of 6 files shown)


FxKu:
Aside from the mentioned PR we also got #1446 that not only covers your implementation (thanks btw! :smiley:) but also allows for specifying cluster name and UID like with cloning. So I guess, I would aim for merging this one soon.

d0x2f:
oh excellent! looking forward to it :)

PR #1521, opened 2021-06-09
Adjust per pool by number of pools

erthalion:
default_pool_size, min_pool_size, reserve_pool_size are per pool values,
that's it per (database, user) pair. Which means it's wrong to calculate
them based on the global maxDBConn, otherwise for the cases when there
are too many users it would result in the configuration conflicting
between maxDBConn and min_pool_size. Adjust those values by the number
of known pools.

This not necessarily a perfect solution, because load could be
distributed between databases & users unevenly in which case some of
those parameters will cause rarely used pools being of the same size as
the active ones.


FxKu:
unit test fail:
```
--- FAIL: TestConnectionPoolerCreateDeletion (0.00s)
panic: runtime error: integer divide by zero [recovered]
	panic: runtime error: integer divide by zero
goroutine 77 [running]:
testing.tRunner.func1.1(0x1b685e0, 0x2d74dd0)
	/usr/local/go/src/testing/testing.go:1072 +0x30d
testing.tRunner.func1(0xc0000d0d80)
	/usr/local/go/src/testing/testing.go:1075 +0x41a
panic(0x1b685e0, 0x2d74dd0)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
github.com/zalando/postgres-operator/pkg/cluster.(*Cluster).getConnectionPoolerEnvVars(0xc000012000, 0xc000371760, 0xc00037f260, 0xc000357c1c)
	/workspace/pkg/cluster/connection_pooler.go:169 +0x625
	```

Issue #1285, opened 2021-01-05
Set Loadbalancer IP


timmy59100:
I couldn't find an option to set the loadBalancerIP for the  loadbalancer: "enableMasterLoadBalancer".
It would be similiar to allowedSourceRanges but one option per Loadbalancer.

christiancorbin:
Agree this would be a good feature

yevon:
I was just searching for this, nice to see this is going to be implemented!

necipakca:
This is not relevant anymore. Cause loadBalancerIP is already [deprecated in 1.24](https://kubernetes.io/docs/concepts/services-networking/service/#loadbalancer)
You can use service annotations!

PR #1528, opened 2021-06-17
Add support for manually setting load balancer IPs

jamesmcdonald:
This adds the settings `masterLoadBalancerIP` and `replicaLoadBalancerIP` to the `postgresql` CRD. Setting either of these options will cause `loadBalancerIP` to be set on the generated service.

Solves #1285

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -314,6 +314,8 @@ spec:
                                     type: array
                                     items:
                                       type: string
+              masterLoadBalancerIP:
+                type: string
               numberOfInstances:
                 type: integer
                 minimum: 0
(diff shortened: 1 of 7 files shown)


jamesmcdonald:
This rebased version isn't tested yet; I'm not sure how it will interact with the replica pooler load balancer stuff.

yoshihikoueno:
@jamesmcdonald 
Hi, this PR looks really useful and I'd appreciate it if this gets merged.
Is there still something that needs to be solved?

jamesmcdonald:
We don't actually use static IPs in our setup any more because we switched to external-dns, so I haven't chased it up or kept the PR up-to-date with upstream.

Suzamax:
@jamesmcdonald other people do use static IPs. It should be merged tbh

Sharofiddin:
We also need this feature, please merge this.

PR #1535, opened 2021-06-23
refactor restarting instances

FxKu:
refactored restart of instances by moving them into a loop over pods. Same for patching Postgres with `checkAndSetGlobalPostgreSQLConfiguration`

From:
```
checkAndSetGlobalPostgreSQLConfiguration()
-> loop over pods and patch Patroni config (Postgresql.Parameters)
-> return true if bootstrap only parameters are defined in manifest to trigger restart

restartInstances()
-> call Patroni's `config` endpoint to get `ttl` duration for post restart wait
-> loop over pods and call Patroni's `restart` endpoint
```

To:
```
for pod in pods:
  syncPostgreSQLConfiguration(pod)

func syncPostgreSQLConfiguration(pod)
  call Patroni's `config` endpoint

  restartRequired := checkAndSetPostgreSQLConfiguration( pass config for comparison )
    -> patch Patroni config (Postgresql.Parameters)
    -> return true if bootstrap only parameters are defined in manifest *and differ from effective config* to trigger restart

  if restartRequired:
    -> call Patroni's `restart` endpoint
```

This PR also adds the following functionality:
- extends `checkAndSetGlobalPostgreSQLConfiguration` to compare Patroni config with `c.Spec.Patroni`, too:
- gathers all options to set in one config object and send it to a new `c.patroni.SetConfig` endpoint which supports the full config hierarchy compared to `c.patroni.SetPostgresParameters` which is only for `postrgesql.parameters`
- extend (and rename) `test_decrease_max_connections` e2e test (or create a new one) for config changes


e2e/tests/k8s_api.py
```diff
@@ -496,6 +503,13 @@ def exec_with_kubectl(self, pod, cmd):
                               stdout=subprocess.PIPE,
                               stderr=subprocess.PIPE)
 
+    def patroni_rest(self, pod, path):
```
sdudoladov, line 506: duplicate of [246-252](https://github.com/zalando/postgres-operator/pull/1535/files#diff-2fb572115844fab37e322804b14ff5809f4e3e0f2c12f7add8e5509532403a31R246-R252) ?

e2e/tests/test_e2e.py
```diff
@@ -858,6 +858,7 @@ def test_multi_namespace_support(self):
         try:
             k8s.create_with_kubectl("manifests/complete-postgres-manifest.yaml")
             k8s.wait_for_pod_start("spilo-role=master", self.test_namespace)
+            k8s.wait_for_pod_start('spilo-role=replica')
```
sdudoladov, line 861: why does waiting or the master uses the `self.test_namespace` and waiting for the replica does not?

e2e/tests/test_e2e.py
```diff
@@ -1419,52 +1421,79 @@ def test_zzzz_cluster_deletion(self):
...
 
-        def get_max_connections():
+        try:
+            k8s.api.custom_objects_api.patch_namespaced_custom_object(
+                "acid.zalan.do", "v1", "default", "postgresqls", "acid-minimal-cluster", pg_patch_config)
+            
+            self.eventuallyEqual(lambda: self.k8s.get_operator_state(), {"0": "idle"}, "Operator does not get in sync")
+
+            def compare_config():
+                effective_config = k8s.patroni_rest(masterPod.metadata.name, "config")
```
sdudoladov, line 1464: but that only tells you if the Patroni config was updated, not if the change [was actually applied on the PG level](https://github.com/zalando/postgres-operator/pull/1535/files#diff-0b56893556c9ee8b6a3f37f0e8984f343da0fd397154b370938a433fa4aa475dR1427-R1428).
Patroni has[ the `pending_restart` flag](https://github.com/zalando/patroni/blob/75e52226a848d64317ac98264234e62706fd762c/docs/rest_api.rst) to indicate such situation
```bash
$ curl -s http://localhost:8008/patroni | jq .
{
  "pending_restart": true,
```

pkg/cluster/sync.go
```diff
@@ -381,21 +386,52 @@ func (c *Cluster) syncStatefulSet() error {
 	// Apply special PostgreSQL parameters that can only be set via the Patroni API.
 	// it is important to do it after the statefulset pods are there, but before the rolling update
 	// since those parameters require PostgreSQL restart.
-	instancesRestartRequired, err = c.checkAndSetGlobalPostgreSQLConfiguration()
+	pods, err = c.listPods()
 	if err != nil {
-		return fmt.Errorf("could not set cluster-wide PostgreSQL configuration options: %v", err)
+		c.logger.Infof("could not list pods of the statefulset: %v", err)
```
sdudoladov, line 391: that is arguably still an error because the Sync will fail to complete without the list o pods

FxKu: The error was not replaced but moved into the loop. I took inspiration from line 276 where it's also just an info log line. The function would not fail with an empty list of pods. I've change the log level to warning.

e2e/tests/k8s_api.py
FxKu: yes, there's a lot of duplicate code. Don't understand why. I thought to better do it, too :smile: 
Someone should tidy this part up.

e2e/tests/test_e2e.py
FxKu: an oversight, indeed. Might explain why the test fails atm

e2e/tests/test_e2e.py
FxKu: true. Added a check to query the settings in the database

Jan-M:
:+1:

FxKu:
:+1:

PR #1537, opened 2021-06-25
Add Support for Azure WAL-G Backups

apeschel:
This commit adds support for using an Azure storage account as a backup
location.

It uses the existing GCS functionality as a reference for what to do,
and follows the example set by GCS as closely as possible.

The decision to name the cloud provider key "aws_or_gcp" is unfortunate
while adding support for Azure, but I have left it alone to allow for
this changeset to be backwards compatible.


FxKu:
:+1:

michaelmcmillan:
This is great, thanks! When is 1.7 due?

apeschel:
@FxKu Thank you, much appreciated! Looking forward to further contributions to this project!

PR #1569, opened 2021-07-26
fix searching for users with namespace in name

FxKu:
and simplify e2e test for cross namespace secrets.

Removed codelines have become obsolete, because the username will always contain the namespace at this point. At some stage of #1490 we've extracted the user from the full namespace-qualified name, hence making it necessary to add the namespace here. But, as of now, it would result in an error, because the namespace would be appended again (after the username, which seems strange, too). The name of the corresponding e2e test is changed, too, to be executed sooner and not at the end (which led to overseeing this bug).


FxKu:
:+1:

sdudoladov:
:+1:

PR #1570, opened 2021-07-27
create CDC event stream CRD

FxKu:
The PR introduces a new operator feature to enable change data capture (CDC) streams into Zalando’s distributed event broker [Nakadi](https://nakadi.io/). It utilizes an internal Zalando operator for creating [Debezium](https://debezium.io/documentation/reference/tutorial.html) workflows reading changes from Postgres’ logical replication slots and distribute to Nakadi (or AWS SQS)

From a new section in the manifest to list streams, the operator will create or update a custom resource that can be picked up by the CDC operator. The Postgres Operator user must specify the database tables for which logical decoding is enabled as well as event topics which are defined in the sink (Nakadi). The CDC operator is build around the [outbox pattern](https://debezium.io/blog/2019/02/19/reliable-microservices-data-exchange-with-the-outbox-pattern/). It’s idea is to create a de-normalized table in Postgres that resembles the structure of then event sink and perform logical decoding only on this “outbox” table. Either the app or the database is then responsible to copy change data from original tables to the outbox table. The Postgres Operator assumes that the outbox table is called like the specified table, appended with event type and `_outbox` suffix.

ToDo:
- [x] ~fix code generation, atm it has been created manually.~ Running `./hack/update-codegen.sh` only updates acid API
- [x] ~do we need to sync DB resources? Might be the job of the CDC operator to report if specified databases and tables do not exist~ **The operator should not deal with creating tables**
- [x] should we patch the Postgres manifest, setting wal_level to `logical` and creating a logical replication slot under Patroni section? Or is it up to the user to configure it before enabling `streams`? **yes and no: the operator changes the Postgres config without the need to prepare the manifest**
- [x] operator should create a replication user to be used for all event streams


pkg/cluster/streams.go
```diff
@@ -0,0 +1,261 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var outboxTableNameTemplate config.StringTemplate = "{table}_{eventtype}_outbox"
+
+func (c *Cluster) createStreams() error {
+	c.setProcessName("creating streams")
+
+	fes := c.generateFabricEventStream()
+	_, err := c.KubeClient.FabricEventStreams(c.Namespace).Create(context.TODO(), fes, metav1.CreateOptions{})
```
sdudoladov, line 24: Is the Zalando-internal custom operator assumed to always listen to all namespaces ? in other words, does `FabricEventStreams(c.Namespace)` always make sense ?

pkg/cluster/streams.go
```diff
@@ -0,0 +1,261 @@
...
+			Name:        c.credentialSecretNameForCluster(user, c.Name),
+			UserKey:     "username",
+			PasswordKey: "password",
+		},
+	}
+}
+
+func (c *Cluster) getLogicalReplicationSlot(database string) string {
+	for slotName, slot := range c.Spec.Patroni.Slots {
+		if slot["type"] == "logical" && slot["database"] == database {
```
sdudoladov, line 204: what if there are logical slots unrelated to CDC?

pkg/cluster/streams.go
```diff
@@ -0,0 +1,261 @@
...
+				"database": stream.Database,
+				"plugin":   "wal2json",
+				"type":     "logical",
+			}
+			slots[constants.EventStreamSourceSlotPrefix+"_"+stream.Database] = slot
+		}
+	}
+
+	if len(slots) > 0 {
+		c.logger.Debugf("setting wal level to 'logical' in Postgres configuration")
```
sdudoladov, line 82: It pays to log why this change happens

pkg/cluster/streams.go
```diff
@@ -0,0 +1,261 @@
...
+			c.logger.Debugf("creating logical replication slot %q in database %q", slotName, slot["database"])
+		}
+		desiredPostgresConfig["slots"] = slots
+	} else {
+		return nil
+	}
+
+	pods, err := c.listPods()
+	if err != nil || len(pods) == 0 {
+		return err
```
sdudoladov, line 93: is it intended to return `nil` in case `len(pods) == 0` ?
that is not necessary because the combination of [line 95](https://github.com/zalando/postgres-operator/pull/1570/files#diff-6592d6884b6c11bdc6a11828daa2592c9c23b29865c49a30a55f6be5c7602ff7R95) and [line 110](https://github.com/zalando/postgres-operator/pull/1570/files#diff-6592d6884b6c11bdc6a11828daa2592c9c23b29865c49a30a55f6be5c7602ff7R110) already does it

pkg/cluster/streams.go
```diff
@@ -0,0 +1,261 @@
...
+		podName := util.NameFromMeta(pods[i].ObjectMeta)
+		effectivePostgresConfig, err := c.patroni.GetConfig(&pod)
+		if err != nil {
+			c.logger.Warningf("could not get Postgres config from pod %s: %v", podName, err)
+			continue
+		}
+
+		_, err = c.checkAndSetGlobalPostgreSQLConfiguration(&pod, effectivePostgresConfig, desiredPostgresConfig)
+		if err != nil {
+			c.logger.Warningf("could not set PostgreSQL configuration options for pod %s: %v", podName, err)
```
sdudoladov, line 105: Just to double check: here it is assumed the operator will retry setting the `effectivePostgresConfig` on the next Sync, right ?

pkg/cluster/streams.go
```diff
@@ -0,0 +1,261 @@
...
+
+func (c *Cluster) getStreamConnection(database, user string) zalandov1alpha1.Connection {
+	return zalandov1alpha1.Connection{
+		Url:      fmt.Sprintf("jdbc:postgresql://%s.%s/%s?user=%s&ssl=true&sslmode=require", c.Name, c.Namespace, database, user),
+		SlotName: c.getLogicalReplicationSlot(database),
+		DBAuth: zalandov1alpha1.DBAuth{
+			Type:        constants.EventStreamSourceAuthType,
+			Name:        c.credentialSecretNameForCluster(user, c.Name),
+			UserKey:     "username",
+			PasswordKey: "password",
```
sdudoladov, line 197: These are dummy values or ?

pkg/cluster/streams.go
FxKu: Yes

pkg/cluster/cluster.go
```diff
@@ -890,6 +899,10 @@ func (c *Cluster) Delete() {
 	defer c.mu.Unlock()
 	c.eventRecorder.Event(c.GetReference(), v1.EventTypeNormal, "Delete", "Started deletion of new cluster resources")
 
+	if err := c.deleteStreams(); err != nil {
+		c.logger.Warningf("could not delete event streams: %v", err)
```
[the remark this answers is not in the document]
FxKu: I've added an OwnerReference to create a dependency to the statefulSet - like we do with connection pooler

pkg/cluster/cluster.go
```diff
@@ -1043,6 +1056,23 @@ func (c *Cluster) initSystemUsers() {
...
+		username := constants.EventStreamSourceSlotPrefix + constants.UserRoleNameSuffix
+		streamUser := spec.PgUser{
+			Origin:    spec.RoleConnectionPooler,
+			Name:      username,
+			Namespace: c.Namespace,
+			Flags:     []string{constants.RoleFlagLogin, constants.RoleFlagReplication},
+			Password:  util.RandomPassword(constants.PasswordLength),
+		}
+
+		if _, exists := c.pgUsers[username]; !exists {
```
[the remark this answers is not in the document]
FxKu: have to test which role is chosen in the end. At least we have a conflict handling for redundant roles

pkg/cluster/streams.go
FxKu: the FES CRD is created in the cluster's namespace. You think somebody might wish to create it in a separate namespace?

pkg/cluster/streams.go
FxKu: in this case, it would be referenced here. Maybe I can also check for the wal2json plugin. Hm or better rely on a fixed naming schema...

pkg/cluster/streams.go
FxKu: no, this is what FES uses, too

pkg/cluster/streams.go
FxKu: returning a warning now

docs/reference/cluster_manifest.md
```diff
@@ -513,3 +513,33 @@ Those parameters are grouped under the `tls` top-level key.
...
+  database configuration (setting `wal_level: logical`, creating logical
+  replication slots, using output plugin `wal2json` and creating a dedicated
+  replication user). Required.
+
+* **tables**
+  Defines a map of table names and event types. The CDC operator is following
+  the [outbox pattern](https://debezium.io/blog/2019/02/19/reliable-microservices-data-exchange-with-the-outbox-pattern/)
+  meaning changes are only consumed from an extra table that already has the
+  structure of the event in the target sink. The operator will assume that this
+  outbox table is called like `<table>_<event_type>_outbox`. Required.
```
ePaul, line 537: From reading this, I don't understand what exactly is done by this operator, or the CDC operator, and what is the responsibility of the application.
Does this create the outbox table on the DB managed with this? Does this set up some logic (per triggers?) to replicate data from the origin table into the outbox table, or is this the responsibility of the application? 
If the latter, what's the purpose of having the original table mentioned at all?

cjbooms: It's confusing because it is not correct. Application-level terminology from the Outbox Pattern has been misinterpreted and applied to the the DB layer. From a Postgres CDC point of view, change events are emitted from a single table, there is no "extra" table. 

A builder persists a json event payload into a column _on any declared table_. Any change to a row on that table is captured via Logical Replication, with `after` state of the column eventually being emitted onto Nakadi via FES CDC.

ePaul:
```suggestion
* **tables**
  Defines a map of (outbox) table names and event types. The CDC operator is following
  the [outbox pattern](https://debezium.io/blog/2019/02/19/reliable-microservices-data-exchange-with-the-outbox-pattern/).
  This means that the application will put events into a column in the outbox table
  in the structure of the target event type, and the CDC operator will capture them
  shortly after the transaction is committed. Required.
```
(Of course, the implementation also needs to change accordingly.)

I guess we need to define the column name too (either document it, if it's fixed, or have it be configurable here)?

cjbooms: Reads much better. 
In the FES CRD, we have it convention-over-configuration for the column name:
```
    PostgresWalToGenericNakadiEventData:
      type: object
      properties:
        payloadColumn:
          type: string
          description: Name of the column holding the Json payload for the Nakadi event.
          example: my_business_event
          default: "payload"
```

Jan-M: I see some confusion, but I already talked to both ACID/Nakadi to setup internal example repo, with Postgers clusters, Nakadi event type, I think this helps. We were too detached.

charts/postgres-operator/crds/postgresqls.yaml
```diff
@@ -469,6 +469,36 @@ spec:
...
+                    filter:
+                      type: object
+                      additionalProperties:
+                        type: string
+                    tables:
+                      type: object
+                      additionalProperties:
+                        type: object
+                        required:
+                          - evenType
```
cjbooms, line 494: typo

pkg/cluster/cluster.go
```diff
@@ -361,6 +361,10 @@ func (c *Cluster) Create() error {
 	// something fails, report warning
 	c.createConnectionPooler(c.installLookupFunction)
 
+	if err = c.syncStreams(); err != nil {
+		return fmt.Errorf("could not create streams: %v", err)
```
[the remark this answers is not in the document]
FxKu: It's the last step before major version upgrade. I think error makes sense because it's the desired user setup that could not be fulfilled.

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+		return fmt.Errorf("could not update event stream custom resource: %v", err)
+	}
+
+	return nil
+}
+
+func (c *Cluster) deleteStreams() error {
+	c.setProcessName("deleting event streams")
+
+	// check if stream CRD is installed before trying a delete
```
sdudoladov, line 52: that should be a condition for calling `deleteStreams()`

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+				}
+			}
+			publications[slotName] = streamTables
+		}
+	}
+
+	if len(slots) > 0 {
+		desiredPatroniConfig.Slots = slots
+	} else {
+		return nil
```
sdudoladov, line 119: This line in combination with the `defer` above likely logs  a confusing error `"could not close database connection` where there was no connection opened in the first place and in fact there was no error

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+
+	// if streams are defined wal_level must be switched to logical
+	desiredPgParameters := map[string]string{"wal_level": "logical"}
```
sdudoladov, line 123: 1. that is yet another constant to be defined,
2. and it is required rather than desired

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+
+	// apply config changes in pods
+	pods, err := c.listPods()
+	if err != nil || len(pods) == 0 {
+		c.logger.Warningf("could not list pods of the statefulset: %v", err)
```
sdudoladov, line 128: the error message does not cover the case `len(pods) == 0`

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+func (c *Cluster) syncPostgresConfig() error {
+	slots := make(map[string]map[string]string)
+	publications := make(map[string]map[string]acidv1.StreamTable)
+	createPublications := make(map[string]string)
+	alterPublications := make(map[string]string)
+
+	defer func() {
+		if err := c.closeDbConn(); err != nil {
```
sdudoladov, line 84: where exactly is that DB connection opened ?

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+			return fmt.Errorf("could not init database connection")
+		}
+
+		currentPublications, err := c.getPublications()
+		if err != nil {
+			return fmt.Errorf("could not get current publications: %v", err)
+		}
+
+		tableNames := make([]string, len(tables))
+		i := 0
```
sdudoladov, line 160: why is this counter needed ?

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+			createPublications[publication] = tableList
+		} else if currentTables != tableList {
+			alterPublications[publication] = tableList
+		}
+
+		if len(createPublications)+len(alterPublications) == 0 {
+			return nil
+		}
+
+		for publicationName, tables := range createPublications {
```
sdudoladov, line 180: well, there is seemingly no way  golang-native way to combine maps, but 
```golang
for publicationName, tables := range alterPublications {
			if err = c.executeAlterPublication(publicationName, tables); err != nil {
				c.logger.Warningf("%v", err)
			}
		}
```
still may be a small function to reduce code duplication

pkg/cluster/streams.go
```diff
@@ -0,0 +1,349 @@
...
+				EventStreamFlow:   streamFlow,
+				EventStreamSink:   streamSink,
+				EventStreamSource: streamSource})
+		}
+	}
+
+	return &zalandov1.FabricEventStream{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       constants.EventStreamSourceCRDKind,
+			APIVersion: "zalando.org/v1",
```
sdudoladov, line 217: should not that be a constant defined elsewhere instead of the litreal ?

pkg/cluster/streams.go
FxKu: I see what you mean. I copied it from the syncing database to [create or alter](https://github.com/zalando/postgres-operator/blob/3ce0b1e7fa236e8d54fbf066a5fb45fa11c041fd/pkg/cluster/sync.go#L916).

pkg/cluster/streams.go
FxKu: to fill the tableNames slice a bit quicker. I [read](https://stackoverflow.com/questions/21362950/getting-a-slice-of-keys-from-a-map) that append (which we usually use :smiley: ) is slower.

pkg/cluster/streams.go
FxKu: I would prefer not to do that in the Delete() function of Cluster, to keep the code there cleaner. I think, it's fine here.

pkg/cluster/streams.go
FxKu: Renamed. Not sure, if I want to create constants for all literals :confused:

sdudoladov:
:+1:

achrafsahnoun1:
Hey, I want to include the streams section in the cluster manifet of postgres, I didnt find any example on how to write the tables that we want to include, this is one of my trials:
 # Enables change data capture streams for defined database tables
  streams:
  - applicationId: test-app
    database: postgres
    tables:
      data.state_pending_outbox:
        eventType: test-app.status-pending
      data.state_approved_outbox:
        eventType: test-app.status-approved
      data.orders_outbox:
        eventType: test-app.order-completed
        idColumn: o_id
        payloadColumn: o_payload
      data.links_outbox:
        eventType: test-app.order-completed
        idColumn: id
        payloadColumn: payload
///////////////////////////////
In this case I want to include the links table which is in the test database, I checked if there are any publication created from the operator for that table since it should create it automatically but there is no one created. Thank you in advance.

PR #1575, opened 2021-08-02
fix(ui): update ingress api versions in helm chart for newer k8s vers…

### LINKED ISSUE #1574: UI-Ingress is Deprecated
- **Which image of the operator are you using?** v1.6.3 helm chart from github repository
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** OVH managed Kubernetes

I setup the postgres-operator-ui using helm. In the `values.yaml` I have setup an nginx ingress with the following configuration:

```yaml
ingress:
  enabled: true
  annotations:
    kubernetes.io/ingress.class: nginx
    kubernetes.io/tls-acme: "true"
  hosts:
    - host: ui.yourname.xyz
      paths: [""]
  tls:
    - secretName: postgres-operator-ui-tls
      hosts:
        - ui.yourname.xyz
```

During the installation I get this deprecation warnings:

```
W0802 18:44:38.032843   17478 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0802 18:44:38.938036   17478 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0802 18:44:39.062773   17478 warnings.go:70] networking.k8s.io/v1beta1 Ingress is deprecated in v1.19+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

I expect the chart installing without the deprecation warnings.

I think the fix is relative easy. Just updating the apiVersion in the [ingress template ](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator-ui/templates/ingress.yaml) and double-check if there are any incompatibilities with the new version.  The old version will not longer supported with K8S release 1.22.

More Information: https://kubernetes.io/docs/reference/using-api/deprecation-guide/

markus-codes:
fixes #1574 

I have added an additional condition which applies the newer `networking.k8s.io/v1` ingress configuration if the cluster version is greater than `1.19.0`.

Maybe it should be considered to check which configuration is used based on `Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress"` instead of comparing the cluster version. I used the second approach, because it was already implemented like this.

### DIFF
--- a/charts/postgres-operator-ui/templates/ingress.yaml
+++ b/charts/postgres-operator-ui/templates/ingress.yaml
@@ -1,7 +1,10 @@
 {{- if .Values.ingress.enabled -}}
 {{- $fullName := include "postgres-operator-ui.fullname" . -}}
 {{- $svcPort := .Values.service.port -}}
-{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+
+{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1
+{{- else if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: networking.k8s.io/v1beta1
 {{- else -}}
 apiVersion: extensions/v1beta1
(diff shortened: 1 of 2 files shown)


FxKu:
Thanks. Can you also change it in the example manifest for UI [here](https://github.com/zalando/postgres-operator/blob/master/ui/manifests/ingress.yaml). Just use the newest api version. No need for backwards compatibility here.

markus-codes:
done 😊

FxKu:
:+1:

Jan-M:
:+1:

FxKu:
Thanks @markus-codes for your contribution.

PR #1577, opened 2021-08-04
Pass AWS region to Spilo

### LINKED ISSUE #1550: aws_region not passed to Spilo

- **Which image of the operator are you using?** e.g. registry.opensource.zalan.do/acid/postgres-operator:v1.6.3
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AWS K8s | Azure
- **Type of issue?** [Bug report, question, feature request, etc.] Bug report

The aws_region parameter doesn't appear to actually get passed to spilo, and spilo instead appears to poll the AWS API to check the region for a node.

Other postgres-operator settings are passed to Spilo via an environment variable like so:

https://github.com/zalando/postgres-operator/blob/ebb3204cdd7002742499c85b1df15b43a68f005b/pkg/cluster/k8sres.go#L789-L793

But there is no equivalent logic for AWS_REGION.

Would it be okay to just add the logic for passing this to Spilo?

apeschel:
Pass `aws_region` operator setting to Spilo via `AWS_REGION` environment variable.

fixes #1550

### DIFF
--- a/pkg/cluster/k8sres_test.go
+++ b/pkg/cluster/k8sres_test.go
@@ -498,7 +498,7 @@ func TestCloneEnv(t *testing.T) {
 			subTest: "generated s3 path, AWS region",
 			cloneOpts: &acidv1.CloneDescription{
 				ClusterName:  "test-cluster",
-				S3WalPath:    "s3://some/path/",
+				S3WalPath:    "",
 				EndTimestamp: "somewhen",
 			},
 			env: v1.EnvVar{
--- a/pkg/cluster/k8sres_test.go
+++ b/pkg/cluster/k8sres_test.go
@@ -492,13 +492,26 @@ func TestCloneEnv(t *testing.T) {
 				Name:  "CLONE_TARGET_TIME",
 				Value: "somewhen",
 			},
-			envPos: 4,
+			envPos: 5,
+		},
+		{
+			subTest: "generated s3 path, AWS region",
+			cloneOpts: &acidv1.CloneDescription{
+				ClusterName:  "test-cluster",
+				EndTimestamp: "somewhen",
+			},
+			env: v1.EnvVar{
+				Name:  "CLONE_AWS_REGION",
+				Value: "test-region",
+			},
+			envPos: 2,
 		},
 	}
 
 	var cluster = New(
 		Config{
 			OpConfig: config.Config{
+				AWSRegion:      "test-region",
 				WALES3Bucket:   "wale-bucket",
 				ProtectedRoles: []string{"admin"},
 				Auth: config.Auth{
--- a/pkg/cluster/k8sres_test.go
+++ b/pkg/cluster/k8sres_test.go
@@ -498,6 +498,7 @@ func TestCloneEnv(t *testing.T) {
 			subTest: "generated s3 path, AWS region",
 			cloneOpts: &acidv1.CloneDescription{
 				ClusterName:  "test-cluster",
+				S3WalPath:    "s3://some/path/",
 				EndTimestamp: "somewhen",
 			},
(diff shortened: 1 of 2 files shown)


apeschel:
I tried adding a test here, but I'm not 100% sure I did it correctly.

FxKu:
@apeschel there is no field `AWSRegion` in `CloneDescription` struct. You have to define it in the cluster config of that test:

```
	var cluster = New(
		Config{
			OpConfig: config.Config{
				AWSRegion:      "test-region",
				WALES3Bucket:   "wale-bucket",
				...
```

Then you have to add the env var `CLONE_AWS_REGION` to all tests.

apeschel:
@FxKu I'll do that, thanks for the guidance!

FxKu:
any update @apeschel ?

apeschel:
@FxKu I made the suggested change, but I'm still seeing the same error. I'm trying to figure out if I can run the tests locally.

apeschel:
@FxKu I managed to get the unit test working through trial and error. Is there a good way to tell what the `envPos` is supposed to be?

Unfortunately, the e2e tests are failing now, and I'm not sure why. I tried getting them running locally, but haven't had any success.

pkg/cluster/k8sres_test.go
```diff
@@ -498,13 +498,26 @@ func TestCloneEnv(t *testing.T) {
...
 				Value: "somewhen",
 			},
-			envPos: 4,
+			envPos: 5,
+		},
+		{
+			subTest: "generated s3 path, AWS region",
+			cloneOpts: &acidv1.CloneDescription{
+				ClusterName:  "test-cluster",
+				EndTimestamp: "somewhen",
```
FxKu, line 507:
```suggestion
				ClusterName:  "test-cluster",
				S3WalPath:    "s3://some/path/",
				EndTimestamp: "somewhen",
```

FxKu:
Thanks. With this PR you're introducing one more environment variable at the second position in the list of variables, hence expected `envPos` is 2 and +1 for all subsequent ones.

A diff in env variables will also trigger a rolling update on all your clusters.

pkg/cluster/k8sres_test.go
```diff
@@ -498,13 +498,27 @@ func TestCloneEnv(t *testing.T) {
...
 				Value: "somewhen",
 			},
-			envPos: 4,
+			envPos: 5,
+		},
+		{
+			subTest: "generated s3 path, AWS region",
+			cloneOpts: &acidv1.CloneDescription{
+				ClusterName:  "test-cluster",
+				S3WalPath:    "s3://some/path/",
```
FxKu, line 507:
```suggestion
				S3WalPath:    "",
```

apeschel:
@FxKu Looks like the e2e tests are still failing, any insight into what's going on?

apeschel:
@FxKu After being distracted with other work for a while, I can pick this back up. I still need to figure out what is going on with the tests. If there is anyone who can help out, it would be appreciated!

FxKu:
@apeschel after some more internal discussion we came to the conclusion that we do not want this. Atm, the `aws_region` config option set's the region for EBS volumes. It might not necessarily be the same region then your S3. Therefore, if we want the operator to pass AWS_REGION to the pods there should rather be a new option like `wal_s3_region` or something (and rename the other option to `ebs_aws_region` :smile: ). But it will not stop there, e.g. look at all the options we have for [logical backups](https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md#logical-backup). I think, as of now, people should continue to use the `pod_environment_configmap` approach.

I will add a not to the docs before doing the next release. Thanks again for your contribution and bringing this topic up.

PR #994, opened 2020-05-27
WIP: Allow NodePort type for master service

k0da:
This commit introduces new property to a postgresql spec
serviceNodePort. It enable NodePort type service for master role with
automatic port allocation if value is 0. Othewise it takes desired port
number like 30025.

closes #751

Signed-off-by: Dinar Valeev <k0da@opensuse.org>

PR #1584, opened 2021-08-16
WIP: Allow NodePort type for master service

ockl:
Try to fix abandoned PR #994


FxKu:
When introducing a new field to the postgres CRD type, we have to update:
1. CRD code validation
2. Rerun `./hack/update-codegen.sh`.
3. Extend documentation (reference docs, maybe user docs, too)

That is also missing in the original PR (which I forgot to comment :disappointed: )

gricey432:
@ockl what happened to this one? Is it in a position for someone else to approach again?

PR #1589, opened 2021-08-22
Fix empty resources spec field failing schema validation

jgillich:
In Go, when a struct field is not set, it becomes a struct with
default values for all fields. These default values are included
during serialization. This causes issues with schema validation
where optional fields cannot be omitted because default values
are considered invalid.

This patch addresses this issue for `Resources` fields on several
types by using a pointer value.

### DIFF
--- a/pkg/apis/acid.zalan.do/v1/postgresql_type.go
+++ b/pkg/apis/acid.zalan.do/v1/postgresql_type.go
@@ -27,7 +27,7 @@ type PostgresSpec struct {
 	PostgresqlParam `json:"postgresql"`
 	Volume          `json:"volume,omitempty"`
 	Patroni         `json:"patroni,omitempty"`
-	Resources       `json:"resources,omitempty"`
+	*Resources      `json:"resources,omitempty"`
 
 	EnableConnectionPooler        *bool             `json:"enableConnectionPooler,omitempty"`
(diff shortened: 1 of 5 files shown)


FxKu:
LGTM @jgillich 
Maybe we can use this PR to go one step further and remove the `required` flag from resources in the CRD schema validation.
[Here](https://github.com/zalando/postgres-operator/blob/f1055335432c31caaae5d44f6176e42db009c765/manifests/postgresql.crd.yaml#L436), [here](https://github.com/zalando/postgres-operator/blob/f1055335432c31caaae5d44f6176e42db009c765/charts/postgres-operator/crds/postgresqls.yaml#L403) and [here](https://github.com/zalando/postgres-operator/blob/f1055335432c31caaae5d44f6176e42db009c765/pkg/apis/acid.zalan.do/v1/crds.go#L589). Plus, remove for connection pooler as well.

FxKu:
Speaking of the pooler. Can you do the same change on the pooler [struct](https://github.com/zalando/postgres-operator/blob/f0815fc5bd4c8e5c89dae3d16b83a0ff03d6302c/pkg/apis/acid.zalan.do/v1/postgresql_type.go#L226) too @jgillich?

jopadi:
👍

FxKu:
:+1:

PR #1602, opened 2021-08-31
Fix malformed YAML in the manifest

michaelmcmillan:
The ports section should be a list. Without this fix  you'll trigger the following error:

```
Warning  Create  2m38s  postgres-operator  json: cannot unmarshal object into Go struct field Sidecar.spec.sidecars.ports of type []v1.ContainerPort
```


FxKu:
:+1:

Jan-M:
:+1:

PR #1616, opened 2021-09-14
fix operator 1.7.0 helm package

FxKu:
updating the 1.7.0 helm chart with the latest fix from #1603


FxKu:
:+1:

RafiaSabih:
:+1:

stanislav-zaprudskiy:
Thank you for the update! This would also be super-useful if you just increased the patch version instead of overwriting the existing version.

david-l-riley:
^ this; not updating the patch version can cause a lot of caching problems with Helm. Once a version gets released, it shouldn't change.

Issue #1358, opened 2021-02-10
cloning existing cluster - Failed to bootstrap cluster - Clone failed


arbe:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.6.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AWS K8s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report I think

I have some WAL backups in S3, all working fine, I can see base backups twice a day created, and all the related wal files.

I tried to setup a cluster cloning an existing one from the backup, by declaring a new Postgres CRD:

```
apiVersion: "acid.zalan.do/v1"
kind: postgresql
metadata:
  name: mydb-postgresql-clone
spec:
  clone:
    uid: "633b6c86-a4d8-460d-8a54-fbf4128b74c8"
    cluster: "mydb-postgresql"
    timestamp: "2021-02-08T13:05:00+01:00"

  teamId: "mydb"
  volume:
    size: 2Gi
  numberOfInstances: 2
  postgresql:
    version: "13"
```

Once I create this in the cluster, and the postgres pods are up, I can see from the logs it can't find the backups

```
2021-02-10 13:19:34,231 INFO: Running custom bootstrap script: envdir "/run/etc/wal-e.d/env-clone-simplrapi-postgresql" python3 /scripts/clone_with_wale.py --recovery-target-time="2021-02-08T13:05:00+01:00"
2021-02-10 13:19:34,407 INFO: Trying s3://some-s3-name/spilo/mydb-postgresql/633b6c86-a4d8-460d-8a54-fbf4128b74c8/wal/13/ for clone
INFO: 2021/02/10 13:19:34.544441 No backups found
2021-02-10 13:19:34,547 INFO: Trying s3://some-s3-name/spilo/mydb-postgresql/633b6c86-a4d8-460d-8a54-fbf4128b74c8/wal/12/ for clone
INFO: 2021/02/10 13:19:34.678884 No backups found
2021-02-10 13:19:34,683 INFO: Trying s3://some-s3-name/spilo/mydb-postgresql/633b6c86-a4d8-460d-8a54-fbf4128b74c8/wal/11/ for clone
INFO: 2021/02/10 13:19:34.861163 No backups found
[...]

Issue #1132, opened 2020-09-06
How to configure WAL S3 archiving? 


yaroslavkasatikov:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** 
          image: registry.opensource.zalan.do/acid/postgres-operator:v1.5.0

- **Where do you run it - cloud or metal? Kubernetes or OpenShift?**
  AWS. Openshift 4.5
- **Are you running Postgres Operator in production?** 
no
- **Type of issue?** 
question:

Hello team, I can't understand how to enable WAL S3 archiving and didn't find straight forward documentation for it.

Here is my steps:

I added into  operatorconfigurations.acid.zalan.do this lines:

```
configuration:
  aws_or_gcp:
    aws_endpoint: s3.eu-west-1.amazonaws.com
    aws_region: eu-west-1
    wal_s3_bucket: zalando-bucketname
```

here is full output for my operatorconfgurations:

```
apiVersion: acid.zalan.do/v1                                                                                                                                                       [186/1894]
configuration:
  aws_or_gcp:
    aws_endpoint: s3.eu-west-1.amazonaws.com
    aws_region: eu-west-1
    wal_s3_bucket: zalando-bucketname
  connection_pooler:
    connection_pooler_default_cpu_limit: "1"
    connection_pooler_default_cpu_request: 500m
    connection_pooler_default_memory_limit: 100Mi
    connection_pooler_default_memory_request: 100Mi
    connection_pooler_image: registry.opensource.zalan.do/acid/pgbouncer:master-7
    connection_pooler_mode: transaction
    connection_pooler_number_of_instances: 2
  debug:
    debug_logging: true
    enable_database_access: true
  docker_image: registry.opensource.zalan.do/acid/spilo-12:1.6-p3
  etcd_host: ""
  kubernetes:
    cluster_domain: cluster.local
    cluster_labels:
      application: spilo
    cluster_name_label: cluster-name
    enable_init_containers: true
    enable_pod_antiaffinity: false
    enable_pod_disruption_budget: true
    enable_sidecars: true
    master_pod_move_timeout: 20m
    oauth_token_secret_name: postgresql-operator
    pdb_name_format: postgres-{cluster}-pdb
    pod_antiaffinity_topology_key: kubernetes.io/hostname
    pod_environment_configmap: **zalando-operator/env**
    pod_management_policy: ordered_ready
    pod_role_label: spilo-role
    pod_service_account_name: postgres-pod
[...]

PR #1621, opened 2021-09-16
Update username.md

foxmeyson:
I had the opportunity to analyze the case with cluster recovery from the S3 bucket, and until now I have received many questions about this case. Users do not consider the documentation to be explicit enough for such a case and face difficulties. Therefore, I propose to add this description to the documentation.

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -766,6 +766,403 @@ spec:
     s3_force_path_style: true
 ```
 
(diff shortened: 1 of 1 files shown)


FxKu:
Hey @foxmeyson. Thanks for sharing you insights here and sorry I did not respond sooner. Your additions reads more like a blog post to me rather than material for the documentation. I will see if I can take some hints from it for the admin or user page as some folks still having trouble with the setup #1132 #1358 #1690.

PR #1623, opened 2021-09-20
Fix logical backup script sigpipe problem

amirmd76:
Sometimes the `head -n` pipe in the logical backup script (when getting the cluster host ip), causes SIGPIPE which will cause the script to exit with code 141, since `head` will only read the first line and cause a SIGPIPE signal to `jq`. I added an extra pipe to `tee` to resolve this issue.


sdudoladov:
:+1:

FxKu:
:+1:

PR #1631, opened 2021-10-04
Update user.md

elmehdisaniss:
The name of the created secret with the minimal postgres deployment is "postgres.acid-minimal-cluster.credentials.postgresql.acid.zalan.do" so when trying the command export "PGPASSWORD=$(kubectl get secret postgres.acid-minimal-cluster.credentials -o 'jsonpath={.data.password}' | base64 -d)"   base64 complains about invalid input and when running the command "kubectl get secret postgres.acid-minimal-cluster.credentials -o 'jsonpath={.data.password}'" it complains about not finding the resource. it worked when I used the full name, i.e. "postgres.acid-minimal-cluster.credentials.postgresql.acid.zalan.do".

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -88,7 +88,7 @@ which was generated when creating the `acid-minimal-cluster`. As non-encrypted
 connections are rejected by default set the SSL mode to `require`:
 
 ```bash
-export PGPASSWORD=$(kubectl get secret postgres.acid-minimal-cluster.credentials -o 'jsonpath={.data.password}' | base64 -d)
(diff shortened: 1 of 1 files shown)


FxKu:
Thanks @elmehdisaniss . Can you update the example in the [configmap manifest](https://github.com/zalando/postgres-operator/blob/d3183604a78120126345fcafa9621a2b8b0f24ff/manifests/configmap.yaml#L117), too?

Should be: "{username}.{cluster}.credentials.{tprkind}.{tprgroup}"

PR #1635, opened 2021-10-06
improve Patroni config sync

FxKu:
the current way of syncing and comparing Patroni config has a few issues:

1. returned Patroni config can be [empty but not nil](https://github.com/zalando/postgres-operator/blob/d3183604a78120126345fcafa9621a2b8b0f24ff/pkg/cluster/sync.go#L502), e.g. during restore which can crash the operator
2. by default `synchronous_mode` and `synchronous_mode_strict` are not set in Spilo and create a [diff](https://github.com/zalando/postgres-operator/blob/d3183604a78120126345fcafa9621a2b8b0f24ff/pkg/cluster/sync.go#L537) with desired state even when not set, because the default is `false`
3. [slot comparison](https://github.com/zalando/postgres-operator/blob/d3183604a78120126345fcafa9621a2b8b0f24ff/pkg/cluster/sync.go#L534) does not work even when they are semantically the same because go types are different (`map[string]map[string]string` vs. `map[string]interface{}`). Moreover, the slot settings should not be compared as a whole. It should only be checked if specified slots exist and if they differ.


FxKu:
:+1:

PR #1643, opened 2021-10-12
fix TestNodeAffinity in e2e tests

FxKu:
updating the NodeAffinity in the PostgresSpec again should retrigger another rolling update leading to two pods on two different workers - so pod distribution might be skipped


Jan-M:
:+1:

PR #1656, opened 2021-10-22
Add POD_NAME environment variable for the Spilo Pod

bleleve:
Hello everyone,

As part of our cluster deployment we need the POD_NAME environment variable which is initialized on the sidecar but not on the Spilo pod. This envvar may be useful to others.

Best regards.


bleleve:
Closed due to renaming of the source branch.

PR #1666, opened 2021-11-02
Add POD_NAME environment variable for the Spilo Pod

bleleve:
Hello everyone,

As part of our cluster deployment we need the POD_NAME environment variable which is initialized on the sidecar but not on the Spilo pod. This envvar may be useful to others.

Best regards.


FxKu:
Have nothing against the change so thanks @bleleve for contributing. But it would trigger a rolling update of all your clusters. This is similar to e.g. the change proposed in #1577. I think, I would incl. both in the next 1.8.0 release with a proper statement in the release notes.

PR #1667, opened 2021-11-02
Allow distinction between local and bootstrap PostgreSQL parameters

bleleve:
Hello everyone,

Currently the management of PostgreSQL parameters is quite rigid. The distinction between "bootstrap" and "local" parameters is hard coded https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L413. This PR allow to define the type of parameter in the PostgreSQL manifest.

```yaml
kind: postgresql
metadata:
  name: acid-test-cluster
spec:
  ...
  postgresql:
  version: "14"
  parameters:  # Expert section
    shared_buffers:
      type: local
      value: "32MB"
    max_connections:
      type: bootstrap
      value: "10"
    log_statement:
      type: bootstrap
      value: "all"
  ...
```

Best regards.


FxKu:
Hi @bleleve and thanks for yet another contribution.

I would vote against this change. Being rigid is actually one of our operator principles. We do not aim to provide endless configuration options because in the end users tend to rather shoot themselves in the foot with it. We have seen this at Zalando and sometimes doubt that it was a good idea to make Postgres configuration parameters configurable from the manifest.

Another point is that the current structure resembles the Postgres configuration within the Pod returned by Patroni's API (e.g. when you call `patronictl show-config`). This makes it easy to compare what's specified under `postgresql.parameters` - also code wise. Your change would require rewriting several code passages, plus we have to guarantee backwards compatibility. Ultimately, I don't really see what we gain from this change. Makes lives for users and developers unnecessarily more complex.

bleleve:
Hi @FxKu and thanks.

Although I understand the desire to simplify the task on the user side, it is really too limiting. The way the function isBootstrapOnlyParameter() is coded implies that all parameters other than those set in the list are global and synchronized across the cluster via the Patroni DCS.

On our side we use custom parameters to pass variables to postgres, some of them are indeed global but others are linked to the Spilo pod (for example POD_NAME, see my other PR).

Moreover, the `postgres.parameters` section in the complete manifest is indicated as being reserved for experts.

Having an operator too restrictive and standard makes it lose interest compared to a managed db.

My 2 coppers ;)

CyberDem0n:
The concept of global dynamic and local configurations exists in Patroni because it could run anywhere and nodes could vary by their specs and usage patterns.
In this regard on K8s all pods are supposed to have the same configuration, hence most of the parameters should be in the dynamic configuration, because it allows changing them with minimal efforts (whatever parameter ends up in patroni.yaml requires a rolling restart of all pods!).

bleleve:
> 
> 
>[...] on K8s all pods are supposed to have the same configuration [...]

This is true, but there are special cases such as for example the insertion in the id of the pod index in which the data was generated (useful in the case of sharding).

If I understood correctly the distribution between the different projects, the parameters are defined by Spilo in an optimal way. This PR affects a section already reserved for "expert" use. It seems relevant to me, but I understand the need not to address out of the scope cases. 

Best regards.

CyberDem0n:
> (useful in the case of sharding)

Sharding is just a set of _independent_ clusters, that have own configurations

bleleve:
> Sharding is just a set of _independent_ clusters, that have own configurations

It's an architectural choice

CyberDem0n:
From the perspective K8s and postgres-operator all pods that belong to the same cluster (StatefulSet) are equal.
This is also an architectural choice that should be respected.

PR #1669, opened 2021-11-03
Add support for logical backups to Azure

stgrace:
This PR adds support for logical backups to be pushed to Azure

In the backup script used to create a logical backup and push to cloud, I had to temporarily write the logical backup to a file in tmp, because Azure's cli cannot by upload data from stdin to a file in a storage account.
There is an extension that can do so, but this extension only works with strings, and fails when you pipe the result of the compress to it. (https://github.com/Azure/azure-cli-extensions/blob/main/src/storage-blob-preview/azext_storage_blob_preview/_help.py#L172) 

I also added an option to add a secret as env variables, like for the postgres pod. It makes the cronjobs easier to use in my opinion, and it was also not possible to mount the config file for AZ in the cronjob pod, because azure tries to write to the config folder, and k8s mounts it into a readonly directory.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -173,6 +173,9 @@ spec:
                   enable_init_containers:
                     type: boolean
                     default: true
+                  enable_cross_namespace_secret:
+                    type: boolean
+                    default: false
                   enable_pod_antiaffinity:
                     type: boolean
                     default: false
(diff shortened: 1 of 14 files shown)


stgrace:
Any thoughts on this?

FxKu:
Thanks @stgrace . Lots of new environment variables. `LOGICAL_BACKUP_STORAGE_BLOB` looks unused. Do we need CONTAINER, when we have the storage account? For backup/restore we use only storage account, too.

Could you move the secret part in it's own PR since it does not only affect Azure backups? Makes it easier in the end to track changes when new features were introduced :smiley:

PR #1672, opened 2021-11-04
Add valueFrom capability for PostgreSQL parameters

bleleve:
Hi,

This PR allow the use of valueFrom into `spec.postgresql.parameters`, e.g.:
```yaml
apiVersion: acid.zalan.do/v1
kind: postgresql
metadata:
  name: my-cluster
spec:
  ...
  postgresql:
    version: '13'
    parameters:
      'custom.my_setting_1':
        value: 'TEST'
      'custom.my_setting_2':
        valueFrom:
          configMapKeyRef:
            name: custom-pg-config
            key: MY_SETTING_2	
      'custom.my_setting_3':
        valueFrom:
          secretKeyRef:
            name: custom-pg-secrets
            key: MY_SETTING_3	  
  ...
```

It is useful in the context of a CI/CD, and also in combination with the PR https://github.com/zalando/postgres-operator/pull/1667.

Best regards.


FxKu:
@bleleve like I said in #1667 at the moment there's no intention to change the type of the parameters field from `map[string]string` to `map[string]Object`. Maybe we can think of another way to pass custom config options to Postgres config...

bleleve:
Hi @FxKu,

There is no problem, I have adapted the operator to my needs, it is up to you to take what seems relevant to you.

Take care of yourself.

FxKu:
I believe, #1794 goes in this direction as well, so I'm closing this PR.

PR #1676, opened 2021-11-08
Provide a bit more flexibility for where images live

bkrugerp99:
Proposed helm chart changes would allow to just update image tags and not have to touch the rest of the values.

This just mirrors how the operator image is derived.  Very useful if you have code to mirror images locally to your own docker repos.  (If you use something like [docker-mirror](https://github.com/seatgeek/docker-mirror))

### DIFF
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -1,8 +1,27 @@
+# Operator image
 image:
   registry: registry.opensource.zalan.do
   repository: acid/postgres-operator
   tag: v1.7.1
   pullPolicy: "IfNotPresent"
+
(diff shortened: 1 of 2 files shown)


charts/postgres-operator/values.yaml
```diff
@@ -1,9 +1,28 @@
...
+logical_image:
+  registry: registry.opensource.zalan.do
+  repository: acid/logical-backup
+  tag: v1.7.1
+
+# Spilo docker image
+spilo_image:
+  registry: registry.opensource.zalan.do
+  repository: acid/spilo-
+  tag: 14:2.1-p3
```
FxKu, line 18:
```suggestion
  repository: acid/spilo-14
  tag: 2.1-p3
```

PR #1688, opened 2021-11-14
Adds required Azure backup env variable WALG_AZ_PREFIX to documentation

bloever:
Current documentation in the postgres-operator fails to mention that WALG_AZ_PREFIX enables Azure Backups. This is a simple fix for the documentation so others won't be stranded.

Spilo image env variables: (https://github.com/zalando/spilo/blob/master/ENVIRONMENT.rst)

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -885,6 +885,7 @@ data:
   USE_WALG_BACKUP: "true"
   USE_WALG_RESTORE: "true"
   CLONE_USE_WALG_RESTORE: "true"
+  WALG_AZ_PREFIX: "azure://container-name/$(SCOPE)/$(PGVERSION)" # Enables Azure Backups (SCOPE = Cluster name) (PGVERSION = Postgres version) 
 ```
 
 3. Setup your operator configuration values. With the `psql-backup-creds`


FxKu:
`WALG_AZ_PREFIX` is generated by spilo if `wal_az_storage_account` is configured. I think, that's why @apeschel left this variable out in #1537. You can overwrite it, but not sure if everything still works then.

bloever:
My experience was that there was nothing in `/run/etc/wal-e.d/env` before i manually configured `WALG_AZ_PREFIX`. `wal_az_storage_account` was set from the start. 

I guess this is the relevant part in the Spilo image?
(https://github.com/zalando/spilo/blob/b2b1c6988c81ed4705c08b1902712251b86a4f75/postgres-appliance/scripts/configure_spilo.py#L819)

FxKu:
Ok, just learned that I mixed this variable up with the scope prefix I guess. So it might indeed be needed.
:+1:

FxKu:
:+1:

Jan-M:
:+1:

FxKu:
Thanks again @bloever

PR #1692, opened 2021-11-17
Add log messages for usernames

### LINKED ISSUE #1684: Question about creating secrets with credentials in other namespaces
Hi!

This is sort of a question, and maybe a bug. I'm not quite sure. I'm trying the operator in Minikube (in an Intel Macbook Pro, using Hyperkit as hypervisor). The operator's version is 1.7.1

My question is about defining users in a different namespace than the one in which the cluster is created. When you define the user, you must add it in the manifest as the following:

```yaml
spec:
  users:
  #users with secret in dfferent namespace
   appspace.db_user:
    - createdb
```

What I'm seeing is that the created username is `appspace.db_user` instead of `db_user`. Is this the expected behavior? If yes, let me tell you that maybe you should include the following information when using this. When defining databases, the owner of it should also include the namespace, if not the operator will log that the user does not exists.

Thank you very much!

RafiaSabih:
If enable_cross_namespaced_secret is set then emit log messages that the created username is with the namespace.

fixes #1684

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -141,14 +141,18 @@ other roles.
 
 To define the secrets for the users in a different namespace than that of the
 cluster, one can set `enable_cross_namespace_secret` and declare the namespace
-for the secrets in the manifest in the following manner,
+for the secrets in the manifest in the following manner (note, that it has to
+be reflected in the `database` section, too),
 
 ```yaml
 spec:
   users:
-  #users with secret in dfferent namespace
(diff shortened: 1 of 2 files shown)


FxKu:
:+1:

FxKu:
:+1:

Jan-M:
Reminds me we should look at levels of messages.

Jan-M:
:+1:

PR #1715, opened 2021-12-02
make sure upgrade script runs on the master

FxKu:
fixes looping over pods to get the master before performing the major version upgrade. If replicas come later `masterPod` can become a replica because of `= &pod`


FxKu:
:+1:

Jan-M:
:+1:

PR #1717, opened 2021-12-03
move to go 1.17.4 and update dependencies

FxKu:
- bump go `1.16.9` -> `1.17.4`
- bump k8s dependencies `1.22.3` -> `1.22.4`
- using `go get -d` (use of just `go get` is deprecated within go modules)
- switch to `go install` for mockgen and kind
- re-init go.mod and go.sum
- update kubectl-pg plugin
- adding `//go:build !ignore_autogenerated` to zz_generated.deepcopy.go file


FxKu:
It seems mockgen does not support 1.17 yet. So we have to wait :disappointed:

dmvolod:
> It seems mockgen does not support 1.17 yet.

It seems to `go install github.com/golang/mock/mockgen@v1.6.0` need to be utilized for golang 1.16+ instead of `go get`

FxKu:
:+1:

jocielepadilha:
👍

Jan-M:
:+1:

PR #1724, opened 2021-12-15
handle case when Patroni returns that lag is unknown

FxKu:
In GitHub e2e tests we see the following error:
```
cannot unmarshal string into Go struct field ClusterMember.members.lag of type int
```

Turns out Patroni can return the string value `"unknown"` for replica lag - e.g. in case there's no leader. The operator must deal with this by using the `IntOrString` type by `intstr` lib.

Will take the chance to rename the field simply to `Lag`.


manifests/postgresql.crd.yaml
```diff
@@ -462,11 +462,11 @@ spec:
                 type: integer
               standby:
                 type: object
-                required:
-                  - s3_wal_path
                 properties:
                   s3_wal_path:
                     type: string
+                  gs_wal_path:
+                    type: string
```
FxKu, line 469: leftover from #1446

pkg/util/util.go
```diff
@@ -322,6 +323,20 @@ func testNil(values ...*int32) bool {
...
+// Convert int to IntOrString type
+func ToIntStr(val int) *intstr.IntOrString {
+	b := intstr.FromInt(val)
+	return &b
+}
+
+// Get int from IntOrString and return max int if string
+func IntFromIntStr(intOrStr intstr.IntOrString) int {
+	if intOrStr.Type == 1 {
+		return 32 << (^int(0) >> 32 & 1) // return max int
```
Jan-M, line 335: Is there no constant for max int in go?

FxKu: indeed: `math.MaxInt`

FxKu:
:+1:

Jan-M:
:+1:

PR #1729, opened 2021-12-23
toleration diff and nodeReadinessLabel merge with manifest matchExpressions

FxKu:
When adding tolerations to the manifest it will simply be ignored since the comparison between running and desired stateful set does not yield a mismatch. This PR will introduce a rolling update trigger for toleration changes just like with `nodeAffinity`. Only slice length is compared for now.

With nodeAffinity we spotted another problem: If somebody specified `nodeAffinity` in the manifest and has configured `node_readiness_label` the resulting affinity on the pods will have two `matchExpressions` section in the `nodeSelectorTerms`.

```yaml
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: manifestlabelkey
            ...
        - matchExpressions:
          - key: myreadinesslabelkey
            ...
```

Multiple `matchExpressions` are evaluated with OR condition. Thus, pods could be scheduled on nodes having either one of the specified expressions. This might be fine for some users. However, we want the readiness label to be evaluated always together with other specified affinities. Therefore, we will introduce another option to choose if the config affinity should get merged under one `matchExpressions` (AND) or stay in it's own (OR).

Finally, this PR also tries to re-enable the taint based eviction and node readiness label tests which use the features touched here.


docs/reference/operator_parameters.md
```diff
@@ -343,8 +343,13 @@ configuration they are grouped under the `kubernetes` key.
   considered `ready`. The operator uses values of those labels to detect the
   start of the Kubernetes cluster upgrade procedure and move master pods off
   the nodes to be decommissioned. When the set is not empty, the operator also
-  assigns the `Affinity` clause to the Postgres pods to be scheduled only on
-  `ready` nodes. The default is empty.
+  assigns the `nodeAffinity` clause to the Postgres pods to be scheduled only
+  on `ready` nodes. If a `nodeAffinity` is specified in the postgres cluster
+  manifest as well the `nodeSelectorTerms` will get merged. If the
```
sdudoladov, line 348: `nodeSelectorTerms` will get merged with what ? `nodeAffinity` ?

docs/reference/operator_parameters.md
```diff
@@ -343,8 +343,13 @@ configuration they are grouped under the `kubernetes` key.
   considered `ready`. The operator uses values of those labels to detect the
   start of the Kubernetes cluster upgrade procedure and move master pods off
   the nodes to be decommissioned. When the set is not empty, the operator also
-  assigns the `Affinity` clause to the Postgres pods to be scheduled only on
-  `ready` nodes. The default is empty.
+  assigns the `nodeAffinity` clause to the Postgres pods to be scheduled only
+  on `ready` nodes. If a `nodeAffinity` is specified in the postgres cluster
+  manifest as well the `nodeSelectorTerms` will get merged. If the 
+  `nodeAffinity` of the manifest contains only one `matchExpressions` slice
+  the node readiniess label expressions will be moved there (AND condition).
```
sdudoladov, line 350: 1. typo in `readiniess`
2. `be moved there` reads a bit misleading. I assume the original manifest is not modified, right ? But internally in the operator the `matchExpressions` slice will contain the node readiness label

docs/user.md
```diff
@@ -703,6 +703,10 @@ spec:
           - pci
 ```
 
+If you need to define a `nodeAffinity` for all your Postgres clusters use the
+`node_readiness_label` configuration option, which allows you to define a list
```
sdudoladov, line 707: is the "operator configuration option" meant here  ?

pkg/cluster/cluster.go
```diff
@@ -427,13 +432,11 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 	}
 
 	if !reflect.DeepEqual(c.Statefulset.Spec.Template.Annotations, statefulSet.Spec.Template.Annotations) {
-		match = false
```
sdudoladov, line 430: why is `match`no longer necessary in all these conditions ?

pkg/cluster/k8sres.go
```diff
@@ -352,8 +352,22 @@ func nodeAffinity(nodeReadinessLabel map[string]string, nodeAffinity *v1.NodeAff
...
-			nodeAffinityCopy.RequiredDuringSchedulingIgnoredDuringExecution = &v1.NodeSelector{
-				NodeSelectorTerms: append(nodeAffinityCopy.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms, nodeReadinessSelectorTerm),
+			// if there are multiple node selector terms specified, append the node readiness label expressions (OR condition)
+			if len(nodeAffinityCopy.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms) > 1 {
+				manifestTerms := nodeAffinityCopy.RequiredDuringSchedulingIgnoredDuringExecution.NodeSelectorTerms
+				manifestTerms = append(manifestTerms, nodeReadinessSelectorTerm)
+				nodeAffinityCopy.RequiredDuringSchedulingIgnoredDuringExecution = &v1.NodeSelector{
+					NodeSelectorTerms: manifestTerms,
+				}
+				// if there's just one term defined merge it with the readiness label term (AND condition)
```
sdudoladov, line 362: suboptimal comment indent, it is not immediately clear it is a comment on the else branch

sdudoladov (commented):
could you add 2 examples (for `OR and `AND`) to the admin docs showcasing the different  behavior ?

pkg/cluster/cluster.go
FxKu: wherever `needsRollUpdate` or `needsReplace` are set to `true` we do not need to do the same for `match` because in the end we do:
```
	if needsRollUpdate || needsReplace {
		match = false
	}
```

FxKu:
:+1:

sdudoladov:
:+1:

Issue #1393, opened 2021-03-03
Configurable CRD creation by operator


ant31:
- **Which image of the operator are you using?** e.g. 
registry.opensource.zalan.do/acid/postgres-operator:v1.6.1

- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** [AWS K8s | GCP ... | Bare Metal K8s]
Baremetal k8s
- **Are you running Postgres Operator in production?** [yes | no] 
yes
- **Type of issue?** 
feature request.

Currently it's not possible to start the operator without a ClusterRoleBinding because it register itself the `operatorconfigurations.acid.zalan.do` which requires cluster wide permission. 
It also makes it more complex to run multiple operator in different namespaces. 

The CRD has to be registered only  once, when deploying an additional operator instance in a namespace there CRD definition should already exist. 

Currently both the usage AND the creation of the `operatorconfigurations.acid.zalan.do` are both parameterize with the same envvar. 

I propose to separate it in two vars or at least to add one: 
"REGISTER_OPERATORCONFIGURATION:  true"
So it can be turned off
 
Only need to add a condition here (from what I can tell):
https://github.com/zalando/postgres-operator/blob/41858a702ce678ee05351f58d942628f2b0f4426/pkg/controller/controller.go#L310

ant31:
I can make the PR if the change is ok

FxKu:
The operator creates two CRDs: `postgresqls.acid.zalan.do` and `operatorconfigurations.acid.zalan.do`. I think there was a proposal to make this configurable before. We then would need different RBAC depending on what the user chooses (no `create` privilege for CRDs and yes - role binding). For the name, I would vote for something like EnableCRDCreation (registration? installation?) to be aligned with the other flags we have.

Are you aware that you can already run two and more operators next to each other using a [CONTROLLER_ID](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#operators-with-defined-ownership-of-certain-postgres-clusters)? But, I think even then the RBAC is a valid point.

ant31:
Also it's a good practice to give least privileged roles when possible.

I prefer 'registration' 
Yes I am using controller_id but in multiple namespace.

ant31:
fixed in #1733

PR #1733, opened 2022-01-05
Add configuration flag for disabling operator CRD creation/update

dpeckett:
In many applications having the operator dynamically creating/updating CRDs isn't desirable eg. #1393.

This PR introduces a new configuration flag `enable_crd_registration` that can be used to disable the automatic CRD creation/update process. Also included is a modification to the operator chart to allow dropping CRD create/update RBAC permissions when CRD registration is disabled.


imusmanmalik:
+1

microwaves:
+1 very needed option to avoid conflicts with ArgoCD

FxKu:
Can you swap registration and validation flags to follow alphabetical sorting? The rest looks fine.

dpeckett:
Thanks, updated.

FxKu:
Ah two more missing pieces.
1. We use to copy the OperatorConfiguration into an internal config struct. You need to add this new option [here](https://github.com/zalando/postgres-operator/blob/b4155bc8fb468590677fb1e808708884d744d694/pkg/controller/operator_config.go#L35).

2. Can you update code generation (`make codegen`) and commit the changed pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go file?

FxKu:
:+1:

Jan-M:
:+1:

ant31:
awesome, thank you @dpeckett

Issue #1737, opened 2022-01-10
Multiple non-matching CRD definitions


apeschel:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** e.g. registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AWS K8s | Azure
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug Report

There are two locations in this repository where CRDs are defined:

1. https://github.com/zalando/postgres-operator/tree/master/charts/postgres-operator/crds
2. https://github.com/zalando/postgres-operator/tree/master/manifests

However, it seems that the CRDs in the second location have not been maintained. The `operatorconfiguration` CRD is missing the `enable_cross_namespace_secret` key.

https://github.com/zalando/postgres-operator/blob/master/manifests/operatorconfiguration.crd.yaml
https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/crds/operatorconfigurations.yaml

Perhaps a symlink could be used to ensure there is only a single copy of these files?

PR #1738, opened 2022-01-10
Unify CRD Definitions

apeschel:
There are duplicate CRD definitions in the Chart and Manifests
directories. This commit changes the manifests files into symlinks and
points them at the chart. It also matches the filename in both the
manifests and charts directories.


apeschel:
Fixes https://github.com/zalando/postgres-operator/issues/1737

dmvolod:
Seems to `manifests` folder should be single source of truth, and `/charts/postgres-operator/crds/` could be a link. Please refer **Updating manifest files** inside the https://github.com/zalando/postgres-operator/blob/master/docs/developer.md

PR #1749, opened 2022-01-20
Password rotation in secrets

FxKu:
For now, see new chapter in admin docs for details.

TL;DR: Operator can rotate usernames and passwords in K8s secrets. Either extra users are created or only password is replaced (e.g. for users doing one time operations on pod start).

New config options:
- enable_password_rotation, false
- password_rotation_interval: 90 (days)
- password_rotation_user_retention: 180 (days)

New manifest options:
- usersWithPasswordRotation
- usersWithInPlacePasswordRotation

ToDos:
- [x] reset secrets when feature is disabled
- [x] sane cleanup cycle of rotated users even when features is disabled


sdudoladov:
:+1:

PR #1753, opened 2022-01-22
Feature: Create customized Secrets via Postgreqsl.userSecret

### LINKED ISSUE #1746: Feature-Request: Templating Secrets

- **Where do you run it - cloud or metal? Kubernetes or OpenShift?**  Bare Metal K3s
- **Type of issue?**  feature request

# Templating Secrets

I'm currently facing the issue to get the Postgres-Operator work in combination with the AWX-Operator. The AWX-Operator requires a postgres-configuration-secret like this:
```yaml
stringData:
  host: <external ip or url resolvable by the cluster>
  port: <external port, this usually defaults to 5432>
  database: <desired database name>
  username: <username to connect as>
  password: <password to connect with>
  sslmode: prefer
  type: unmanaged
```

But the Secret generated by the postgres-operator only contains the username & password. So either I have to copy or modify the secret. Both things i would like avoid.

Since I guess that there are more cases of software requiring e.g. different secret keys, templating the secret just like `secret_name_template` would be awsome. This way I could easily define something like this:
```yaml
secret_template:
  host:  {{ service-name }}
  port: 5432
  database: {{ database }}
  username:  {{ username }}
  password: {{ password }}
  sslmode: prefer
  type: unmanaged
```

Nold360:
Fixes #1746
Note: I'm not a dev & barely coded any go, so be kind to me 

### What does this PR provides
I completely reworked my first attempt, because I found that there is something similar implemented in the [infrastructure roles](https://postgres-operator.readthedocs.io/en/latest/user/#infrastructure-roles-secret)

This PR adds the `Postgresql.userSecret` Spec. It allows to change the secret key names of the username and password. Furthermore it allows to add additional key-value pairs to the created secrets. 

Example:
``` yaml
kind: postgresql
metadata:
  name: acid-minimal-cluster
spec:
  [...]
  userSecret:
    passwordkey: pass
    userkey: user
    additionalkeys:
      database: foobar
      foo: bar
      port: "5432"
```

The resulting secrets would look like this:
```yaml
kind: Secret
data:
  user: <username>
  pass: <password>
  database: foobar
  foo: bar
  port: 5432
```

### Username & Password Keys
Many parts in the code rely on the the `username` & `password` key inside the user secrets. I hope i found & fixed them all.

### Also..
 - Values are string only
 - Tests are missing - i have no idea how to write them, plz help me
 - Documentation will be added
 
Most importantly: I don't know which side effects this might cause, that's why I'm opening this PR as RFC.


jahknem:
Will anyone merge this? It is exactly what I need and its somewhat sad to see that the solution has been coded but not merged since 1.5 years..

FxKu:
@jahknem @Nold360 sorry this was never reviewed. Porbably nobody liked the idea in the first place. Why adding more and more stuff to the cluster manifest? This is not our intention with this Postgres Operator. Our goal is to keep the manifest schema as small as possible and move all configuration options to the gloabal config.

And I think, this also applies here. It looks to me that some other utility is providing the secrets for you so you would probably need the user secret format for all your Postgres clusters. Adding config means for setting the password and user key sounds good. For additional data you should use the `pod_environment_secret`option.

Will close this PR now. Sorry again you had to wait so long for an answer. And feel free to continue the discussion if you disagree.

PR #1755, opened 2022-01-25
Compare container ports in a smarter way

### LINKED ISSUE #1384: Sidecar port default protocol are buggy
This is a duplicate post of https://github.com/zalando/postgres-operator/issues/1291  — which was never fixed. The core problem after investigation is that the `protocol` in the ports declaration of side-cars isn't compared in a way that counts `protocol: null` === `protocol: TCP` despite TCP being the default if protocol in the Go k8s struct is nil.

This is a good first issue:

1. Create a unit test that validates the bug (by failing) when comparing the protocol property of a ports array (of a side-car or otherwise)
2. Special-case the defaults: this could potentially include other defaults if they exist for this record

Or alternatively; semantic comparison with defaults is probably already implemented in the k8s go SDK: call this SDK's comparison function instead.

---
After getting advice, this is a cross post with https://github.com/zalando/spilo/issues/536 — it would seem the fault is with the operator and not Spilo.

- **Which image of the operator are you using?** `registry.opensource.zalan.do/acid/postgres-operator:v1.6.0`
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** GCP
- **Type of issue?** bug report

For me it's enough to define a sidecar like this, to trigger the behaviour:

```yaml

  sidecars:
  - name: exporter

    env:
    - name: "DATA_SOURCE_URI"
      value: app-analytics-db/analytics?sslmode=require

    - name: "DATA_SOURCE_USER"
      valueFrom:
        secretKeyRef:
          name: postgres.app-analytics-db.credentials
          key: username

    - name: "DATA_SOURCE_PASS"
      valueFrom:
        secretKeyRef:
          name: postgres.app-analytics-db.credentials
          key: password

    - name: PG_EXPORTER_WEB_LISTEN_ADDRESS
      value: ":9114"

    - name: PG_EXPORTER_CONSTANT_LABELS
      value: app=analytics-db,component=postgres

    ports:
    - name: http-prom
      containerPort: 9114

    resources:
      limits:
        cpu: 500m
        memory: 256M
      requests:
        cpu: 100m
        memory: 200M
```

Causes it to fail over every thirty minutes:

image

Some logs:

```
time="2021-01-06T17:24:36Z" level=debug msg="metadata.annotation are different" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:36Z" level=debug msg="-  zalando-postgres-operator-rolling-update-required: false" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:36Z" level=debug msg="+  zalando-postgres-operator-rolling-update-required: true" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:36Z" level=info msg="reason: new statefulset containers's exporter (index 1) ports do not match the current one" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:36Z" level=debug msg="updating statefulset" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:36Z" level=debug msg="patching statefulset annotations" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:37Z" level=debug msg="patching statefulset annotations" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:37Z" level=debug msg="calling Patroni API on a pod app/app-analytics-db-0 to set the following Postgres options: map[wal_level:logical]" cluster-name=app/app-analytics-db pkg=cluster worker=1
time="2021-01-06T17:24:37Z" level=debug msg="making PATCH http request: http://10.4.7.76:8008/config" cluster-name=app/app-analytics-db pkg=cluster worker=1
```

uzzz:
This PR makes the comparison of containers' ports a bit smarter: it takes into account default protocol (TCP) and ignores different ports order.

Fixes #1384.

### DIFF
--- a/pkg/cluster/cluster_test.go
+++ b/pkg/cluster/cluster_test.go
@@ -2,9 +2,10 @@ package cluster
 
 import (
 	"fmt"
-	"github.com/stretchr/testify/assert"
 	"reflect"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
 
 	"github.com/sirupsen/logrus"
 	acidv1 "github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1"
--- a/pkg/cluster/cluster_test.go
+++ b/pkg/cluster/cluster_test.go
@@ -1091,170 +1091,101 @@ func TestValidUsernames(t *testing.T) {
 	}
 }
 
-func TestCluster_compareContainers(t *testing.T) {
+func TestClusterComparePorts(t *testing.T) {
 	testCases := []struct {
-		name                    string
-		setA                    []v1.Container
-		setB                    []v1.Container
-		expectedNeedsRollUpdate bool
-		expectedReasons         []string
+		name     string
+		setA     []v1.ContainerPort
+		setB     []v1.ContainerPort
+		expected bool
 	}{
 		{
-			name: "different set sizes",
-			setA: []v1.Container{
-				{Name: "container1"},
-			},
-			setB: []v1.Container{
-				{Name: "container1"},
-				{Name: "container2"},
-			},
-			expectedNeedsRollUpdate: true,
-			expectedReasons: []string{
-				"new statefulset description's length does not match the current ones",
-			},
-		},
-		{
-			name: "different container name",
-			setA: []v1.Container{
-				{Name: "container1"},
-			},
-			setB: []v1.Container{
-				{Name: "container2"},
-			},
-			expectedNeedsRollUpdate: true,
-			expectedReasons: []string{
-				"new statefulset description's container1 (index 0) name does not match the current one",
-			},
-		},
-		{
 			name: "different ports",
-			setA: []v1.Container{
+			setA: []v1.ContainerPort{
 				{
-					Ports: []v1.ContainerPort{
-						{
-							Name:          "metrics",
-							ContainerPort: 9187,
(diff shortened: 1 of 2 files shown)


pkg/cluster/cluster_test.go
```diff
@@ -1088,3 +1090,171 @@ func TestValidUsernames(t *testing.T) {
 		}
 	}
 }
+
+func TestCluster_compareContainers(t *testing.T) {
```
FxKu, line 1094: Could you reduce the tests to just `TestComparePorts` cases?

uzzz: Sure. Done

pkg/cluster/cluster_test.go
```diff
@@ -1088,3 +1090,102 @@ func TestValidUsernames(t *testing.T) {
 		}
 	}
 }
+
+func TestClusterComparePorts(t *testing.T) {
```
FxKu, line 1094:
```suggestion
func TestComparePorts(t *testing.T) {
```

jopadi:
👍

FxKu:
:+1:

FxKu:
Thanks @uzzz for your contribution.

Issue #1758, opened 2022-01-26
Insufficient permissions when kubernetes_use_configmaps is true


stgrace:
When settings `kubernetes_use_configmaps` to true in the values of the operator chart, some endpoint permissions are dropped and configmap permissions are added.
But when trying to create psql cluster using these permissions, the following error is thrown:

```
time="2022-01-26T20:13:08Z" level=error msg="could not create cluster: could not create master endpoint: could not create master endpoint: endpoints is forbidden: User \"system:serviceaccount:postgres-operator:postgres-operator\" cannot create resource \"endpoints\" in API group \"\" in the namespace \"postgres-operator\"" cluster-name=postgres-operator/acid-minimal-cluster pkg=controller worker=0
```

Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** e.g. registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Azure AKS
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report

Some general remarks when posting a bug report:
- Please, check the operator, pod (Patroni) and postgresql logs first. When copy-pasting many log lines please do it in a separate GitHub gist together with your Postgres CRD and configuration manifest.
- If you feel this issue might be more related to the [Spilo](https://github.com/zalando/spilo/issues) docker image or [Patroni](https://github.com/zalando/patroni/issues), consider opening issues in the respective repos.

PR #1759, opened 2022-01-26
Fix chart

stgrace:
See the following issue for more info: #1758


FxKu:
:-1: 

We should better tackle the source of this problem and not adjust the RBAC. If `kubernetes_use_configmaps` is set to true it should create config maps instead of endpoints [here](https://github.com/zalando/postgres-operator/blob/fe340192cabc9757062073530f8599a5c076ff2a/pkg/cluster/cluster.go#L269). The option was introduced since endpoints do not exist in OpenShift.

stgrace:
Even better, I'll close this issue then

Issue #587, opened 2019-06-13
Spilo pods don't define liveness, readiness probes


multi-io:
...which leads to situations where e.g. if Patroni incessantly tries and fails to start postgres, this won't be visible in the pod's status (and hence in generic monitoring setups that don't specifically track e.g. the DB's replica count).

At the minimum I'd think the liveness probe should track the supervisord status (pod is live iff all services are running) and the readiness probe should track the Patroni state of the member (pod is ready iff the state is "running").

Crazy-Hopper:
The absence of readiness probe impairs the PodDisruptionBudget logic, btw.

So I've defined those probes as follows. Maybe it's somewhat an overkill.

```
         livenessProbe:
          exec:
            command:
            - /bin/sh
            - -c
            - |-
              /usr/local/bin/supervisorctl status &&
              /usr/bin/pg_isready &&
              /usr/bin/curl -s localhost:8008 | /bin/grep '"state": "running"'
          failureThreshold: 8
          initialDelaySeconds: 60
          periodSeconds: 30
          successThreshold: 1
          timeoutSeconds: 5

        readinessProbe:
          exec:
            command:
            - /bin/sh
            - -c
            - |-
              /usr/local/bin/supervisorctl status &&
              /usr/bin/pg_isready &&
              /usr/bin/curl -s localhost:8008 | /bin/grep '"state": "running"'
          failureThreshold: 3
          initialDelaySeconds: 45
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 5
```

baddlan:
https://github.com/zalando/patroni/issues/1159

FxKu:
@neelasha-09 let's continue here.

As support for probes has been [merged](https://github.com/zalando/patroni/pull/1590) for Patroni they will likely land somewhen later this in the operator, too.

chitoku-k:
@FxKu Hi, would you mind looking at this issue? It would be pretty helpful if we could add probe logics to Pods.

Thanks in advance.

PR #1760, opened 2022-01-27
do not create endpoints when use config maps

### LINKED ISSUE #1740: When using ConfigMaps operator stiil need permission to create endpoints

- **Which image of the operator are you using?** 'registry.opensource.zalan.do/acid/postgres-operator:v1.7.1'
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** OpenShift 4.6

When using ConfigMaps, the operator still needs permission to create endpoints.

The workaround is to manually add "create" permission to "endpoints" after installing the operator.

I saw a previous issue with the same problem https://github.com/zalando/postgres-operator/issues/1606 but didn't see any feedback from the issue's author.

Error in operator pod
```
time="2022-01-12T15:27:42Z" level=error msg="could not create cluster: could not create master endpoint: could not create master endpoint: endpoints is forbidden: User \"system:serviceaccount:postgres-operator:postgres-operator\" cannot create resource \"endpoints\" in API group \"\" in the namespace \"evg-test-2\"" cluster-name=evg-test-2/acid-test pkg=controller worker=6
```

My postres-operator config
```
apiVersion: acid.zalan.do/v1
configuration:
  ...
  kubernetes_use_configmaps: true
  ...
```

### LINKED ISSUE #1758: Insufficient permissions when kubernetes_use_configmaps is true

evsasha:
This PR disable create endpoins when option `kubernetes_use_configmaps` set to true.
  
I did not change RBAC, the operator works correctly with existing rules
```
  - verbs:
      - get
    apiGroups:
      - ''
    resources:
      - endpoints
```
and also works without them

Tested on OpenShift 4.6.

Fix #1740, Fix #1758

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -260,18 +260,20 @@ func (c *Cluster) Create() error {
 
 	for _, role := range []PostgresRole{Master, Replica} {
 
-		if c.Endpoints[role] != nil {
-			return fmt.Errorf("%s endpoint already exists in the cluster", role)
-		}
-		if role == Master {
-			// replica endpoint will be created by the replica service. Master endpoint needs to be created by us,
-			// since the corresponding master service does not define any selectors.
-			ep, err = c.createEndpoint(role)
-			if err != nil {
-				return fmt.Errorf("could not create %s endpoint: %v", role, err)
+		if !c.patroniKubernetesUseConfigMaps() {
+			if c.Endpoints[role] != nil {
+				return fmt.Errorf("%s endpoint already exists in the cluster", role)
 			}
-			c.logger.Infof("endpoint %q has been successfully created", util.NameFromMeta(ep.ObjectMeta))
-			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Endpoints", "Endpoint %q has been successfully created", util.NameFromMeta(ep.ObjectMeta))
+			if role == Master {
+				// replica endpoint will be created by the replica service. Master endpoint needs to be created by us,
+				// since the corresponding master service does not define any selectors.
+				ep, err = c.createEndpoint(role)
+				if err != nil {
+					return fmt.Errorf("could not create %s endpoint: %v", role, err)
+				}
+				c.logger.Infof("endpoint %q has been successfully created", util.NameFromMeta(ep.ObjectMeta))
+				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Endpoints", "Endpoint %q has been successfully created", util.NameFromMeta(ep.ObjectMeta))
+			}
 		}
 
 		if c.Services[role] != nil {
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -43,7 +43,7 @@ var (
 	alphaNumericRegexp    = regexp.MustCompile("^[a-zA-Z][a-zA-Z0-9]*$")
 	databaseNameRegexp    = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")
 	userRegexp            = regexp.MustCompile(`^[a-z0-9]([-_a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-_a-z0-9]*[a-z0-9])?)*$`)
-	patroniObjectSuffixes = []string{"config", "failover", "sync"}
+	patroniObjectSuffixes = []string{"config", "failover", "sync", "leader"}
 )
 
 // Config contains operator-wide clients and configuration used from a cluster. TODO: remove struct duplication.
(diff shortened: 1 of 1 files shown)


evsasha:
I think i finded one more problem after testing.

Operator do not delete configMaps with role prefix (leader, replica), only endpoints:

https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L911-L917

Therefore, configMaps remain after the cluster is removed.

evsasha:
I tested this fix and get correct output on cluster delete event `test444` in namespace `intsess`
```
time="2022-01-27T15:49:29Z" level=debug msg="removing leftover Patroni objects (endpoints / services and configmaps)" cluster-name=intsess/intsess-test444 pkg=cluster worker=1
time="2022-01-27T15:49:29Z" level=debug msg="deleting Patroni cluster object \"configmap\" with name \"intsess/intsess-test444-config\"" cluster-name=intsess/intsess-test444 pkg=cluster worker=1
time="2022-01-27T15:49:29Z" level=debug msg="deleting Patroni cluster object \"configmap\" with name \"intsess/intsess-test444-leader\"" cluster-name=intsess/intsess-test444 pkg=cluster worker=1
```

FxKu:
I would expect a replacement ConfigMap to be created in case `kubernetes_use_configmaps` is enabled. So instead of `CreateEndpoint`, call `CreateConfigMap`. Same for Delete. What does OpenShift need here?

evsasha:
@FxKu, what you suggest is a bit more complex than my solution:

1. I do not quite understand at what point configmaps are created. I need help with this point.

2. Currently configmaps are deleted by [`deletePatroniClusterConfigMaps()`](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L1517-L1528)
Do we need different function for delete like `deleteConfigmap()`?

3. I don't see configmaps in [type kubeResources](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L59-L67). Shall we provide additional types?

FxKu:
@evsasha sorry for the late reply. I've just learned that the operator does not need to create the ConfigMaps, since Patroni does it for you. And deletion code is already in place as you mentioned. So is setting selector for the master service. Then there's only one last thing to do - defining a [readinessProbe](https://github.com/zalando/patroni/blob/0057f9018bc0b8f78cd5c53d8a6ae7a06141e24d/kubernetes/openshift-example/templates/template_patroni_ephemeral.yml#L157-L166) for the postgres container in the statefulSet template. I think, it makes to add it [after](https://github.com/zalando/postgres-operator/blob/d6641a9b0fb83797bb17bc1f3dcb10635b2ac230/pkg/cluster/k8sres.go#L1240) the generation of the `spiloContainer`.

evsasha:
> Then there's only one last thing to do - defining a [readinessProbe](https://github.com/zalando/patroni/blob/0057f9018bc0b8f78cd5c53d8a6ae7a06141e24d/kubernetes/openshift-example/templates/template_patroni_ephemeral.yml#L157-L166) for the postgres container in the statefulSet template. I think, it makes to add it [after](https://github.com/zalando/postgres-operator/blob/d6641a9b0fb83797bb17bc1f3dcb10635b2ac230/pkg/cluster/k8sres.go#L1240) the generation of the `spiloContainer`.

[FxKu](https://github.com/FxKu) I don't agree with you at this point. Adding probes is not a goal of this PR. I spent some time for research and didn't get the proper result (i couldn't get it to work). I propose not to make this change in that PR.

Moreover, there is a separate task for this: https://github.com/zalando/postgres-operator/issues/587

FxKu:
@evsasha the readinessProbe we need only in case patroniKubernetesUseConfigMaps is true. But, I'm fine doing it in a separate PR.

FxKu:
:+1:

jopadi:
👍

Issue #1765, opened 2022-01-31
DB pod in pending state --> Eventually all 3 DB pods are in pending states ( down ). "Rolling upgrade"


Samusername:
Following seems serious. Unless it is some configuration problem.

ConfigMap setups.
 - configTarget: "ConfigMap"
 - kubernetes_use_configmaps: true

synchronous_mode: true and synchronous_mode_strict: true.
Spilo: spilo-14:2.1-p3
postgres-operator: v1.7.1
PostgreSQL: 13.4
helm version 3
Usage: In production pipe.

Testing scenario: intentionally force a pod to wait in pending state.

helm upgrade

First all are ok:

default             rdbms-pg-cluster-0                                           2/2     Running            0          104m    x.y.93.250    systemnode06   <none>           <none>
default             rdbms-pg-cluster-1                                           2/2     Running            0          102m    x.y.61.40     systemnode09   <none>           <none>
default             rdbms-pg-cluster-2                                           2/2     Running            0          101m    x.y.60.234    systemnode07   <none>           <none>

default             rdbms-pg-cluster-0                                           2/2     Running            0          112m    x.y.93.250    systemnode06   <none>           <none>
default             rdbms-pg-cluster-1                                           2/2     **Terminating**        0          110m    x.y.61.40     systemnode09   <none>           <none>
default             rdbms-pg-cluster-2                                           2/2     Running            0          108m    x.y.60.234    systemnode07   <none>           <none>

That pod was in pending state. This is an intentional test case, where pods can end up in pending state.

Then 3rd pod was restarted by postgres-operator (?) / 3rd pod vanished:

default             rdbms-pg-cluster-0                                           2/2     Running            0          161m    x.y.93.250    systemnode06   <none>           <none>
default             rdbms-pg-cluster-1                                           0/2     Pending            0          48m     <none>            <none>          <none>           <none>

At that point:

| Member             | Host           | Role    | State   | TL | Lag in MB |
+--------------------+----------------+---------+---------+----+-----------+
| rdbms-pg-cluster-0 | x.y.93.250 | Leader  | running |  1 |           |
| rdbms-pg-cluster-1 |                | Replica |         |    |   unknown |
+--------------------+----------------+---------+---------+----+-----------+

After a while, also Leader pod ended up in Pending state:

default             rdbms-pg-cluster-0                                           0/2     **Pending**             0          24m     <none>            <none>          <none>           <none>
[...]

PR #1700, opened 2021-11-23
choose switchover candidate based on lag and role

FxKu:
Change 1: When attempting a switchover, try it only on a replica that has no replication lag. In case synchronous replication is enabled there will be a replica with `sync_standby` role which should be chosen instead for switchover.

Change 2: In case the switchover is not executed or failing, return an error instead of emitting a `Warning`. Before we were always continuing by replacing the master pod anyway - producing downtime. Now, the master keeps running on an outdated pod. But, the Pod will have the rolling-update flag and the operator will notice it on the next sync and try another failover. It can be, that the failover just took longer, e.g. because taking a CHECKPOINT took longer than Patroni API waits to respond to the user. In this ideal case only a replica would be recreated. In worst case, a switchover might be retried on each sync. But this should be fixed by the admin anyway.

fixes #1686 
fixes #109
fixes #600

PR #1767, opened 2022-02-01
do not recreate pods if previous Patroni API calls fail

FxKu:
In #1341 we moved the rolling update flag from statefulset to single pods. From there on the list of pods to recreate was gathered in `syncStatefulset` which was then passed to `recreatedPods`. Our `isSafeToRecreatePods` [check](https://github.com/zalando/postgres-operator/blob/a78a619e909a1c64d4917bf6a426bcaff18dbc39/pkg/cluster/pod.go#L427) within `recreatePods` only examined the pods to recreate.

This could lead to a situation demonstrated in #1765. A change in resources leads to unschedulable instances. The rolling updates times out, but on next sync the operator will continue with the other instances that [have the update flag](https://github.com/zalando/postgres-operator/blob/a78a619e909a1c64d4917bf6a426bcaff18dbc39/pkg/cluster/sync.go#L319).

The log lines from the issue make the problem very well visible:
```
log lines from syncStatefulSet

time="2022-01-31T12:23:07Z" level=debug msg="syncing statefulsets" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:07Z" level=debug msg="found rolling update flag on pod \"rdbms-pg-cluster-0\"" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:07Z" level=debug msg="found rolling update flag on pod \"rdbms-pg-cluster-2\"" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:07Z" level=debug msg="2 / 3 pod(s) still need to be rotated" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:07Z" level=debug msg="making GET http request: http://x.y.93.250:8008/config" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=debug msg="could not get member data of  pod default/rdbms-pg-cluster-1 - skipping possible restart attempt:  is not a valid IP" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=debug msg="making GET http request: http://x.y.60.234:8008/patroni" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=debug msg="making GET http request: http://x.y.93.250:8008/patroni" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=debug msg="performing rolling update" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=info msg="there are 2 pods in the cluster to recreate" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0

log lines from isSafeToRecreatePods

time="2022-01-31T12:23:19Z" level=debug msg="name=rdbms-pg-cluster-0 phase=Running ip=x.y.93.250" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=debug msg="name=rdbms-pg-cluster-2 phase=Running ip=x.y.60.234" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=debug msg="making GET http request: http://x.y.93.250:8008/patroni" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
time="2022-01-31T12:23:19Z" level=debug msg="making GET http request: http://x.y.60.234:8008/patroni" cluster-name=default/rdbms-pg-cluster pkg=cluster worker=0
```

 At least, since #1700 (will be in `v1.8.0`) the master [should not get recreated](https://github.com/zalando/postgres-operator/blob/a78a619e909a1c64d4917bf6a426bcaff18dbc39/pkg/cluster/pod.go#L466) when there's no [healthy switchover target](https://github.com/zalando/postgres-operator/blob/a78a619e909a1c64d4917bf6a426bcaff18dbc39/pkg/cluster/pod.go#L506) (and number of instances is > 1).

We should return to run `isSafeToRecreatePods` on all pods and in fact we already do so by calling Patroni config (only once) and [member data endpoint](https://github.com/zalando/postgres-operator/blob/a78a619e909a1c64d4917bf6a426bcaff18dbc39/pkg/cluster/sync.go#L470) when attempting restarts. See also log lines above. 

`restart_instance` does not return an error yet, but this would be changed by the PR. Thus we can remove `isSafeToRecreatePods` function and only add a flag in `syncStatefulSet`. I took the retry logic from it as well as the `State == "creating replica"` check and moved those extra code parts to separate functions in `Pod.go` to make `syncStatefulSet` less verbose.

### DIFF
--- a/pkg/cluster/pod.go
+++ b/pkg/cluster/pod.go
@@ -13,6 +13,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
+	acidv1 "github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1"
 	"github.com/zalando/postgres-operator/pkg/spec"
 	"github.com/zalando/postgres-operator/pkg/util"
 	"github.com/zalando/postgres-operator/pkg/util/patroni"
@@ -347,6 +348,54 @@ func (c *Cluster) MigrateReplicaPod(podName spec.NamespacedName, fromNodeName st
 	}
 
 	return nil
+}
+
+func (c *Cluster) getPatroniConfig(pod *v1.Pod) (acidv1.Patroni, map[string]string, error) {
+	var (
+		patroniConfig acidv1.Patroni
+		pgParameters  map[string]string
+	)
+	podName := util.NameFromMeta(pod.ObjectMeta)
+	err := retryutil.Retry(1*time.Second, 5*time.Second,
+		func() (bool, error) {
+			var err error
+			patroniConfig, pgParameters, err = c.patroni.GetConfig(pod)
+
+			if err != nil {
+				return false, err
+			}
+			return true, nil
+		},
+	)
+
+	if err != nil {
+		return acidv1.Patroni{}, nil, fmt.Errorf("could not get Postgres config from pod %s: %v", podName, err)
+	}
+
+	return patroniConfig, pgParameters, nil
+}
+
+func (c *Cluster) getPatroniMemberData(pod *v1.Pod) (patroni.MemberData, error) {
+	var memberData patroni.MemberData
+	err := retryutil.Retry(1*time.Second, 5*time.Second,
+		func() (bool, error) {
+			var err error
+			memberData, err = c.patroni.GetMemberData(pod)
+
+			if err != nil {
+				return false, err
+			}
+			return true, nil
+		},
+	)
+	if err != nil {
+		return patroni.MemberData{}, fmt.Errorf("could not get member data: %v", err)
+	}
+	if memberData.State == "creating replica" {
+		return patroni.MemberData{}, fmt.Errorf("replica currently being initialized")
+	}
+
+	return memberData, nil
 }
 
 func (c *Cluster) recreatePod(podName spec.NamespacedName) (*v1.Pod, error) {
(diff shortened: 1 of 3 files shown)


pkg/cluster/pod.go
```diff
@@ -349,6 +350,54 @@ func (c *Cluster) MigrateReplicaPod(podName spec.NamespacedName, fromNodeName st
...
+	if err != nil {
+		return acidv1.Patroni{}, nil, fmt.Errorf("could not get Postgres config from pod %s: %v", podName, err)
+	}
+
+	return patroniConfig, pgParameters, nil
+}
+
+func (c *Cluster) getPatroniMemberData(pod *v1.Pod) (patroni.MemberData, error) {
+	var memberData patroni.MemberData
+	err := retryutil.Retry(1*time.Second, 5*time.Second,
```
dmvolod, line 380: There are a lot of places with the same timeout/retry combination (`1*time.Second, 5*time.Second`).
@FxKu what do you think about to make it configurable as it could be depends of the env performance or/and utilize some constants for them?

FxKu: Makes sense. I will create two timeout options for Patroni API calls. Maybe in a follow up PR so that this one doesn't get too polluted.

jopadi:
👍

FxKu:
:+1:

Issue #1689, opened 2021-11-15
major version upgrade failed


monotek:
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
- **Where do you run it - cloud or metal? Syseleven (Openstack)
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** [Bug report, question, feature request, etc.] Bug

We just tried to update the major postgres version of one of our patroni clusters.

For this we've update the following fileds in the cr:

spec.dockerImage: 
from: "registry.opensource.zalan.do/acid/spilo-12:1.6-p5" to "registry.opensource.zalan.do/acid/spilo-13:2.1-p1"

spec.postgresql.version:
from "12" to "13"

In the first place everything looked fine as the pods were successfully restarted but the operator logs point out that it fails:

```
time="2021-11-15T16:26:43Z" level=debug msg="making GET http request: http://172.25.78.171:8008/patroni" cluster-name=auth/auth-postgres-db pkg=cluster
time="2021-11-15T16:26:43Z" level=info msg="healthy cluster ready to upgrade, current: 120008 desired: 130000" cluster-name=auth/auth-postgres-db pkg=cluster
time="2021-11-15T16:26:43Z" level=info msg="triggering major version upgrade on pod auth-postgres-db-2 of 3 pods" cluster-name=auth/auth-postgres-db pkg=cluster
time="2021-11-15T16:26:45Z" level=error msg="major version upgrade failed: could not execute: command terminated with exit code 1" cluster-name=auth/auth-postgres-db pkg=cluster
time="2021-11-15T16:26:45Z" level=info msg="cluster has been synced" cluster-name=auth/auth-postgres-db pkg=controller worker=0
time="2021-11-15T16:26:45Z" level=info msg="recieved add event for already existing Postgres cluster" cluster-name=auth/auth-postgres-db pkg=controller worker=0
```

What can we do to get the upgrade to run?

FxKu:
Can you check Postgres logs (not pod logs) of the master where the upgrade was attempted? Sometimes an update can fail if there is data corruption.

monotek:
Where can i find non pod / postgresql logs?
Are they saved inside the container?

FxKu:
[...]

PR #1770, opened 2022-02-03
major version upgrade for rootless and ocp : solving #1689

neelasha-09:
We tested the solution in 3 places 
1. Openshift (with reduced perms)
2. AWS (with access runAsNonRoot: false )
3. Pure Vanila k8s

At this stage, we cannot change spilo image instead. 
Reason: PGO does not call a wrapper shell script (sh), but directly the python process. If it were a shell, we could have added this logic there. But now, it's too late, and only place to do it at this time is PGO.

The solution is conditionally applying solution of #1689 when root-less

### DIFF
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/zalando/postgres-operator/pkg/spec"
 	"github.com/zalando/postgres-operator/pkg/util"
@@ -106,13 +107,35 @@ func (c *Cluster) majorVersionUpgrade() error {
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Starting major version upgrade on pod %s of %d pods", masterPod.Name, numberOfPods)
 			upgradeCommand := fmt.Sprintf("/usr/bin/python3 /scripts/inplace_upgrade.py %d 2>&1 | tee last_upgrade.log", numberOfPods)
 
-			result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
-			if err != nil {
-				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
-				return err
+			// ######## checking if the spilo image runs with root or non-root (check for user id=0)
+			c.logger.Infof("checking if the spilo image runs with root or non-root (check for user id=0)")
+
+			resultIdCheck, errIdCheck := c.ExecCommand(podName, "/bin/bash", "-c", "/usr/bin/id -u")
+			if errIdCheck != nil {
+				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Pre-Upgrade step (checking user with id -u) for Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, errIdCheck)
+				return errIdCheck
 			}
 
-			c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
+			resultIdCheck = strings.TrimSuffix(resultIdCheck, "\n")
+
+			if resultIdCheck != "0" {
+				c.logger.Infof("User id was identified as: %s, hence default user is non-root already", resultIdCheck)
+				result, err := c.ExecCommand(podName, "/bin/bash", "-c", upgradeCommand)
+				if err != nil {
+					c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
+					return err
+				}
+				c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
+			} else {
+				c.logger.Infof("User id was identified as: %s, using su to reach the postgres user", resultIdCheck)
+				result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
+				if err != nil {
+					c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
+					return err
+				}
+				c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
+			}
+
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d finished", c.currentMajorVersion, desiredVersion)
 		}
 	}
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -107,7 +107,7 @@ func (c *Cluster) majorVersionUpgrade() error {
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Starting major version upgrade on pod %s of %d pods", masterPod.Name, numberOfPods)
 			upgradeCommand := fmt.Sprintf("/usr/bin/python3 /scripts/inplace_upgrade.py %d 2>&1 | tee last_upgrade.log", numberOfPods)
 
-			// ######## checking if the spilo image runs with root or non-root (check for user id=0)
+			
 			c.logger.Debugf("checking if the spilo image runs with root or non-root (check for user id=0)")
 
 			resultIdCheck, errIdCheck := c.ExecCommand(podName, "/bin/bash", "-c", "/usr/bin/id -u")
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -117,27 +117,23 @@ func (c *Cluster) majorVersionUpgrade() error {
 			}
 
 			resultIdCheck = strings.TrimSuffix(resultIdCheck, "\n")
-
+            result, err := c.ExecCommand(podName, "/bin/bash")
 			if resultIdCheck != "0" {
 				c.logger.Infof("User id was identified as: %s, hence default user is non-root already", resultIdCheck)
-				result, err := c.ExecCommand(podName, "/bin/bash", "-c", upgradeCommand)
-				if err != nil {
-					c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
-					return err
-				}
-				c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
+				result, err = c.ExecCommand(podName, "/bin/bash", "-c", upgradeCommand)
 			} else {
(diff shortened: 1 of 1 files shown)


pkg/cluster/majorversionupgrade.go
```diff
@@ -106,13 +107,35 @@ func (c *Cluster) majorVersionUpgrade() error {
...
+
+			resultIdCheck, errIdCheck := c.ExecCommand(podName, "/bin/bash", "-c", "/usr/bin/id -u")
+			if errIdCheck != nil {
+				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Pre-Upgrade step (checking user with id -u) for Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, errIdCheck)
+				return errIdCheck
+			}
+
+			resultIdCheck = strings.TrimSuffix(resultIdCheck, "\n")
+
+			if resultIdCheck != "0" {
```
FxKu, line 121: on set the command here ("/bin/bash" or "/bin/su") so you don't repeat all the code in this if else block

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,13 +107,35 @@ func (c *Cluster) majorVersionUpgrade() error {
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Starting major version upgrade on pod %s of %d pods", masterPod.Name, numberOfPods)
 			upgradeCommand := fmt.Sprintf("/usr/bin/python3 /scripts/inplace_upgrade.py %d 2>&1 | tee last_upgrade.log", numberOfPods)
 
-			result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
-			if err != nil {
-				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
-				return err
+			// ######## checking if the spilo image runs with root or non-root (check for user id=0)
+			c.logger.Infof("checking if the spilo image runs with root or non-root (check for user id=0)")
```
FxKu, line 111: This should rather be on the Debug level. Comment above is obsolete.

neelasha-09:
@FxKu : Could you please review and update ?

FxKu:
@neelasha-09 well you addressed one comment. What about the other? There's some redundant code where only the command 
"/bin/bash" vs. "/bin/su" differ. It can be simplified.

neelasha-09:
@FxKu 
execCommand has different number of args in each case
so we cannot simply have a string with 2 values, it's two different execCommands.
We agree with modifying the code if that is requried.

neelasha-09:
@FxKu Can you please suggest how can it be done ?

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,13 +107,35 @@ func (c *Cluster) majorVersionUpgrade() error {
...
+				c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
+			} else {
+				c.logger.Infof("User id was identified as: %s, using su to reach the postgres user", resultIdCheck)
+				result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
+				if err != nil {
+					c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
+					return err
+				}
+				c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
 			}
```
FxKu, line 137:
```suggestion
			if resultIdCheck != "0" {
				c.logger.Infof("User id was identified as: %s, hence default user is non-root already", resultIdCheck)
				result, err := c.ExecCommand(podName, "/bin/bash", "-c", upgradeCommand)
			} else {
				c.logger.Infof("User id was identified as: %s, using su to reach the postgres user", resultIdCheck)
				result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
			}
			
			if err != nil {
				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
				return err
			}
			c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
```

make sure indentation (go.fmt) is correct. Sometimes github suggestions can mess it up.

FxKu, line 137:
```suggestion
			if resultIdCheck != "0" {
				c.logger.Infof("User id was identified as: %s, hence default user is non-root already", resultIdCheck)
				result, err := c.ExecCommand(podName, "/bin/bash", "-c", upgradeCommand)
			} else {
				c.logger.Infof("User id was identified as: %s, using su to reach the postgres user", resultIdCheck)
				result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
			}

			if err != nil {
				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
				return err
			}
			c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
```

neelasha-09:
@FxKu : Please check and approve the PR

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,15 +107,33 @@ func (c *Cluster) majorVersionUpgrade() error {
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Starting major version upgrade on pod %s of %d pods", masterPod.Name, numberOfPods)
 			upgradeCommand := fmt.Sprintf("/usr/bin/python3 /scripts/inplace_upgrade.py %d 2>&1 | tee last_upgrade.log", numberOfPods)
 
-			result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
+			
+			c.logger.Debugf("checking if the spilo image runs with root or non-root (check for user id=0)")
+
+			resultIdCheck, errIdCheck := c.ExecCommand(podName, "/bin/bash", "-c", "/usr/bin/id -u")
+			if errIdCheck != nil {
+				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Pre-Upgrade step (checking user with id -u) for Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, errIdCheck)
```
FxKu, line 115: we should change the event type here to `v1.EventTypeWarning`

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,15 +107,33 @@ func (c *Cluster) majorVersionUpgrade() error {
...
+				return errIdCheck
+			}
+
+			resultIdCheck = strings.TrimSuffix(resultIdCheck, "\n")
+            result, err := c.ExecCommand(podName, "/bin/bash")
```
FxKu, line 120:
```suggestion
```
This line looks out of place here.

pkg/cluster/majorversionupgrade.go
FxKu, line 115:
```suggestion
				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeWarning, "Major Version Upgrade", "checking user id to run upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, errIdCheck)
```
we should change the event type here to v1.EventTypeWarning

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,15 +107,33 @@ func (c *Cluster) majorVersionUpgrade() error {
...
+				result, err = c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
+			}
 			if err != nil {
 				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, err)
 				return err
 			}
-
 			c.logger.Infof("upgrade action triggered and command completed: %s", result[:100])
+
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Upgrade from %d to %d finished", c.currentMajorVersion, desiredVersion)
```
FxKu, line 134:
```suggestion
			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeWarning, "Major Version Upgrade", "upgrade from %d to %d finished", c.currentMajorVersion, desiredVersion)
```

FxKu:
@neelasha-09 please fix the errors (e.g. define result variable etc.) so that unit tests will run.

neelasha-09:
@FxKu the command at line 120 was added to define the variable as I'm not sure  of the type of the variables to be defined for `result` and `err`
`result, err := c.ExecCommand(podName, "/bin/bash")`
Please confirm the same

FxKu:
You will know it if you check what [ExecCommand](https://github.com/zalando/postgres-operator/blob/b300fca364466485ffc07432e9b2a4ae95930177/pkg/cluster/exec.go#L19) function returns? Since `err` is already defined above is enough to just define `var result string`

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,15 +107,31 @@ func (c *Cluster) majorVersionUpgrade() error {
...
 		}
+
```
FxKu, line 134: can you remove these new lines. Have you run go.fmt over your code?

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,15 +107,31 @@ func (c *Cluster) majorVersionUpgrade() error {
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Starting major version upgrade on pod %s of %d pods", masterPod.Name, numberOfPods)
 			upgradeCommand := fmt.Sprintf("/usr/bin/python3 /scripts/inplace_upgrade.py %d 2>&1 | tee last_upgrade.log", numberOfPods)
 
-			result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
+			
+			c.logger.Debugf("checking if the spilo image runs with root or non-root (check for user id=0)")
+			resultIdCheck, errIdCheck := c.ExecCommand(podName, "/bin/bash", "-c", "/usr/bin/id -u")
+			if errIdCheck != nil {
+				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeWarning, "Major Version Upgrade", "Checking user id to run upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, errIdCheck)
```
FxKu, line 114:
```suggestion
				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Major Version Upgrade", "Checking user id to run upgrade from %d to %d FAILED: %v", c.currentMajorVersion, desiredVersion, errIdCheck)
```

pkg/cluster/majorversionupgrade.go
```diff
@@ -105,14 +106,28 @@ func (c *Cluster) majorVersionUpgrade() error {
...
+			}
 
-			result, err := c.ExecCommand(podName, "/bin/su", "postgres", "-c", upgradeCommand)
+			resultIdCheck = strings.TrimSuffix(resultIdCheck, "\n")
+                        var result string
```
FxKu, line 117:
```suggestion
			var result string
```

CyberDem0n:
:+1:

FxKu:
:+1:

FxKu:
Thanks @neelasha-09 for your contribution!

Issue #1776, opened 2022-02-10
Retry fetching the pod_environment_secret


snarlysodboxer:
- **Which image of the operator are you using?** The latest, I'm looking at the codebase.
- **Where do you run it - Metal. Kubernetes.
- **Type of issue?** - Bug report.

Sometimes secrets don't show up for a few moments, for example if using [ExternalSecrets](https://external-secrets.io/v0.4.1/).

The [current code](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L896-L902) tries once to retrieve the `pod_environment_secret` secret, and from my testing, only tries again after 1 hour.

Any reason not to use the retryutil and add a few retries here? I may be able to help, depending on how involved.

PR #1777, opened 2022-02-10
Retry when getting the pod_environment_secret

snarlysodboxer:
See https://github.com/zalando/postgres-operator/issues/1776.

Not sure if this is how you would want this implemented, but it didn't take long so I thought I'd get the conversation going. It's unclear to me how retries here might play into retries at higher levels.

Should the interval or timeout be configurable? Should it retry indefinitely?

### DIFF
--- a/docs/developer.md
+++ b/docs/developer.md
@@ -208,6 +208,13 @@ dlv connect 127.0.0.1:DLV_PORT
 
 ## Unit tests
 
+Prerequisites:
+
+```bash
+make deps
+make mocks
+```
+
 To run all unit tests, you can simply do:
 
 ```bash
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -8,11 +8,13 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	appsv1 "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	policybeta1 "k8s.io/api/policy/v1beta1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -24,6 +26,7 @@ import (
 	"github.com/zalando/postgres-operator/pkg/util/config"
 	"github.com/zalando/postgres-operator/pkg/util/constants"
 	"github.com/zalando/postgres-operator/pkg/util/k8sutil"
+	"github.com/zalando/postgres-operator/pkg/util/retryutil"
 	batchv1 "k8s.io/api/batch/v1"
 	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -893,12 +896,30 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
 		return secretPodEnvVarsList, nil
 	}
 
-	secret, err := c.KubeClient.Secrets(c.Namespace).Get(
-		context.TODO(),
-		c.OpConfig.PodEnvironmentSecret,
-		metav1.GetOptions{})
+	secret := &v1.Secret{}
+	var notFoundErr error
+	err := retryutil.Retry(c.OpConfig.ResourceCheckInterval, c.OpConfig.ResourceCheckTimeout,
+		func() (bool, error) {
+			var err error
+			secret, err = c.KubeClient.Secrets(c.Namespace).Get(
+				context.TODO(),
+				c.OpConfig.PodEnvironmentSecret,
+				metav1.GetOptions{})
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					notFoundErr = err
+					return false, nil
+				}
+				return false, err
+			}
+			return true, nil
+		},
+	)
+	if notFoundErr != nil && err != nil {
+		err = errors.Wrap(notFoundErr, err.Error())
+	}
 	if err != nil {
-		return nil, fmt.Errorf("could not read Secret PodEnvironmentSecretName: %v", err)
+		return nil, errors.Wrap(err, "could not read Secret PodEnvironmentSecretName")
 	}
 
 	for k := range secret.Data {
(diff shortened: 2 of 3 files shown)


pkg/cluster/k8sres.go
```diff
@@ -893,10 +895,20 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
 		return secretPodEnvVarsList, nil
 	}
 
-	secret, err := c.KubeClient.Secrets(c.Namespace).Get(
-		context.TODO(),
-		c.OpConfig.PodEnvironmentSecret,
-		metav1.GetOptions{})
+	secret := &v1.Secret{}
+	err := retryutil.Retry(10*time.Second, 3*time.Minute,
```
dmvolod, line 899: @snarlysodboxer what do you thinks about to don't hardcode these values and utilize configurable values from the `Resources.ResourceCheckInterval` and `Resources.ResourceCheckTimeout`

FxKu: I was thinking about this as well. The values here are not too far off from the defaults of `ResourceCheckInterval` and `ResourceCheckTimeout` so I would vote to use them @snarlysodboxer .

pkg/cluster/k8sres.go
```diff
@@ -7,6 +7,7 @@ import (
 	"path"
 	"sort"
 	"strings"
+	"time"
```
FxKu, line 11:
```suggestion

```

FxKu: Not used. You don't have go.fmt enabled in your IDE?

snarlysodboxer: I do, but it runs a little slow sometimes on really large files like this one. Still can't believe I missed this though!

I've now fixed it.

FxKu: It's still there @snarlysodboxer

snarlysodboxer: I'm so embarrassed. Fixed for real, for real, this time.

FxKu:
:+1:

snarlysodboxer:
> Ok, one unit test is failing because the intervals must be set for it. Please add the following two lines 

Done.

I had a lot of trouble trying to run the tests. I couldn't get them to pass even in the master branch. `GO111MODULE=on go test ./...` returns `undefined: mocks.NewMockVolumeResizer` errors.

pkg/cluster/k8sres.go
```diff
@@ -893,10 +894,20 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
...
+	secret := &v1.Secret{}
+	err := retryutil.Retry(c.OpConfig.ResourceCheckInterval, c.OpConfig.ResourceCheckTimeout,
+		func() (bool, error) {
+			var err error
+			secret, err = c.KubeClient.Secrets(c.Namespace).Get(
+				context.TODO(),
+				c.OpConfig.PodEnvironmentSecret,
+				metav1.GetOptions{})
+			if err != nil {
+				return false, err
```
dmvolod, line 906: `err` shouldn't be returned here, otherwise `Retry` will not work and existed immediately.
Probably only `apierrors.IsNotFound` could be ignored only

snarlysodboxer: Excellent observation. I changed it to check for the is not found error, and to wrap it when needed. I also added a test.

snarlysodboxer:
I was able to get the unit tests running after some trouble. - I added to the `docs/developers.md` to hopefully help others avoid the problems I ran into.

docs/developer.md
```diff
@@ -208,6 +208,13 @@ dlv connect 127.0.0.1:DLV_PORT
 
 ## Unit tests
 
+Prerequisites:
+
+```bash
+make deps
+make mocks
```
dmvolod, line 215: It seems tp we need to automate a bit `mockgen` installation with for the local env
`go install github.com/golang/mock/mockgen@latest`

snarlysodboxer: I'm sorry, I don't quite follow. Are you saying add `go install github.com/golang/mock/mockgen@latest` to these prerequisites?

I notice that `make deps` calls `make tools`, and `make tools` [installs mockgen](https://github.com/zalando/postgres-operator/blob/master/Makefile#L89) version 1.6.0.

dmvolod: Oh, yes, sorry missed that deps installing this tools. You are right.

pkg/cluster/k8sres_test.go
```diff
@@ -788,16 +795,31 @@ func TestPodEnvironmentSecretVariables(t *testing.T) {
...
+			subTest: "API error during PodEnvironmentSecret retrieval",
+			opConfig: config.Config{
+				Resources: config.Resources{
+					PodEnvironmentSecret:  testPodEnvironmentSecretNameAPIError,
+					ResourceCheckInterval: time.Duration(3),
+					ResourceCheckTimeout:  time.Duration(10),
 				},
 			},
-			err: fmt.Errorf("could not read Secret PodEnvironmentSecretName: Secret PodEnvironmentSecret not found"),
+			err: fmt.Errorf(`could not read Secret PodEnvironmentSecretName: Secret PodEnvironmentSecret API error`),
```
FxKu, line 814: are back ticks required here instead of double quotes?

snarlysodboxer: Not required. Was a left over from when the message had double quotes in it, like on like 803. I've now updated.

pkg/cluster/k8sres_test.go
```diff
@@ -788,16 +795,31 @@ func TestPodEnvironmentSecretVariables(t *testing.T) {
 			subTest: "Secret referenced by PodEnvironmentSecret does not exist",
 			opConfig: config.Config{
 				Resources: config.Resources{
-					PodEnvironmentSecret: "idonotexist",
+					PodEnvironmentSecret:  "idonotexist",
+					ResourceCheckInterval: time.Duration(3),
+					ResourceCheckTimeout:  time.Duration(10),
+				},
+			},
+			err: fmt.Errorf(`could not read Secret PodEnvironmentSecretName: still failing after 3 retries: secret.core "idonotexist" not found`),
```
FxKu, line 803:
```suggestion
			err: fmt.Errorf("could not read Secret PodEnvironmentSecretName: still failing after 3 retries: secret.core "idonotexist" not found"),
```

FxKu, line 803:
```suggestion
			err: fmt.Errorf("could not read Secret PodEnvironmentSecretName: still failing after 3 retries: secret.core %q not found", "idonotexist"),
```

FxKu: Would make sense to have a constant for the name, check #1794

snarlysodboxer: I've added [this commit](https://github.com/zalando/postgres-operator/pull/1777/commits/4ea0954efcfda03010a800148436375cdd17524c), is that what you were looking for?

FxKu:
:+1:

jopadi:
👍

PR #1789, opened 2022-02-21
Copy pod_environment_secret to the cluster namespaces

### LINKED ISSUE #1272: Feature request: Copy environment secret to cluster namespaces

Since #946 you can specify Spilo Pod environment variables through a K8s Secret, by specifying the name of the Secret in the operator configuration:

```
configuration:
  kubernetes:
    pod_environment_secret: spilo-env
```

## Problem

However, the Secret must live in the same namespace as the Postgresql cluster. If you want to use this features for multiple Postgresql clusters in different namespaces, you must manually create the `spilo-env` Secret in each namespace. It would be nice if there was a feature such that you need to create the Secret only once.

## Possible Solution

I suggest the following as a possible solution.

1. We should allow to optionally specify a namespace in `pod_environment_secret`, e.g. `pod_environment_secret: postgres-operator/spilo-env`.
2. If a namespace is specified, then for each Postgresql cluster, the operator should copy the Secret to the cluster namespace.

dmvolod:
Fixes #1272

### DIFF
--- a/manifests/postgresql-operator-default-configuration.yaml
+++ b/manifests/postgresql-operator-default-configuration.yaml
@@ -79,7 +79,7 @@ configuration:
     pdb_name_format: "postgres-{cluster}-pdb"
     pod_antiaffinity_topology_key: "kubernetes.io/hostname"
     # pod_environment_configmap: "default/my-custom-config"
-    # pod_environment_secret: "my-custom-secret"
+    # pod_environment_secret: "default/my-custom-secret"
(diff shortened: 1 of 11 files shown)


FxKu:
The original author of this feature restricted it to the namespace for a [reason](https://github.com/zalando/postgres-operator/pull/946#discussion_r440727300). Not sure how I should feel about copying a secret with credentials around. @frittentheke fyi.

PR #1794, opened 2022-02-22
Ability to set pod environment variables on cluster resource

### LINKED ISSUE #907: Disable WAL archiving on cluster

### LINKED ISSUE #1209: Can we configure s3 bucket per pg cluster?
Can we configure s3 bucket per pg cluster?

Currently the s3 bucket for `WAL_S3_BUCKET` and  `LOGICAL_BACKUP_S3_BUCKET` come from operator config.
```
if c.OpConfig.WALES3Bucket != "" {
		envVars = append(envVars, v1.EnvVar{Name: "WAL_S3_BUCKET", Value: c.OpConfig.WALES3Bucket})
		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
}
```

```
               { 
			Name:  "LOGICAL_BACKUP_S3_BUCKET",
			Value: c.OpConfig.LogicalBackup.LogicalBackupS3Bucket,
		},
```
We need to configure bucket per cluster. I tried to define them in `pod_environment_configmap` but looks like operator config takes precedence.
```
        // add vars taken from pod_environment_configmap and pod_environment_secret first
	// (to allow them to override the globals set in the operator config)
	if len(customPodEnvVarsList) > 0 {
		envVars = append(envVars, customPodEnvVarsList...)
	}
```
Is there anyway we could configure s3 bucket per cluster and overwrite s3 bucket defined in operator config??

### LINKED ISSUE #1422: Can we configure separate s3 bucket per pg cluster
Hi Team,

I am new in operator need some clarification how zalando operator works. here i am deploying zalando operator by usig crd based configuration and create custom config also which having my s3 information and that config inject in crd config,then deploy operator after all these create manifest cluster and my s3 information coming as a environment variable in a pod then wal backup started to my s3 all is working fine. 
Now my requirement is to create another manifest cluster in different namespace by using same operator and take s3 backup in different s3 bucket but it's moving in same s3 bucket that we have passed in our previous config so please anyone help me out how can we take backup in different s3 bucket for another manifest cluster or how to inject config map in our second cluster manifest.

Thanks in advance..

_Originally posted by @vishantsolanki in https://github.com/zalando/postgres-operator/issues/1419#issuecomment-806451852_

### LINKED ISSUE #1566: Feature request: ability to set pod_environment_secret and pod_environment_configmap on cluster resource
- **Which image of the operator are you using?** e.g. registry.opensource.zalan.do/acid/postgres-operator:v1.6.3
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AWS K8s / Bare Metal K8s

We need configuration flexibility for backup and logging configuration on a cluster basis. Use cases:
- per cluster backup conf (WAL_S3_BUCKET)
- per cluster logging conf (LOG_SHIP_SCHEDULE)
- per cluster vars for custom images

Similar requests:
https://github.com/zalando/postgres-operator/issues/907
https://github.com/zalando/postgres-operator/issues/1209
https://github.com/zalando/postgres-operator/issues/1422

And more in comments on other issues.

It is a relatively small change with a big impact on operator usability.

### LINKED ISSUE #1807: How to add some additional env in pg pod?

- **Which image of the operator are you using?**  registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** OpenShift
- **Type of issue?** question

As I know, I had added a common pod envs by:
```yaml
pod_environment_configmap: "postgresql-pod-config"
```
In this configmap, I add some backup envs about wal-g and it works.
But in some cases, I start a pg database for test or dev, I don't want to backup this database log and data. I want to reset these common envs. Or another case, I want to backup a database to another bucket or s3 url.

I didn't find a solution to this problem in the document. So I want to ask whether the operator does not support this case at present? Should we add this feature?

dmvolod:
Fixes #1566
Closes #1209
Closes #1422
Closes #907
Closes #1807

Please pay attention, that this PR implementing an individual `[]EnvVar` option to the cluster CR and has higher priority to the cluster `pod_environment_secret` and `pod_environment_configmap` options.

TODO (after approach approve)

- [x] Doc changes
- [x] Unit testing

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -708,11 +708,10 @@ to the Postgres StatefulSet/pods.
 
 ### For individual cluster
 
-It is possible to set environment variables in the Spilo pods directly as parameters or
-as a link to secrets for each individual cluster, in addition to or instead of the
-global parameters given in the examples above.
-
-In order to do this, you need to configure the cluster parameters, as in the example below
+It is possible to define environment variables directly in the Postgres cluster
+manifest to configure it individually. The variables must be listed under the
+`env` section in the same way you would do for [containers](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/).
+Global parameters served from a custom config map or secret will be overridden.
 
 ```yaml
 apiVersion: "acid.zalan.do/v1"
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -729,11 +729,6 @@ spec:
           key: minio_secret_key
 ```
 
-If global parameters are set at the level of the entire controller in the form of
-a ConfigMap or a Secret, then the individual cluster parameters will have a higher
-priority and the same keys will be overridden, and different ones will be merged into
-a single array of environment variable parameter values.
-
 ## Limiting the number of min and max instances in clusters
 
 As a preventive measure, one can restrict the minimum and the maximum number of
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -706,6 +706,35 @@ data:
 The key-value pairs of the Secret are all accessible as environment variables
 to the Postgres StatefulSet/pods.
 
+### For individual cluster
+
+It is possible to set environment variables in the Spilo pods directly as parameters or
+as a link to secrets for each individual cluster, in addition to or instead of the
+global parameters given in the examples above.
+
+In order to do this, you need to configure the cluster parameters, as in the example below
+
+```yaml
+apiVersion: "acid.zalan.do/v1"
+kind: postgresql
+metadata:
+  name: acid-test-cluster
+spec:
+  env:
+  - name: wal_s3_bucket
+    value: my-custom-bucket
+  - name: minio_secret_key
+      valueFrom:
+        secretKeyRef:
+          name: my-custom-secret
+          key: minio_secret_key
+```
+
+If global parameters are set at the level of the entire controller in the form of
+a ConfigMap or a Secret, then the individual cluster parameters will have a higher
+priority and the same keys will be overridden, and different ones will be merged into
+a single array of environment variable parameter values.
+
 ## Limiting the number of min and max instances in clusters
 
 As a preventive measure, one can restrict the minimum and the maximum number of
--- a/manifests/postgresql.crd.yaml
+++ b/manifests/postgresql.crd.yaml
@@ -194,6 +194,12 @@ spec:
                 type: boolean
               enableShmVolume:
                 type: boolean
+              env:
+                type: array
+                nullable: true
+                items:
+                  type: object
+                  x-kubernetes-preserve-unknown-fields: true
               init_containers:
                 type: array
                 description: deprecated
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -769,7 +769,13 @@ func (c *Cluster) generateSpiloPodEnvVars(
 	cloneDescription *acidv1.CloneDescription,
 	standbyDescription *acidv1.StandbyDescription,
 	customPodEnvVarsList []v1.EnvVar) []v1.EnvVar {
-	envVars := []v1.EnvVar{
+	envVars := make([]v1.EnvVar, 0)
+
+	if len(c.Spec.Env) > 0 {
+		envVars = append(envVars, c.Spec.Env...)
+	}
+
+	envVars = append(envVars, []v1.EnvVar{
 		{
 			Name:  "SCOPE",
 			Value: c.Name,
@@ -842,7 +848,8 @@ func (c *Cluster) generateSpiloPodEnvVars(
 			Name:  "HUMAN_ROLE",
 			Value: c.OpConfig.PamRoleName,
 		},
-	}
+	}...)
+
 	if c.OpConfig.EnablePgVersionEnvVar {
 		envVars = append(envVars, v1.EnvVar{Name: "PGVERSION", Value: c.GetDesiredMajorVersion()})
 	}
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -844,6 +844,10 @@ func (c *Cluster) generateSpiloPodEnvVars(
 		},
 	}
 
+	if c.OpConfig.EnableSpiloWalPathCompat {
+		envVars = append(envVars, v1.EnvVar{Name: "ENABLE_WAL_PATH_COMPAT", Value: "true"})
+	}
+
 	if c.OpConfig.EnablePgVersionEnvVar {
 		envVars = append(envVars, v1.EnvVar{Name: "PGVERSION", Value: c.GetDesiredMajorVersion()})
 	}
@@ -1100,16 +1104,6 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
 		initContainers = spec.InitContainers
 	}
 
-	spiloCompathWalPathList := make([]v1.EnvVar, 0)
-	if c.OpConfig.EnableSpiloWalPathCompat {
-		spiloCompathWalPathList = append(spiloCompathWalPathList,
-			v1.EnvVar{
-				Name:  "ENABLE_WAL_PATH_COMPAT",
-				Value: "true",
-			},
-		)
-	}
-
 	// fetch env vars from custom ConfigMap
 	configMapEnvVarsList, err := c.getPodEnvironmentConfigMapVariables()
 	if err != nil {
@@ -1123,8 +1117,7 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
 	}
 
 	// concat all custom pod env vars and sort them
-	customPodEnvVarsList := append(spiloCompathWalPathList, configMapEnvVarsList...)
-	customPodEnvVarsList = append(customPodEnvVarsList, secretEnvVarsList...)
+	customPodEnvVarsList := append(configMapEnvVarsList, secretEnvVarsList...)
 	sort.Slice(customPodEnvVarsList,
 		func(i, j int) bool { return customPodEnvVarsList[i].Name < customPodEnvVarsList[j].Name })
 
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -769,13 +769,8 @@ func (c *Cluster) generateSpiloPodEnvVars(
 	cloneDescription *acidv1.CloneDescription,
 	standbyDescription *acidv1.StandbyDescription,
 	customPodEnvVarsList []v1.EnvVar) []v1.EnvVar {
-	envVars := make([]v1.EnvVar, 0)
 
-	if len(c.Spec.Env) > 0 {
-		envVars = append(envVars, c.Spec.Env...)
-	}
-
-	envVars = append(envVars, []v1.EnvVar{
+	envVars := []v1.EnvVar{
 		{
 			Name:  "SCOPE",
 			Value: c.Name,
@@ -848,7 +843,7 @@ func (c *Cluster) generateSpiloPodEnvVars(
 			Name:  "HUMAN_ROLE",
 			Value: c.OpConfig.PamRoleName,
 		},
-	}...)
+	}
 
 	if c.OpConfig.EnablePgVersionEnvVar {
 		envVars = append(envVars, v1.EnvVar{Name: "PGVERSION", Value: c.GetDesiredMajorVersion()})
@@ -881,41 +876,64 @@ func (c *Cluster) generateSpiloPodEnvVars(
 		envVars = append(envVars, c.generateStandbyEnvironment(standbyDescription)...)
 	}
 
+	if len(c.Spec.Env) > 0 {
+		envVars = appendEnvVars(envVars, c.Spec.Env...)
+	}
+
 	// add vars taken from pod_environment_configmap and pod_environment_secret first
 	// (to allow them to override the globals set in the operator config)
 	if len(customPodEnvVarsList) > 0 {
-		envVars = append(envVars, customPodEnvVarsList...)
+		envVars = appendEnvVars(envVars, customPodEnvVarsList...)
 	}
 
 	if c.OpConfig.WALES3Bucket != "" {
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_S3_BUCKET", Value: c.OpConfig.WALES3Bucket})
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_S3_BUCKET", Value: c.OpConfig.WALES3Bucket})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
 	}
 
 	if c.OpConfig.WALGSBucket != "" {
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_GS_BUCKET", Value: c.OpConfig.WALGSBucket})
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_GS_BUCKET", Value: c.OpConfig.WALGSBucket})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
 	}
 
 	if c.OpConfig.WALAZStorageAccount != "" {
-		envVars = append(envVars, v1.EnvVar{Name: "AZURE_STORAGE_ACCOUNT", Value: c.OpConfig.WALAZStorageAccount})
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
-		envVars = append(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "AZURE_STORAGE_ACCOUNT", Value: c.OpConfig.WALAZStorageAccount})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "WAL_BUCKET_SCOPE_PREFIX", Value: ""})
 	}
 
 	if c.OpConfig.GCPCredentials != "" {
-		envVars = append(envVars, v1.EnvVar{Name: "GOOGLE_APPLICATION_CREDENTIALS", Value: c.OpConfig.GCPCredentials})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "GOOGLE_APPLICATION_CREDENTIALS", Value: c.OpConfig.GCPCredentials})
 	}
 
 	if c.OpConfig.LogS3Bucket != "" {
-		envVars = append(envVars, v1.EnvVar{Name: "LOG_S3_BUCKET", Value: c.OpConfig.LogS3Bucket})
-		envVars = append(envVars, v1.EnvVar{Name: "LOG_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
-		envVars = append(envVars, v1.EnvVar{Name: "LOG_BUCKET_SCOPE_PREFIX", Value: ""})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "LOG_S3_BUCKET", Value: c.OpConfig.LogS3Bucket})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "LOG_BUCKET_SCOPE_SUFFIX", Value: getBucketScopeSuffix(string(uid))})
+		envVars = appendEnvVars(envVars, v1.EnvVar{Name: "LOG_BUCKET_SCOPE_PREFIX", Value: ""})
 	}
 
 	return envVars
+}
+
+func appendEnvVars(envs []v1.EnvVar, appEnv ...v1.EnvVar) []v1.EnvVar {
+	jenvs := envs
+	for _, env := range appEnv {
+		if !isEnvVarPresent(jenvs, env.Name) {
+			jenvs = append(jenvs, env)
+		}
+	}
+	return jenvs
+}
+
+func isEnvVarPresent(envs []v1.EnvVar, key string) bool {
+	for _, env := range envs {
+		if env.Name == key {
+			return true
+		}
+	}
+	return false
 }
 
 // deduplicateEnvVars makes sure there are no duplicate in the target envVar array. While Kubernetes already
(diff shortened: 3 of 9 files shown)


dmvolod:
@FxKu could you please to have a look at the fix approach. If it will be suitable, I will add docs and unit-test for it. Thanks in advance.

cdmikechen:
Is it possible to add only some environment variables to pod? for example
```yaml
spec:
  env:
    - name: ENV_1
      value: VALUE_1
```

dmvolod:
> In CRD, Is there possible to add only some environment variables to pod? for example

Not sure, that's good approach as will need to support both setup for predefined ConfigMap/Secret and current one. Common way much better and didn't overload CR if it's not needed

dmvolod:
@FxKu could you please to find a time and have a look at the fix approach. If it will be suitable, I will add docs and unit-test for it. Thanks in advance.

FxKu:
Hey @dmvolod, while it looks elegant at first, I don't think it should be solved like this because only one way is possible - global config map or the need to add config maps for each cluster. The latter will be tedious if you need cluster-specific variables only for a handful of clusters while the rest of your fleet should be configured in the same way. Better have env variables from cluster-specific config maps / secrets to overwrite global ones - like we do it with other options, too. Probably, partly overwriting (in case of same env variables) will be most convenient.

pkg/cluster/k8sres.go
```diff
@@ -941,6 +941,11 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
 	return secretPodEnvVarsList, nil
 }
 
+func (c *Cluster) podEnvironmentObjectNameTemplate(template string) string {
+	objectNameTemplate := config.StringTemplate(template)
+	return objectNameTemplate.Format("cluster", c.clusterName().Name, "tprkind", acidv1.PostgresCRDResourceKind, "tprgroup", acidzalando.GroupName)
```
FxKu, line 946: why should the template contain the CRD kind and group?

dmvolod: I got the same template as operator has for the secret names to be unique name across different types and be a separating. Probably, this options can be omitted if not needed. Also, we could probably add a `CONTROLLER_ID` to have unique name across clusters and controllers.

dmvolod:
> Hey @dmvolod, while it looks elegant at first, I don't think it should be solved like this because only one way is possible - global config map or the need to add config maps for each cluster. The latter will be tedious if you need cluster-specific variables only for a handful of clusters while the rest of your fleet should be configured in the same way. Better have env variables from cluster-specific config maps / secrets to overwrite global ones - like we do it with other options, too. Probably, partly overwriting (in case of same env variables) will be most convenient.

Thanks for feedback, @FxKu . I'm also agree with approach to have a ConfigMap/Secret parameter for each cluster individually if needed and present. If you agree with this approach I will add these options to the CRD and implement logic to overwrite global parameters if CR parameters are present and find in the cluster ConfigMap/Secret.

cdmikechen:
I‘m looking forward to the progress of this PR. 
I have similar needs at present. We need to deploy multiple Postgres to different namespaces and back up to different minio services.

FxKu:
@cdmikechen note that you can run multiple operators next to each other configured differently. They could either watch only one namespace or you define a [CONTROLLER_ID]( https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#operators-with-defined-ownership-of-certain-postgres-clusters) in deployment and manifests to specify ownership over clusters.

cdmikechen:
> @cdmikechen note that you can already run multiple operators next to each other configured differently. They could either watch only one namespace or you define a [CONTROLLER_ID](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#operators-with-defined-ownership-of-certain-postgres-clusters) in deployment and manifests to specify ownership over clusters.

@FxKu
Thanks~ It looks like a good solution for my needs. I used to deploy only one operator to manage all namespaces, which will really bring me some trouble. 
I'll try it later and see if there will be any other problems.

dmvolod:
@FxKu please have a look at the new approach with individual `[]EnvVar` option again and has a higher priority to the global cluster options. Thanks in advance.

gc-jro:
Hi @dmvolod, thanks for this PR.

Do you still plan to add the possibility to refer to a cluster specific secret for environment variables in the postgresql custom resource?

I’ve a slightly different set of requirements, where we maintain our cluster configuration in Git, but aren’t allowed to store confidential information (like wal-g backup settings) in plain text. To circumvent this, we use SealedSecrets [1] to store the settings encrypted in Git. In our k8s cluster the SealedSecrets get decrypted into normal secrets that can be used by the cluster. I’ve also the requirement that we’ve multiple postgresql clusters in the same namespace, and the need that they use different wal-g backup settings.

Therefore, it would be great if in addition to use individual `[]EnvVar` settings in the postgresql CR, there would be an option to specify an individual secret in the postgresql CR that would be used for populating environment variables.
If you aren’t planning to add this option that’s totally fine, then I’ll ask my boss to get some time to patch the functionality in myself.

[1] https://github.com/bitnami-labs/sealed-secrets

dmvolod:
> 

Hi @gc-jro !

Yes, sure. Could you please to have a look at the `EnvVar` struct which is including `EnvVarSource` field and supports referencing to the Secret.
https://github.com/kubernetes/api/blob/master/core/v1/types.go#L1977-L2015

What do you think about it?

gc-jro:
Ah I see. Thanks a lot for clearing that up for me. I really like it, it'll solve all my problems :smile:

I'll just have to define something like
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: wal-g-envs
  namespace: yyy
data:
  BACKUP_NUM_TO_RETAIN: xxx
  ...
---
kind: postgresql
metadata:
  name: wal01-postgres-cluster
  namespace: yyy
spec:
  teamId: "wal01"
  volume:
    size: 20Gi
  env:
    - name: BACKUP_NUM_TO_RETAIN
      valueFrom:
        secretKeyRef:
          key: BACKUP_NUM_TO_RETAIN
          name: wal-g-envs
  ...
```
to forward the values from the secret to the database. Very neat.

Thank you very much. I can't wait to give it a spin once its merged into master.

FxKu:
Thank @dmvolod ! Like the new approach much better. As you already pointed out, docs and unit test are the remaining bits to get this merged.

pkg/cluster/k8sres.go
```diff
@@ -1000,20 +1010,31 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
...
-				},
-			}})
+				}})
+		}
 	}
 
 	return secretPodEnvVarsList, nil
 }
 
+func isEnvVarExist(envs []v1.EnvVar, key string) bool {
```
sdudoladov, line 1029: Minor naming thing: `is` would read better with an adjective, for instance `isEnvVarPresent`

pkg/cluster/k8sres.go
```diff
@@ -969,6 +975,10 @@ func (c *Cluster) getPodEnvironmentConfigMapVariables() ([]v1.EnvVar, error) {
 func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
 	secretPodEnvVarsList := make([]v1.EnvVar, 0)
 
+	if len(c.Spec.Env) > 0 {
+		secretPodEnvVarsList = append(secretPodEnvVarsList, c.Spec.Env...)
+	}
+
 	if c.OpConfig.PodEnvironmentSecret == "" {
 		return secretPodEnvVarsList, nil
 	}
```
sdudoladov, line 984: What is the expected outcome of  this piece of code when one has in the manifest
```yaml
spec:
  env:
...
  - name: minio_secret_key
      valueFrom:
        secretKeyRef:
          name: my-custom-secret
          key: minio_secret_key
```
but ` c.OpConfig.PodEnvironmentSecret` is not set ?

dmvolod: This should return `[]v1.EnvVar` from the Postgresql CR. i.e. 
```
  - name: minio_secret_key
      valueFrom:
        secretKeyRef:
          name: my-custom-secret
          key: minio_secret_key
```

pkg/cluster/k8sres.go
dmvolod: Good point, thanks, will change

manifests/postgresql.crd.yaml
```diff
@@ -194,6 +194,12 @@ spec:
                 type: boolean
               enableShmVolume:
                 type: boolean
+              env:
+                type: array
+                nullable: true
+                items:
+                  type: object
+                  x-kubernetes-preserve-unknown-fields: true
```
FxKu, line 202: please, add it also in the [chart](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/crds/postgresqls.yaml).

docs/administrator.md
```diff
@@ -706,6 +706,35 @@ data:
 The key-value pairs of the Secret are all accessible as environment variables
 to the Postgres StatefulSet/pods.
 
+### For individual cluster
+
+It is possible to set environment variables in the Spilo pods directly as parameters or
+as a link to secrets for each individual cluster, in addition to or instead of the
+global parameters given in the examples above.
```
FxKu, line 713:
```suggestion
It is possible to define environment variables directly in the Postgres cluster
manifest to configure it individually. The variables must be listed under the
`env` section in the same way you would do for [containers](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/).
Global parameters served from a custom config map or secret will be overridden.
```

docs/administrator.md
```diff
@@ -706,6 +706,35 @@ data:
...
+
+In order to do this, you need to configure the cluster parameters, as in the example below
```
FxKu, line 715:
```suggestion
It is possible to define environment variables directly in the Postgres cluster
manifest to configure it individually. The variables must be listed under the
`env` section in the same way you would do for [containers](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/).
Global parameters served from a custom config map or secret will be overridden.
```

docs/administrator.md
```diff
@@ -706,6 +706,35 @@ data:
...
+      valueFrom:
+        secretKeyRef:
+          name: my-custom-secret
+          key: minio_secret_key
+```
+
+If global parameters are set at the level of the entire controller in the form of
+a ConfigMap or a Secret, then the individual cluster parameters will have a higher
+priority and the same keys will be overridden, and different ones will be merged into
+a single array of environment variable parameter values.
```
FxKu, line 736: I think, this paragraph is obsolete after my suggestion from above.

FxKu:
I feel like this needs more unit tests. As per [docs](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#custom-pod-environment-variables) we currently say:
"In general the Operator will give preference to the globally configured variables, to not have the custom ones interfere with core functionality. Variables with the 'WAL_' and 'LOG_' prefix can be overwritten though, to allow backup and log shipping to be specified differently."

Might be worth to update this paragraph, too. Here's the corresponding [code](https://github.com/zalando/postgres-operator/blob/0dc370f15d5fd348937dd4eb631852e55df33e57/pkg/cluster/k8sres.go#L877), but I find it also weird that we keep appending. There's no `isEnvVarPresent` check happening. So another good test case.

Overall probably a bit confusing. Variables under `spec.env` override the global `pod_environment_configmap/secret` and partly override the global config, but then some global variables take precedence over individual cluster variables.

manifests/postgresql.crd.yaml
dmvolod: Yes, thanks, will do

docs/administrator.md
dmvolod: Ok. good point, done

dmvolod:
> I feel like this needs more unit tests. As per [docs](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#custom-pod-environment-variables) we currently say: "In general the Operator will give preference to the globally configured variables, to not have the custom ones interfere with core functionality. Variables with the 'WAL_' and 'LOG_' prefix can be overwritten though, to allow backup and log shipping to be specified differently."
> 
> Might be worth to update this paragraph, too. Here's the corresponding [code](https://github.com/zalando/postgres-operator/blob/0dc370f15d5fd348937dd4eb631852e55df33e57/pkg/cluster/k8sres.go#L877), but I find it also weird that we keep appending. There's no `isEnvVarPresent` check happening. So another good test case.
> 
> Overall probably a bit confusing. Variables under `spec.env` override the global `pod_environment_configmap/secret` and partly override the global config, but then some global variables take precedence over individual cluster variables.

Yes, very good point. In this case, what do you think to move cluster `v1.EnvVar` logic later after the `getPodEnvironmentConfigMapVariables` and `getPodEnvironmentSecretVariables` calls and got a higher priority for the cluster parameters. Or we could keep logic "as is" and just adding more tests and descriptions?

dmvolod:
> I see that we even have a function to take care of duplicate env vars - so we might not even need the `isEnvVarPresent` check. I have [extended](https://gist.github.com/FxKu/0b4e0641f1c846a2a0edbc2c6ee6a0cc) `TestPodEnvironmentConfigMapVariables` with the behavior that reflects the docs. But's it's failing. Guess `deduplicateEnvVars` is doing something wrong then.

Thanks for pointing me. Will fix it and add an additional tests.

FxKu:
I got it working now. See my gist: https://gist.github.com/FxKu/0b4e0641f1c846a2a0edbc2c6ee6a0cc
Because I changed the expected list of []v1.EnvVar in `TestPodEnvironmentConfigMapVariables` the comparison with `getPodEnvironmentConfigMapVariables` result did not work anymore. But maybe it should be a test on its own together with fetching variables for the secrets and test clearly what overrides what (and also make it clearer in the docs, but I can help there).

dmvolod:
> I got it working now. See my gist: https://gist.github.com/FxKu/0b4e0641f1c846a2a0edbc2c6ee6a0cc
> Because I changed the expected list of []v1.EnvVar in `TestPodEnvironmentConfigMapVariables` the comparison with `getPodEnvironmentConfigMapVariables` result did not work anymore. But maybe it should be a test on its own together with fetching variables for the secrets and test clearly what overrides what (and also make it clearer in the docs, but I can help there).

Thanks for test, @FxKu 
Probably, we should move EnvVar logic to the `generateSpiloPodEnvVars` and utilize another tests for that. What do you think about it?

pkg/cluster/k8sres.go
```diff
@@ -939,6 +939,10 @@ func deduplicateEnvVars(input []v1.EnvVar, containerName string, logger *logrus.
 func (c *Cluster) getPodEnvironmentConfigMapVariables() ([]v1.EnvVar, error) {
 	configMapPodEnvVarsList := make([]v1.EnvVar, 0)
 
+	if len(c.Spec.Env) > 0 {
+		configMapPodEnvVarsList = append(configMapPodEnvVarsList, c.Spec.Env...)
+	}
```
FxKu, line 944: Lets to do this only once in `generateStatefulSet`. First get c.Spec.Env because it takes precedence over all other global configs. With the result from `getPodEnvironmentConfigMapVariables` and `getPodEnvironmentSecretVariables` we can do the `isEnvVarPresent` in  `generateStatefulSet`, too. This way, both functions will just do what their names tell us they are supposed to do.

pkg/cluster/k8sres.go
```diff
@@ -969,6 +975,10 @@ func (c *Cluster) getPodEnvironmentConfigMapVariables() ([]v1.EnvVar, error) {
 func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
 	secretPodEnvVarsList := make([]v1.EnvVar, 0)
 
+	if len(c.Spec.Env) > 0 {
+		secretPodEnvVarsList = append(secretPodEnvVarsList, c.Spec.Env...)
+	}
```
FxKu, line 980: see comment above. Maybe we should variables from Secrets should take precedence over variables from config map.

pkg/cluster/k8sres.go
```diff
@@ -960,7 +964,9 @@ func (c *Cluster) getPodEnvironmentConfigMapVariables() ([]v1.EnvVar, error) {
 		}
 	}
 	for k, v := range cm.Data {
-		configMapPodEnvVarsList = append(configMapPodEnvVarsList, v1.EnvVar{Name: k, Value: v})
+		if !isEnvVarPresent(configMapPodEnvVarsList, k) {
+			configMapPodEnvVarsList = append(configMapPodEnvVarsList, v1.EnvVar{Name: k, Value: v})
+		}
```
FxKu, line 969: With the result from `getPodEnvironmentConfigMapVariables` we can do the `isEnvVarPresent` in  `generateStatefulSet`, too. This way, the functions will just do what its name tells us it's supposed to do.

pkg/cluster/k8sres.go
```diff
@@ -1000,20 +1010,31 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
 	}
 
 	for k := range secret.Data {
-		secretPodEnvVarsList = append(secretPodEnvVarsList,
-			v1.EnvVar{Name: k, ValueFrom: &v1.EnvVarSource{
-				SecretKeyRef: &v1.SecretKeySelector{
-					LocalObjectReference: v1.LocalObjectReference{
-						Name: c.OpConfig.PodEnvironmentSecret,
+		if !isEnvVarPresent(secretPodEnvVarsList, k) {
```
FxKu, line 1013: Same here. Let's do this in `generateStatefulSet`

pkg/cluster/k8sres.go
dmvolod: Already did it, what do you think about `generateSpiloPodEnvVars` location?

pkg/cluster/k8sres.go
```diff
@@ -1000,20 +1010,31 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
...
 }
 
+func isEnvVarPresent(envs []v1.EnvVar, key string) bool {
+	for _, env := range envs {
+		if env.Name == key {
+			return true
+		}
+	}
+	return false
+}
```
FxKu, line 1036: Actually, it might be even better to do `isEnvVarPresent` in [generateSpiloPodEnvVars](https://github.com/zalando/postgres-operator/blob/43e18052c427e255afaa985a046130fc47cf9264/pkg/cluster/k8sres.go#L880), too, instead of appending duplicate variables and resolving it later with `deduplicateEnvVars`.

pkg/cluster/k8sres.go
FxKu: Yes, also a good idea, to move just everything from `generateStatefulSet` to `generateSpiloPodEnvVars`

pkg/cluster/k8sres.go
dmvolod: Fix logic, please review.

pkg/cluster/k8sres.go
dmvolod: Fix logic, please review.

pkg/cluster/k8sres.go
dmvolod: Fix logic, please review.

pkg/cluster/k8sres.go
dmvolod: Fix logic, please review.

pkg/cluster/k8sres.go
```diff
@@ -769,7 +769,13 @@ func (c *Cluster) generateSpiloPodEnvVars(
 	cloneDescription *acidv1.CloneDescription,
 	standbyDescription *acidv1.StandbyDescription,
 	customPodEnvVarsList []v1.EnvVar) []v1.EnvVar {
-	envVars := []v1.EnvVar{
+	envVars := make([]v1.EnvVar, 0)
+
+	if len(c.Spec.Env) > 0 {
+		envVars = append(envVars, c.Spec.Env...)
+	}
```
FxKu, line 776: Appending c.Spec.Env should happen after the hard-coded set of variables, because we don't want them to get overridden by custom variables.

pkg/cluster/k8sres_test.go
```diff
@@ -131,45 +131,52 @@ func TestGenerateSpiloPodEnvVars(t *testing.T) {
...
 			envVarValue:    "some_path_to_credentials",
 		},
 	}
+	expectedValuesS3Bucket := []ExpectedValue{
+		{
+			envIndex:       0,
+			envVarConstant: "WAL_S3_BUCKET",
+			envVarValue:    "custom-s3-bucket",
+		},
+	}
```
FxKu, line 179: custom variables should not be pre-pended so this addition is also obsolete.

pkg/cluster/k8sres_test.go
```diff
@@ -207,10 +215,57 @@ func TestGenerateSpiloPodEnvVars(t *testing.T) {
...
+				Spec: acidv1.PostgresSpec{
+					Env: []v1.EnvVar{
+						{
+							Name:  "WAL_S3_BUCKET",
+							Value: "custom-s3-bucket",
+						},
+					},
+				},
+			},
+		},
```
FxKu, line 263: The test should incl. 3 kinds of variables - like in my gist.
1. A custom variable called like a hard-coded variable that cannot be overridden (I uses `KUBERNETES_SCOPE_LABEL`)
2. A custom variable called like a hard-coded variable that can be overridden (like WAL_S3_BUCKET)
3. And a custom variable that's not in the list of hard-coded variables. And there we can test, if a variable from c.Spec.Env overwrites on from environment ConfigMap and/or Secret.

dmvolod: Good point, thanks will implement this options.

pkg/cluster/k8sres.go
dmvolod: Ok, good point, thanks. WIll implement

FxKu:
Ok aside from missing unit tests there's one problem that has to be fixed. Right now we only append env vars and rely on `deduplicateEnvVars` to come up with the right set in the end. That is, it will take only the first mention of a variable into account except when the variable starts with `WAL_` or `LOG_`.

This logic will not work anymore now that we have cluster-specific `Env`.
1. It MUST NOT override our hard-coded set of variables from `generateSpiloPodEnvVars` up until this [line](https://github.com/zalando/postgres-operator/blob/43e18052c427e255afaa985a046130fc47cf9264/pkg/cluster/k8sres.go#L868). Only append after that.
2. MUST override the PodEnvironmentConfigMap and/or Secret. So we append before that. But then `WAL_` or `LOG_` settings from PodEnvironmentConfigMap and/or Secret would override `c.Spec.Env` because of the logic in `deduplicateEnvVars`. 

That's why I suggested that we should not simply append to the list of envVars but only do so if the key does not exists yet. So basically what you partly did with `isEnvVarPresent`. But instead of adding this check everywhere, we better write an append function for EnvVar lists that incl. this check.

dmvolod:
> That's why I suggested that we should not simply append to the list of envVars but only do so if the key does not exists yet. So basically what you partly did with `isEnvVarPresent`. But instead of adding this check everywhere, we better write an append function for EnvVar lists that incl. this check. Finally, `deduplicateEnvVars` can be removed.

Does it means that we need to utilize new `appendEnvVars` function everywhere inside the `generateSpiloPodEnvVars`?

FxKu:
Yes, inside `generateSpiloPodEnvVars`. The flow should be:

1. envVars := []v1.EnvVar{...}
2. append spiloCompathWalPathList
3. append c.Spec.Env
4. fetch from getPodEnvironmentSecretVariables and append
5. fetch from getPodEnvironmentConfigMapVariables and append
6. fetch from generateCloneEnvironment and append
7. fetch from generateStandbyEnvironment and append
8. append WAL_ LOG_ AZURE_ etc. variables

I don't see why anything related to spilo env vars should happen outside of `generateSpiloPodEnvVars`. Noticed that `deduplicateEnvVars` is also used for sidecars. So maybe it can stay then.

pkg/cluster/k8sres_test.go
dmvolod: Implemented tests, please review.

dmvolod:
> Yes, inside `generateSpiloPodEnvVars`. The flow should be:
> 
> 1. envVars := []v1.EnvVar{...}
> 2. append spiloCompathWalPathList
> 3. append c.Spec.Env
> 4. fetch from getPodEnvironmentSecretVariables and append
> 5. fetch from getPodEnvironmentConfigMapVariables and append
> 6. fetch from generateCloneEnvironment and append
> 7. fetch from generateStandbyEnvironment and append
> 8. append WAL_ LOG_ AZURE_ etc. variables
> 
> I don't see why anything related to spilo env vars should happen outside of `generateSpiloPodEnvVars`. Noticed that `deduplicateEnvVars` is also used for sidecars. So maybe it can stay then.

Thanks for idea, please have a look at the code. Added functions and changed from plain `append` to the new one.

pkg/cluster/k8sres_test.go
dmvolod: This is approach from the following point
- A custom variable called like a hard-coded variable that can be overridden (like WAL_S3_BUCKET)

FxKu:
:+1:

sdudoladov:
:+1:

FxKu:
Thanks @dmvolod for your contribution. Good collaboration!

dmvolod:
Thanks @FxKu and @sdudoladov for review and ideas.

PR #1796, opened 2022-02-23
minor fixes to password rotation

FxKu:
1. When the next rotation date is set the period should be added to the current time, not to the time in the secret. In case the timestamp in the secret would be very old, it would need a few unnecessary rotations until we reach the current time.
2. The timestamp in the secret should also store the timestamp. We can just use Go's time.RFC3339 constant and do not worry about the format. As location we hardcode to UTC.
3. The complete manifest was giving wrong examples for password rotation options.


sdudoladov:
:+1:

jopadi:
👍

FxKu:
:+1:

PR #1797, opened 2022-02-26
Fixing toleration fields requirements

preved911:
Hello. Tolerations subfileds aren't required in all cases. I think that required subfields list should be removed from crd.

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -495,10 +495,6 @@ spec:
                 type: array
                 items:
                   type: object
-                  required:
-                    - key
-                    - operator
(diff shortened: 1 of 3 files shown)


FxKu:
There is a third place you would have to remove it: https://github.com/zalando/postgres-operator/blob/fb8a6c7a6823bb143bbfd6b5ec2e75d8de4b40d9/pkg/apis/acid.zalan.do/v1/crds.go#L796

preved911:
> There is a third place you would have to remove it

thanks, removed

FxKu:
:+1:

jopadi:
👍

FxKu:
Thanks @preved911 for your contribution.

PR #1799, opened 2022-02-28
LoadBalancer toggles for master and replica pooler pods

### LINKED ISSUE #1774: annotation change is not taken by connectionPooler pod

- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** GCP kubernetes
- **Type of issue?** [Bug report]

Hello

We are using postgres operator with pgbouncer activated, Pooler actually get the same annotation as postgres pod. 

When we update the podAnnotations part of the postgres ressources the pooler is not update unless we desactivate it and recreate it.

```
  enableConnectionPooler: true
  enableReplicaConnectionPooler: true
```

it's a big problem as if we disable and reenable our pooler it's make unacceptable downtime 

Thanks for your help

FxKu:
Continuation from #943.
Closes #943
Fixes #1774

Adding LB support for pooler pods. Required implementing update logic for pooler services.


pkg/cluster/resources.go
```diff
@@ -275,19 +275,15 @@ func (c *Cluster) createService(role PostgresRole) (*v1.Service, error) {
...
-func (c *Cluster) updateService(role PostgresRole, newService *v1.Service) error {
+func (c *Cluster) updateService(role PostgresRole, oldService *v1.Service, newService *v1.Service) (*v1.Service, error) {
 	var (
 		svc *v1.Service
 		err error
 	)
 
 	c.setProcessName("updating %v service", role)
 
-	if c.Services[role] == nil {
```
sdudoladov, line 286: why is this check removed ?

pkg/cluster/k8sres.go
```diff
@@ -99,15 +99,15 @@ func (c *Cluster) serviceAddress(role PostgresRole) string {
 	return ""
 }
 
-func (c *Cluster) servicePort(role PostgresRole) string {
+func (c *Cluster) servicePort(role PostgresRole) int32 {
```
sdudoladov, line 102: that is better renamed `generateServicePort` - otherwise the logic of returning `5432` when a service does not exist is vauge

pkg/cluster/k8sres.go
```diff
@@ -99,15 +99,15 @@ func (c *Cluster) serviceAddress(role PostgresRole) string {
...
 
 	if exist {
-		return fmt.Sprint(service.Spec.Ports[0].Port)
+		return service.Spec.Ports[0].Port
 	}
 
-	c.logger.Warningf("No service for role %s", role)
-	return ""
+	c.logger.Warningf("No service for role %s - defaulting to port 5432", role)
+	return 5432
```
sdudoladov, line 110: [that is a constant](https://github.com/zalando/postgres-operator/blob/695ad44caffa7c03e19a1b59796d93942d8cb70c/pkg/cluster/k8sres.go#L41)

pkg/cluster/k8sres.go
FxKu: Yes something like `getServicePort` would be better. But, now it's also in line with the other functions like `serviceName`, `serviceAddress`. I think, I would keep the name.

jopadi:
👍

FxKu:
:+1:

PR #1805, opened 2022-03-02
grant db owners to cron_admin

FxKu:
Recently we introduced the `cron_admin` role in Spilo. The idea is to make it possible for database owners to set up cron jobs with manual management of privileges. The default is empty, so behavior stays the same. `cron_admin` is also added to protected role names.

This PR also adds forgotten options about password rotation to crds.go and values.yaml file.


Jan-M:
:+1:

FxKu:
:+1:

PR #1809, opened 2022-03-08
Setting up a standby cluster with AWS

DadiAnas:
In case you are using AWS, please make sure that the AWS user you provided its credentials with `STANDBY_AWS_ACCESS_KEY_ID`, `STANDBY_AWS_SECRET_ACCESS_KEY` and `STANDBY_AWS_REGION` has permissions to deal with s3, and that you have created a role with following policies:

Policy:

```json
{
    "Statement": [
        {
            "Action": [
                "ec2:Describe*",
                "ec2:Describe*",
                "ec2:ModifyVolumeAttribute"
            ],
            "Effect": "Allow",
            "Resource": "*"
        }
    ],
    "Version": "2012-10-17"
}
```

Role:

```json
{
    Version = "2012-10-17"
    Statement = [
      {
        Action = "sts:AssumeRole"
        Effect = "Allow"
        Sid    = ""
        Principal = {
          Service = "ec2.amazonaws.com"
        }
      },
    ]
  }
```

Then, add the name of the role name to `kube_iam_role` [param](https://github.com/zalando/postgres-operator/blob/c10d30903e049bc75ce29e0a9342ff45434deeb5/manifests/configmap.yaml#L52) or the annotation `iam.amazonaws.com/role: "postgres-operator-role"` to the operator deployment.

### DIFF
--- a/docs/user.md
+++ b/docs/user.md
@@ -818,6 +818,48 @@ spec:
     gs_wal_path: "gs://<bucketname>/spilo/<source_db_cluster>/<UID>/wal/<PGVERSION>"
 ```
 
+In case you are using AWS, please make sure that the AWS user you provided its credentials with `STANDBY_AWS_ACCESS_KEY_ID`, `STANDBY_AWS_SECRET_ACCESS_KEY` and `STANDBY_AWS_REGION` has permissions to deal with s3, and that you have created a role with following policies:
+
+Policy:
+
+```json
+{
+    "Statement": [
+        {
+            "Action": [
+                "ec2:Describe*",
+                "ec2:Describe*",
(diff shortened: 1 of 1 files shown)


FxKu:
Always open to improve docs. This would be better suited for the admin docs, I believe. Can you check [this paragraph](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#using-aws-s3-or-compliant-services) if something can be added there? I would prefer yaml over json examples. We do not list the policy yet.

For the user docs it might make sense to add just one sentence that cloning / standby from s3 only works if continuous archiving is set up correctly.

PR #1811, opened 2022-03-14
Pooler sync fix

FxKu:
This PR goes back to always sync pooler components, since updates on services are now supported with #1799. 

Error handling in `syncConnectionPooler` is wrong. The GET request to find the pooler service [defines err](https://github.com/zalando/postgres-operator/blob/69254abeba0b9991bb765c2673f87a09a5f8dc5d/pkg/cluster/connection_pooler.go#L930) for the if block, but the latter `if !k8sutil.ResourceNotFound(err)` check is using the last state of `err` from the surrounding scope. In case the postgresql resource has annotations defined, `err` likely has the value `<nil>` from [updating the pooler deployment](https://github.com/zalando/postgres-operator/blob/69254abeba0b9991bb765c2673f87a09a5f8dc5d/pkg/cluster/connection_pooler.go#L923). If not, `err` would still be `NotFound` from the [deployment check](https://github.com/zalando/postgres-operator/blob/69254abeba0b9991bb765c2673f87a09a5f8dc5d/pkg/cluster/connection_pooler.go#L842), hence `if !k8sutil.ResourceNotFound(err)` check works as expected.

The PR fixes another minor issue: A standby cluster could not have a connection pooler enabled because the operator would try to install the pooler user an schema in the DB cluster, although they should already be present (if the source clusters uses the pooler, too). Now, pooler installation is skipped for standby clusters.


FxKu:
:+1:

PR #1817, opened 2022-03-17
fix comparison of event stream array

FxKu:
streams cannot simply be compared with DeepEqual since the order of event streams returned by the K8s API is not deterministic. This PR provides a new sameStreams function to avoid updating FES resources on potentially every sync.

Another fix is applied, if some fields are omitted in the Postgres manifest like `idColumn`, `payloadColumn` or `maxBatchSize`, these fields should be nil in the FES resource. Therefore, they are turned into pointers.


pkg/util/k8sutil/k8sutil.go
```diff
@@ -37,6 +37,14 @@ func Int32ToPointer(value int32) *int32 {
 	return &value
 }
 
+func UInt32ToPointer(value uint32) *uint32 {
+	return &value
+}
+
+func StringToPointer(str string) *string {
```
dmvolod, line 44: Please note, that `k8s.io/utils/pointer/pointer.go` package already contains these kind of functions.

FxKu: Ah good to know :smiley:

FxKu:
:+1:

PR #1819, opened 2022-03-18
fix unit test and improve stability in e2e test

FxKu:
TestInitAdditionalOwnerRoles needed to be fixed because DeepEqual does not work properly when there is a nested slice where order can change. Now, only compare MemberOf field of pgUser.

For e2e test, remove waiting for operator to get in sync before waiting for a failover. I have seen multiple times in logs of failed e2e tests that failover happend and only then the operator was in sync. Thus, waiting for another failover never finishes. Sometimes in works, though.


jopadi:
👍

FxKu:
:+1:

PR #1822, opened 2022-03-23
refactor GenerateResourceRequirements and provide unit tests

FxKu:
This PR does the following:
- fixing a [bug](https://github.com/zalando/postgres-operator/blob/38db48c7f085be6834a626d496d5bdece62f8a9d/pkg/cluster/k8sres.go#L1017) introduced by #1819 
- lift the requirement for specifying full resources section
- provide unit test for `GenerateResourceRequirements`
- set request to limits inside `GenerateResourceRequirements` with an extra function
- for `enforceMinResourceLimits`
  - move it to `k8sres.go` next to other resource functions
  - call it only from `GenerateResourceRequirements`
  - takes up `v1.ResourceRequirements` instead of PostgresSpec


pkg/cluster/util.go
```diff
@@ -594,3 +594,12 @@ func trimCronjobName(name string) string {
 	}
 	return name
 }
+
+func parseResourceRequirements(resourcesRequirement v1.ResourceRequirements) (acidv1.Resources, error) {
+	var resources acidv1.Resources
+	resourcesJSON, _ := json.Marshal(resourcesRequirement)
```
sdudoladov, line 600: so the marshalling error is ignored but the unmarshalling on the line below is not ?

pkg/cluster/k8sres.go
```diff
@@ -194,6 +223,48 @@ func fillResourceList(spec acidv1.ResourceDescription, defaults acidv1.ResourceD
...
+
+	var specRequests, specLimits acidv1.ResourceDescription
+
+	if resources == nil {
+		specRequests = acidv1.ResourceDescription{}
+		specLimits = acidv1.ResourceDescription{}
+	} else {
+		specRequests = resources.ResourceRequests
+		specLimits = resources.ResourceLimits
+	}
```
sdudoladov, line 240: that can be slightly shorter, sth along:
```golang
specRequests := acidv1.ResourceDescription{}
specLimits := acidv1.ResourceDescription{}

if resources != nil {
		specRequests = resources.ResourceRequests
		specLimits = resources.ResourceLimits
}

```

PR #1825, opened 2022-03-28
define readinessProbe on statefulSet

FxKu:
With #1760 merged, we should add a readinessProbe to the statefultSet because the ConfigMaps option defines a selector on the master service. Label selectors incl. readiness probes. We will add it either way.

#1760 also introduced a new suffix for deleting Patroni cluster objects which will result in an error for the new "leader" suffix if endpoints are used. Instead, the operator should just l log a warning and try to delete as much as it can. Another bug was introduced by not deleting all services if endpoints are used.

Other changes:
- user variables when specifying container ports
- check configmaps option when syncing ClusterStatus not omit endpoints
- remove endpoints from RBAC (provide RBAC manifest for OpenShift)

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -76,20 +76,13 @@ func (c *Cluster) statefulSetName() string {
 	return c.Name
 }
 
-func (c *Cluster) configMapName(role PostgresRole) string {
+func (c *Cluster) endpointName(role PostgresRole) string {
 	name := c.Name
-	if role == Master {
-		name = name + "-leader"
-	}
 	if role == Replica {
 		name = name + "-repl"
 	}
 
 	return name
-}
-
-func (c *Cluster) endpointName(role PostgresRole) string {
-	return c.serviceName(role)
 }
 
 func (c *Cluster) serviceName(role PostgresRole) string {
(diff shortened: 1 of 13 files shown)


pkg/cluster/k8sres.go
```diff
@@ -1058,6 +1059,22 @@ func extractPgVersionFromBinPath(binPath string, template string) (string, error
 	return fmt.Sprintf("%v", pgVersion), nil
 }
 
+func generateSpiloReadinessProbe() *v1.Probe {
+	return &v1.Probe{
+		Handler: v1.Handler{
+			HTTPGet: &v1.HTTPGetAction{
+				Path: "/readiness",
+				Port: intstr.IntOrString{IntVal: patroni.ApiPort},
```
beneso, line 1067: Hi, thanks for the commit!
I have a question: are you sure of the port? Container is never ready with this probe, HTTP 503. While if we try the same path with port 8080, we get HTTP 200.

~~~
root@acid-grafana-postgres-0:/home/postgres# curl -vvvv $POD_IP:8008/readiness
*   Trying 10.0.2.66...
* TCP_NODELAY set
* Connected to 10.0.2.66 (10.0.2.66) port 8008 (#0)
> GET /readiness HTTP/1.1
> Host: 10.0.2.66:8008
> User-Agent: curl/7.58.0
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 503 Service Unavailable
< Server: BaseHTTP/0.6 Python/3.6.7
< Date: Tue, 31 May 2022 15:43:49 GMT
< Content-Type: application/json
< 
* Closing connection 0
<json>
~~~

vs port 8080:

~~~
root@acid-grafana-postgres-0:/home/postgres# curl -vvvv $POD_IP:8080/readiness
*   Trying 10.0.2.66...
* TCP_NODELAY set
* Connected to 10.0.2.66 (10.0.2.66) port 8080 (#0)
> GET /readiness HTTP/1.1
> Host: 10.0.2.66:8080
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Tue, 31 May 2022 15:43:44 GMT
< Content-Length: 3375
<json>
~~~

Thanks!

PR #1828, opened 2022-03-28
HPA integration (rebase included)

AnisTigrini:
This is the Horizontal Pod Autoscaler feature Including the rebase.

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -447,6 +447,92 @@ spec:
                 items:
                   type: object
                   x-kubernetes-preserve-unknown-fields: true
+              horizontalPodAutoscalerV1:
+                type: object
+                required:
+                  - kind
+                  - apiVersion
+                  - spec
+                properties:
+                  kind:
+                    type: string
+                    enum:
+                      - HorizontalPodAutoscaler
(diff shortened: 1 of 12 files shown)


AnisTigrini:
Hi there, I decided to delete the previous PR related to the HPA integration since it was based on an old version of Zalando and created this PR instead which is much cleaner.

So the goal of the Project was to have an Horizontal Pod Autoscaled which would be controlled directly by the Postgres Operator. 
With this PR, you should be able to create/update/delete a Horizontal Pod Autoscaler directly from the Custom Resource Definition file you submit to create PostgreSQL pods. 

When it comes to the HPA, I split them in HPAV1 and HPAV2 since the APIs are a bit different. This may seem a bit ugly at first, but I am open to changing my code if you guys see more generic ways of doing things. 

I tried as much as possible to stay generic and avoid hard coding things. But the HPA should be used to target the StatefulSet created by Zalando's CRD.

AnisTigrini:
Hello guys hope you are well, just following up on this

gesanderson:
Hi guys, I just wanted to let you know that we're also interested by this feature and are looking forward to using it !

paretl:
Hey, I'm very interested by this feature as well. It could be very helpful to manage the number of pods used for our replicas

PR #1831, opened 2022-03-30
Fix deletion of event streams resources

FxKu:
The delete code was still reflecting an earlier state of #1570 where only one event stream resource was possible per cluster. In fact, there can now exists multiple as per application ID. This PR introduces a cluster field to store those IDs to loop over them after cluster deletion.


jopadi:
👍

FxKu:
:+1:

PR #1837, opened 2022-04-05
bump pooler image with pgBouncer 1.17.0 and auth_type md5

### LINKED ISSUE #1750: Connection pooler does not support SCRAM-SHA authentication
After turning on SCRAM-SHA authentication and letting postgres-operator re-sync the password, our applications are unable to connected to the connection pooler. At the same time, this error appears in the connection pooler log: 

```
2022-01-21 06:53:19.947 UTC [8] ERROR S-0x558ea0528100: xxx/xxx@10.11.209.70:5432 cannot do SCRAM authentication: password is SCRAM secret but client authentication did not provide SCRAM keys
```

We suspect this is due to the `auth_type` in pgbouncer.ini is hard-coded to `plain`, so the client applications don't sent the necessary information to pgbouncer.

  - registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
  - registry.opensource.zalan.do/acid/pgbouncer:master-19
  - Bare Metal K8s
  - Yes
- **Type of issue?** 
  - Bug report

FxKu:
Fixes #1750 

We changed auth_type in our pgBouncer image from `plain` to `md5`. According to pgBouncer [docs](https://www.pgbouncer.org/config.html#authentication-settings) if "md5 is configured and a user has a SCRAM secret, then SCRAM authentication is used automatically instead."

Added a paragraph how one can configure using secrets with `scram-sha-256` hashing.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -588,7 +588,7 @@ spec:
                     default: "pooler"
                   connection_pooler_image:
                     type: string
-                    default: "registry.opensource.zalan.do/acid/pgbouncer:master-19"
+                    default: "registry.opensource.zalan.do/acid/pgbouncer:master-22"
                   connection_pooler_max_db_connections:
                     type: integer
                     default: 60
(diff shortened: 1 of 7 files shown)


FxKu:
:+1:

FxKu:
:+1:

sdudoladov:
:+1:

PR #1848, opened 2022-04-12
refactor spilo env var generation

FxKu:
Do everything related to env var of DB pods inside `GenerateSpiloPodEnvVars` and extend unit tests. Plus, I shuffled around unit tests in k8sres_test so that tests covering the same topics are next to each other.

incl. fix for setting cron_admin when there's only one owner to grant it to

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -603,9 +603,9 @@ spec:
 
 The operator will assign a set of environment variables to the database pods
 that cannot be overridden to guarantee core functionality. Only variables with
-'WAL_' and 'LOG_' prefixes can be customized, to allow backup and log shipping
-to be specified differently. There are three ways to specify extra environment
-variables (or override existing ones) for database pods:
+'WAL_' and 'LOG_' prefixes can be customized to allow for backup and log
+shipping to be specified differently. There are three ways to specify extra
+environment variables (or override existing ones) for database pods:
 
 * [Via ConfigMap](#via-configmap)
 * [Via Secret](#via-secret)
@@ -975,7 +975,7 @@ generated automatically. `WALG_S3_PREFIX` is identical to `WALE_S3_PREFIX`.
 `SCOPE` is the Postgres cluster name.
 
 :warning: If both `AWS_REGION` and `AWS_ENDPOINT` or `WALE_S3_ENDPOINT` are
-defined, backups with WAL-E will fail. You can fix it by switching to WAL-G
+defined backups with WAL-E will fail. You can fix it by switching to WAL-G
 with `USE_WALG_BACKUP: "true"`.
 
 ### Google Cloud Platform setup
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -601,15 +601,38 @@ spec:
 
 ## Custom Pod Environment Variables
 
-It is possible to configure a ConfigMap as well as a Secret which are used by
-the Postgres pods as an additional provider for environment variables. One use
-case is a customized Spilo image configured by extra environment variables.
-Another case could be to provide custom cloud provider or backup settings.
+The operator will assign a set of environment variables to the database pods
+that cannot be overridden to guarantee core functionality. Only variables with
+'WAL_' and 'LOG_' prefixes can be customized, to allow backup and log shipping
+to be specified differently. There are three ways to specify extra environment
+variables (or override existing ones) for database pods:
 
-In general the Operator will give preference to the globally configured
-variables, to not have the custom ones interfere with core functionality.
-Variables with the 'WAL_' and 'LOG_' prefix can be overwritten though, to
-allow backup and log shipping to be specified differently.
+* [Via ConfigMap](#via-configmap)
+* [Via Secret](#via-secret)
+* [Via Postgres Cluster Manifest](#via-postgres-cluster-manifest)
+
+The first two options must be referenced from the operator configuration
+making them global settings for all Postgres cluster the operator watches.
+One use case is a customized Spilo image that must be configured by extra
+environment variables. Another case could be to provide custom cloud
+provider or backup settings.
+
+The last options allows for specifying environment variables individual to
+every cluster via the `env` section in the manifest. For example, if you use
+individual backup locations for each of your clusters. Or you want to disable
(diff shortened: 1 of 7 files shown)


pkg/cluster/k8sres.go
```diff
@@ -960,12 +977,14 @@ func (c *Cluster) getPodEnvironmentConfigMapVariables() ([]v1.EnvVar, error) {
 				metav1.GetOptions{})
 		}
 		if err != nil {
-			return nil, fmt.Errorf("could not read PodEnvironmentConfigMap: %v", err)
+			return configMapPodEnvVarsList, fmt.Errorf("could not read PodEnvironmentConfigMap: %v", err)
```
FxKu, line 980: return empty list instead of nil so that appending works nicely in generateSpiloEnvVars

pkg/cluster/k8sres.go
```diff
@@ -1945,30 +1927,20 @@ func (c *Cluster) generateStandbyEnvironment(description *acidv1.StandbyDescript
...
-				{
-					Name:  "STANDBY_WALE_GS_PREFIX",
-					Value: description.GSWalPath,
-				},
-				{
-					Name:  "STANDBY_GOOGLE_APPLICATION_CREDENTIALS",
-					Value: c.OpConfig.GCPCredentials,
-				},
-			}
-			result = append(result, envs...)
```
FxKu, line 1971: We cannot do similar to what we do in cloning. There users have the choice to overwrite the WAL location with S3WalPath but can stick to operator config to generate the path. However, in standby there's only one way possible: Explicit mention of WAL path. Therefore, we should not use `c.OpConfig.GCPCredentials` here but must rely on the user to define it. This way, `GCPCredentials` from config could also be overridden which is otherwise not possible.

pkg/cluster/k8sres.go
```diff
@@ -1104,23 +1124,6 @@ func (c *Cluster) generateStatefulSet(spec *acidv1.PostgresSpec) (*appsv1.Statef
...
-	secretEnvVarsList, err := c.getPodEnvironmentSecretVariables()
-	if err != nil {
-		return nil, err
-	}
-
-	// concat all custom pod env vars and sort them
-	customPodEnvVarsList := append(configMapEnvVarsList, secretEnvVarsList...)
-	sort.Slice(customPodEnvVarsList,
-		func(i, j int) bool { return customPodEnvVarsList[i].Name < customPodEnvVarsList[j].Name })
-
```
FxKu, line 1123: moved inside of `generateSpiloPodEnvVars`. Plus, if fetching configMap or secret fails there will  be a warning but no error returned

docs/administrator.md
```diff
@@ -601,15 +601,38 @@ spec:
...
+environment variables. Another case could be to provide custom cloud
+provider or backup settings.
+
+The last options allows for specifying environment variables individual to
+every cluster via the `env` section in the manifest. For example, if you use
+individual backup locations for each of your clusters. Or you want to disable
+WAL archiving for a certain cluster by setting `WAL_S3_BUCKET`, `WAL_GS_BUCKET`
+or `AZURE_STORAGE_ACCOUNT` to an empty string.
+
+The operator will give precendence to environment variables in the following
```
sdudoladov, line 626: 1. typo: `preceNdence` -> `precedence`
2. From this sentence the exact ordering is imo unclear. It should mention that 1 has the lowest priority and an env var with the same name defined at each subsequent level will override the value from the previous level.

docs/user.md
```diff
@@ -784,6 +784,14 @@ metadata:
   uid: efd12e58-5786-11e8-b5a7-06148230260c
 ```
 
+If your source cluster uses a WAL location different from the global
+configuration you can specify the full path under `s3_wal_path`. For
+[Google Cloud Plattform](administrator.md#google-cloud-platform-setup)
```
sdudoladov, line 789: typo: `PlaTtform`  -> `Platform`

docs/user.md
```diff
@@ -793,6 +801,7 @@ spec:
     uid: "efd12e58-5786-11e8-b5a7-06148230260c"
     cluster: "acid-minimal-cluster"
     timestamp: "2017-12-19T12:40:33+01:00"
+    s3_wal_path: "s3://<bucketname>/spilo/<source_db_cluster>/<UID>/wal/<PGVERSION>"
```
sdudoladov, line 804: please change to `"s3://custom/path/to/bucket"`

docs/user.md
```diff
@@ -864,22 +873,28 @@ the PostgreSQL version between source and target cluster has to be the same.
...
+[custom pod environment variable](administrator.md#custom-pod-environment-variables).
+It is not set from the config to allow for overridding.
+
 ```yaml
 spec:
   standby:
     gs_wal_path: "gs://<bucketname>/spilo/<source_db_cluster>/<UID>/wal/<PGVERSION>"
 ```
 
+For a remote primry you specify the host address and optionally the port.
```
sdudoladov, line 895: typo: primry -> primAry

pkg/cluster/k8sres.go
```diff
@@ -944,7 +961,7 @@ func (c *Cluster) getPodEnvironmentConfigMapVariables() ([]v1.EnvVar, error) {
 	configMapPodEnvVarsList := make([]v1.EnvVar, 0)
 
 	if c.OpConfig.PodEnvironmentConfigMap.Name == "" {
-		return configMapPodEnvVarsList, nil
+		return nil, nil
```
sdudoladov, line 964: Why ? nil slice and empty slice are different things. In that case it does not seem to have any effect, but in general case might change some behaviour

pkg/cluster/k8sres.go
```diff
@@ -974,7 +993,7 @@ func (c *Cluster) getPodEnvironmentSecretVariables() ([]v1.EnvVar, error) {
 	secretPodEnvVarsList := make([]v1.EnvVar, 0)
 
 	if c.OpConfig.PodEnvironmentSecret == "" {
-		return secretPodEnvVarsList, nil
+		return nil, nil
```
sdudoladov, line 996: the same "nil vs empty" question here

FxKu:
:+1:

sdudoladov:
:+1:

sdudoladov:
:+1:

FxKu:
:+1:

PR #1851, opened 2022-04-14
do not call EBS api when there are no pvs

FxKu:
Only minor adjustments to improve syncing volumes when resize mode is `ebs`.


pkg/cluster/volumes.go
```diff
@@ -451,11 +454,6 @@ func quantityToGigabyte(q resource.Quantity) int64 {
 }
 
 func (c *Cluster) executeEBSMigration() error {
-	if !c.OpConfig.EnableEBSGp3Migration {
-		return nil
-	}
-	c.logger.Infof("starting EBS gp2 to gp3 migration")
-
```
FxKu, line 458: `c.OpConfig.EnableEBSGp3Migration` is already checked before calling `executeEBSMigration`. The log line is quite confusing because the function is executed every sync and at this point it's not decided yet, that a migration is actually happening.

pkg/cluster/volumes.go
```diff
@@ -167,7 +170,7 @@ func (c *Cluster) populateVolumeMetaData() error {
 		return err
 	}
 
-	if len(currentVolumes) != len(c.EBSVolumes) {
+	if len(currentVolumes) != len(c.EBSVolumes) && len(c.EBSVolumes) > 0 {
```
FxKu, line 173: On first run c.EBSVolumes will be empty at this point. We already logged the size of known volumes above. Therefore, we should print this debug message only on subsequent sync runs and only if there is a difference.

sdudoladov:
:+1:

FxKu:
:+1:

Issue #1203, opened 2020-11-09
How can I change the Patroni log level?


boonen:
For my production postgres clusters I'd like to limit log messages to a reasonable amount. Therefore I want to get rid off all the Patroni log messages that are logged - about every 10 seconds - at the `INFO` level, e.g.:

```
11/9/2020 5:27:22 PM 2020-11-09 16:27:22,345 INFO: Lock owner: cluster-name-0; I am cluster-name-0
11/9/2020 5:27:22 PM 2020-11-09 16:27:22,401 INFO: no action.  i am the leader with the lock
```

I found in the Patroni documentation that I can set the environment variable `PATRONI_LOG_LEVEL`, but I don't know how I can achieve this when Patroni is managed with the Postgres Operator (I use version `v1.5.0`). I tried setting it in the `ConfigMap` that I also use to configure *WAL backups*, but that hadn't any effect. Can anybody point me in the right direction?

Besides these Patroni messages I also see the following message about every 30 seconds:

```
2020-11-09 16:27:21.335 46 LOG {ticks: 0, maint: 0, retry: 0}
```

I don't know where it is coming from and how to suppress these message. Again, any help is appreciated.

Jan-M:
`PATRONI_LOG_LEVEL` check the docs, we allow I think to inject custom environment variables into the statefulset.

@CyberDem0n may know the answer to part two.

stephan2012:
Just raised https://github.com/zalando/patroni/issues/1800 which is closely related, I think.

boonen:
@Jan-M @CyberDem0n I have tried your suggestion to pass the environment variable `PATRONI_LOG_LEVEL` to the `StatefulSet`, but unfortunately this has no result. It seems that Spilo does not configure Patroni using the supplied environment variable. Do you have any hints on how to get this working?

I'm using the latest build of Spilo (from master) which runs Patroni `2.0.0`. Output of `env | grep PATRONI` in the Spilo `Pod` returns the following information:

```
PATRONI_LOG_LEVEL=WARNING
PATRONIVERSION=2.0.0
```

DrissiReda:
Did you manage to set the patroni env vars? I also tried the variable `PATRONI_LOG_FORMAT` which is completely ignored.

@boonen

mhaley-tignis:
A little late, but yes, most environment variables are ignored so that secret information is not leaked. Spilo does this, see the issue I raised here https://github.com/zalando/spilo/issues/718.

[...]

PR #1857, opened 2022-04-18
allow ability to adjust patroni log setting (#1203)

mhaley-tignis:
The operator produces its own `SPILO_CONFIGURATION` based on the information provided in the postgresql spec. This PR adds a `patroni.log` section to the spec to allow one to set log configuration for patroni. My end goal is to use this new feature to have patroni output to a log file within a container. I will then use another container in the same pod to scrape those process logs (promtail) and send them to our log storage (loki).

If you like these changes, I will go ahead and update the docs and example too.

Related: I have a PR up for spilo too so that we can configure logs via environment variables (something that this operator already supports) see https://github.com/zalando/spilo/pull/719.

### DIFF
--- a/manifests/postgresql.crd.yaml
+++ b/manifests/postgresql.crd.yaml
@@ -322,6 +322,29 @@ spec:
                     type: object
                     additionalProperties:
                       type: string
+                  log:
+                    type: object
+                    properties:
+                      level:
+                        type: string
+                      traceback_level:
+                        type: string
+                      format:
+                        type: string
+                      dateformat:
+                        type: string
+                      max_queue_size:
+                        type: string
+                      dir:
+                        type: string
+                      file_num:
+                        type: string
+                      file_size:
+                        type: string
+                      loggers:
+                        type: array
+                        items:
+                          type: string
                   loop_wait:
                     type: integer
                   maximum_lag_on_failover:
--- a/pkg/apis/acid.zalan.do/v1/postgresql_type.go
+++ b/pkg/apis/acid.zalan.do/v1/postgresql_type.go
@@ -157,6 +157,19 @@ type Resources struct {
 	ResourceLimits   ResourceDescription `json:"limits,omitempty"`
 }
 
+// PatroniLog contains Patroni-specific log configuration
+type PatroniLog struct {
+	Level          *string  `json:"level,omitempty"`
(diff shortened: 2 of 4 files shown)


mhaley-tignis:
@CyberDem0n perhaps maybe you can take a look at this PR?

mhaley-tignis:
@sdudoladov @Jan-M @CyberDem0n @FxKu @jopadi hello? Can anyone lend me a hand? I would like to move this PR along so I don't need to continue to maintain my own fork of the library. Thank you!

jehof:
Is this MR obsolete with the addition of https://github.com/zalando/postgres-operator/pull/1794 in release 1.8 of the operator? I may be wrong but now you should be able to define patroni configuration using Environment Variables and these should have a higher precedence.

mhaley-tignis:
> Is this MR obsolete with the addition of #1794 in release 1.8 of the operator? I may be wrong but now you should be able to define patroni configuration using Environment Variables and these should have a higher precedence.

I was always able to define patroni config using environment variables, but they are ignored by the spilo startup scripts (see https://github.com/zalando/spilo/pull/719). 

The suggestion https://github.com/zalando/spilo/issues/718#issuecomment-1099253455 to set the SPILO_CONFIGURATION is not possible because this operator does not let you set that yourself. Thus, the PR to expand the options available to include the ability to set your own log configuration too.

I either need this PR or this https://github.com/zalando/spilo/pull/719 one merged in order to be able to set patroni log settings.

mhaley-tignis:
One last attempt. @sdudoladov @Jan-M @CyberDem0n @FxKu @jopadi, is someone able to have a look here? Thanks in advance!

JulesLalu:
I am interested in this feature too, it's a shame that no one has reviewed this PR since it was opened more than a year and a half ago..

mhaley-tignis:
@FxKu I noticed you made a release a few week ago. Any chance you have a few moments to review this PR?

D1StrX:
Almost 1.5y later, when is this option enrolled?

mhaley-tignis:
We have plan to move off of postgres. Our main issues is how postgres handles HA scenarios. It only supports active/passive mode, meaning there is always downtime during a failure scenario. We are considering something like cockroachDB that uses a consensus architecture and does not suffer from the same limitation.

PR #1862, opened 2022-04-21
reverse membership for additional owner roles

FxKu:
While it can also be useful to grant roles to all database owners our intention with #1805 was actually the opposite.

This PR reverts a couple of changes done in #1805 like introducing a `RoleOriginSpilo` role type. Do not create a pgUser for additional owner roles - it's enough to add them in the `memberOf` field of existing DB owners. In case somebody already started using this feature, the operator will handle the reversed membership and revoke additional roles from database owners again. The latter is implemented as part of `ExecuteSyncRequests`. When altering membership of db owners fails (e.g. because of the wrong membership):

1. check for matches between additional owners and database owner's groups (`memberOf`)
2. try to revoke database owner from additional owner role. The latter is not stored in any pgUser map. So wrong membership cannot be checked - we just try the revoke)

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -164,9 +164,13 @@ class EndToEndTestCase(unittest.TestCase):
            Test granting additional roles to existing database owners
         '''
         k8s = self.k8s
+
+        # first test - wait for the operator to get in sync and set everything up
+        self.eventuallyEqual(lambda: k8s.get_operator_state(), {"0": "idle"},
+            "Operator does not get in sync")
         leader = k8s.get_cluster_leader_pod()
 
-        # produce wrong membership from v1.8.0
+        # produce wrong membership for cron_admin
         grant_dbowner = """
             GRANT bar_owner TO cron_admin;
         """
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -164,6 +164,13 @@ class EndToEndTestCase(unittest.TestCase):
            Test granting additional roles to existing database owners
         '''
         k8s = self.k8s
+        leader = k8s.get_cluster_leader_pod()
+
+        # produce wrong membership from v1.8.0
+        grant_dbowner = """
+            GRANT bar_owner TO cron_admin;
+        """
+        self.query_database(leader.metadata.name, "postgres", grant_dbowner)
 
         # enable PostgresTeam CRD and lower resync
         owner_roles = {
@@ -175,7 +182,6 @@ class EndToEndTestCase(unittest.TestCase):
         self.eventuallyEqual(lambda: k8s.get_operator_state(), {"0": "idle"},
                              "Operator does not get in sync")
 
-        leader = k8s.get_cluster_leader_pod()
         owner_query = """
             SELECT a2.rolname
               FROM pg_catalog.pg_authid a
@@ -188,6 +194,8 @@ class EndToEndTestCase(unittest.TestCase):
         """
         self.eventuallyEqual(lambda: len(self.query_database(leader.metadata.name, "postgres", owner_query)), 3,
             "Not all additional users found in database", 10, 5)
+
+        print('Operator log: {}'.format(k8s.get_operator_log()))
 
     @timeout_decorator.timeout(TEST_TIMEOUT_SEC)
     def test_additional_pod_capabilities(self):
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -161,7 +161,7 @@ class EndToEndTestCase(unittest.TestCase):
     @timeout_decorator.timeout(TEST_TIMEOUT_SEC)
     def test_additional_owner_roles(self):
         '''
-           Test adding additional member roles to existing database owner roles
+           Test granting additional roles to existing database owners
         '''
         k8s = self.k8s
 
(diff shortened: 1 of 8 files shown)


e2e/tests/test_e2e.py
```diff
@@ -175,16 +182,15 @@ def test_additional_owner_roles(self):
...
                              "Operator does not get in sync")
 
-        leader = k8s.get_cluster_leader_pod()
         owner_query = """
             SELECT a2.rolname
               FROM pg_catalog.pg_authid a
               JOIN pg_catalog.pg_auth_members am
                 ON a.oid = am.member
-               AND a.rolname = 'cron_admin'
+               AND a.rolname IN ('zalando', 'bar_owner', 'bar_data_owner')
```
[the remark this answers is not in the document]
FxKu: we have `zalando` as an owner in our [minimal manifest](https://github.com/zalando/postgres-operator/blob/8b6664f1a29e2cc3d87ea074dd5f96368e8d9d2e/manifests/minimal-postgres-manifest.yaml#L17). `zalandos` role is not a DB owner.

pkg/util/users/users.go
```diff
@@ -272,6 +296,16 @@ func quoteMemberList(user spec.PgUser) string {
 	return strings.Join(memberof, ",")
 }
 
+func revokeRole(groupRole, role string, db *sql.DB) error {
```
sdudoladov, line 300: iirc we used to claim operator never revokes roles. Please check docs/comments to ensure it is no longer the case.

pkg/util/users/users.go
```diff
@@ -120,6 +118,13 @@ func (strategy DefaultUserSyncStrategy) ExecuteSyncRequests(requests []spec.PgSy
 			if err := strategy.alterPgUser(request.User, db); err != nil {
 				reqretries = append(reqretries, request)
 				errors = append(errors, fmt.Sprintf("could not alter user %q: %v", request.User.Name, err))
+				// check if additional owners are misconfigured as members to a database owner
```
sdudoladov, line 122: that line well-deserved a comment with the PR number so that a reader can find more context on why the misconfiguration was ever possible

pkg/util/users/users.go
```diff
@@ -152,6 +158,25 @@ func (strategy DefaultUserSyncStrategy) ExecuteSyncRequests(requests []spec.PgSy
 	return nil
 }
 
+func resolveOwnerMembership(dbOwner spec.PgUser, additionalOwners []string, db *sql.DB) error {
+	errors := make([]string, 0)
+	for _, groupRole := range dbOwner.MemberOf {
+		for _, additionalOwner := range additionalOwners {
+			if additionalOwner == groupRole {
```
sdudoladov, line 165: I don't quite get the logic here. So we want to go from the current buggy setup
```
 db_owner               |                           | {}
 cron_admin             | Cannot login              | {db_owner}
 ```
to the one with the corrected membership. For that part of the code that means simply revoking the wrong role as in 
```
 db_owner               |                           | {}
 cron_admin             | Cannot login              | {}
 ```
For that simple example, the `dbOwner.MemberOf` will be empty, and the  `REVOKE` won't get executed ?

maybe the vars are mis-named ?

pkg/util/users/users.go
FxKu: Docs say, it does not revoke when things are removed from the manifest ;)

pkg/util/users/users.go
FxKu: actually the `for _, groupRole := range dbOwner.MemberOf {` loop is not needed. Whenever there are additional owner roles configured the operator should try to revoke db owners from them

pkg/util/users/users.go
```diff
@@ -272,6 +293,16 @@ func quoteMemberList(user spec.PgUser) string {
 	return strings.Join(memberof, ",")
 }
 
+func revokeRole(groupRole, role string, db *sql.DB) error {
+	revokeStmt := fmt.Sprintf(revokeFromUserSQL, groupRole, role)
```
sdudoladov, line 297: a better name for `revokeFromUserSQL` would be `revokeGroupRoleFromRoleSQL`, for example the `cron_admin` is not a user in PG jargon because it is a non-login role.

pkg/util/users/users.go
```diff
@@ -152,6 +158,21 @@ func (strategy DefaultUserSyncStrategy) ExecuteSyncRequests(requests []spec.PgSy
 	return nil
 }
 
+func resolveOwnerMembership(dbOwner spec.PgUser, additionalOwners []string, db *sql.DB) error {
+	errors := make([]string, 0)
+	for _, additionalOwner := range additionalOwners {
+		if err := revokeRole(dbOwner.Name, additionalOwner, db); err != nil {
```
sdudoladov, line 164: Does revoking an additional role from the role that lacks it generate an error in Go ?
sample psql session:
```
postgres=# revoke cron_admin from pooler;
WARNING:  role "pooler" is not a member of role "cron_admin"
REVOKE ROLE
```

pkg/util/users/users.go
```diff
@@ -120,6 +119,13 @@ func (strategy DefaultUserSyncStrategy) ExecuteSyncRequests(requests []spec.PgSy
 			if err := strategy.alterPgUser(request.User, db); err != nil {
 				reqretries = append(reqretries, request)
 				errors = append(errors, fmt.Sprintf("could not alter user %q: %v", request.User.Name, err))
+				// check if additional owners are misconfigured as members to a database owner (check #1862 for details)
+				// resolve it by revoking the database owner from the additional owner role
+				if request.User.IsDbOwner && len(strategy.AdditionalOwnerRoles) > 0 {
```
sdudoladov, line 124: I think the condition mismatches the comment: the `if` condition is more general.
What the current condition says and the code in the `true` branch does:

```
 Every time we failed (`err != nil` at line 119) to modify a db owner role (`if request.User.IsDbOwner`)
 and there are additional roles defined (`&& len(strategy.AdditionalOwnerRoles) > 0`), revoke all additional roles (lines 163-166) from the db owner.
```
e.g. if a Db is transiently not accessible for a second, we might get some future unrelated additional role removed for no reason.

pkg/util/users/users.go
FxKu: Oh, good point. But actually, even better if Postgres does not return an error if there's no wrong membership in place.

sdudoladov:
:+1:

FxKu:
:+1:

PR #1876, opened 2022-04-29
removing inner goroutine in cluster.Switchover

### LINKED ISSUE #1867: Missing subscriber channel for pooler pods
How to replicate:
1. Create cluster with 2 spilo and pooler pods.
2. Trigger a switchover by changing the manifest

```
panic: send on closed channel

goroutine 366 [running]:
github.com/zalando/postgres-operator/pkg/cluster.(*Cluster).processPodEvent(0xc00064a000, {0x1bdc360, 0xc000a0ef00})
	/workspace/pkg/cluster/cluster.go:1037 +0x205
k8s.io/client-go/tools/cache.(*FIFO).Pop(0xc000e5b8c0, 0xc000a2ff90)
	/workspace/vendor/k8s.io/client-go/tools/cache/fifo.go:300 +0x1fc
github.com/zalando/postgres-operator/pkg/cluster.(*Cluster).processPodEventQueue(0xc00064a000, 0x0)
	/workspace/pkg/cluster/cluster.go:1054 +0x65
created by github.com/zalando/postgres-operator/pkg/cluster.(*Cluster).Run
	/workspace/pkg/cluster/cluster.go:1045 +0x77
```

The operator pod comes back immediately and finishes the rolling update as expected (since it finds the corresponding annotations on the pods to be rotated). It seems that subscribers are missing if there is more than one pooler pod.

FxKu:
fixes #1867

Switchover function has this inner go routine which is not need. I guess, it was once introduced with the idea to have two go routines that wait for each other when sending and consuming evens on the podEvent channels. But even with a WaitGroup added later, changes for a race condition between closing the channel (consumer side) and sending events were still high as the operator did panic in most of my tests. And because of the fact that we wait anyway for the go routine to [finish](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L1513) the need concurrency in this part of the code looks questionable. 

With the go routine removed from Switchover, chances for hitting the race condition are already a lot lower. To fully eliminate the possibility of running into it, the PR moves c.podSubscribersMu.RUnlock() after the event handling. Thus, it will always send on the channel first before [unregisterPodSubscriber called from Switchover](https://github.com/zalando/postgres-operator/blob/264e66d0a1ef35f194ccd1fcaac48d069656b04f/pkg/cluster/cluster.go#L1511) can close it. Since unbuffered channels are usually blocking when there's no receiver (Switchover might already be done), a select block with a do nothing default case has to be provided. See this [go example](https://gobyexample.com/non-blocking-channel-operations).

The PR changes the check and timeout setting for deleting and patching Pods. `PatroniAPICheck*` config values are used because their defaults correspond to the hard coded values of 1s and 5s that we used before. But `PatroniAPICheck*` constants should only be used when calling the Patroni REST API instead.

The PR also fixes WaitGroup when PostgresTeam informer is enabled. Thanks @dmvolod for the hint in #1874 .

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1032,16 +1032,19 @@ func (c *Cluster) processPodEvent(obj interface{}) error {
 		return fmt.Errorf("could not cast to PodEvent")
 	}
 
+	// can only take lock when (un)registerPodSubscriber is finshed
 	c.podSubscribersMu.RLock()
 	subscriber, ok := c.podSubscribers[spec.NamespacedName(event.PodName)]
 	if ok {
 		select {
 		case subscriber <- event:
 		default:
-			// we end up here when there is no receiver of the channel
-			// avoiding a deadlock: https://gobyexample.com/non-blocking-channel-operations
+			// ending up here when there is no receiver on the channel (i.e. waitForPodLabel finished)
+			// avoids blocking channel: https://gobyexample.com/non-blocking-channel-operations
 		}
 	}
+	// hold lock for the time of processing the event to avoid race condition
+	// with unregisterPodSubscriber closing the channel (see #1876)
 	c.podSubscribersMu.RUnlock()
 
 	return nil
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1033,11 +1033,12 @@ func (c *Cluster) processPodEvent(obj interface{}) error {
 	}
 
 	c.podSubscribersMu.RLock()
+	defer c.podSubscribersMu.RUnlock()
+
 	subscriber, ok := c.podSubscribers[spec.NamespacedName(event.PodName)]
 	if ok {
 		subscriber <- event
 	}
-	c.podSubscribersMu.RUnlock()
 
 	return nil
 }
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1033,9 +1033,8 @@ func (c *Cluster) processPodEvent(obj interface{}) error {
 	}
 
 	c.podSubscribersMu.RLock()
-	defer c.podSubscribersMu.RUnlock()
-
 	subscriber, ok := c.podSubscribers[spec.NamespacedName(event.PodName)]
+	c.podSubscribersMu.RUnlock()
 	if ok {
 		subscriber <- event
 	}
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1034,10 +1034,10 @@ func (c *Cluster) processPodEvent(obj interface{}) error {
 
 	c.podSubscribersMu.RLock()
 	subscriber, ok := c.podSubscribers[spec.NamespacedName(event.PodName)]
-	c.podSubscribersMu.RUnlock()
 	if ok {
 		subscriber <- event
 	}
+	c.podSubscribersMu.RUnlock()
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


FxKu:
:+1:

pkg/cluster/pod.go
```diff
@@ -67,7 +67,7 @@ func (c *Cluster) markRollingUpdateFlagForPod(pod *v1.Pod, msg string) error {
 		return fmt.Errorf("could not form patch for pod's rolling update flag: %v", err)
 	}
 
-	err = retryutil.Retry(c.OpConfig.PatroniAPICheckInterval, c.OpConfig.PatroniAPICheckTimeout,
+	err = retryutil.Retry(c.OpConfig.ResourceCheckInterval, c.OpConfig.ResourceCheckTimeout,
```
sdudoladov, line 70: what about the timeouts ?  PatroniAPICheckTimeout = 5s , ResourceCheckTimeout = 10 minutes

pkg/cluster/pod.go
```diff
@@ -151,12 +151,13 @@ func (c *Cluster) unregisterPodSubscriber(podName spec.NamespacedName) {
...
 	defer c.podSubscribersMu.Unlock()
 
-	if _, ok := c.podSubscribers[podName]; !ok {
+	ch, ok := c.podSubscribers[podName]
+	if !ok {
 		panic("subscriber for pod '" + podName.String() + "' is not found")
 	}
 
-	close(c.podSubscribers[podName])
 	delete(c.podSubscribers, podName)
```
sdudoladov, line 159: close() before delete() looks a bit safer here. (what does Goalng do to `ch` when the entry that contains it is removed from  c.podSubscribers ?)

pkg/cluster/pod.go
```diff
@@ -399,11 +400,12 @@ func (c *Cluster) getPatroniMemberData(pod *v1.Pod) (patroni.MemberData, error)
...
 
 func (c *Cluster) recreatePod(podName spec.NamespacedName) (*v1.Pod, error) {
+	stopCh := make(chan struct{})
 	ch := c.registerPodSubscriber(podName)
 	defer c.unregisterPodSubscriber(podName)
-	stopChan := make(chan struct{})
+	defer close(stopCh)
 
-	err := retryutil.Retry(c.OpConfig.PatroniAPICheckInterval, c.OpConfig.PatroniAPICheckTimeout,
+	err := retryutil.Retry(c.OpConfig.ResourceCheckInterval, c.OpConfig.PodDeletionWaitTimeout,
```
sdudoladov, line 408: again the timeout values is increased from 5s to 10 minutes

pkg/cluster/cluster.go
```diff
@@ -1034,10 +1034,15 @@ func (c *Cluster) processPodEvent(obj interface{}) error {
...
 	if ok {
-		subscriber <- event
+		select {
+		case subscriber <- event:
+		default:
+			// we end up here when there is no receiver of the channel
+			// avoiding a deadlock: https://gobyexample.com/non-blocking-channel-operations
+		}
 	}
+	c.podSubscribersMu.RUnlock()
```
sdudoladov, line 1045: What is the reason to protect channel operations by an unrealted mutex ?

pkg/cluster/pod.go
FxKu: we can go back to hard-coded timeouts again. Trying to patch for 10 minutes might too long, I agree.

pkg/cluster/pod.go
FxKu: PatroniAPICheckTimeout is the wrong timeout and to me PodDeletionWaitTimeout looked more meaning full. But since it's not a waiting but deleting maybe we should go back to the previous hard-coded value of 5 seconds.

pkg/cluster/cluster.go
FxKu: podSubscribersMu will also be locked when closing the channel in c.unregisterPodSubscriber. This should prevent the panic when writing to a closed channel

sdudoladov:
:+1:

FxKu:
:+1:

PR #1884, opened 2022-05-06
bump to v1.8.1

### LINKED ISSUE #1882: Helm2 support dropped, but not removed from Chart.yaml?
- **Which image of the operator are you using?** postgres-operator:v1.8.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AWS/GCP/Azure
- **Type of issue?** Inquiry/possible bug report.

I see in [Release 1.8.0](https://github.com/zalando/postgres-operator/releases/tag/v1.8.0) that helm 2 support is dropped (listed as a breaking change).

Shouldn't the [apiVersion](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/Chart.yaml#L1) be bumped to v2, then?

According to the [docs](https://helm.sh/docs/topics/charts/), "The apiVersion field should be v2 for Helm charts that require at least Helm 3. Charts supporting previous Helm versions have an apiVersion set to v1 and are still installable by Helm 3."

Just curious if there is a reason apiVersion remains at v1 if that isn't the intended path forward.

FxKu:
* bump operator to 1.8.1
* raising apiVersion of helm charts to only support helm3+ (closes #1882)
* quoting users in REVOKE statement just in case it might be needed
* adding comments to stream types


CyberDem0n:
:+1:

FxKu:
:+1:

PR #1888, opened 2022-05-11
signal processPodEvent routine to close PodEvent channel

FxKu:
This is a continuation of #1876 that already improved on the situation that operator runs into a panic during switchover. However, there's still no guarantee that `Switchover` closes the PodEvent channel when calling `unregisterPodSubscriber` while the `processPodEvent` routine still sends something.

The idea of this PR is to send a new PodEvent type to the `processPodEvent` routine and make it responsible for closing the channel. Unfortunately, the race condition remains the same, but instead of a panic we will face a deadlock. Switchover method can send the special END event which will be added to [processPodEventQueue](https://github.com/zalando/postgres-operator/blob/a77d5df158593ce1f10afec5de3caec5698422f7/pkg/cluster/cluster.go#L1050), but there might still be more events in the queue that will then be send to the channel. However, the consumer (in this case waitForPodLabel) is already finished. Now the go routine is blocked and will no proceed.

One solution could be to define a buffered channel when we do the switchover. Buffered channels do not block go routines when events sent to a channel are not consumed. However, using the correct size for a channel might be a challenge, since it's not known how long the actual switchover will take and how many events are really sent during one second. And when the channel is full we have a deadlock again.

### DIFF
--- a/pkg/cluster/pod.go
+++ b/pkg/cluster/pod.go
@@ -136,8 +136,9 @@ func (c *Cluster) deletePods() error {
 
 func (c *Cluster) deletePod(podName spec.NamespacedName) error {
 	c.setProcessName("deleting pod %q", podName)
-	ch := c.registerPodSubscriber(podName)
-	defer c.unregisterPodSubscriber(podName)
+	ch := c.registerUnbufferedPodSubscriber(podName)
+	// send special end event to trigger removal of PodEvent channel by processPodEvent
+	defer c.ReceivePodEvent(PodEvent{PodName: types.NamespacedName(podName), EventType: PodEventEnd})
 
 	if err := c.KubeClient.Pods(podName.Namespace).Delete(context.TODO(), podName.Name, c.deleteOptions); err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


pkg/cluster/pod.go
```diff
@@ -399,11 +415,13 @@ func (c *Cluster) getPatroniMemberData(pod *v1.Pod) (patroni.MemberData, error)
 }
 
 func (c *Cluster) recreatePod(podName spec.NamespacedName) (*v1.Pod, error) {
-	ch := c.registerPodSubscriber(podName)
-	defer c.unregisterPodSubscriber(podName)
-	stopChan := make(chan struct{})
+	stopCh := make(chan struct{})
+	ch := c.registerUnbufferedPodSubscriber(podName)
+	// send special end event to trigger removal of PodEvent channel by processPodEvent
+	defer c.ReceivePodEvent(PodEvent{PodName: types.NamespacedName(podName), EventType: PodEventEnd})
```
dmvolod, line 421: @FxKu a bit confusing with the function name, as it sends event, not receives.

FxKu: I also stumbled over the name. But maybe I can get around sending artificial podEvents, after all, see #1891

PR #1891, opened 2022-05-13
Use second channel for podSubscribers to trigger its removal by cluster go routine

FxKu:
Another continuation of #1876. It takes the idea of #1888 sending a poison pill to the `processPodEvent` function inside the cluster go routine. But instead of using a podEvent that has to queue up, we are using a second channel next to the podEvents channel of cluster.podSubscribers map. The latter is not a map of channels anymore, but a map of `podSubscriber` struct consisting of two channels.

When `waitForPodLabel` is finished with consuming podEvents it will write on the second channel which will trigger the removal of podSubscribers. So a next podEvent should not find the subscriber anymore and not write on a closed channel.

fixes: #1867
closes #1876
closes #1888

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1046,8 +1046,8 @@ func (c *Cluster) processPodEvent(obj interface{}) error {
 		select {
 		case <-subscriber.stopEvent:
 			c.unregisterPodSubscriber(podName)
+		case subscriber.podEvents <- event:
 		default:
-			subscriber.podEvents <- event
 		}
 	}
 	// hold lock for the time of processing the event to avoid race condition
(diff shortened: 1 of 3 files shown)


pkg/cluster/cluster.go
```diff
@@ -1032,11 +1037,18 @@ func (c *Cluster) processPodEvent(obj interface{}) error {
...
 	c.podSubscribersMu.RLock()
-	subscriber, ok := c.podSubscribers[spec.NamespacedName(event.PodName)]
+	subscriber, ok := c.podSubscribers[podName]
 	c.podSubscribersMu.RUnlock()
 	if ok {
-		subscriber <- event
+		select {
+		case <-subscriber.stopEvent:
+			c.unregisterPodSubscriber(podName)
+		default:
```
Jan-M, line 1049: This is likely not a good default, it breaks the select semantics, you go into blocking wait for input to my understanding

pkg/cluster/cluster.go
```diff
@@ -1501,49 +1513,25 @@ func (c *Cluster) Switchover(curMaster *v1.Pod, candidate spec.NamespacedName) e
...
-		case podLabelErr <- func() (err2 error) {
-			_, err2 = c.waitForPodLabel(ch, stopCh, &role)
-			return
-		}():
-		}
+	subscriber := c.registerPodSubscriber(candidate)
+	defer func() {
+		subscriber.stopEvent <- struct{}{}
 	}()
+	defer close(stopCh)
```
Jan-M, line 1521: do we still need or use stopCh?

FxKu:
closing since #1876 already solved #1767

PR #1902, opened 2022-05-24
bring back CLONE_WAL_BUCKET_SCOPE_PREFIX

FxKu:
In #1848 setting the clone bucket prefix env variable was mistakenly removed.


FxKu:
:+1:

sdudoladov:
:+1:

farodin91:
This change prevents that i can set this value in the config map.

PR #1849, opened 2022-04-13
ETCDV3 support for DCS

AnisTigrini:
Adding environnement variables for etctdv3.
Integrating etcdv3.
Updating docs.

PR #1919, opened 2022-06-02
DCS - Etcd version customisation

Lucasgranet:
Hello,

Similar to #1849 
Allow change etcd version. 
Allow `v2` (default) and `v3` (which are currently supported by Patroni) 

I'm not agree with the author (of #1849) to bring username and password at operator level. (they should be defined through Postgres CR)

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -91,6 +91,12 @@ spec:
               etcd_host:
                 type: string
                 default: ""
+              etcd_host_version:
+                type: string
+                default: "v2"
+                enum:
+                  - "v2"
+                  - "v3"
               ignore_instance_limits_annotation_key:
                 type: string
               kubernetes_use_configmaps:
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -35,6 +35,8 @@ configGeneral:
   enable_spilo_wal_path_compat: false
   # etcd connection string for Patroni. Empty uses K8s-native DCS.
   etcd_host: ""
+  # etcd version v2 (default) or v3
+  etcd_host_version: v2
   # Spilo docker image
   docker_image: registry.opensource.zalan.do/acid/spilo-14:2.1-p6
 
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -97,6 +97,11 @@ Those are top-level keys, containing both leaf keys and groups.
   Patroni native Kubernetes support is used. The default is empty (use
   Kubernetes-native DCS).
 
+* **etcd_host_version**
+  Etcd version string for Patroni. Available versions are `v2` (default) or `v3`.
+  See [Patroni documentation](https://patroni.readthedocs.io/en/latest/SETTINGS.html#etcd)
+  for more information.
+
 * **kubernetes_use_configmaps**
   Select if setup uses endpoints (default), or configmaps to manage leader when
   DCS is kubernetes (not etcd or similar). In OpenShift it is not possible to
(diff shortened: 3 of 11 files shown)


charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -91,6 +91,9 @@ spec:
               etcd_host:
                 type: string
                 default: ""
+              etcd_host_version:
+                type: string
+                default: "v2"
```
dmvolod, line 96: @Lucasgranet please pay attention that's able to set any Patroni environment variable via `Env` section of the `Postgresql` spec now.

dmvolod, line 96: @Lucasgranet please pay attention that's able to set any Patroni environment variable via `Env` section of the `Postgresql` spec now.

dmvolod: https://postgres-operator.readthedocs.io/en/latest/administrator/#via-postgres-cluster-manifest

dmvolod: https://postgres-operator.readthedocs.io/en/latest/administrator/#via-postgres-cluster-manifest

Lucasgranet: Hello @dmvolod,

We need this change because it's the only way to disable the use of Kubernetes as DCS for Patroni.
Current code force the usage of etcdv2 (which is now deprecated) that's why I suggest this change to support etcdv3.

See: https://github.com/zalando/postgres-operator/pull/1919/files#diff-e9c5b05262207b16b959659256e773d9158a1710b89d1b73eca6453888563130R867

dmvolod: Thanks for explanation. Have you mind to add enum validation for possible options on the CRD level?

Lucasgranet: Added

AnisTigrini: Since we seem to be able to set username and password from Postgres CR, I agree, this seems to be the better solution.
Thanks @Lucasgranet !

Lucasgranet:
Hello,

Any news ?

Lucas,

Lucasgranet:
Hello,

Rebased on master.

Lucas,

thcdrt:
Hello,

@Jan-M , @sdudoladov, @CyberDem0n, @FxKu, @idanovinda, @jopadi could you take a look please ? :)

Thanks,
Thomas

thcdrt:
Hello,

@Jan-M , @sdudoladov, @CyberDem0n, @FxKu, @idanovinda, @jopadi could you take a look please ? :)

Thanks,
Thomas

gc-jro:
This would be a really helpful feature.

Lucasgranet:
Hello,

Rebased on master.

etcd-v2 storage backend will be removed in etcd 3.6 which encourage the use of etcd-v3.

Lucas,

mdoulabi1:
I noticed that this PR (#1849) was closed in favor of #1919, but #1919 was also closed and never merged.

At this point etcd v2 is effectively EOL, and newer etcd releases (especially 3.6+) no longer support the v2 API at all (the --enable-v2 flag has been removed).
This makes it impossible to use external etcd clusters with the operator unless etcd v3 is supported.

For our production setup we must use an external etcd cluster (separate from Kubernetes control-plane etcd), so using Kubernetes-native DCS is not an option.

Could you please clarify:

Why etcd v3 support was not merged?

Is there an official roadmap/plan to support etcd3 as Patroni DCS in the operator?

Should users maintain a fork, or is there a recommended workaround?

Happy to help with testing or providing feedback if needed.

Thanks!

PR #1943, opened 2022-06-28
annotation to bypass globally configured instance limits

FxKu:
In some situations it might be desired to scale beyond globally configured instance limits for single clusters.
The idea is to toggle this via an annotation in the cluster manifest. Just like with delete protection the key name is configurable. It is empty by default which also disables the feature.

When set, users have to add the annotation with value "true" to ignore the configured instance limits.


FxKu:
:+1:

PR #1953, opened 2022-07-06
allow in place pw rotation of system users

FxKu:
First approach to support password rotation for system users except `postgres` superuser. The tricky part: We can make the operator change the password, but the pods have to be replaced immediately afterwards to keep the system (replication) running.

Therefore, the idea of this PR is to tag pods with rolling update annotation from within the `updateSecret` method. Because the secret is only updated at the end of `updateSecret` we cannot rotate pods immediately. For pooler pods this requires some extra login while syncing pooler objects: Listing pods, check for annotation and delete if found.

ToDos for future PRs:
- System users can only be rotated when mentioned in the manifest. Should the global toggle `enable_password_rotation` incl. system users as well, or should there be an extra config option instead?
- Passwords can also be rotated if we mount the secrets as volumes and provide a config structure within the secret. `updateSecret` can update that structure and only call the Patroni reload endpoint, without replacing the pods.


FxKu:
:+1:

Jan-M:
:+1:

Issue #1703, opened 2021-11-24
Postgres operator does not recover instance if K8S API temporarily unavailable


monotek:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?**  registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Syseleven Metakube (Openshift)
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report

We're using spilo image registry.opensource.zalan.do/acid/spilo-13:2.1-p1 and postgres version 13.

Unfortunately our manageg kubernetes provider has not the best availabilty for the k8s api. 
If the api is not available temporarily, one, multiple or all patroni cluster members fail and can't recover. 
If cluster status is checked with `patronictl list` the members of the cluster claiming "start failed".

The reason they can't (re)start seems to be, that somehow the pvc permissions get messed up, when the api problem happens:
```
2021-11-24 03:18:18 UTC [159287]: [1-1] 619daefa.26e37 0     FATAL:  data directory "/home/postgres/pgdata/pgroot/data" has invalid permissions
2021-11-24 03:18:18 UTC [159287]: [2-1] 619daefa.26e37 0     DETAIL:  Permissions should be u=rwx (0700) or u=rwx,g=rx (0750).
```

If the instace is an replica you can fix the problem by `patronictl reinit` or restarting the pod.

Unfortunatly this can also happen for the leader, where only a `k -n namespcae delete po -l application=spilo --force --grace-period=0` can recover the cluster.

The main problem here is that the patroni cluster is not able to recover from such issue by itself. 
If the leader is in "start failed" status, i would expect that one of the available replicas becomes the new master.

I also don't understand, why a not available K8S API leads to changes in the file permissions of the cluster member at all?

This is the complete postgres log of such case:

```
2021-11-24 03:17:00.804 31 LOG {ticks: 0, maint: 0, retry: 0}
2021-11-24 03:17:07,516 INFO: no action. I am (auth-postgres-db-1) the leader with the lock
2021-11-24 03:17:17,518 INFO: no action. I am (auth-postgres-db-1) the leader with the lock
2021-11-24 03:17:27,512 INFO: no action. I am (auth-postgres-db-1) the leader with the lock
2021-11-24 03:17:30.820 31 LOG {ticks: 0, maint: 0, retry: 0}
[...]

PR #1972, opened 2022-07-25
feat: Add configurable liveness probe

davidspek:
Closes: https://github.com/zalando/postgres-operator/issues/58, https://github.com/zalando/postgres-operator/issues/1703

This PR adds a new field to the Operator configuration parameters and the Postgres spec allowing users too specify a liveness probe for the Spilo container. This is necessary too resolve errors with Postgres going down if communication with the K8s API temporarily fails, as described in [this issue](https://github.com/zalando/postgres-operator/issues/1703).

~~The Postgres CRD managed by the operator doesn't seem to have the new field added, trying to fix that still. Also, it doesn't seem like the statefulset is updated when the liveness probe is changed which still needs fixing.~~

The CRD managed by the operator now includes the new field and changes to the liveness and readiness probes now properly update the statefulset.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -260,6 +260,139 @@ spec:
                     type: array
                     items:
                       type: string
+                  liveness_probe:
+                    description: Probe describes a health check to be performed against
+                      a container to determine whether it is alive or ready to receive
+                      traffic.
+                    properties:
+                      exec:
+                        description: One and only one of the following should be specified.
+                          Exec specifies the action to take.
+                        properties:
+                          command:
+                            description: Command is the command line to execute inside
+                              the container, the working directory for the command  is
+                              root ('/') in the container's filesystem. The command
+                              is simply exec'd, it is not run inside a shell, so traditional
+                              shell instructions ('|', etc) won't work. To use a shell,
+                              you need to explicitly call out to that shell. Exit
+                              status of 0 is treated as live/healthy and non-zero
+                              is unhealthy.
+                            items:
+                              type: string
+                            type: array
+                        type: object
+                      failureThreshold:
+                        description: Minimum consecutive failures for the probe to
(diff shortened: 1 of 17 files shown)


FxKu:
LivenessProbes are [dangerous](https://srcco.de/posts/kubernetes-liveness-probes-are-dangerous.html). :fire: :fire: :fire:
I'm fine with a toggle so that anybody, who wants to risk it, can go with them. But, I'd prefer just one toggle and the operator creating a probe with a setup we think works for most cases. Like I proposed it for the ReadinessProbe here #2004 . I don't want to have a full spec option in the CRD manifests. Chances of misuse would be even higher then.

monotek:
Imho you've did not got the point of the article.

Liveness probes are not dangerous per se. You just have to use them right.

I even would go so far to say that a deployment which can't properly crash and restart automatically because of a missing liveness probe is just broken, because that's one of the essential things using a container in kubernetes.

Therefore it should be configurable at least even if not used by default.

FxKu:
I would still favor one pre-configured liveness probe that works for the operator and have only one toggle for users to switch.

DavidSpek:
We actually developed this as we didn’t want to use patroni-selfheal, and have been using it in prod on multiple cluster for 2 months now without issues and solving the problem we had. The main problem with a single liveness probe toggle is that it will likely not work in many scenarios, similarly to the readiness probe. One example is if a replica needs to clone a very large DB from the master, the probes would need to be configurable for this use case specifically. 
We were actually planning on making the readiness probe configurable as well because of this.

In terms of errors configuring the probes, they have the same type and validation as Kubernetes so that should take care of part of it. We can also add docs, examples, etc to inform people how to configure the probe and give a good starting point they can use.

DavidSpek:
@FxKu friendly bump since I believe this is a pretty essential feature for the operator.

aleskandro:
Hi, having this PR and feature merged seems a very good improvement. Any luck that it will happen?

theBNT:
still valid :>

PR #1976, opened 2022-07-26
Allow to override clone and standby description

farodin91:
Fixes #1902 unable to override scope prefix

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -628,11 +628,11 @@ order (e.g. a variable defined in 4. overrides a variable with the same name
 in 5.):
 
 1. Assigned by the operator
-2. Clone section (with WAL settings from operator config when `s3_wal_path` is empty)
-3. Standby section
-4. `env` section in cluster manifest
-5. Pod environment secret via operator config
-6. Pod environment config map via operator config
+2. `env` section in cluster manifest
+3. Pod environment secret via operator config
+4. Pod environment config map via operator config
+5. Clone section (with WAL settings from operator config when `s3_wal_path` is empty)
+6. Standby section
 7. WAL and logical backup settings from operator config
 
 ### Via ConfigMap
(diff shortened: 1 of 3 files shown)


farodin91:
@FxKu  Would you like to review?

farodin91:
@sdudoladov Would you like to review?

FxKu:
@farodin91 the clone and standby description would be overridable if you move [line 910 - 916](https://github.com/zalando/postgres-operator/blob/b91b69c736adc6aa2bdf82e0fd2a81e6e2bf2614/pkg/cluster/k8sres.go#L910) before line 939. Then cluster-specfic env var as well as environment configmpa or secret come first.

You would have to adjust unit test though, because order of variables and overriding behavior changes.

farodin91:
@FxKu  I updated the behavior.

farodin91:
@FxKu Do you have time to look into it?

FxKu:
@farodin91 yes. LGTM. Will merge next week when I‘m back from vcation.

FxKu:
@farodin91 can you rebase the PR, please? #2045 introduced some minor changes to env var generation. Thank you :)

farodin91:
@FxKu Rebased.

FxKu:
@farodin91 I gave this a second thought and now having doubts that this is right approach. Global config (like from pod_environment_configmap/secret) should not override local config (like `clone` and `standby` sections). Especially for [CloneDescription](https://github.com/zalando/postgres-operator/blob/70f3ee8e36bb00e7a23fedd1f0354f7faa8b8b5a/pkg/apis/acid.zalan.do/v1/postgresql_type.go#L194) we already have fields like `S3AccessKeyId` where one would expect that it overrides CLONE_AWS_ACCESS_KEY_ID from a global config map.

In that regard, having `spec.Env` before `generateCloneEnvironment` etc. is fine. This would also allow you to override the  CLONE_WAL_BUCKET_SCOPE_PREFIX. I know, doing this in every manifest is not what you aim for.

Why do you want to set the prefix in the first place? Note, that it has to be empty when wal path is set (manifest or config). This was the behavior ever since except for version 1.8.0 and 1.8.1 where it was forgotten by mistake. Is it [empty](https://github.com/zalando/postgres-operator/blob/v1.8.2/pkg/cluster/k8sres.go#L1862) in your case? That's the only case I see now, where we should not set it to `""`. Are there more fields you want to override from what is set in `generateCloneEnvironment`?

farodin91:
For me, it just CLONE_WAL_BUCKET_SCOPE_PREFIX. I could built a code section to set empty if not set, at the end.

We use the prefix for the cluster name.

Issue #1829, opened 2022-03-29
Feature: include load balancer costs into cost calculation


ePaul:
- **Type of issue?** feature request
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AWS K8s (the Zalando installation)
- **Which image of the operator are you using?** _(not sure how to find this out)_
- **Are you running Postgres Operator in production?** yes

The Operator UI on its "#/list" page shows a list of postgresql cluster managed by the operator, with an estimation of monthly costs.
These costs [are calculated](https://github.com/zalando/postgres-operator/blob/master/ui/app/src/postgresqls.tag.pug#L230) based on volume size (+ throughput/IOPs), memory + CPU requests and number of pods.

But there is another contribution to the total costs: The load balancers which are enabled by [`enableMasterLoadBalancer` or `enableReplicaLoadBalancer`](https://postgres-operator.readthedocs.io/en/latest/administrator/#load-balancers-and-allowed-ip-ranges) cost (with AWS's prices today) each 21.60 $/month, which (for smaller DBs) can easily exceed the other costs.

**Request**: Include the load balancer costs in the cost estimations shown in the #/list page.

PR #1977, opened 2022-07-27
[UI] include load balancer in the monthly cost calculation

idanovinda:
Update the monthly cost calculation by including the load balancer cost. It costs 0.03 per hour per load balancer. 

Price source: https://aws.amazon.com/elasticloadbalancing/pricing/

Based on issue #1829

### DIFF
--- a/ui/operator_ui/main.py
+++ b/ui/operator_ui/main.py
@@ -525,7 +525,8 @@ def get_postgresqls():
             'namespaced_name': namespace + '/' + name,
             'full_name': namespace + '/' + name + ('/' + uid if uid else ''),
             'status': status,
-            'num_elb': spec.get('enableMasterLoadBalancer', 0) + spec.get('enableReplicaLoadBalancer', 0),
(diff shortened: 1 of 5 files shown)


ui/operator_ui/main.py
```diff
@@ -523,6 +525,7 @@ def get_postgresqls():
             'namespaced_name': namespace + '/' + name,
             'full_name': namespace + '/' + name + ('/' + uid if uid else ''),
             'status': status,
+            'num_elb': spec.get('enableMasterLoadBalancer', 0) + spec.get('enableReplicaLoadBalancer', 0),
```
FxKu, line 528: maybe you can also add support for `enableMasterPoolerLoadBalancer` and `enableReplicaPoolerLoadBalancer` so that they are included here? :)

idanovinda: is the cost the same for the pooler load balancer?

Jan-M: Yes

FxKu:
:+1:

idanovinda:
👍

Issue #1983, opened 2022-08-02
Switchover (during a Node drain) fails randomly in synchronous mode


gdemonet:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** `registry.opensource.zalan.do/acid/postgres-operator:v1.8.2`
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K8s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report

---

**Context**: This problem closely relates to #1686. In our scenario, we have a 3 nodes cluster, with one synchronous replica and another asynchronous.

**Symptom**: The problem arises when we perform a rolling upgrade of our K8s cluster, which upgrades nodes one by one, while they are drained. Draining the node where the PostgreSQL master is running is prevented (as expected) by the disruption budget defined by `postgres-operator`, and waits indefinitely.

We can observe these logs in the operator (and they loop, until we fix the situation by performing a manual failover):

```
time="2022-07-29T08:30:18Z" level=info msg="starting process to migrate master pod \"foo-auth/foo-postgres-0\"" cluster-name=foo-auth/foo-postgres pkg=cluster
time="2022-07-29T08:30:18Z" level=debug msg="Waiting for any replica pod to become ready" cluster-name=foo-auth/foo-postgres pkg=cluster
time="2022-07-29T08:30:18Z" level=debug msg="Found 2 running replica pods" cluster-name=foo-auth/foo-postgres pkg=cluster
time="2022-07-29T08:30:18Z" level=info msg="check failed: pod \"foo-auth/foo-postgres-1\" is already on a live node" cluster-name=foo-auth/foo-postgres pkg=cluster
time="2022-07-29T08:30:18Z" level=debug msg="switching over from \"foo-postgres-0\" to \"foo-auth/foo-postgres-1\"" cluster-name=foo-auth/foo-postgres pkg=cluster
time="2022-07-29T08:30:18Z" level=debug msg="subscribing to pod \"foo-auth/foo-postgres-1\"" cluster-name=foo-auth/foo-postgres pkg=cluster
time="2022-07-29T08:30:18Z" level=debug msg="making POST http request: http://10.233.247.14:8008/failover" cluster-name=foo-auth/foo-postgres pkg=cluster
[...]

PR #1984, opened 2022-08-02
BugFix: Switchover (during a Node drain) fails randomly in synchronous mode

JBWatenbergScality:
Change1: MigrateMasterPod now use getSwitchOverCandidate to select a new masterCandidate
Change2: Removed unused masterCandidate (replaced by getSwitchOverCandidate)
Fixes: #1983

### DIFF
--- a/pkg/cluster/pod.go
+++ b/pkg/cluster/pod.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"math/rand"
 	"sort"
 	"strconv"
 	"time"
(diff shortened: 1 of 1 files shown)


pkg/cluster/pod.go
```diff
@@ -283,13 +252,19 @@ func (c *Cluster) MigrateMasterPod(podName spec.NamespacedName) error {
...
 	} else {
 		c.logger.Warningf("migrating single pod cluster %q, this will cause downtime of the Postgres cluster until pod is back", c.clusterName())
 	}
 
+	masterCandidatePod, err := c.KubeClient.Pods(masterCandidateName.Namespace).Get(context.TODO(), masterCandidateName.Name, metav1.GetOptions{})
+
+	if err != nil {
+		return fmt.Errorf("could not get master candidate pod: %v", err)
+	}
+
```
FxKu, line 267: getSwitchoverCandidate only picks a replica when state is `running`. I don't think we need another API call here.

JBWatenbergScality: Indeed but then I will need to change `getSwitchoverCandidate` to make it return a `Pod` instead of a `NamespacedName` as below methods require a Pod. Another alternative would be to create a new `getSwitchoverCandidatePod` and refactor  `getSwitchoverCandidate` to use that one if we want to keep the old signature. Which approach would you prefer ?

FxKu: ah I see. I think in this case, it's fine then. We would have to do the call anyway - whether it's here or in `getSwitchoverCandidate`. And the call does not happen each sync, so it's fine.

JBWatenbergScality:
Hi @FxKu , I just rebased my branch as I saw the [End-2-End tests](https://github.com/zalando/postgres-operator/runs/7766246685?check_suite_focus=true#logs) were failing for reasons that doesn't seems to be related to my change. Not sure it will help through. Should I investigate those or assume that it is a flaky/ignore them ?

Jan-M:
:+1:

FxKu:
:+1:

FxKu:
Thanks @JBWatenbergScality for this nice contribution!

PR #1986, opened 2022-08-03
allows you to use variables in values .Values.configGeneral

alexey-gavrilov-flant:


alexey-gavrilov-flant:
#1988

charts/postgres-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: postgres-operator
-version: 1.8.2
+version: 1.8.3
```
FxKu, line 3: will change it for the next release

FxKu: can you remove this change

charts/postgres-operator/templates/operatorconfiguration.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Release.Name }}
 configuration:
-{{ toYaml .Values.configGeneral | indent 2 }}
+{{ tpl (toYaml .Values.configGeneral) . | indent 2 }}
```
FxKu, line 13: would it make sense to change it also for the other config sections?

alfsch: yes it would make sense

alfsch:
We're waiting for this chart feature, since it makes it possible to configure the image repositories of sidecars stage/deployment specific. With helm it's not possible to override arrays with --set and the sidecars are an array.

Currently we're patching the helm tarball with above change to make the images repositories changeable.

We have different image repositories for test, int, prod stages....

alfsch:
Created an issue #2114

charts/postgres-operator/Chart.yaml
FxKu:
```suggestion
version: 1.8.2
```

charts/postgres-operator/templates/operatorconfiguration.yaml
FxKu: could you do the same for `configLoadBalancer` and `configKubernetes` - I think it might only be required there

FxKu:
Please add this tpl function to the other sections mentioned and then we can merge.

charts/postgres-operator/templates/operatorconfiguration.yaml
alfsch: From my point of view it's currently only needed in configKubernetes since this section contains an array. I'll create a separate pull request for that.

FxKu:
Will close in favor for #2115. But thanks @alexey-gavrilov-flant for raising it!

alfsch:
@FxKu you have also to merge this pull, since I don't inclued this change in mine

PR #1993, opened 2022-08-10
Fix postgresql CRD for more safely validation

hornwind:
This change will allow to immediately discard invalid values when parsing a logicalBackupSchedule string.

Previously used regular expression `^(\d+|\*)(/\d+)?(\s+(\d+|\*)(/\d+)?){4}$` may pass invalid values to cron, like `"00 03 * * 7"`
Although `"00 03 * * 7"` may be valid on some systems, but this value will throw an error when applied to kubernetes.
https://en.wikipedia.org/wiki/Cron

FYI: kubernetes uses cron implementation from go module "github.com/robfig/cron/v3"
https://github.com/robfig/cron/blob/v3/spec.go
https://github.com/robfig/cron/blob/v3/parser.go


bwrobc:
In my view this validation should just be removed, as it will fail many valid schedule strings using intervals, named days, ranges, etc. and it would be nearly impossible to properly validate them. See #2323.

PR #2001, opened 2022-08-16
Make teamId in cluster name optional

FxKu:
Teams of cluster change. Although the name of the cluster is immutable, it would still be helpful to change teamId field without the need to create a PostgresTeam resource.

On start up the operator calls `add_cluster` for all Postgresql resource it finds. When the new `enable_team_id_clustername_prefix` is enabled and cluster name does not start with team ID, `add_cluster` will return an error message and the cluster is not added to the operator's internal cluster list.

The Postgresql resource itself still gets created via K8s API. We have no controller in between to block this. Because the resource is there we can patch it's `PostgresClusterStatus` field in status subresource to `invalid`. Each sync on the cluster will try to call add_cluster again because it does not exist in the operator#s internal list. When the option is turned off and operator is restarted, the cluster is synced like very other cluster.

This PR keeps `teamId` as an `required` field of the cluster manifest (because we want to keep cluster ownership from teams) and it deprecates the cluster's internal `ClusterName` field which used to store the cluster name without the `teamId` prefix. We did not use the `ClusterName` anywhere else.

### DIFF
--- a/docs/reference/cluster_manifest.md
+++ b/docs/reference/cluster_manifest.md
@@ -53,8 +53,7 @@ Those parameters are grouped under the `metadata` top-level key.
 These parameters are grouped directly under  the `spec` key in the manifest.
 
 * **teamId**
-  name of the team the cluster belongs to. Changing it after the cluster
-  creation is not supported. Required field.
+  name of the team the cluster belongs to. Required field.
 
 * **numberOfInstances**
   total number of  instances for a given cluster. The operator parameters
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -94,8 +94,8 @@ Those are top-level keys, containing both leaf keys and groups.
 
 * **enable_team_id_clustername_prefix**
   To lower the risk of name clashes between clusters of different teams you
-  can turn on this flag and the operator will only allow new clusters that start
-  with the `teamId` (from `spec`) plus `-`. Default is `false`.
+  can turn on this flag and the operator will sync only clusters where the
+  name starts with the `teamId` (from `spec`) plus `-`. Default is `false`.
 
 * **etcd_host**
   Etcd connection string for Patroni defined as `host:port`. Not required when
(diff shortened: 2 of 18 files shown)


docs/reference/cluster_manifest.md
```diff
@@ -53,8 +53,7 @@ Those parameters are grouped under the `metadata` top-level key.
 These parameters are grouped directly under  the `spec` key in the manifest.
 
 * **teamId**
-  name of the team the cluster belongs to. Changing it after the cluster
-  creation is not supported. Required field.
+  name of the team the cluster belongs to. Required field.
```
Jan-M, line 56: Should inform about settings and implications. it may still be impossible.

docs/reference/operator_parameters.md
```diff
@@ -92,6 +92,11 @@ Those are top-level keys, containing both leaf keys and groups.
 * **enable_spilo_wal_path_compat**
   enables backwards compatible path between Spilo 12 and Spilo 13+ images. The default is `false`.
 
+* **enable_team_id_clustername**
```
Jan-M, line 95: enable_team_id_clustername_prefix

pkg/apis/acid.zalan.do/v1/marshal.go
```diff
@@ -110,15 +110,9 @@ func (p *Postgresql) UnmarshalJSON(data []byte) error {
 	}
 	tmp2 := Postgresql(tmp)
 
-	if clusterName, err := extractClusterName(tmp2.ObjectMeta.Name, tmp2.Spec.TeamID); err != nil {
-		tmp2.Error = err.Error()
-		tmp2.Status = PostgresStatus{PostgresClusterStatus: ClusterStatusInvalid}
-	} else if err := validateCloneClusterDescription(tmp2.Spec.Clone); err != nil {
-
+	if err := validateCloneClusterDescription(tmp2.Spec.Clone); err != nil {
```
Jan-M, line 113: how is this related to clone? maybe better function name? or is this indeed about cloning?

FxKu: yes, this is a validation of the clone section. Existed also before.

docs/reference/cluster_manifest.md
FxKu: it's now possible to change it later. When `enable_team_id_clustername_prefix` is enabled you will only get an error on sync that cluster name does not match the { TEAM-ID }-{... format with status being set to inavlid.

FxKu:
:+1:

idanovinda:
👍

Issue #2003, opened 2022-08-17
pod-0 unhealthy --> Restart all 3 pods --> 2 healthy pods are not launching/starting at all --> DB cluster remains "stuck"/down


Samusername:
Hi!
postgres-operator: v1.8.1
spilo: 2.1-p6
Patroni: 2.1.4
synchronous_mode: true
synchronous_mode_strict: true
We use "ConfigMap" setup instead of endpoints setup.

rdbms-pg-cluster-0  Replica    ---   We broke this up intentionally. We moved WAL files etc. to break it up.
rdbms-pg-cluster-1  Leader
rdbms-pg-cluster-2  Sync Standby

After we broke rdbms-pg-cluster-0:
If we delete all 3 pods, then DB cluster does not start up:
It ends up in following state:

kubectl get pods -A | grep rdbms
namespace0                                        **rdbms-pg-cluster-0**                                                1/2     Running            0          43m
namespace0                                        rdbms-pg-operator-6c8b55d586-c28xp                                1/1     Running            0          21h

Only one pod is shown / launching.
rdbms-pg-cluster-1 nor rdbms-pg-cluster-2 are not even shown in the list of pods.

Q: How should such be tried to be solved (or fixed)?

In postgres-operator or with pod_management_policy?

Reference:
pod_management_policy, ordered_ready (default), (or parallel).
https://opensource.zalando.com/postgres-operator/docs/reference/operator_parameters.html

We hesitate using "parallel" in pod_management_policy.

<details><summary>Logs from postgres-operator. Click to expand</summary>

time="2022-08-17T04:14:57Z" level=info msg="SYNC event has been queued" cluster-name=namespace0/rdbms-pg-cluster pkg=controller worker=0
time="2022-08-17T04:14:57Z" level=info msg="there are 1 clusters running" pkg=controller
time="2022-08-17T04:14:57Z" level=info msg="Creating the role binding \"postgres-pod\" in the \"namespace0\" namespace" pkg=controller
time="2022-08-17T04:14:57Z" level=warning msg="pods and/or Patroni may misfunction due to the lack of permissions: could not create role binding \"postgres-pod\" : cannot bind the pod service account \"postgres-pod\" defined in the configuration to the cluster role in the \"namespace0\" namespace: clusterroles.rbac.authorization.k8s.io \"postgres-pod\" not found" pkg=controller
[...]

Issue #1978, opened 2022-07-28
Cluster fails to restart if lock is held by other pod than the first one in the statefulset


christiaangoossens:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.8.2
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** k0s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug

Some general remarks when posting a bug report:
- Please, check the operator, pod (Patroni) and postgresql logs first. When copy-pasting many log lines please do it in a separate GitHub gist together with your Postgres CRD and configuration manifest.
- If you feel this issue might be more related to the [Spilo](https://github.com/zalando/spilo/issues) docker image or [Patroni](https://github.com/zalando/patroni/issues), consider opening issues in the respective repos.

---

Cluster fails to restart if the lock is held by pod-1, while the statefulset is restarting. Pod-0 will start, recognise that pod-1 holds the lock and wait forever, eventually falling into a state where it knows its not up to date and thus waits forever.

Due to using podManagementPolicy Ordered, instead of Parallel, pod-1 will never be started, and thus this configuration stays broken.

In my case this happened when a node failed, there was a failover to pod-1 and then both died.

christiaangoossens:
```
2022-07-28 07:02:08,735 - bootstrapping - INFO - Figuring out my environment (Google? AWS? Openstack? Local?)
2022-07-28 07:02:09,417 - bootstrapping - INFO - Looks like your running aws
2022-07-28 07:02:09,480 - bootstrapping - INFO - Configuring bootstrap
2022-07-28 07:02:09,480 - bootstrapping - INFO - Configuring crontab
2022-07-28 07:02:09,480 - bootstrapping - INFO - Skipping creation of renice cron job due to lack of SYS_NICE capability
2022-07-28 07:02:09,481 - bootstrapping - INFO - Configuring standby-cluster
2022-07-28 07:02:09,481 - bootstrapping - INFO - Configuring log
2022-07-28 07:02:09,481 - bootstrapping - INFO - Configuring pam-oauth2
2022-07-28 07:02:09,482 - bootstrapping - INFO - Writing to file /etc/pam.d/postgresql
2022-07-28 07:02:09,482 - bootstrapping - INFO - Configuring patroni
[...]

PR #2004, opened 2022-08-22
add toggle to turn off readiness probes

FxKu:
Addresses: #2003, #1978

Having statefulsets with readiness probes was requested a couple of times in the past. We merged it with #1825. Now they are enabled by default, which turns out to be not the greatest idea when `pod_management_policy` is set to `ordered_ready`. What could happen?

Imagine a cluster with `pod-0` and `pod-1` where the first instance (replica) is broken (e.g. missing a WAL file to restore, disk full etc.) but the master is doing fine. Now the master pod must be replaced (maintenance or it dies for whatever reason). The first pod is failing the readiness probe blocking the second pod from coming back and restore the former master.

Without the readiness probe, it's enough for K8s to see that the first pod is in `Running` state ignoring the actual state of the cluster member. This helped us in the past that masters could repair themselves.

Changing the `pod_management_policy` is not as easy as it seems, because the operator does not compare it on sync. This PR will include the policy as well as the readiness probe of the container into the stateful set comparison to trigger a replacement of the stateful set (and rolling update of pods if the probe changes).

With `ordered_ready` being the default policy, the readiness probe should be disabled by default although this is a breaking change to v1.8. Therefore, a new toggle is added to keep the existing behavior.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -211,6 +211,9 @@ spec:
                   enable_pod_disruption_budget:
                     type: boolean
                     default: true
+                  enable_readiness_probe:
+                    type: boolean
+                    default: false
                   enable_sidecars:
                     type: boolean
                     default: true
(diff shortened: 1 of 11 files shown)


OlleLarsson:
Any updates regarding this @FxKu?

idanovinda:
👍

zlcnju:
why not use parallel as the default policy?

PR #2009, opened 2022-08-24
fetch pooler and fes_user system user only when corresponding features are used

FxKu:
If somebody unintentionally defines reserved system users `pooler` or `fes_user` in the manifest, the [updateSecret](https://github.com/zalando/postgres-operator/blob/e11edcdcde897ec16e6a9903f97a6cad2e96eee2/pkg/cluster/sync.go#L720) function will use the systemUsers map. But if the pooler or stream feature are not used, no corresponding PgUser object will exist in that map. The PgUser used in password rotation will be empty ultimatly leading to this error:

```
could not sync roles: error executing sync statements: could not execute sync requests for users: could not create user "": pq: zero-length delimited identifier at or near """"
```

The bug was introduced with #1953 .

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1136,8 +1136,8 @@ func (c *Cluster) initSystemUsers() {
 			Password:  util.RandomPassword(constants.PasswordLength),
 		}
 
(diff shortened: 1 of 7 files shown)


idanovinda:
👍

PR #2011, opened 2022-08-25
use old LB DNS format when teamId prefix is disabled

FxKu:
With #2001 merged we dropped the teamId prefix constraint for cluster names and deprecated the `ClusterName` field of the `Postgresql` type assuming it was not used somewhere else. Turns out, it is used for LoadBalancer service's external DNS name and ClusterStatus API.

This PR will set this field again when clusters are added to the operator's internal cluster list.

If we would continue to use the Spec.ClusterName for the LB DNS names they would break when disabling the team Id constraint. Therefore, we should continue to use old format when the cluster name starts with the team ID. external-dns supports multiple names per annotation, separated by comma.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -424,10 +424,10 @@ spec:
                     default: "Cluster"
                   master_dns_name_format:
                     type: string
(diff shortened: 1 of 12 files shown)


Jan-M:
:+1:

FxKu:
:+1:

PR #2015, opened 2022-08-26
deprecate ClusterName field of Postgresql type and remove team from REST endpoints

FxKu:
With #2001 we made the teamId prefix optional. The prefix was formerly checked during unmarshaling of the cluster manifest. It was calling [ExtractClusterName](https://github.com/zalando/postgres-operator/blob/89375186b3fd3df2e787f9db5a85c13166729581/pkg/apis/acid.zalan.do/v1/util.go#L49) function. On success, the PostgresSpec field ClusterName would receive the cluster name without teamId.

Leaving the field just empty led to a bug with LoadBalancer services. DNS names missing the cluster name because `c.Spec.ClusterName` was not set. So we tried to [set it](https://github.com/zalando/postgres-operator/blob/89375186b3fd3df2e787f9db5a85c13166729581/pkg/controller/postgresql.go#L179) again with #2011. `cl.ClusterName` is wrong - it should be `cl.Spec.ClusterName`. Nevertheless, it is still empty when accessing it in [masterDNSName](https://github.com/zalando/postgres-operator/blob/89375186b3fd3df2e787f9db5a85c13166729581/pkg/cluster/util.go#L533).

This PR now suggest to deprecate the `ClusterName` field like it was originally planned in #2001. The teamId prefix option should only check if clusters start with the team Id and block them if not. Even when prefix constraint is enabled the DNS name will take the full name, because we changed the DNS template from {clustername}.{team}.{hostedzone} to {clustername}.{namespace}.{hostedzone}. There's no point in cutting of the teamId prefix from the cluster name anymore.

To provide backwards compatibilty #2011 already introduced a second DNS name that resembles the old format {cluster-with-no-team-prefix}.{team}.{hostedzone}.

Another oversight so far, has been the Operator REST API that [prepends the teamID](https://github.com/zalando/postgres-operator/blob/89375186b3fd3df2e787f9db5a85c13166729581/pkg/controller/logs_and_api.go#L22) for status, history and logs endpoints. In the API server's internal `clusterNamesPerTeam` the teamId must also remain in the name.


pkg/cluster/util.go
```diff
@@ -507,38 +507,38 @@ func (c *Cluster) roleLabelsSet(shouldAddExtraLabels bool, role PostgresRole) la
...
 
 	return dnsString
 }
 
 func (c *Cluster) masterDNSName() string {
 	return strings.ToLower(c.OpConfig.MasterDNSNameFormat.Format(
-		"cluster", c.Spec.ClusterName,
+		"cluster", c.Name,
 		"namespace", c.Namespace,
+		"team", c.teamName(),
```
FxKu, line 533: If somebody still has the old {clustername}.{team}.{hostedzone} template in the config, we'll make sure it continues to work.

pkg/apiserver/apiserver_test.go
```diff
@@ -5,9 +5,9 @@ import (
 )
 
 const (
-	clusterStatusTest        = "/clusters/test-id/test_namespace/testcluster/"
-	clusterStatusNumericTest = "/clusters/test-id-1/test_namespace/testcluster/"
-	clusterLogsTest          = "/clusters/test-id/test_namespace/testcluster/logs/"
+	clusterStatusTest        = "/clusters/test_namespace/testcluster/"
+	clusterStatusNumericTest = "/clusters/test_namespace-1/testcluster/"
```
Jan-M, line 9: this looks odd with mix of - and _

FxKu:
:+1:

Issue #2014, opened 2022-08-26
Enhance documentation for "PostgresTeam/Teams API" - automatic role removal


DaveDef:
Hello,

What I need to achieve is to enable ```enable_team_member_deprecation: true``` in order to rename roles and append a configured suffix to their name as it is described in documentation:

https://github.com/zalando/postgres-operator/blob/master/docs/user.md#teams-api-roles

So far I was not lucky to make it work and I still have some questions regarding the implementation of this feature, for example:

- Is enabled ``` Teams API``` a dependency for ```PostgresTeam```? In other words, Do I need to setup and enable ```Teams API``` just to use ```PostgresTeam``` resource?
  - If yes, during the deployment of ```Teams API``` I need to setup "fake" teams-api which has a deployment manifest where image is mentioned: ```image: ikitiki/fake-teams-api:latest```. The image is nowhere to be found. Do I need to build it?
  - When I apply ```fake-teams-api.yaml``` I got error saying the manifest is not valid. I had to add ```matchLabels:``` keyword with label under the ```spec```.
  
- Is there anyone who was able to achieve automatic removal of roles via ```PostgresTeam``` resources and has some step-by-step example or demo to share?

Thank you.

FxKu:
The Teams API should not be a requirement for PostgresTeam CRD to work. In the [code](https://github.com/zalando/postgres-operator/blob/4c07494ac731775d100d773987adb575c5b12581/pkg/cluster/util.go#L271), it is checked after additional members are fetched.

The fake-teams-api is indeed missing a selector. Maybe you can create a PR to add:
```
  selector:
    matchLabels:
      name: fake-teams-api
```
before `template`.

When you remove a role from the PostgresTeam CRD it should be reflected on the next operator sync and append the deprecation prefix.

DaveDef:
Hello,

I was able to figure it out in the end.
It didn't work for me in manual deployment (maybe some missconfiguration on my side) but it did work via helm chart deployment.

Just worth to mention from my experiences:

1. If you are deploying via helm chart make sure to set ```enable_postgres_team_crd: true``` and ```enable_team_member_deprecation: true```  in ```values.yaml```

2. For testing purposes it is good to set ```resync_period: 30m``` to lower values. Otherwise the operator will update database after 30mins by default.

3. By default env variable in ```deployment``` is set to:

```
Environment:
      POSTGRES_OPERATOR_CONFIGURATION_OBJECT:  postgres-operator
```

[...]

PR #2029, opened 2022-08-31
use correct keys in updateSecret

FxKu:
The `updateSecret` code was using wrong keys when treating pooler and FES users accidentally adding more empty `cluster.systemUsers` leading to a similar error like in #2009 

I have extended the unit test for `updateSecret` to cover it.

### DIFF
--- a/manifests/fake-teams-api.yaml
+++ b/manifests/fake-teams-api.yaml
@@ -4,6 +4,9 @@ metadata:
   name: fake-teams-api
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      name: fake-teams-api
   template:
     metadata:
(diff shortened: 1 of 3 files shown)


manifests/fake-teams-api.yaml
```diff
@@ -4,6 +4,9 @@ metadata:
   name: fake-teams-api
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      name: fake-teams-api
```
FxKu, line 9: addresses: https://github.com/zalando/postgres-operator/issues/2014

FxKu:
:+1:

jopadi:
👍

PR #2034, opened 2022-09-05
create streams only after postgres instances were restarted

FxKu:
When `streams` are specified in the manifest the operator updates the Postgres config and starts creating DB publications and stream CRDs. The stream operator will notice the CRDs and start with creating deployments to consume from logical slots. But it will find the Postgres config not yet updated and crash a few times until the operator restarts the Pods on next sync.

When we know that instances must be restarted in `syncPostgresConfig` we could call Patroni's restart endpoint. But it would happen anyway when the statefulset is synced. Therfore, we can just quit `syncStreams` and wait for it.


FxKu:
:+1:

idanovinda:
👍

PR #2043, opened 2022-09-15
Fix password rotation

FxKu:
### Current user and secret generation flow

How it looks like without rotation

1. [initUsers](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/cluster.go#L196): Generates list of desired users (manifest roles, system users, team members etc.) with new [random passwords](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/cluster.go#L1302)
2. [syncSecrets](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L628)
a. [generateUserSecrets](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/k8sres.go#L1669): Generates a list of secrets where `username` and `password` in secret are taken from the [pgUser](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/k8sres.go#L1724) generated in 1.
b. Operator tries to create the secret or calls `updateSecret` when it already exists (although an actual update might not happen when there are no changes).
3. Inside [updateSecrets](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L690)
a. query K8s API using the names of generates secrets to retrieve the real username and password values
<--- rotation disabled, it would otherwise happen here --->
b. In the PgUser list [override](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L778) the random password from 1. with the actual password found in the secret
4. [syncRoles](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L882):
a. Takes the PgUser list from 1 and [queries](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L938) the database to compare later.
b. [Compares password](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/util/users/users.go#L61) of pgUser with password found in database and triggers an ALTER statement on diff.
c. All accumulated ALTER requests are executed

### Bug report

What happens when password rotation is enabled:

Everything except an extra step between 3a and 3b is the same. When rotation is enabled the operator will try to find a rotation date in the secret and sets one if it does not exist. When the rotation interval has passed a new rotation user will specified under the `username` key and a new `nextRotation` date is set.

:lady_beetle: : All other PgUser fields for the rotation users are [copied](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L828) from the unrotated user incl. the current password. A few lines later a new password is introduced for the secret struct.

:lady_beetle::lady_beetle: : As step 3b always overrides the generated password with the value from the secret, the rotation will result in a new password in the PgUser struct of the unrotated user. Thus, its password is updated in the database in 4c.

This makes the secret useless as it contains credentials that will not work for new pods. Running pods will also fail connecting to the database because the password of the used DB user has changed.

Fyi: Rotation users are added [straight from the `syncSecrets` ](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L659) method. We do not wait until syncRoles, because rotation users are not part of the PgUser list generated in 1.

### Possible solutions

The PR suggest to override the username in step 3b [like it's done with the password](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/cluster/sync.go#L778). Like this we could also remove code from `syncSecrets` to create rotation users. `syncRoles` would take care of it because it will find the rotation users in the PgUser list and incl. them in the database query. Thus creation would then be triggered [here](https://github.com/zalando/postgres-operator/blob/d209612b18122ae42250bd0517533fd7cbad5c66/pkg/util/users/users.go#L52).

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -818,10 +818,11 @@ func (c *Cluster) rotatePasswordInSecret(
 
 	// update password and next rotation date if configured interval has passed
 	if currentTime.After(nextRotationDate) {
-		newPassword := []byte(util.RandomPassword(constants.PasswordLength))
 		// create rotation user if role is not listed for in-place password update
 		if !util.SliceContains(c.Spec.UsersWithInPlaceSecretRotation, secretUsername) {
-			secret.Data["username"] = []byte(fmt.Sprintf("%s%s", secretUsername, currentTime.Format("060102")))
+			rotationUsername := fmt.Sprintf("%s%s", secretUsername, currentTime.Format("060102"))
+			secret.Data["username"] = []byte(rotationUsername)
+			c.logger.Infof("updating username in secret %s and creating rotation user %s in the database", secretName, rotationUsername)
 			// whenever there is a rotation, check if old rotation users can be deleted
 			*retentionUsers = append(*retentionUsers, secretUsername)
 		} else {
@@ -858,10 +859,10 @@ func (c *Cluster) rotatePasswordInSecret(
 
 			// when password of stream user is rotated in place, it should trigger rolling update in FES deployment
 			if roleOrigin == spec.RoleOriginStream {
-				c.logger.Warnf("secret of stream user %s changed", constants.EventStreamSourceSlotPrefix+constants.UserRoleNameSuffix)
+				c.logger.Warnf("password in secret of stream user %s changed", constants.EventStreamSourceSlotPrefix+constants.UserRoleNameSuffix)
 			}
 		}
-		secret.Data["password"] = newPassword
+		secret.Data["password"] = []byte(util.RandomPassword(constants.PasswordLength))
 		secret.Data["nextRotation"] = []byte(nextRotationDateStr)
 		updateSecretMsg = fmt.Sprintf("updating secret %s due to password rotation - next rotation date: %s", secretName, nextRotationDateStr)
 	}
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -638,7 +638,6 @@ func (c *Cluster) syncSecrets() error {
 	c.logger.Info("syncing secrets")
 	c.setProcessName("syncing secrets")
 	generatedSecrets := c.generateUserSecrets()
-	rotationUsers := make(spec.PgUserMap)
 	retentionUsers := make([]string, 0)
 	currentTime := time.Now()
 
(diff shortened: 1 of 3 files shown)


pkg/cluster/sync.go
```diff
@@ -927,8 +909,8 @@ func (c *Cluster) syncRoles() (err error) {
 	}
 
 	// copy map for ProduceSyncRequests to include also system users
-	for userName, pgUser := range c.pgUsers {
-		newUsers[userName] = pgUser
```
FxKu, line 931: key must be the same as name of the user like it will be set for `dbUsers` list. Otherwise, ProduceSyncRequests will try to create the rotation user again, although it already exists.

pkg/cluster/sync.go
```diff
@@ -642,29 +641,14 @@ func (c *Cluster) syncSecrets() error {
...
-		}
-		pgSyncRequests := c.userSyncStrategy.ProduceSyncRequests(spec.PgUserMap{}, rotationUsers)
-		if err = c.userSyncStrategy.ExecuteSyncRequests(pgSyncRequests, c.pgDb); err != nil {
-			return fmt.Errorf("error creating database roles for password rotation: %v", err)
-		}
-		if err := c.closeDbConn(); err != nil {
-			c.logger.Errorf("could not close database connection after creating users for password rotation: %v", err)
-		}
-	}
-
```
FxKu, line 667: instead of creating rotation users inside `updateSecret` we can use the existing logic called from `syncRoles` function

jopadi:
👍

FxKu:
:+1:

PR #2072, opened 2022-10-10
fix team member deprecation

FxKu:
#1953 introduced a bug in the user sync code. An extra pgUserMap [newUsers](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/sync.go#L942) was added to include systemUsers when passed to ProduceSyncRequests function. The info if a user is deprecated is overridden after newUsers variable is created, but only stored in the cluster's pgUsers variable.

Thus, ProduceSyncRequests does not know which users can be considered deprecated. The following bug can now happen:
1. Team members are marked as deprecated when they appear in the [pgUsersCache](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/sync.go#L932) but do not exist in desired c.pgUsers. This works.
2. Team members can re-join the team and the operator should rename the user back and grant LOGIN again. But it creates a new users.
2.1. [readPgUsersFromDatabase](https://github.com/zalando/postgres-operator/blob/84fe38a06908aa8adbe580041c133db84faf7121/pkg/cluster/database.go#L235) finds the deprecated user counterpart to a user that is desired in c.pgUsers. 
2.2. c.pgUsers is [updated](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/sync.go#L955) based on this finding, but not newUsers variable which is passed to ProduceSyncRequests.
2.3. ProduceSyncRequests [skips users marked as deleted](https://github.com/zalando/postgres-operator/blob/84fe38a06908aa8adbe580041c133db84faf7121/pkg/util/users/users.go#L47), but the info isn't found.
2.4. Since the new desired user does not exist in the database it is [created](https://github.com/zalando/postgres-operator/blob/84fe38a06908aa8adbe580041c133db84faf7121/pkg/util/users/users.go#L52).

Both the LOGIN team user and its deprecated NOLOGIN counterpart now exist in the database. The next time the member moves the operator will fail to rename the user because either of its versions already exist.

This PR fixes the bug and also enables the operator to deal with the erroneous setup of "redundant" roles. If rename requests fails the operator will only log the error but will not fail the sync.

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -104,18 +104,19 @@ func (c *Cluster) Sync(newSpec *acidv1.Postgresql) error {
 	if !(c.databaseAccessDisabled() || c.getNumberOfInstances(&newSpec.Spec) <= 0 || c.Spec.StandbyCluster != nil) {
 		c.logger.Debug("syncing roles")
 		if err = c.syncRoles(); err != nil {
-			err = fmt.Errorf("could not sync roles: %v", err)
-			return err
+			// remember all cached users in c.pgUsers
+			for cachedUserName, cachedUser := range c.pgUsersCache {
+				c.pgUsers[cachedUserName] = cachedUser
+			}
+			c.logger.Errorf("could not sync roles: %v", err)
 		}
 		c.logger.Debug("syncing databases")
 		if err = c.syncDatabases(); err != nil {
-			err = fmt.Errorf("could not sync databases: %v", err)
-			return err
+			c.logger.Errorf("could not sync databases: %v", err)
 		}
 		c.logger.Debug("syncing prepared databases with schemas")
 		if err = c.syncPreparedDatabases(); err != nil {
-			err = fmt.Errorf("could not sync prepared database: %v", err)
-			return err
+			c.logger.Errorf("could not sync prepared database: %v", err)
 		}
 	}
 
@@ -933,10 +934,7 @@ func (c *Cluster) syncRoles() (err error) {
 		}
 	}
 
-	// copy map for ProduceSyncRequests to include also system users
-	for userName, pgUser := range c.pgUsers {
-		newUsers[userName] = pgUser
-	}
+	// search also for system users
 	for _, systemUser := range c.systemUsers {
 		userNames = append(userNames, systemUser.Name)
 		newUsers[systemUser.Name] = systemUser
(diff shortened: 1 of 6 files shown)


pkg/util/users/users.go
```diff
@@ -82,22 +83,28 @@ func (strategy DefaultUserSyncStrategy) ProduceSyncRequests(dbUsers spec.PgUserM
...
+				_, originalUserWanted := newUsers[originalName]
+				_, originalUserAlreadyExists := dbUsers[originalName]
+				if !originalUserWanted || originalUserAlreadyExists {
+					continue
+				}
+				// a deleted dbUser has no NOLOGIN flag, so we can add the LOGIN flag
+				dbUser.Flags = append(dbUser.Flags, constants.RoleFlagLogin)
 			} else {
+				// user found in database and not wanted in newUsers - replace LOGIN flag with NOLOGIN
 				dbUser.Flags = util.StringSliceReplaceElement(dbUser.Flags, constants.RoleFlagLogin, constants.RoleFlagNoLogin)
```
FxKu, line 103: When a member leaves the team revoking the LOGIN right will still succeed but renaming will fail because the deprecated user already exists.

pkg/util/users/users.go
```diff
@@ -82,22 +83,28 @@ func (strategy DefaultUserSyncStrategy) ProduceSyncRequests(dbUsers spec.PgUserM
...
-			userFlags = append(userFlags, dbUser.Flags...)
 			if dbUser.Deleted {
-				dbUser.Flags = util.StringSliceReplaceElement(dbUser.Flags, constants.RoleFlagNoLogin, constants.RoleFlagLogin)
+				// * user with deletion suffix and NOLOGIN found in database
+				// grant back LOGIN and rename only if original user is wanted and does not exist in database
+				originalName := strings.TrimSuffix(name, strategy.RoleDeletionSuffix)
+				_, originalUserWanted := newUsers[originalName]
+				_, originalUserAlreadyExists := dbUsers[originalName]
+				if !originalUserWanted || originalUserAlreadyExists {
+					continue
```
FxKu, line 97: When a member joins again but original role already exist the operator will not even try to ALTER the deprecated user.

pkg/cluster/sync.go
```diff
@@ -950,13 +948,21 @@ func (c *Cluster) syncRoles() (err error) {
 	// update pgUsers where a deleted role was found
 	// so that they are skipped in ProduceSyncRequests
 	for _, dbUser := range dbUsers {
-		if originalUser, exists := deletedUsers[dbUser.Name]; exists {
-			recreatedUser := c.pgUsers[originalUser]
+		originalUsername, foundDeletedUser := deletedUsers[dbUser.Name]
+		// check if original user does not exist in dbUsers
+		_, originalUserAlreadyExists := dbUsers[originalUsername]
+		if foundDeletedUser && !originalUserAlreadyExists {
```
FxKu, line 954: With this PR the operator can revoke LOGIN from users but is [allowed to fail the renaming](https://github.com/zalando/postgres-operator/pull/2072/files#r991361190). This means, both users with and without suffix cannot login anymore. When the member joins the team again [altering the deprecated user is skipped](https://github.com/zalando/postgres-operator/pull/2072/files#r991363521), but LOGIN should be granted back to the role without the suffix. [ProduceSyncRequests](https://github.com/zalando/postgres-operator/blob/84fe38a06908aa8adbe580041c133db84faf7121/pkg/util/users/users.go#L70) can do this, but only if we do not mark the user as deleted here.

jopadi:
👍

FxKu:
:+1:

PR #2075, opened 2022-10-12
Fix upgrade command

hughcapet:
There is a regression caused by https://github.com/zalando/postgres-operator/pull/2056: upgrade run using postgres user was always failing because of an invalid option.  `-o pipefail` was moved directly to upgradeCommand to fix this.
\+ minor formatting corrections added


Jan-M:
:+1:

FxKu:
:+1:

PR #2083, opened 2022-10-19
skip db sync on failed initUsers during UPDATE

FxKu:
During UPDATE events initUsers can fail leading to a wrong c.pgUsers list and user syncing.
On SNYC events we [bail out immediately](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/sync.go#L51), however on UPDATE we carry on syncing everything we can. While it makes sense to sync K8s objects (stateful set, PDB, cron job) we better skip the database parts.

This PR also refactors small parts to better decide if initUsers is needed during UPDATE in the first place. For example, only do it if connectionPooler has been enabled or pooler user has changed. The stream user has been overlooked so far.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -227,6 +227,10 @@ func (c *Cluster) initUsers() error {
 	}
 
 	if err := c.initHumanUsers(); err != nil {
+		// remember all cached users in c.pgUsers
+		for cachedUserName, cachedUser := range c.pgUsersCache {
+			c.pgUsers[cachedUserName] = cachedUser
+		}
 		return fmt.Errorf("could not init human users: %v", err)
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/cluster/cluster.go
```diff
@@ -785,32 +790,39 @@ func (c *Cluster) Update(oldSpec, newSpec *acidv1.Postgresql) error {
...
-	if !sameUsers || !sameRotatedUsers || needConnectionPooler {
-		c.logger.Debugf("initialize users")
-		if err := c.initUsers(); err != nil {
-			c.logger.Errorf("could not init users: %v", err)
-			updateFailed = true
-		}
-
-		c.logger.Debugf("syncing secrets")
+	// Users
+	func() {
```
FxKu, line 794: putting this code into a func() to skip secrets sync

jopadi:
👍

PR #2085, opened 2022-10-21
create publication before creating logical replication slot

FxKu:
When `streams` feature is used, we were seeing error messages about [non-existing publication](https://www.postgresql.org/message-id/CADK3HH%2B9uPg%3DanB-thAMojVhsZk3tYWMwux4YmHB2%2BByJXcE3w%40mail.gmail.com), although the operator created them. Turns out, the operator created the logical replication slots first and was syncing publications after it waited for Patroni to add the slots. Before we merged #2034 we could have been lucky that slots were added later. But with the current code we will always run into this issue which can only be fixed by dropping the slot manually in the DB.

Publications have to be created first. There's also a nice [blog post](https://elephanttamer.net/?p=50) on this topic.

### DIFF
--- a/pkg/cluster/streams.go
+++ b/pkg/cluster/streams.go
@@ -315,6 +315,17 @@ func (c *Cluster) syncStreams() error {
 		return nil
 	}
 
+	// create publications to each created slot
+	c.logger.Debug("syncing database publications")
+	for publication, tables := range publications {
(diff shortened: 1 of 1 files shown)


idanovinda:
👍

FxKu:
:+1:

PR #2089, opened 2022-10-24
Allow drop slots when it gets deleted from the manifest

idanovinda:
### Problem
Users can [specify replication slots](https://github.com/zalando/postgres-operator/blob/master/manifests/complete-postgres-manifest.yaml#L122) in the Postgres manifest and they are [added](https://github.com/zalando/postgres-operator/blob/4786f53f033154009051c2b1dc300be48c55790a/pkg/cluster/sync.go#L572) to the database via [Patroni REST API](https://github.com/zalando/postgres-operator/blob/4786f53f033154009051c2b1dc300be48c55790a/pkg/cluster/sync.go#L628). When the slot is removed from the manifest, it won't get deleted in the database. Here we follow the convention of the operator never dropping database objects because manifest edits could have been wrong accidentally.

However, with logical replication slots the situation is a little different. If they are not used, diskspace will grow until it's exhausted. Our users are mostly not aware of this and think the slot will be gone when it's not desired anymore in the manifest. Some users check the database and still find the slot. So they call `pg_drop_replication_slot` function to drop it only to be surprised seconds later that the slot is recreated. 

### Solution
The operator should drop replication slots when they are removed from the manifest by updating the Patroni config. So far, our code only adds new slots or slots that differ from existing slots to the [configToSet](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/sync.go#L579). Patroni's config endpoint only merges the patches - it does not override the slots section.

To drop the replication slots that get deleted from the manifest, we can set the unused slots to null in the current Patroni's configuration and using `PATCH /config`.

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -526,7 +526,7 @@ class EndToEndTestCase(unittest.TestCase):
             """ % (slot_to_remove)
             
             self.eventuallyEqual(lambda: self.query_database(replica.metadata.name, "postgres", deleted_slot_query)[0], 0,
-                "The replication slot cannot be deleted")       
+                "The replication slot cannot be deleted", 10, 5)       
 
         except timeout_decorator.TimeoutError:
             print('Operator log: {}'.format(k8s.get_operator_log()))
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -402,6 +402,9 @@ class EndToEndTestCase(unittest.TestCase):
                     "slots": {
                         "test_slot": {
                             "type": "physical"
+                        },
+                        "test_slot_2": {
+                            "type": "physical"
                         }
                     },
                     "ttl": 29,
@@ -496,6 +499,34 @@ class EndToEndTestCase(unittest.TestCase):
                 "Previous max_connections setting not applied on master", 10, 5)
             self.eventuallyEqual(lambda: self.query_database(replica.metadata.name, "postgres", setting_query)[0], lower_max_connections_value,
                 "Previous max_connections setting not applied on replica", 10, 5)
+
+            # delete test_slot_2 from config
+            slot_to_remove = "test_slot_2"
+            pg_patch_slots = {
+                "spec": {
+                    "patroni": {
+                        "slots": {
+                            "test_slot": {
+                                "type": "physical"
+                            }
+                        }
+                    }
+                }
+            }
+
+            k8s.api.custom_objects_api.patch_namespaced_custom_object(
+                "acid.zalan.do", "v1", "default", "postgresqls", "acid-minimal-cluster", pg_patch_slots)
+            
+            self.eventuallyEqual(lambda: k8s.get_operator_state(), {"0": "idle"}, "Operator does not get in sync")
+
+            deleted_slot_query = """
+                SELECT count(*)
+                  FROM pg_replication_slots
+                 WHERE slot_name = '%s';
+            """ % (slot_to_remove)
+            
+            self.eventuallyEqual(lambda: self.query_database(replica.metadata.name, "postgres", deleted_slot_query)[0], 0,
+                "The replication slot cannot be deleted")       
 
         except timeout_decorator.TimeoutError:
             print('Operator log: {}'.format(k8s.get_operator_log()))
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -511,12 +511,15 @@ class EndToEndTestCase(unittest.TestCase):
 
             # patch new slot via Patroni REST
             patroni_slot = "test_patroni_slot"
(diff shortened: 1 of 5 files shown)


pkg/cluster/sync.go
```diff
@@ -568,8 +568,17 @@ func (c *Cluster) checkAndSetGlobalPostgreSQLConfiguration(pod *v1.Pod, effectiv
...
 
+	slotsToSet := make(map[string]interface{})
+	// check if there is any slot deletion
+	for slotName, effectiveSlot := range effectivePatroniConfig.Slots {
+		if desiredSlot, exists := desiredPatroniConfig.Slots[slotName]; exists {
+			if reflect.DeepEqual(effectiveSlot, desiredSlot) {
+				continue
+			}
+		}
+		slotsToSet[slotName] = nil
```
FxKu, line 579: this will also remove slots that were already existing in Postgres/Patroni config but not in the manifest, right? We should only remove slots when they were specified in the manifest before. I think we would need to cache all slots from oldSpec on UPDATE events and only do `slotsToSet[slotName] = nil` when slotName exists in that cache.

e2e/tests/test_e2e.py
```diff
@@ -494,6 +497,34 @@ def compare_config():
             self.eventuallyEqual(lambda: self.query_database(replica.metadata.name, "postgres", setting_query)[0], lower_max_connections_value,
                 "Previous max_connections setting not applied on replica", 10, 5)
 
+            # delete test_slot_2 from config
```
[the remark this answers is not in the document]
hughcapet: Another test case I can think of is the situation @FxKu mentioned: slots existing in the Patroni config before but not in the desired manifest

\+ It seems we lack extension of `TestCheckAndSetGlobalPostgreSQLConfiguration` unit test for the new feature 😇

pkg/cluster/sync.go
```diff
@@ -625,7 +625,7 @@ func (c *Cluster) checkAndSetGlobalPostgreSQLConfiguration(pod *v1.Pod, effectiv
 	}
 
 	// check if there exist only config updates that require a restart of the primary
-	if !util.SliceContains(restartPrimary, false) && len(configToSet) == 0 {
+	if len(restartPrimary) > 0 && !util.SliceContains(restartPrimary, false) && len(configToSet) == 0 {
```
FxKu, line 628: if there's nothing to set `requiresMasterRestart` would become `true` here which is wrong

e2e/tests/test_e2e.py
```diff
@@ -395,12 +395,20 @@ def test_config_update(self):
...
                      }
                  },
-                 "patroni": {
+                "users": {
+                    "test_user": []
+                },
+                "databases": {
+                    "foo": "test_user",
+                    "bar": "test_user"
+                },
```
FxKu, line 408: Do not create users and databases here but use existing ones. e2e test uses the minimal manifest so we already have `foo` (owner: zalando) and `bar` (owner: bar_owner). With test_user we get an extra secret which leads to the next test failing.

e2e/tests/test_e2e.py
```diff
@@ -497,6 +507,97 @@ def compare_config():
...
+                "acid.zalan.do", "v1", "default", "postgresqls", "acid-minimal-cluster", pg_delete_slot_patch)
+
+            self.eventuallyEqual(lambda: k8s.get_operator_state(), {"0": "idle"}, "Operator does not get in sync")
+            self.eventuallyTrue(compare_config, "Postgres config not applied")
+
+            deleted_slot_query = """
+                SELECT slot_name
+                  FROM pg_replication_slots
+                 WHERE slot_name = '%s';
+            """ % (slot_to_remove)
```
FxKu, line 577: might be worth to define a function for querying a slot and not repeat code

e2e/tests/test_e2e.py
```diff
@@ -395,12 +395,20 @@ def test_config_update(self):
...
                  },
-                 "patroni": {
+                "users": {
+                    "test_user": []
+                },
+                "databases": {
+                    "foo": "test_user",
+                    "bar": "test_user"
+                },
+                "patroni": {
```
FxKu, line 409:
```suggestion
                "patroni": {
```

Do not create users and databases here but use existing ones. e2e test uses the minimal manifest so we already have foo (owner: zalando) and bar (owner: bar_owner). With test_user we get an extra secret which leads to the next test failing.

e2e/tests/test_e2e.py
```diff
@@ -395,12 +395,13 @@ def test_config_update(self):
             "spec": {
                 "postgresql": {
                     "parameters": {
-                        "max_connections": new_max_connections_value
+                        "max_connections": new_max_connections_value,
+                        "wal_level": "logical"
                      }
                  },
-                 "patroni": {
+                "patroni": {
```
FxKu, line 402:
```suggestion
                 "patroni": {
```

FxKu:
:+1:

idanovinda:
👍

PR #2091, opened 2022-10-25
skip slots where publication sync failed

FxKu:
If tables for streams are not found in the database (maybe because of a typo in the table name) the publication sync will fail. The operator only returns a warning and continues to create the logical replication slot. This leads to the same error that #2085 already tried to solve. Even when creation of publications succeeds in future sync cycles, Postgres will return an error that they do not exist because the slots have been created before.

This PR suggests to only sync slots where publication sync succeeded.


FxKu:
:+1:

PR #2109, opened 2022-11-15
first sync wal_level then publications

FxKu:
Right now, the operator tries to change wal_level and adding extra slots in one config sync. However, publications cannot be created correctly until wal_level is switched. The publication will be "empty", meaning there's no entry in `publication_tables` catalog.

On subsequent sync cycles the operator cannot find the created publications because it uses a JOIN with `publication_tables`. With a LEFT JOIN it would find the empty state and fix it with ALTER PUBLICATION.

To avoid this situation in the first place, the PR suggests to do two config syncs. After changing the wal_level we need to wait for a Postgres restart on the next sync. After that creating publication, replication slots and stream resources should work in one go.


FxKu:
:+1:

idanovinda:
👍

PR #2117, opened 2022-11-24
Set securityContext for backup container

### LINKED ISSUE #1306: Run logical_backup as non-root

addreas:
Picking up from where #1407 left off, this fixes the same issue and also makes the refactoring changes suggested by @FxKu.

resolves #1306


FxKu:
👍

idanovinda:
👍

Issue #2122, opened 2022-12-02
ARM image support


lengrongfu:
We can adds workflow for multiarch builds (linux/amd64, linux/arm64) being published to ghcr.io, such as https://github.com/zalando/spilo/pull/790/files

mprimeaux:
@lengrongfu Any help needed with the PR?

PR #2127, opened 2022-12-06
add build arm image pipeline

lengrongfu:
issue: https://github.com/zalando/postgres-operator/issues/2122

### DIFF
--- a/.github/workflows/publish-ghcr-container.yaml
+++ b/.github/workflows/publish-ghcr-container.yaml
@@ -1,0 +1,51 @@
+name: Publish multiarch images on ghcr.io
+on:
+  push:
+    tags:
+      - '*'
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ${{ github.repository }}
+
+jobs:
+  docker:
+    name: Build and push postgres-operator multiarch images
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: image name
+        run: |
+          VERSION=`git describe --tags --always --dirty`
+          TAG=${VERSION}
+          IMAGE=${{ env.REGISTRY }}/acid/postgres-operator:${TAG}
+          echo "NAME=$IMAGE" >> $GITHUB_OUTPUT
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+
+      - name: Login to GHCR
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run unit tests
+        run: make test
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          platforms: |
+            linux/amd64
+            linux/arm64
+          file: ./docker/Dockerfile-multi-arch
+          build-args: [base-image=alpine:3.15]
+          tags: "${{ steps.image.outputs.NAME }}"
--- a/docker/Dockerfile-multi-arch
+++ b/docker/Dockerfile-multi-arch
@@ -1,0 +1,23 @@
+ARG base_image=alpine:3.15
+FROM golang:1.18.3 as build
+WORKDIR /workspace
+ENV GO111MODULE=on \
+    GOPROXY=https://goproxy.cn,direct \
+    BINARY=postgres-operator
+COPY . .
+RUN CGO_ENABLED=0 go build -o build/linux/${BINARY} ./cmd/main.go
+
+FROM ${base_image}
+LABEL maintainer="Team ACID @ Zalando <team-acid@zalando.de>"
+# We need root certificates to deal with teams api over https
+RUN apk --no-cache add curl \
+    apk --no-cache add ca-certificates
+
+COPY --from=build /workspace/build/linux /
+
+RUN addgroup -g 1000 pgo
+RUN adduser -D -u 1000 -G pgo -g 'Postgres Operator' pgo
+
+USER 1000:1000
+
+ENTRYPOINT ["/postgres-operator"]
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,9 @@ docker: ${DOCKERDIR}/${DOCKERFILE} docker-context
 	echo "git describe $(shell git describe --tags --always --dirty)"
 	cd "${DOCKERDIR}" && docker build --rm -t "$(IMAGE):$(TAG)$(CDP_TAG)$(DEBUG_FRESH)$(DEBUG_POSTFIX)" -f "${DOCKERFILE}" .
 
+build-multi-arch:
+	docker buildx build --platform linux/arm64,linux/amd64 --push -t "$(IMAGE):$(TAG)" -f ./${DOCKERDIR}/Dockerfile-multi-arch .
+
 indocker-race:
 	docker run --rm -v "${GOPATH}":"${GOPATH}" -e GOPATH="${GOPATH}" -e RACE=1 -w ${PWD} golang:1.17.3 bash -c "make linux"
 


lengrongfu:
@FxKu Hi, PTAL.

docker/Dockerfile-multi-arch
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.18.3 as build
+WORKDIR /workspace
+ENV GO111MODULE=on \
+    GOPROXY=https://goproxy.cn,direct \
+    BINARY=postgres-operator
+COPY . .
+RUN GOOS=linux GOARCH=$(go env GOARCH) CGO_ENABLED=0 go build -o build/linux/${BINARY} ./cmd/main.go
+
+FROM registry.opensource.zalan.do/library/alpine-3.15:latest
```
hughcapet, line 9: it is actually an amd-only image. You better use the official one
But what is the reason for creating a separate Dockerfile for a multi-arch build?

.github/workflows/publish-ghcr-container.yaml
```diff
@@ -0,0 +1,45 @@
...
+        uses: docker/setup-qemu-action@v2
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+      - name: Login to GHCR
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
```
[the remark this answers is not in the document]
FxKu: indeed, pls add unit tests

.github/workflows/publish-ghcr-container.yaml
```diff
@@ -0,0 +1,45 @@
+name: Publish multiarch images on ghcr.io
+on:
+  push:
+    branches:
+      - master
```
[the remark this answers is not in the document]
FxKu: what do you suggest? only on every tagged release? Later would be fine with me

.github/workflows/publish-ghcr-container.yaml
```diff
@@ -0,0 +1,45 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          platforms: |
+            linux/amd64
+            linux/arm64
+          file: ./docker/Dockerfile-multi-arch
+          tags: "${{ steps.image.outputs.NAME }}"
```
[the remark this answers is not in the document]
lengrongfu: I don't see any difference between them.

docker/Dockerfile-multi-arch
lengrongfu: 1. `registry.opensource.zalan.do/library/alpine-3.15:latest` I'm not sure if it's the same as `docker.io/library/alpine-3.15:latest`
2. Dockerfile-multi-arch was created because I didn't want to break existing code.

.github/workflows/publish-ghcr-container.yaml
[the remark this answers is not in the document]
lengrongfu: Done. add `make test` step.

.github/workflows/publish-ghcr-container.yaml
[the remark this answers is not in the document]
hughcapet: yes, I think pushing on tag is enough
At least, it is what we do in Spilo

.github/workflows/publish-ghcr-container.yaml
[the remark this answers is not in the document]
hughcapet: new line

docker/Dockerfile-multi-arch
hughcapet: it is not the same but we can make image configurable via build arg and use different base images for different purposes

lengrongfu: About using different base images I don't quite understand what you want to do, can you give more hints? 
@hughcapet

lengrongfu: @hughcapet Hi, PTAL.

docker/Dockerfile-multi-arch
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.18.3 as build
+WORKDIR /workspace
+ENV GO111MODULE=on \
+    GOPROXY=https://goproxy.cn,direct \
+    BINARY=postgres-operator
+COPY . .
+RUN GOOS=linux GOARCH=$(go env GOARCH) CGO_ENABLED=0 go build -o build/linux/${BINARY} ./cmd/main.go
+
+FROM {base_image}
```
verbotenj, line 9: I think you would need ARG BASE_IMAGE=alpine:3.15 because the official image has amd/arm support.
https://github.com/zalando/spilo/blob/master/postgres-appliance/Dockerfile#L18

docker/Dockerfile-multi-arch
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.18.3 as build
```
verbotenj, line 1: I would move code into a single Dockerfile

docker/Dockerfile-multi-arch
```diff
@@ -0,0 +1,23 @@
...
+WORKDIR /workspace
+ENV GO111MODULE=on \
+    GOPROXY=https://goproxy.cn,direct \
+    BINARY=postgres-operator
+COPY . .
+RUN CGO_ENABLED=0 go build -o build/linux/${BINARY} ./cmd/main.go
+
+FROM {base_image}
```
verbotenj, line 9: If someone clones, the project won't have {base_image} defined and will need help figuring out where it comes from. If you use, ARG will help with understanding.
Instead of  sed -i '' 's/{base_image}/${base_arm_image}/g' ./docker/Dockerfile-multi-arch --
use build-args from build-push-action

.github/workflows/publish-ghcr-container.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          platforms: |
+            linux/amd64
+            linux/arm64
+          file: ./docker/Dockerfile-multi-arch
+          tags: "${{ steps.image.outputs.NAME }}"
```
verbotenj, line 53: If you add `build-args: [base-image=alpine:3.15]`. You can control ARG base-image in Dockerfile

docker/Dockerfile-multi-arch
lengrongfu: Thanks for the tip.

.github/workflows/publish-ghcr-container.yaml
lengrongfu: Thanks for the tip.

.github/workflows/publish-ghcr-container.yaml
```diff
@@ -0,0 +1,44 @@
...
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          platforms: |
+            linux/amd64
+            linux/arm64
+          file: ./docker/Dockerfile-multi-arch
+          build-args: [base-image=alpine:3.15]
+          tags: "${{ steps.image.outputs.NAME }}"
```
verbotenj, line 44: I notice `steps.image` was removed where you were getting tags. Maybe you want to explore
https://github.com/docker/metadata-action or keep
```      - name: image name
         id: image
         run: |
           VERSION=`git describe --tags --always --dirty`
           TAG=${VERSION}
           IMAGE=${{ env.REGISTRY }}/acid/postgres-operator:${TAG}
           echo "NAME=$IMAGE" >> $GITHUB_OUTPUT```

Comment (lengrongfu) [reply .github/workflows/publish-ghcr-container.yaml:51]: 👍

**Comment by verbotenj** [comment]
@lengrongfu @FxKu PTAL

**Comment by mprimeaux** [comment]
Just checking to see if I can help in any way to support this merge. Much thanks in advance.

File: docker/Dockerfile-multi-arch [hunk 10]
```diff
@@ -0,0 +1,23 @@
+FROM golang:1.18.3 as build
+WORKDIR /workspace
+ENV GO111MODULE=on \
+    GOPROXY=https://goproxy.cn,direct \
+    BINARY=postgres-operator
+COPY . .
+RUN CGO_ENABLED=0 go build -o build/linux/${BINARY} ./cmd/main.go
+
+ARG base-image
```
verbotenj, line 9: ARG base-image will need to be defined at the top of the file to be defined as global and will be available in FROM.
I recommend setting the default value for the `base-image=alpine:3.15` in the Dockerfile, which allows it to be used locally with ease.

docker/Dockerfile-multi-arch
```diff
@@ -0,0 +1,23 @@
...
+FROM ${base-image}
+LABEL maintainer="Team ACID @ Zalando <team-acid@zalando.de>"
+# We need root certificates to deal with teams api over https
+RUN apk --no-cache add curl
+RUN apk --no-cache add ca-certificates
```
verbotenj, line 14: You could merge RUN in one command

verbotenj:
Another thing to consider is that in the operator helm chart, there are multiple images that can be specified.
Example: https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values.yaml#L351
Dockerfile for it
https://github.com/zalando/postgres-operator/blob/master/docker/logical-backup/Dockerfile
To have everything running only on arm64 arch.

FxKu:
Closed in favor for #2268. Thanks again to all contributors.

PR #2134, opened 2022-12-09
fix migration of single-node clusters

FxKu:
#1984 introduced a bug when fetching the master candidate pod the name is not set for single node clusters. Results in the following error:

```
could not get master candidate pod: resource name may not be empty
```

The pod needs only to be fetched when there's a replica that can become the master. If there's only the master then we already fetched the pod above.


FxKu:
:+1:

sdudoladov:
:+1:

PR #2137, opened 2022-12-14
use random short name for stream CRDs

FxKu:
So far, the name of stream CRDs is concatenated from Postgres cluster name + applicationID if the stream. However, this can produce quite long resource names which leads to errors when creating deployments by the stream operator. It sets a label with the CRD name and that is only allowed to be 63 characters long in K8s.

Postgres cluster names can only be 53 characters long. I think, it still makes sense to re-use the cluster name in the stream CRD, but instead of the applicationID we can append a random string (+ EventStreamSourceSlotPrefix) to stay below 64 characters.

The random string makes the search for a particular stream CRD non-deterministic. A list request is required, followed by comparing the applicationID. Labels must be set to filter for the stream CRDs belonging to the Postgres cluster. By doing so, the cached streamApplicationIds of the cluster are not necessary anymore.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -91,7 +91,6 @@ type Cluster struct {
 	currentProcess      Process
 	processMu           sync.RWMutex // protects the current operation for reporting, no need to hold the master mutex
 	specMu              sync.RWMutex // protects the spec for reporting, no need to hold the master mutex
-	streamApplications  []string
 	ConnectionPooler    map[PostgresRole]*ConnectionPoolerObjects
 	EBSVolumes          map[string]volumes.VolumeProperties
 	VolumeResizer       volumes.VolumeResizer
--- a/pkg/cluster/streams.go
+++ b/pkg/cluster/streams.go
@@ -15,15 +15,16 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func (c *Cluster) createStreams(appId string) error {
+func (c *Cluster) createStreams(appId string) (*zalandov1.FabricEventStream, error) {
 	c.setProcessName("creating streams")
 
 	fes := c.generateFabricEventStream(appId)
-	if _, err := c.KubeClient.FabricEventStreams(c.Namespace).Create(context.TODO(), fes, metav1.CreateOptions{}); err != nil {
-		return err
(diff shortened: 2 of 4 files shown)


pkg/cluster/streams.go
```diff
@@ -15,15 +15,16 @@ import (
...
 	fes := c.generateFabricEventStream(appId)
-	if _, err := c.KubeClient.FabricEventStreams(c.Namespace).Create(context.TODO(), fes, metav1.CreateOptions{}); err != nil {
-		return err
+	streamCRD, err := c.KubeClient.FabricEventStreams(c.Namespace).Create(context.TODO(), fes, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
 	}
 
-	return nil
+	return streamCRD, nil
```
FxKu, line 27: return stream because the name is not deterministic anymore

pkg/cluster/cluster.go
```diff
@@ -91,7 +91,6 @@ type Cluster struct {
 	currentProcess      Process
 	processMu           sync.RWMutex // protects the current operation for reporting, no need to hold the master mutex
 	specMu              sync.RWMutex // protects the spec for reporting, no need to hold the master mutex
-	streamApplications  []string
```
FxKu, line 94: we don't need to remember stream apllicationIDs when we can list all stream CRDs belonging to a Postgres cluster with a K8s API call

pkg/cluster/streams.go
```diff
@@ -184,8 +191,10 @@ func (c *Cluster) generateFabricEventStream(appId string) *zalandov1.FabricEvent
 			Kind:       constants.EventStreamCRDKind,
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name:        fmt.Sprintf("%s-%s", c.Name, appId),
+			// max length for cluster name is 53 so we can only add 10 more characters / numbers
+			Name:        fmt.Sprintf("%s-%s-%s", c.Name, constants.EventStreamSourceSlotPrefix, util.RandomPassword(5)),
 			Namespace:   c.Namespace,
+			Labels:      c.labelsSet(true),
```
FxKu, line 197: use the same labels set what we have for other resources incl. cluster labels like `application: spilo`. Might be debatable but this way it's easier to find stream CRDs created by the Postgres Operator. Alternative: filter by ownerReference, if this is even possible?

jopadi:
👍

FxKu:
:+1:

PR #2139, opened 2022-12-15
Add support for PostgreSQL 15

yoshihikoueno:
Hi,

This PR adds support for PostgreSQL 15 and drops support for 9.5 and 9.6.
As for Spilo image, I've set `spilo-cdp-15:2.1-p253` for now, but I'm willing to change it to non-CDP one once it gets released.

Also, I chose not to touch under `./e2e` as there's currently no spilo image for it as far as I know.

PR #2140, opened 2022-12-15
enable operator support for pg15 and drop support for 9.5 and 9.6

FxKu:
- [x] include 15 in CRDs
- [x] remove 9.5 and 9.6 from operator (keep 9.6 only for backup location)
- [x] bump `minimal_major_version` from "9.6" to "11"
- [x] update some unit test examples to use version 15
- [x] update postgres doc links to 15
- [x] incl. postgres 15 client for logical backup image

ToDo for #2139:
- [ ] provide spilo-15 image
- [ ] update `target_major_version` to 15
- [ ] update README with 15 support
- [ ] bump example manifests to 15
- [ ] include 15 in UI deployment

### DIFF
--- a/ui/operator_ui/main.py
+++ b/ui/operator_ui/main.py
@@ -321,7 +321,7 @@ DEFAULT_UI_CONFIG = {
     'users_visible': True,
     'databases_visible': True,
     'resources_visible': RESOURCES_VISIBLE,
-    'postgresql_versions': ['11', '12', '13', '14', '15'],
+    'postgresql_versions': ['11', '12', '13', '14'],
     'dns_format_string': '{0}.{1}',
     'pgui_link': '',
     'static_network_whitelist': {},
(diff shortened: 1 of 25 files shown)


ui/operator_ui/main.py
```diff
@@ -321,7 +321,7 @@ def index():
     'users_visible': True,
     'databases_visible': True,
     'resources_visible': RESOURCES_VISIBLE,
-    'postgresql_versions': ['11','12','13','14'],
+    'postgresql_versions': ['11', '12', '13', '14'],
```
hughcapet, line 324: why there is no 15 here?  (not sure what this file is about :) )

FxKu: this is for the UI where we should make 15 available only when it#s ready to be used

hughcapet:
👍

FxKu:
:+1:

PR #2146, opened 2022-12-27
Add Support for Custom TLS Certificates in Connection Pooler

jeremie-seguin:
This pull request is heavily inspired by [this PR](https://github.com/zalando/postgres-operator/pull/1232) by @borchero and try to solve [#1230](https://github.com/zalando/postgres-operator/issues/1230)  But do not include the additional annotations on the pgBouncer deployment.

The entrypoint.sh in pgbouncer docker image also need to be updated to use the certificates when present like so:
``` bash
if [ -z "${CONNECTION_POOLER_CLIENT_TLS_CRT}" ]; then
    openssl req -nodes -new -x509 -subj /CN=spilo.dummy.org \
        -keyout /etc/ssl/certs/pgbouncer.key \
        -out /etc/ssl/certs/pgbouncer.crt
else
    ln ${CONNECTION_POOLER_CLIENT_TLS_CRT} /etc/ssl/certs/pgbouncer.crt
    ln ${CONNECTION_POOLER_CLIENT_TLS_KEY} /etc/ssl/certs/pgbouncer.key
fi
```

however the dockerfile of pgbouncer included in the operator does not seems to be public, is it accessible somewhere ? 

If this solution is accepted, certificate renewal will need to be handled either on the operator side with a restart of the pooler deployment or directly on pgbouncer if the reload command is sufficient.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -603,7 +603,7 @@ spec:
                     default: "pooler"
                   connection_pooler_image:
                     type: string
-                    default: "registry.opensource.zalan.do/acid/pgbouncer:master-24"
+                    default: "registry.opensource.zalan.do/acid/pgbouncer:master-26"
                   connection_pooler_max_db_connections:
                     type: integer
                     default: 60
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -396,7 +396,7 @@ configConnectionPooler:
   # db user for pooler to use
   connection_pooler_user: "pooler"
   # docker image
-  connection_pooler_image: "registry.opensource.zalan.do/acid/pgbouncer:master-24"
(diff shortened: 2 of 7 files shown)


FxKu:
`CN=spilo.dummy.org`

Can this be any string? Does it matter or should it be configurable?

jeremie-seguin:
@FxKu The `spilo.dummy.org` is the value currently used in the self signed certificate of pgbouncer:master-24, it will not be used if a `CONNECTION_POOLER_CLIENT_TLS_CRT` is provided to the operator

FxKu:
@jeremie-seguin I've added the change you've suggested to the entrypoint.sh. Can you bump the pgbouncer docker image tag to master-26, pls?

FxKu:
@jeremie-seguin could you maybe cover the new feature in `connection_pooler_test.go`. Just a check that makes sure the volume section is there if `tls` options are specified in the manifest.

jopadi:
👍

FxKu:
:+1:

FxKu:
Thanks @jeremie-seguin . We will add the unit test ourselves.

PR #2150, opened 2022-12-29
Enable logging of all operator statements

Jan-M:
Enable logging of all operator statements


Jan-M:
https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING

PR #2152, opened 2022-12-30
fix syncing of stream CRDs

FxKu:
With #2137 merged, stream CRD names are no longer known beforehand and, therefore, can only be listed using a label selector. The check if a stream CRD does exist and can be updated now happens within a loop over all fetched CRDs. If found that loop is exited - however NOT the loop over application Ids which handles creation of stream CRDs.

This means, stream CRDs are now created on each sync which succeeds because the name now has a random suffix. An extra bool flag is needed per application Id iteration to avoid that.

Unit tests were extended to cover this case.

### DIFF
--- a/pkg/cluster/streams.go
+++ b/pkg/cluster/streams.go
@@ -374,9 +374,12 @@ func (c *Cluster) createOrUpdateStreams() error {
 	}
 
 	for _, appId := range appIds {
+		streamExists := false
+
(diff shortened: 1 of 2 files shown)


jopadi:
👍

jopadi:
👍

FxKu:
:+1:

PR #2155, opened 2023-01-02
polish global config about sharing postgresql-run socket

FxKu:
Follow up to #962 which we have just merged.

- Call the feature `share_pgsocket_with_sidecars` everywhere. Golang name will be `SharePgSocketWithSidecars`.
- Move it to the right alphabetical place in CRD schema
- Cover it also in helm chart
- Extend documentation
- Provide unit test


jopadi:
👍

FxKu:
:+1:

PR #1854, opened 2022-04-15
Bump k8s api to v0.23.5 for operator v1.9.0

dmvolod:
@FxKu could you please to have a look at the update to the new k8s api which is actual right now
Thanks in advance

PR #2158, opened 2023-01-03
Update k8s API for kubectl-pg and regen generated code

dmvolod:
This PR updates k8s API version for the kubectl-pg CLI sub-project, adds `vendor` into the .gitignore and regen generated code with new 2023 year substitution :tada:


FxKu:
I was about to do the same :smiley: 

Can we switch to a higher K8s version than v0.23.5? Maybe v0.25.5?
Maybe we can also bump go version to 1.18.

dmvolod:
> I was about to do the same smiley

Oh, didn't know about this

> 
> Can we switch to a higher K8s version than v0.23.5 (maybe rebase because we merged #1854)? Maybe v0.25.5? Maybe we can also bump go version to 1.18.

Are you talking about whole project update to the k8s API version for whole project or just for the kebectl-pg?
In this case, golang version for the k8s API and client-go should be in sync with parent projects, i.e.
https://github.com/kubernetes/client-go/blob/v0.24.9/go.mod
https://github.com/kubernetes/kubernetes/blob/v1.24.9/go.mod

From support point of view, k8s community guaranties interoperability between client (operator) and k8s servers in the ±1 version.  In this case, k8s API v0.24.9 will support k8s from v1.23.x to v1.25.x.

FxKu:
then let's stick to 0.23.5 because we are only jumping on K8s 1.22

FxKu:
:+1:

jopadi:
👍

PR #2159, opened 2023-01-04
spec.env can override clone and standby variables

FxKu:
Follow up to #1976 but only changes the order for spec.Env overriding variables that would be set when clone or standby section exist. The later still takes precedence over values from global pod_environment_secret/configmap.


idanovinda:
👍

FxKu:
:+1:

PR #2165, opened 2023-01-09
Ineffectual assignment of the envVars for connection pooler

dmvolod:
Ineffectual assignment of the envVars for connection pooler if `spec.TLS` is specified.
@FxKu please have a look at this bug and try to fit it into the 1.9 if possible
Test case can be added, but this case should be covered by the linter

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -18,7 +18,8 @@ trap "cleanup" EXIT SIGINT
 bash "${CODEGEN_PKG}/generate-groups.sh" all \
   "${OPERATOR_PACKAGE_ROOT}/pkg/generated" "${OPERATOR_PACKAGE_ROOT}/pkg/apis" \
   "acid.zalan.do:v1 zalando.org:v1" \
-  --go-header-file "${SCRIPT_ROOT}"/hack/custom-boilerplate.go.txt
+  --go-header-file "${SCRIPT_ROOT}"/hack/custom-boilerplate.go.txt \
+  -o ./
 
 cp -r "${OPERATOR_PACKAGE_ROOT}"/pkg/* "${TARGET_CODE_DIR}"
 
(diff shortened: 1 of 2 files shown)


hack/update-codegen.sh
```diff
@@ -18,7 +18,8 @@ trap "cleanup" EXIT SIGINT
 bash "${CODEGEN_PKG}/generate-groups.sh" all \
   "${OPERATOR_PACKAGE_ROOT}/pkg/generated" "${OPERATOR_PACKAGE_ROOT}/pkg/apis" \
   "acid.zalan.do:v1 zalando.org:v1" \
-  --go-header-file "${SCRIPT_ROOT}"/hack/custom-boilerplate.go.txt
+  --go-header-file "${SCRIPT_ROOT}"/hack/custom-boilerplate.go.txt \
+  -o ./
```
FxKu, line 22: unrelated change. What is it for?

FxKu:
Thanks for raising @dmvolod. We wanted to add a unit test todays and would have stumbled over it, I guess :smiley: 
@jeremie-seguin fyi

hack/update-codegen.sh
dmvolod: This is fix when GOPATH installed explicitly (for example in GoLand IDE and multiple golang version installed) code generation  will be targeted to the GOPATH\src instead of the root project path.

dmvolod:
> We wanted to add a unit test today and would have stumbled over it

Seems to `TestConnectionPoolerPodSpec` need to be refactoring a bit and not ready for easy env vars validation. Probably it can be postponed to next release, as current situation can be handled by the linting step in the future, but complex test solution for env vars need much more time.
@FxKu wdyt about it?

FxKu:
:+1:

jopadi:
👍

FxKu:
Thanks @dmvolod. Yeah we also noticed that pooler unit test need more work than what @jopadi and me expected.

PR #2171, opened 2023-01-13
better backwards compatibility with old DNS name format for LBs

FxKu:
With #2011 we changed the default DNS name format template used when LoadBalancers are enabled. For backwards compatibility we are already appending a second DNS entry when the cluster name still starts with the team Id (mandatory before #2001), because we do not want to break existing connections.

For the old format we only assumed the default template value `"{cluster}.{team}.{hostedzone}"`. There's no option to specify what you have used before. If you have customized the template, all apps would need to be redeployed using the new DNS format. To avoid such a hassle this PR suggest to make the old format configurable with `master|replica_old_dns_name_format` options.

The unit test in cluster_test_go has be updated to also use a template different from the default value.

### DIFF
--- a/pkg/cluster/util.go
+++ b/pkg/cluster/util.go
@@ -545,14 +545,14 @@ func (c *Cluster) replicaDNSName() string {
 }
 
 func (c *Cluster) oldMasterDNSName(clusterName string) string {
-	return strings.ToLower(c.OpConfig.MasterOldDNSNameFormat.Format(
+	return strings.ToLower(c.OpConfig.MasterLegacyDNSNameFormat.Format(
 		"cluster", clusterName,
 		"team", c.teamName(),
 		"hostedzone", c.OpConfig.DbHostedZone))
 }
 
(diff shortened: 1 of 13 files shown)


pkg/cluster/util.go
```diff
@@ -517,36 +517,43 @@ func (c *Cluster) dnsName(role PostgresRole) string {
...
 	return dnsString
 }
 
 func (c *Cluster) masterDNSName() string {
-	return strings.ToLower(c.OpConfig.MasterDNSNameFormat.Format(
+	return strings.ToLower(c.OpConfig.MasterLBDNSNameFormat.Format(
 		"cluster", c.Name,
 		"namespace", c.Namespace,
-		"team", c.teamName(),
 		"hostedzone", c.OpConfig.DbHostedZone))
```
FxKu, line 535: docs say `No other placeholders are allowed!` so we should not have `"team", c.teamName()`. Team for a cluster can change. Should not be part of the DNS anyway.

FxKu:
:+1:

idanovinda:
👍

PR #2175, opened 2023-01-17
do not rotate secrets for standby clusters

FxKu:
Password rotation should not be allowed for standby clusters because rotation users cannot be added anyway.
This is particularly helpful if the source clusters has no rotation enabled (e.g. is not maintained by Postgres Operator).
However, if the secrets are updated for the source cluster, then credentials must be synced manually - like during the initial setup of a standby.


Jan-M:
👍

FxKu:
:+1:

PR #2177, opened 2023-01-17
bump to v1.9.0

FxKu:
* bump operator image in deployment manifest
* bump logical backup default image
* update version in helm charts and create new packages
* drop v1.6.x tgz packages from helm folders


charts/postgres-operator-ui/values.yaml
```diff
@@ -8,7 +8,7 @@ replicaCount: 1
 image:
   registry: registry.opensource.zalan.do
   repository: acid/postgres-operator-ui
-  tag: v1.8.2
+  tag: v1.9.0
```
dmvolod, line 11: @FxKu do you mind that `master` branch should contains `latest` tag and specific version tag should be present inside the chart archive only. Otherwise is not possible to install chart from the current master in the development (with some medication manifests values).
wdyt about it?

dmvolod: Yes, indeed. We could always have the `latest` tag and `imagePullPolicy: Always` on the `master` and create release script to automate this operation.

katlol: Keep in mind! The default `imagePullPolicy` always pulls the `:latest` tag! As such, it should not be specified for this usecase.

https://kubernetes.io/docs/concepts/containers/images/#imagepullpolicy-defaulting

dmvolod: @katlol I know about it, thank you. But for this case, it's better to specified it to simplify release scripts for helm chart if this approach could be implemented.

monotek:
Before thats release it would be realy nice to add an ci step, which pushed the chart to some real helm repo.

This can be achived verya easiyl by using: https://github.com/helm/chart-releaser-action

FxKu:
@monotek we will do it for the next release when we have CI for multi-arch operator image

idanovinda:
👍

FxKu:
:+1:

PR #2178, opened 2023-01-17
UI: fix default cpu memory

idanovinda:
The default memory and CPU that currently showing does not reflect the defined values in the environment variable.


idanovinda:
👍

PR #2183, opened 2023-01-20
copy rolconfig during password rotation

FxKu:
Rotation user inherit all privileges from the original user. But if somebody manually changed the rolconfig of a user this will not be present in the rotation users since Postgres does not inherit it. This PR suggest to add another flag to pgUser struct to notify the syncRoles function that a rotation took place in the syncSecret cycle.

When this happens the original user is added to the list of database roles to query to get the rolconfig and copy it. After that part, the original user must be removed from dbUsers map, because it does not exist in c.pgUsers, which usually trigger role deprecation (adding deletion suffix).

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -1578,7 +1578,7 @@ class EndToEndTestCase(unittest.TestCase):
              WHERE rolname LIKE 'foo_user%';
         """
         self.eventuallyEqual(lambda: len(self.query_database(leader.metadata.name, "postgres", user_query)), 2,
-            "Found incorrect number of rotation users after disbaling password rotation", 10, 5)
+            "Found incorrect number of rotation users after disabling password rotation", 10, 5)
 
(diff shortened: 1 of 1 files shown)


e2e/tests/test_e2e.py
```diff
@@ -1559,7 +1578,7 @@ def test_password_rotation(self):
              WHERE rolname LIKE 'foo_user%';
         """
         self.eventuallyEqual(lambda: len(self.query_database(leader.metadata.name, "postgres", user_query)), 2,
-            "Found incorrect number of rotation users", 10, 5)
+            "Found incorrect number of rotation users after disbaling password rotation", 10, 5)
```
idanovinda, line 1581: disbaling -> disabling

idanovinda, line 1581:
```suggestion
            "Found incorrect number of rotation users after disabling password rotation", 10, 5)
```

FxKu:
:+1:

idanovinda:
👍

PR #2187, opened 2023-01-25
set wal_level for streams in statefulSet sync

FxKu:
The current flow when adding streams in the manifest is as follows:

SNYC 1:
1. [List Pods] `syncStatefulSet` compares desired (manifest) and effective Patroni config and restarts pods
2. [List Pods] `syncStreams` (calling `syncPostgresConfig`) sets `wal_level: logical` which requires a Postgres restart only happening in `syncStatefulSet`, exit `syncStreams` early

SNYC 2:
1. [List Pods] `syncStatefulSet` compares desired (manifest) and effective Patroni config and restarts pods
2. [List Pods] `syncStreams` (calling `syncPostgresConfig`) sets `wal_level: logical` but config is already updated - no restart required anymore, proceed with publications and slots
3. [List Pods] `syncStreams` (calling `syncPostgresConfig`) syncing replication slots in Patroni config

The PR proposes the following refactoring:

SNYC 1:
1. [List Pods] `syncStatefulSet` sets `wal_level: logical` if streams are defined and then compares desired (manifest) and effective Patroni config (calling `syncPatroniConfig`) and restarts pods
2. [List Pods] `syncStreams` expectes `wal_level` to be already configured and calls `syncPatroniConfig` to reflect the slots required for streams in Patroni config

It brings the following advantages:

* 2 instead of 3 list pod calls to K8s API during each SYNC
* stream CRD(s) will be created on the same UPDATE event when `streams` were added to the manifest - no need to wait for next SYNC
* `syncStatefulSet` and `syncStreams` sharing the same code for syncing Postgres config
* collect all error msg on config syncs and log them as a warning
* less code in `syncStatefulSet` due to new functions makes the code easier to understand

### DIFF
--- a/pkg/cluster/streams.go
+++ b/pkg/cluster/streams.go
@@ -79,38 +79,6 @@ func gatherApplicationIds(streams []acidv1.Stream) []string {
 	return appIds
 }
 
-func (c *Cluster) syncPostgresConfig(requiredPatroniConfig acidv1.Patroni) (bool, error) {
-	errorMsg := "no pods found to update config"
-
-	// if streams are defined wal_level must be switched to logical
-	requiredPgParameters := map[string]string{"wal_level": "logical"}
-
-	// apply config changes in pods
-	pods, err := c.listPods()
-	if err != nil {
-		errorMsg = fmt.Sprintf("could not list pods of the statefulset: %v", err)
-	}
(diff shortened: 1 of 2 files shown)


idanovinda:
👍

FxKu:
:+1:

PR #2188, opened 2023-01-25
add pooler suffix to DNS annotation of pooler LoadBalancer service

### LINKED ISSUE #2180: Identical external-dns hostnames for load balancer type services for the master and the master pooler

- **Which image of the operator are you using?** local build from master branch
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Kubernetes 1.24.9 on Talos on OpenStack
- **Type of issue?** Bug or feature request

Hello,

I would like to create a cluster with pooler enabled, load balancer and external-dns.
When I enable this through configuration, the load balancer for the master_pooler has the same hostname set by external-dns annotation than the load balance service pointing to the master.
So, when I resolve the hostname through my DNS, I get 2 records for the same name, one to the master service IP, the other one to mater_pooler service.

As I could work only with master_pooler without direct service to the master, I tried to disable load balancer on master with : 

```
  enable_master_load_balancer: false
  enable_master_pooler_load_balancer: true
  enable_replica_load_balancer: false
  enable_replica_pooler_load_balancer: false
```
The master_pooler load balancer service is well created but without the external-dns annotation.

Am I missing something in the configuration to properly disable master_load_balancer and have only master_pooler balancer with external-dns woriking or to set different hostnames to the master_load_balancer and the master_pooler_load_balancer ?

Thank you,

FxKu:
fixes #2180 

Introduced two new functions: `generatePoolerServiceAnnotations` for the DNS annotation used for LBs and `getCustomServiceAnnotations` which is shared for all services. Because pooler unit tests need quite some refactoring, the annotations are tested in e2e suite.

Functions `masterDNSName` and `replicaDNSName` now take the cluster name as argument, to make them reusable for pooler DNS name resolving

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -810,6 +810,9 @@ Load balancer services can also be enabled for the [connection pooler](user.md#c
 pods with manifest flags `enableMasterPoolerLoadBalancer` and/or
 `enableReplicaPoolerLoadBalancer` or in the operator configuration with
 `enable_master_pooler_load_balancer` and/or `enable_replica_pooler_load_balancer`.
+For the `external-dns.alpha.kubernetes.io/hostname` annotation the `-pooler`
+suffix will be appended to the cluster name used in the template which is
+defined in `master|replica_dns_name_format`.
 
 ## Running periodic 'autorepair' scans of K8s objects
 
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -670,6 +670,20 @@ class EndToEndTestCase(unittest.TestCase):
                              'LoadBalancer',
                              "Expected LoadBalancer service type for replica pooler pod, found {}")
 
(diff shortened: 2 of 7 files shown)


Jan-M:
Risk of breaking existing setups with change of DNS name for pooler LB. Feature added in 1.8. Felix will check.

FxKu:
So far in order to get the DNS entry for the pooler LB one had to enable the normal LB as well because it calls [generateServiceAnnotations] function used also for the other services. This function does not check pooler LBs. So then you end up with two services using the same DNS and you don't know the route a request will take, right?

So not sure, how many have really used the feature because of that? And as the normal LB has to be enabled for the DNS to be there, we can be sure that connections still work via the master|replica service.

FxKu:
:+1:

idanovinda:
👍

PR #2189, opened 2023-01-25
set wal_level config not on empty parameters map

FxKu:
fix regression introduced by #2187

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -403,7 +403,10 @@ func (c *Cluster) syncStatefulSet() error {
 		c.logger.Warnf("could not get list of pods to apply PostgreSQL parameters only to be set via Patroni API: %v", err)
 	}
 
-	requiredPgParameters := c.Spec.Parameters
+	requiredPgParameters := make(map[string]string)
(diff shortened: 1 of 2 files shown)


FxKu:
:+1:

FxKu:
:+1:

pkg/cluster/sync.go
```diff
@@ -403,7 +403,10 @@ func (c *Cluster) syncStatefulSet() error {
 		c.logger.Warnf("could not get list of pods to apply PostgreSQL parameters only to be set via Patroni API: %v", err)
 	}
 
-	requiredPgParameters := c.Spec.Parameters
+	requiredPgParameters := make(map[string]string)
+	for k, v := range c.Spec.Parameters {
+		requiredPgParameters[k] = v
+	}
```
dmvolod, line 409: Probably, `maps.Clone` can be used instead as 
`requiredPgParameters := maps.Clone(c.Spec.Parameters)`

dmvolod: Oh, `maps.Copy` should be used in this case only

idanovinda:
👍

Issue #2116, opened 2022-11-23
WAL bucket scope prefix/suffix missing when configuring WAL_S3_BUCKET locally


haslersn:
When configuring `WAL_S3_BUCKET` via ConfigMap, Secret, or `spec.env` (per PostgreSQL cluster) instead of globally in the operator config, then the operator doesn't configure `WAL_BUCKET_SCOPE_SUFFIX` and `WAL_BUCKET_SCOPE_PREFIX`.
This is due to the following check: https://github.com/zalando/postgres-operator/blob/640581fb4661dbab852dfc0995b5e9f6140fd8f8/pkg/cluster/k8sres.go#L939

This hit us unexpectedly when we wanted to migrate our setup to use a distinct bucket per PostgreSQL cluster. Furthermore, even when using a distinct bucket per PostgreSQL cluster, it's still a good idea to have the cluster UID in the `WAL_BUCKET_SCOPE_SUFFIX`, so that a cluster that gets recreated has a clean state and doesn't conflict with existing WAL backups.

Possible solution: Maybe the operator can unconditionally inject `WAL_BUCKET_SCOPE_SUFFIX` and `WAL_BUCKET_SCOPE_PREFIX`, no matter if `c.OpConfig.WALES3Bucket` is non-empty.

Workaround: In the OperatorConfiguration, set `configuration.aws_or_gcp.wal_s3_bucket: dummy` and overwrite it on a per-PostgreSQL-cluster basis by setting the environment variable `WAL_S3_BUCKET` via ConfigMap, Secret, or `spec.env`.  I only tested it with `spec.env` and that seems to have precedence over the environment variable injected by the operator.

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.8.2

kevinvalk:
NOTE: When using this strategy you cannot restore from clone anymore. No matter what you do, the `dummy` bucket will take precedence over all your configurations efforts of the CLONE_* bucket. The only way I could get this to work was by directly setting the s3_* keys inside `spec.clone`.

```yaml
```yaml
spec:
  clone:
    uid: "efd12e58-5786-11e8-b5a7-06148230260c"
    cluster: "cluster-name"
    timestamp: "2023-12-19T12:40:33+01:00"

    s3_wal_path: "s3://${s3_backup_bucket}/spilo/cluster-name/efd12e58-5786-11e8-b5a7-06148230260c/wal/${pg_version}/"
[...]

PR #2192, opened 2023-01-27
Add cluster UID as environment variable

mkrueger92:
This is useful in case where backups are made.

This improves or is related to issue  #2116.


pkg/cluster/k8sres.go
```diff
@@ -1041,6 +1041,9 @@ func (c *Cluster) generateSpiloPodEnvVars(
 
 	envVars = appendEnvVars(envVars, opConfigEnvVars...)
 
+	// add variable containing the UID of the cluster; usefull for backup paths
+	envVars = append(envVars, v1.EnvVar{Name: "CLUSTER_UID", Value: getBucketScopeSuffix(string(uid))})
```
dmvolod, line 1045: @mkrueger92  probably you can set original `string(uid)` value to avoid confusion with `CLUSTER_ID` name, as `getBucketScopeSuffix(...)` returns `uid` combined with `/` prepared for the `<>_SUFFIX` variables.

mkrueger92: Good point. I will prepare a new commit.

mkrueger92:
Is there anything left to do?

Otherwise I would just close this PR if this change does not fit into the concept.

dmvolod:
@mkrueger92 lets ask for help from the Zalando core team: @FxKu , @idanovinda could you please to have a look at this PR. Thanks in advance.

Jan-M:
Hi! I can understand the value but am surprised it is not there? Default backup paths with Spilo contain the uid already.

baznikin:
Is it possible for single pod to change it's database UID? If true, then pod have to exit and be recreated to see change in env variable. From my very limited experience pods never exit.

mkrueger92:
@baznikin: I am not aware of that an existing pod can easily change the UID.

The use-case behind this PR is that during development when recreating the cluster multiple times backup exist in the external storage but do not match the newly created UID. In globally managed backup configurations this is prevented by using the UID in the path. In locally configured backups (locally in the sense of not globally by the operator but for each cluster individually) the UID (or any other dynamic part, which is stable for all pods of the same cluster but different between clusters) is not available.

Also it does not have negative side-effects having this environment variable set even if it is not used.

mkrueger92:
@Jan-M: Are you still interested in this PR? Otherwise I would close it.

Issue #667, opened 2019-09-11
Database name regexp


krystian-kulgawczuk:
`databaseNameRegexp    = regexp.MustCompile("^[a-zA-Z_][a-zA-Z0-9_]*$")`
`postgres-operator/pkg/cluster/cluster.go`
Is there some kind of reason why this regex looks as it is? 

The pattern seems to follow this from the documentation (not to the dot):
> SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($). Note that dollar signs are not allowed in identifiers according to the letter of the SQL standard, so their use might render applications less portable. The SQL standard will not define a key word that contains digits or starts or ends with an underscore, so identifiers of this form are safe against possible conflict with future extensions of the standard.

But as long as the name is quoted in SQL statement, we should be able to put whatever we like with the exception of \0.
I have this issue that I have to migrate postgres to k8s, but all our databases are named in the following fashion: `[a-z]+-[a-z]+`. 

I'm not very familiar with the operator's code, but if quoted identifiers are used everywhere, then I don't see any reason why the regex shouldn't be expanded.

Jan-M:
We did this on purpose because some issues arose somewhere with "-" in the database name and not having quotes. But valid point, maybe we don't need to enforce this on everyone.

Feel free to move this regex into the configuration ;)

FxKu:
Yes, probably the best option would be another [configuration option](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters), so that everybody can define naming constraints for their databases. I could/should also add a minimal pattern check for the CRD validation (#599).

ClaudiuLSH:
Is there a way to adjust this regex or database name check?

Jan-M:
Not right now, sounds like a good first issue to provide PR moving regex to config

PR #2193, opened 2023-01-27
Add flag to disable database name/schema validation.

flyingcamilo:
The previous setup adheres to the naming convention, and that's a good thing. Unfortunately, however, there are cases in which you have to break with them.

Unfortunately, that is the case with us. We use external software and have no influence on their naming.


FxKu:
Would rather like to see `databaseNameRegexp` being configurable. See #667

flyingcamilo:
@FxKu That sound reasonable. I can do it.

flyingcamilo:
Hey @FxKu, I added in your suggestion, but I have some issues to run the tests. Are they broken, or do I miss something?

flyingcamilo:
Hey @FxKu added in your suggestion.

flyingcamilo:
@FxKu Not sure if the e2e test is failing because of me, but I guess not. Is there anything I can do ?

FxKu:
@flyingcamilo thanks for adding the config options. What would now happen with the old handling? https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L44

I don't see it touched in this PR.

pkg/cluster/cluster.go
```diff
@@ -294,6 +294,10 @@ func (c *Cluster) Create() error {
 		c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Services", "The service %q for role %s has been successfully created", util.NameFromMeta(service.ObjectMeta), role)
 	}
 
+	if err = c.readValidateDatabaseNameRegexp(c.OpConfig.DatabaseNameRegexp); err != nil {
+		return err
+	}
+
```
FxKu, line 300: It should rather place the [existing validation](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/database.go#L419).

pkg/cluster/cluster.go
```diff
@@ -337,6 +341,7 @@ func (c *Cluster) Create() error {
 	// that feature explicitly
 	if !(c.databaseAccessDisabled() || c.getNumberOfInstances(&c.Spec) <= 0 || c.Spec.StandbyCluster != nil) {
 		c.logger.Infof("Create roles")
+
```
FxKu, line 344: please to not introduce unrelated changes - even these are just blank lines

PR #2195, opened 2023-01-29
Adds APP_URL as base tag href to enable loading via ui via ingress ro…

simo9000:
**What**
- Modifies the ui index view and model to bind the APP_URL to the href attribute of an added <base> element.
- Modifies the deployment chart for the ui to allow the APP_URL to be set by values.yaml

**Why**
This change enables configuration to allow the ui to be accessed via ingress rules.

### DIFF
--- a/ui/operator_ui/templates/index.html
+++ b/ui/operator_ui/templates/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
+    <base href="{{app_url}}">
     <title>PostgreSQL Operator UI</title>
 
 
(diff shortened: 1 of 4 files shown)


idanovinda:
👍

FxKu:
:+1:

FxKu:
Thanks @simo9000 ! Good to have this localhost address not hard-coded anymore in the deployment template.

ui/operator_ui/templates/index.html
```diff
@@ -2,6 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
+    <base href="{{app_url}}">
```
FxKu, line 5: @simo9000 unfortunately, this change introduces a breaking change, see #2302. Now everybody has to set appUrl because with the default localhost appUrl as base the app will not load. App URL is only used for authorization. The rest is full links. Maybe most users haven't used it so having it hardcoded in deployment.yaml template was fine. I have to read up on ingress rules, but can you tell us why you need this to be changed?

PR #2216, opened 2023-02-10
minor fix to pooler TLS support

FxKu:
wrong default assignment created in #2146 and raised [here](https://github.com/zalando/postgres-operator/issues/2201#issuecomment-1413798114)

Issue #2201, opened 2023-02-01
Connection Pooler crashes when using custom TLS certificates


heilerich:
Since #2146 has been merged connection pooler pods are going into crash loop on start. We were really looking forward to this feature, but sadly it isn't working at all in our environments. Pinging the authors of that PR @jeremie-seguin @FxKu 

I suspect that I might have missed something here, because it looks to me like the current solution should have never worked anywhere. Anyways, in our deployments we are encountering the following:

The custom TLS certificates are mounted as a secret with a default mode of 0640. This creates two problems:

(1) The secret mount in the container is a separate `tmpfs` filesystem. The entrypoint.sh of the default pooler image (at tag master-26) tries to hardlink the files in the secret mount to the container's root overlay filesystem which produces a `Cross-device link` error. At that point the container will crash.

(2) Even with a fix for the above problem the container will probably still not run because it defaults to running as a non-root user. At the same time the operator does not set a `fsGroup` in the security context for the pods. This will cause the secret to be owned by root and with the operator setting the defaultMode of the secret to `0640` the certificates will not be readable by the pgbouncer process.

Possible solutions:

(1) Update the entrypoint.sh for the default pooler conatainer to use symlinks which will work across file system boundaries (i.e. replace `ln` with `ln -s`

(2) (a) Set a more permissive default mode for the secret mount, (b) use an initContainer running with UID 0 to copy and chown the certificates into a emptyDir volume (c) run pgboucner as UID 0 or (d) set a `fsGroup` for the pod.

- **Which image of the operator are you using?**  registry.opensource.zalan.do/acid/postgres-operator:v1.9.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** KKP, Bare Metal K8s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report

heilerich:
After fixing the above mentioned issues we additionally detected that the default paths for certificate and key files are assigned improper values here

https://github.com/zalando/postgres-operator/blob/30b612489a2a20d968262791857d1db1a85e0b36/pkg/cluster/connection_pooler.go#L351-L356

This results in the key file being linked to the certificate destination and the key destination being linked to the containing folder instead of the key file.

FxKu:
Sad to hear this new feature doesn't work as expected :disappointed: Could you provide a bugfix? My knowledge (and interest) in this whole certificates topic is very limited. This is our entrypoint atm:

```
#!/bin/sh

set -ex

[...]

PR #2219, opened 2023-02-13
fix to pooler TLS support

2tvenom:
last PR #2216 checked and not work. added security context fsGroup for correct load tls cert from mounted volume

### DIFF
--- a/pkg/cluster/connection_pooler.go
+++ b/pkg/cluster/connection_pooler.go
@@ -349,6 +349,9 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 		crtFile := spec.TLS.CertificateFile
 		keyFile := spec.TLS.PrivateKeyFile
 		caFile := spec.TLS.CAFile
+		mountPath := "/tls"
+		mountPathCA := mountPath
+
 		if crtFile == "" {
 			crtFile = "tls.crt"
 		}
@@ -358,17 +361,20 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 		if caFile == "" {
 			keyFile = "ca.crt"
 		}
+		if spec.TLS.CASecretName != "" {
+			mountPathCA = mountPath + "ca"
+		}
 
 		envVars = append(
 			envVars,
 			v1.EnvVar{
-				Name: "CONNECTION_POOLER_CLIENT_TLS_CRT", Value: filepath.Join("/tls", crtFile),
+				Name: "CONNECTION_POOLER_CLIENT_TLS_CRT", Value: filepath.Join(mountPath, crtFile),
 			},
 			v1.EnvVar{
-				Name: "CONNECTION_POOLER_CLIENT_TLS_KEY", Value: filepath.Join("/tls", keyFile),
+				Name: "CONNECTION_POOLER_CLIENT_TLS_KEY", Value: filepath.Join(mountPath, keyFile),
 			},
 			v1.EnvVar{
-				Name: "CONNECTION_POOLER_CLIENT_CA_FILE", Value: filepath.Join("/tls", caFile),
+				Name: "CONNECTION_POOLER_CLIENT_CA_FILE", Value: filepath.Join(mountPathCA, caFile),
 			},
 		)
 
--- a/pkg/cluster/connection_pooler.go
+++ b/pkg/cluster/connection_pooler.go
@@ -359,7 +359,7 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 			keyFile = "tls.key"
 		}
 		if caFile == "" {
-			keyFile = "ca.crt"
+			caFile = "ca.crt"
 		}
 		if spec.TLS.CASecretName != "" {
 			mountPathCA = mountPath + "ca"
--- a/pkg/cluster/connection_pooler.go
+++ b/pkg/cluster/connection_pooler.go
@@ -348,11 +348,15 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 		// Env vars
 		crtFile := spec.TLS.CertificateFile
 		keyFile := spec.TLS.PrivateKeyFile
+		caFile := spec.TLS.CAFile
 		if crtFile == "" {
 			crtFile = "tls.crt"
 		}
 		if keyFile == "" {
 			keyFile = "tls.key"
+		}
+		if caFile == "" {
+			keyFile = "ca.crt"
 		}
 
 		envVars = append(
@@ -362,6 +366,9 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 			},
 			v1.EnvVar{
 				Name: "CONNECTION_POOLER_CLIENT_TLS_KEY", Value: filepath.Join("/tls", keyFile),
+			},
+			v1.EnvVar{
+				Name: "CONNECTION_POOLER_CLIENT_CA_FILE", Value: filepath.Join("/tls", caFile),
 			},
 		)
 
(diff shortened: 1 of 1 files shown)


FxKu:
Thanks @2tvenom . Does it work for you with this fix? Do we need the symbolic links in pgBounder Docker file entrypoint.sh as mentioned [here](https://github.com/zalando/postgres-operator/issues/2201#issuecomment-1413831670)?

2tvenom:
i checked it with next configs

`entrypoint.sh`
```sh
#!/bin/sh

set -ex

if [ "$PGUSER" = "postgres" ]; then
    echo "WARNING: pgbouncer will connect with a superuser privileges!"
    echo "You need to fix this as soon as possible."
fi

if [ -z "${CONNECTION_POOLER_CLIENT_TLS_CRT}" ]; then
    openssl req -nodes -new -x509 -subj /CN=spilo.dummy.org \
        -keyout /etc/ssl/certs/pgbouncer.key \
        -out /etc/ssl/certs/pgbouncer.crt
else
    ln -s /tls/tls.crt /etc/ssl/certs/pgbouncer.crt
    ln -s /tls/tls.key /etc/ssl/certs/pgbouncer.key
    ln -s /tls/ca.crt /etc/ssl/certs/ca.crt
fi

envsubst < /etc/pgbouncer/pgbouncer.ini.tmpl > /etc/pgbouncer/pgbouncer.ini
envsubst < /etc/pgbouncer/auth_file.txt.tmpl > /etc/pgbouncer/auth_file.txt

exec /bin/pgbouncer /etc/pgbouncer/pgbouncer.ini
```

`pgbouncer.ini.tmpl`
```
# vim: set ft=dosini:

[databases]
* = host=$PGHOST port=$PGPORT auth_user=$PGUSER

[pgbouncer]
pool_mode = $CONNECTION_POOLER_MODE
listen_port = $CONNECTION_POOLER_PORT
listen_addr = *
auth_type = md5
auth_file = /etc/pgbouncer/auth_file.txt
admin_users = $PGUSER
stats_users_prefix = robot_
auth_query = SELECT * FROM $PGSCHEMA.user_lookup($1)
logfile = /var/log/pgbouncer/pgbouncer.log
pidfile = /var/run/pgbouncer/pgbouncer.pid

server_tls_sslmode = verify-ca
server_tls_ca_file = /etc/ssl/certs/ca.crt
server_tls_key_file = /etc/ssl/certs/pgbouncer.key
server_tls_cert_file = /etc/ssl/certs/pgbouncer.crt
server_tls_protocols = secure

client_tls_sslmode = verify-ca
client_tls_key_file = /etc/ssl/certs/pgbouncer.key
client_tls_cert_file = /etc/ssl/certs/pgbouncer.crt
client_tls_ca_file = /etc/ssl/certs/ca.crt

log_connections = 0
log_disconnections = 0

# How many server connections to allow per user/database pair.
default_pool_size = $CONNECTION_POOLER_DEFAULT_SIZE

# Add more server connections to pool if below this number. Improves behavior
# when usual load comes suddenly back after period of total inactivity.
#
# NOTE: This value is per pool, i.e. a pair of (db, user), not a global one.
# Which means on the higher level it has to be calculated from the max allowed
# database connections and number of databases and users. If not taken into
# account, then for too many users or databases PgBouncer will go crazy
# opening/evicting connections. For now disable it.
#
# min_pool_size = $CONNECTION_POOLER_MIN_SIZE

# How many additional connections to allow to a pool
reserve_pool_size = $CONNECTION_POOLER_RESERVE_SIZE

# Maximum number of client connections allowed.
max_client_conn = $CONNECTION_POOLER_MAX_CLIENT_CONN

# Do not allow more than this many connections per database (regardless of
# pool, i.e. user)
max_db_connections = $CONNECTION_POOLER_MAX_DB_CONN

# If a client has been in "idle in transaction" state longer, it will be
# disconnected. [seconds]
idle_transaction_timeout = 600

# If login failed, because of failure from connect() or authentication that
# pooler waits this much before retrying to connect. Default is 15. [seconds]
server_login_retry = 5

# To ignore extra parameter in startup packet. By default only 'database' and
# 'user' are allowed, all others raise error.  This is needed to tolerate
# overenthusiastic JDBC wanting to unconditionally set 'extra_float_digits=2'
# in startup packet.
ignore_startup_parameters = extra_float_digits,options,search_path
```

FxKu:
Ok, you have one more line in the entrypoint script `ln -s /tls/ca.crt /etc/ssl/certs/ca.crt` which makes me realize that the operator [does not support ca files](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/connection_pooler.go#L358) for connection pooler like we do with [spilo pods](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L1315). I feel this code should be unified.

2tvenom:
this endpoint show what this worked with `ln -s`
i can check with standard `endpoint.sh`

FxKu:
What I meant is that I can change the entrypoint of our pgBouncer docker image to:
```
...
    ln -s ${CONNECTION_POOLER_CLIENT_TLS_CRT} /etc/ssl/certs/pgbouncer.crt
    ln -s ${CONNECTION_POOLER_CLIENT_TLS_KEY} /etc/ssl/certs/pgbouncer.key
    if [ -z "${CONNECTION_POOLER_CLIENT_CA_FILE}" ]; then
        ln -s ${CONNECTION_POOLER_CLIENT_CA_FILE} /etc/ssl/certs/ca.crt
    fi
...
```

But the operator does not set this `CONNECTION_POOLER_CLIENT_CA_FILE` environment variable, yet.

2tvenom:
@FxKu done

FxKu:
Thanks, but the code looks quite different to what we do with [spilo pods](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L1315). CA should not be set when it's not defined in the manifest. For spilo pods we also use a second volume mount (`tlsca`). Can't remember why we did that back then.

2tvenom:
@FxKu done

pkg/cluster/connection_pooler.go
```diff
@@ -348,20 +348,33 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
...
 		}
 		if keyFile == "" {
 			keyFile = "tls.key"
 		}
+		if caFile == "" {
+			caFile = "ca.crt"
+		}
+		if spec.TLS.CASecretName != "" {
+			mountPathCA = mountPath + "ca"
+		}
```
FxKu, line 366: I believe, we need a second volume mount when introducing CA file support. Hm ... can you remove everything about CA file support from this PR as it looks pretty half baked at the moment and not tested.

pkg/cluster/connection_pooler.go
```diff
@@ -402,6 +415,12 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
 		},
 	}
 
+	if spec.TLS != nil && spec.TLS.SecretName != "" && spec.SpiloFSGroup != nil {
+		podTemplate.Spec.SecurityContext = &v1.PodSecurityContext{
+			FSGroup: spec.SpiloFSGroup,
```
FxKu, line 420: Why should it get the spilo FS group btw? Can it also get it's own? Would an extra config option make sense?

2tvenom: you mean additional fsGroup in CRD pooler block?

pkg/cluster/connection_pooler.go
2tvenom: i will mount CA

pkg/cluster/connection_pooler.go
FxKu: There could be extra settings in the [pooler struct](https://github.com/zalando/postgres-operator/blob/master/pkg/apis/acid.zalan.do/v1/postgresql_type.go#L237), yes. But, on the other hand we're already re-using TLS settings from postgrsql.spec - so I'm fine by copying SpiloFSGroup, too, as it would be convenient to users of this feature. No extra config necessary in the manifest.

Was just wondering if somebody might want this FSGroup setting to be different from spilo...

idanovinda:
👍

FxKu:
Ok @2tvenom. Thanks for your input so far. I will take it from here :smiley:

PR #2225, opened 2023-02-16
Set SecurityContext for connection pooler

addreas:
Set uid/gid based on whats burned into the registry.opensource.zalan.do/acid/pgbouncer image.


FxKu:
I don't see where we set those IDs in the pooler repo. Are these default values? We use:

```
RUN \
    addgroup -S pgbouncer && \
    adduser -S pgbouncer && \

USER pgbouncer:pgbouncer
```

addreas:
I think I had a look in /etc/passwd and /etc/group

FxKu:
Closing because the suggested changes have been reflected by #2255

Issue #2231, opened 2023-02-21
Document Kubernetes supported versions


yajo:
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.9.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** bare metal k3s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** missing docs

I have searched across docs and issues, and it seems there's nothing explaining which k8s versions are supported by each operator release.

https://github.com/zalando/postgres-operator/issues/2094#issuecomment-1291477228 links to upstream k8s version skew policy, but in reality that doesn't mean this operator supports all those versions (AFAICS). For example, https://github.com/zalando/postgres-operator/issues/2098 shows that 1.8.2 didn't really work in k8s 1.24.

Other projects have support matrix tables that allow sysadmins to perform supported and safe cluster upgrades. These are just some random examples:
- https://www.suse.com/suse-rancher/support-matrix/all-supported-versions/rancher-v2-7-1/
- https://cert-manager.io/docs/installation/supported-releases/#supported-releases

It'd be great to have some place where to know which are the supported k8s versions for each postgres-operator release.

Thanks!

FxKu:
Closing after #2232 has been merged.

PR #2232, opened 2023-02-21
Provide support matrix with Pg and K8s version

FxKu:
I think, it's time we can remove the Spilo 12 paragraph from the readme.
Since it has been requested a few times e.g. in #2231


FxKu:
:+1:

idanovinda:
👍

PR #2238, opened 2023-02-23
add c.replicationSlots on sync

FxKu:
Turns out on operator restart existing slots are not added by the [Create code](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L379). I have mistaken ADD events from the log to be calling that code. Only when new clusters are added with slots specified in the manifest they will be added (and dropped when removed).

Newly introduced slots will be added in the [sync code](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/sync.go#L643). But for existing slots the loop over is continued when there is no difference and therefore it is not added to c.replicationSlots. So whenever there happens an operator restart existing slots wont get synced and not be dropped if they are removed from the manifest.

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -413,6 +413,7 @@ func (c *Cluster) syncStatefulSet() error {
 	}
 
 	// sync Patroni config
+	c.logger.Debug("syncing Patroni config")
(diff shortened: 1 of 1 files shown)


FxKu:
:+1:

jopadi:
👍

PR #2245, opened 2023-02-27
sync stateful set when syncing streams during ADD event

FxKu:
In #2187 the flow for configuring Postgres for streams was moved to syncStatefulSet, because it's the only place right now where we restart Postgres on config changes. The restart is required to set wal_level to `logical` before creating publications and replication slots.

However, syncStatefulSet is not executed during [Create](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L244). Whem a new cluster is added with stream section we face the same error as before - creation of publication during syncStreams will fail because wal_level has not been adjusted, yet. It happens only during next SYNC when it's too late.

Therefore, this PR proposed to call syncStatefulSet inside Create method, if streams are defined.

In case the sync of statefulset will fail return an error. I moved setting replicationSlots before the syncing of streams to have it set even when there is this error.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -370,14 +370,20 @@ func (c *Cluster) Create() error {
 	// something fails, report warning
 	c.createConnectionPooler(c.installLookupFunction)
 
+	// remember slots to detect deletion from manifest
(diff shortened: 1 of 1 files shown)


FxKu:
:+1:

idanovinda:
👍

PR #2246, opened 2023-02-28
fix: PG 15 backups working by default

yajo:
## Problem description

In 30b612489a2a20d968262791857d1db1a85e0b36 this update was forgotten. It is needed for logical backups of postgres 15 clusters, as seen in https://github.com/zalando/postgres-operator/issues/1945#issuecomment-1425060560.

The default value documented is different to the one actually used: https://github.com/zalando/postgres-operator/blob/9973262b83507ccb16c097aef9ca56b35a1cbf4a/docs/reference/operator_parameters.md?plain=1#L753-L758

## Linked issues

Fixes https://github.com/zalando/postgres-operator/issues/1945.

@moduon MT-1075

## Checklist

Thanks for submitting a pull request to the Postgres Operator project.
Please, ensure your contribution matches the following items:

- [x] Your go code is [formatted](https://blog.golang.org/gofmt). Your IDE should do it automatically for you.
- [x] You have updated [generated code](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#code-generation) when introducing new fields to the `acid.zalan.do` api package.
- [x] New [configuration options](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters) are reflected in CRD validation, helm charts and sample manifests.
- [x] New functionality is covered by [unit](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#unit-tests) and/or [e2e](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#end-to-end-tests) tests.
- [x] You have checked existing open PRs for possible overlay and referenced them.

PR #2252, opened 2023-03-04
fix: update logical-backup image to support backups of postgres v15

NicolaiSchmid:
The v1.8.0 image does not include Postgres v15, so backups of clusters of that version fail with:
```
/dump.sh: line 31: /usr/lib/postgresql/15/bin/psql: No such file or directory
```


dmvolod:
Hello, @NicolaiSchmid 
It seems to duplicate PR  #2246

NicolaiSchmid:
Then let's get that one merged, shall we!

PR #2255, opened 2023-03-07
set pooler pod security context

FxKu:
This is a follow up PR to #2219 which refactors the way the volume mounts are generated for spilo and pooler pods when spec.TLS is specified. My idea here is to unify have just one function in k8sres.go which generates environment variables and additional volumes to be used for the pod templates. For pooler pods we do not use generatePodTemplate function so the rest of the code is a little different there.

There are a few fixes along the way:
- Do not set CONNECTION_POOLER_CLIENT_CA_FILE when spec.TLS.CAFile is emty
- Make sure we define two volumes for pooler TLS support for `spec.TLS.SecretName` AND `spec.TLS.CASecretName`
- Set not only `FSGroup` of pooler pod securityContext but also `RunAsUser` and `RunAsGroup`. See also #2225 which uses hardcoded values user: 100 and group: 101. Seems that is burned into our pgbouncer image so we should not reuse the spilo config values.

Bumping the pooler image which support of CONNECTION_POOLER_CLIENT_CA_FILE.

### DIFF
--- a/pkg/cluster/connection_pooler.go
+++ b/pkg/cluster/connection_pooler.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"context"
 	"fmt"
-	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


heilerich:
I tested with

`registry.opensource.zalan.do/acid/postgres-operator-test:1954846-pr-2255-5`
`registry.opensource.zalan.do/acid/pgbouncer:master-27`

and could find no apparent problems. Thanks for the effort 👍🏻

idanovinda:
👍

FxKu:
:+1:

cdmikechen:
@FxKu 
Setting `runAsUser` directly is not allowed on, for example, Openshift. 
So I think we should change the serviceaccount of the pooler to `postgres-pod` as well, otherwise the pooler will not be created correctly.

```
Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```

Issue #2198, opened 2023-01-31
PostgreSQL extensions are installed in a random order


justabaka:
- registry.opensource.zalan.do/acid/postgres-operator:v1.8.2
- AWS EKS
- Production: no
- Feature request

There are certain extensions that have other extensions as dependencies, e.g. I want to install `postgis`, `postgis_raster` and `postgis_topology` extensions. The `postgis_topology` requires `postgis` but it is _not always_ installed first. Yes, sometimes it seems to be happening in a correct way, but the whole process feels completely random and may take a lot of attempts and time.

postgresql cluster crd:
```
  preparedDatabases:
    test:
      extensions:
        postgis: public
        postgis_raster: public
        postgis_topology: topology
```

postgres-operator log messages:
```
time="2023-01-31T14:45:26Z" level=info msg="creating extension \"postgis_topology\" schema \"topology\"" cluster-name=default/test pkg=cluster worker=0
time="2023-01-31T14:45:26Z" level=error msg="could not sync prepared databases: could not execute create extension: pq: required extension \"postgis\" is not installed" cluster-name=default/test pkg=cluster worker=0
```

In order to make it work all the time from the first try there should be a strict order of installation.

FxKu:
That's because of the way we [loop over](https://github.com/zalando/postgres-operator/blob/30b612489a2a20d968262791857d1db1a85e0b36/pkg/cluster/sync.go#L1206) database objects to sync. The order of the DB query is [alphabetical](https://github.com/zalando/postgres-operator/blob/30b612489a2a20d968262791857d1db1a85e0b36/pkg/cluster/database.go#L40), but the go loop makes it random, I believe.

Ideally we should not return an error on failure but just collect error message and continue with the next extension. This way your problem would go away automatically on subsequent syncs. We do this already for database users but not here.

underdark456:
Hello, same behavior on 1.11.0
Has this not been fixed?

PR #2262, opened 2023-03-13
continue syncing databases and extensions on err

FxKu:
addresses #2198

For extensions it's possible they are dependent on one another. Thus, it could happen that syncing an extensions can fail because another extension listed in the manifest has not been installed yet. The go loop iterates in random order, so sometimes it might work but mostly it won't.

This PR tries to improve the sync of databases, prepared databases and schemas and extension. Taking the idea that is also used in other parts to collect error messages rather then erroring out immediately. In some places the loop must be continued.


jopadi:
👍

FxKu:
:+1:

PR #2273, opened 2023-03-23
fix: make map in generateUserSecrets with correct size

genofire:
just a small finding during reading


FxKu:
Interesting. Never stumbled over an error related to secret generation. But the code change make sense.

dmvolod:
Seems to most of the cases like this will be covered by linter with additional configuration, still waiting to be merged in #2135

FxKu:
:+1:

idanovinda:
👍

PR #2290, opened 2023-04-17
Fix some errors be ignored

drivebyer:
In `Create()` function, some errors not been check. It will cause cluster status to unexpected.

For example:
https://github.com/zalando/postgres-operator/blob/0e7beb5fe50920cef82ce6e908ac52a9feaa3aa0/pkg/cluster/cluster.go#L271-L273


FxKu:
I don't get what this code change improves? Can you explain in a bit more detail? I've had a quick look at the `Create()` code and it looked fine to me.

FxKu:
:+1:

jopadi:
👍

PR #2076, opened 2022-10-13
Add Patroni failsafe_mode parameter

hughcapet:
Add support for the [upcoming Patroni feature](https://github.com/zalando/patroni/pull/2379)

PR #2297, opened 2023-04-19
fix(chart): Add patroni configuration in operatorconfiguration

thcdrt:
Hello,

When trying to enable patroni DCS failsafe mode, I found out that in [PR 2076](https://github.com/zalando/postgres-operator/pull/2076/files#diff-21b9f2ed91da790a93b909fbb8b429192e4b516f0ebb0ba787c12bb13074531a) the parameter `failsafe_mode` was added to Helm chart values file, but this configuration was never applied inside the templates.

I tested this fix and it's working as expected.

Thomas


charts/postgres-operator/templates/operatorconfiguration.yaml
```diff
@@ -40,4 +40,6 @@ configuration:
 {{ toYaml .Values.configLoggingRestApi | indent 4 }}
   connection_pooler:
 {{ toYaml .Values.configConnectionPooler | indent 4 }}
+patroni:
+  {{ toYaml .Values.configPatroni | indent 4 }}
```
FxKu, line 44:
```suggestion
  patroni:
{{ toYaml .Values.configPatroni | indent 4 }}
```

thcdrt: Indeed, thanks

manifests/postgresql-operator-default-configuration.yaml
```diff
@@ -209,5 +209,5 @@ configuration:
     connection_pooler_number_of_instances: 2
     # connection_pooler_schema: "pooler"
     # connection_pooler_user: "pooler"
-  # patroni:
-    # failsafe_mode: "false"
+  patroni:
+   failsafe_mode: "false"
```
FxKu, line 213:
```suggestion
   failsafe_mode: false
```

FxKu:
:+1:

FxKu:
Thanks @thcdrt for the quick update. I've noticed a few more issues around the config option and will follow up in another PR.

PR #2298, opened 2023-04-20
rename failsafe_mode option in opconfig CRD

FxKu:
There is an inconsistency in the naming of the global config option for Patroni's failsafe mode. `enable_patroni_failsafe_mode` in configmap and `failsafe_mode` in OpeartorConfiguration CRD. In #2297 we noticed that this option is still missing in the helm chart templates, but the difference in the naming would make it a bit harder to add it to the both configmap and CRD templates.

Therefore, this PR will change the name in the CRD to `enable_patroni_failsafe_mode` as well albeit being listed under the `patroni` key. It also adds documentation which was still missing in the config reference. As the next release will the first version to support failsafe_mode by default, I do not think this renaming would break the setups of people out there.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -670,7 +670,7 @@ spec:
               patroni:
                 type: object
                 properties:
-                  failsafe_mode:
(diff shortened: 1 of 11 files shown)


jopadi:
👍

FxKu:
:+1:

PR #2323, opened 2023-05-16
Remove logicalBackupSchedule pattern spec from the postgresql CRD.

bwrobc:
The pattern only validates the most basic cron schedules, and fail with the use of intervals, lists, ranges, etc.

### DIFF
--- a/manifests/postgresql.crd.yaml
+++ b/manifests/postgresql.crd.yaml
@@ -215,7 +215,6 @@ spec:
                   x-kubernetes-preserve-unknown-fields: true
               logicalBackupSchedule:
                 type: string
-                pattern: '^(\d+|\*)(/\d+)?(\s+(\d+|\*)(/\d+)?){4}$'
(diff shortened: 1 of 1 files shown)


bwrobc:
FWIW [see here](https://stackoverflow.com/questions/14203122/create-a-regular-expression-for-cron-statement) for a some attempts to create a comprehensive cron schedule regex.

FxKu:
I don't get what are you trying to achieve here? This is about scheduling pg_dump. Do you really need all the cron flexibility there? Normally, our users use WAL archiving for backups and recovery, and logical backups only if they want to keep backups for a longer time since we use retention with the base backups. So therefore, being able to specify points in time with the schedule is good enough. What requirements do you have?

bwrobc:
It's a while ago now, but as I recall I was running a frequent schedule for test purposes and wanted to limit it to office hours on weekdays.

PR #2342, opened 2023-06-02
check pg version

### LINKED ISSUE #1678: Logical backup fails after major version upgrade unless manual removal of CronJobs
  - registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
  - Bare Metal K8s
  - Yes
- **Type of issue?**
  - Bug report or possibly a feature request

I have an instance of Postgres Operator running with the major version upgrade enabled:

```yaml
apiVersion: acid.zalan.do/v1
kind: OperatorConfiguration
metadata:
  name: postgresql-operator-default-configuration
configuration:
  major_version_upgrade:
    major_version_upgrade_mode: full
```

When I change the spec.postgresql.version filed in a postgresql resource from 13 to 14, the PostgreSQL instance gets successfully updated; however, the CronJobs for logical backup that belongs to that PostgreSQL instance would not get updated and the next logical backup fails with the following error:

```
pg_dumpall: error: aborting because of server version mismatch
pg_dumpall: error: server version: 14.0 (Ubuntu 14.0-1.pgdg18.04+1); pg_dumpall version: 13.4 (Ubuntu 13.4-4.pgdg18.04+1)
```

It seems that it's because the `PG_VERSION` filed in the spec.jobTemplate.spec.template.spec.container[0].env is not updated after the major version upgrade. After a manual removal of such CronJobs, reconciliation automatically creates a new CronJob resource whose PG_VERSION has a new version and solves this issue. It would be nice if Postgres Operator has an automatic update for that field too.

Thanks in advance!

farodin91:
Fixes #1678

### DIFF
--- a/pkg/util/k8sutil/k8sutil.go
+++ b/pkg/util/k8sutil/k8sutil.go
@@ -228,6 +228,16 @@ func getJobImage(cronJob *batchv1.CronJob) string {
 	return cronJob.Spec.JobTemplate.Spec.Template.Spec.Containers[0].Image
 }
 
+func getPgVersion(cronJob *batchv1.CronJob) string {
+	envs := cronJob.Spec.JobTemplate.Spec.Template.Spec.Containers[0].Env
+	for _, env := range envs {
+		if env.Name == "PG_VERSION" {
+			return env.Value
+		}
+	}
+	return ""
+}
+
 // SameLogicalBackupJob compares Specs of logical backup cron jobs
 func SameLogicalBackupJob(cur, new *batchv1.CronJob) (match bool, reason string) {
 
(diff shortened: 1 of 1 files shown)


farodin91:
@FxKu  Would you like to review? Thank you.

pkg/util/k8sutil/k8sutil.go
```diff
@@ -243,6 +253,13 @@ func SameLogicalBackupJob(cur, new *batchv1.CronJob) (match bool, reason string)
 			newImage, curImage)
 	}
 
+	newPgVersion := getPgVersion(new)
+	curPgVersion := getPgVersion(cur)
+	if newPgVersion != curPgVersion {
+		return false, fmt.Sprintf("new job's env PG_VERSION %q does not match the current one %q",
+			newImage, curImage)
```
bwrobc, line 260: Shouldn't this be `newPgVersion, curPgVersion` ?

FxKu:
I'm fine with this change. @sdudoladov any objections?

Having a unit test would be nice though :smiling_face_with_tear:

sdudoladov:
:+1:

FxKu:
:+1:

PR #2343, opened 2023-06-02
Enable specifying PVC retention policy for auto deletion

idanovinda:
### Problem
Users can specify the number of instances in the manifest, both scaling up and scaling down. It creates the PVCs when there is a scaling up event. However, where there is a scaling down event, the control plane doesn't automatically delete the PVCs created for StatefulSets that are no longer needed. It requires manual checking and deletion to clean up the unused PVCs. As a result, unused PVCs continue to occupy storage resources, leading to potential inefficiencies and unnecessary costs.

### Solution
In Kubernetes v1.23, a new feature has been introduced to address the issue of unused PVCs (see [docs](https://kubernetes.io/blog/2021/12/16/kubernetes-1-23-statefulset-pvc-auto-deletion/)). The operator should set the retention policy in the Statefulsets of every clusters. 

To specify the auto-deletion policy in the configmap, we can set:
```
persistent_volume_claim_retention_policy: "when_deleted:delete,when_scaled:delete"
```
while in the CRD configuration:
```
persistent_volume_claim_retention_policy: 
  when_deleted: "delete"
  when_scaled: "delete"
```
The default values are `"retain"` for `"when_deleted"` and `"when_scaled"`.


pkg/cluster/cluster.go
```diff
@@ -409,6 +409,12 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 		reasons = append(reasons, "new statefulset's pod management policy do not match")
 	}
 
+	if !reflect.DeepEqual(c.Statefulset.Spec.PersistentVolumeClaimRetentionPolicy, statefulSet.Spec.PersistentVolumeClaimRetentionPolicy) {
```
FxKu, line 412: we have a failing e2e test - so somehow this comparison must lead to a statefulset replacement. Can the order of the two fields be swapped? Maybe we need an explicit comparison method for the retention policy. Or some side is simply not set and nil.

FxKu:
As we've already discussed in person, you could try to extend the scaling e2e test with checks if PVCs get removed or not

SanjeevChoubey:
@idanovinda @FxKu this feature needed for us too , when we can expect this to be merged?

SanjeevChoubey:
> @idanovinda @FxKu this feature needed for us too , when we can expect this to be merged?

Can I pitch in and get the done, as we need this feature ASAP.

e2e/tests/test_e2e.py
```diff
@@ -1200,6 +1200,67 @@ def check_version_14():
...
+                "persistent_volume_claim_retention_policy": "when_deleted:retain,when_scaled:delete"
+            }
+        }
+        k8s.update_config(patch_scaled_policy_delete)
+        self.eventuallyEqual(lambda: k8s.get_operator_state(), {"0": "idle"}, "Operator does not get in sync")
+
+        pg_patch_scale_down_instances = {
+            'spec': {
+                    'numberOfInstances': 1
+                }
```
FxKu, line 1226: indentation looks slightly off here

e2e/tests/test_e2e.py
```diff
@@ -1200,6 +1200,67 @@ def check_version_14():
...
+        # decrease the number of instances
+        k8s.api.custom_objects_api.patch_namespaced_custom_object(
+            'acid.zalan.do', 'v1', 'default', 'postgresqls', 'acid-minimal-cluster', pg_patch_scale_down_instances)
+        self.eventuallyEqual(lambda: k8s.get_operator_state(), {"0": "idle"},"Operator does not get in sync")
+        self.eventuallyEqual(lambda: k8s.count_pvcs_with_label(cluster_label), 1, "PVCs is not deleted when scaled down")
+
+        pg_patch_scale_up_instances = {
+            'spec': {
+                    'numberOfInstances': 2
+                }
```
FxKu, line 1237: same here

FxKu:
:+1:

idanovinda:
👍

PR #2348, opened 2023-06-06
Fix basic installation on non default namespaces

j4m3s-s:
Manifests that are used with the default installation don't have a namespace (so it uses the current context's one).

It's my first time using this software so if it's possible I've misread things ~.

### DIFF
--- a/manifests/operator-service-account-rbac.yaml
+++ b/manifests/operator-service-account-rbac.yaml
@@ -2,8 +2,6 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: postgres-operator
-  namespace: default
-
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole


FxKu:
Can you also check other mentions of `namespace: default` in our example manifests? I think in the same file you changed the role binding is referencing the service account in `default` namespace. So I guess this has to be updated, too.

Note, these manifests where only meant as an example which users can edit to their own wishes incl. the namespace. But if it makes integration easier we can also remove it, of course.

j4m3s-s:
Hi @FxKu !
Thanks for replying.

I followed the "quickstart" installation which mentions : 
```
kubectl apply -k github.com/zalando/postgres-operator/manifests
```

All the concerned manifest by this kustomize installation do not reference the default namespace otherwise.
The only other reference to this serviceAccount is `minimal-fake-pooler-deployment.yaml` which doesn't mention the namespace either. So it should be all good to merge.

PR #2349, opened 2023-06-06
Update charts URL to an existing location

### LINKED ISSUE #2346: URL https://opensource.zalando.com/postgres-operator/charts/postgres-operator/  and ...-ui leads to 404 on github
The URL's 
https://opensource.zalando.com/postgres-operator/charts/postgres-operator/
and 
https://opensource.zalando.com/postgres-operator/charts/postgres-operator-ui/

lead's to 404 in github.

avandras:
Fixes #2346 (if I understood it right).

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -92,13 +92,13 @@ They only work only with Helm v3. Helm v2 support was dropped with v1.8.0.
 
 ```bash
 # add repo for postgres-operator
-helm repo add postgres-operator-charts https://opensource.zalando.com/postgres-operator/charts/postgres-operator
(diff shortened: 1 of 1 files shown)


avandras:
Anyone?

thiagowfx:
I don't think this change is correct:

```
helm repo add postgres-operator-charts https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator
heError: looks like "https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator" is not a valid chart repository or cannot be reached: error converting YAML to JSON: yaml: line 207: mapping values are not allowed in this context
```

PR #2352, opened 2023-06-13
Add service account name in connection pooler

cdmikechen:
Base on PR https://github.com/zalando/postgres-operator/pull/2255
Setting runAsUser directly is not allowed on, for example, Openshift.
So we should change the serviceaccount of the pooler to `postgres-pod` as well, otherwise the pooler will not be created correctly.

```
Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]
```


FxKu:
:+1:

idanovinda:
👍

Issue #1663, opened 2021-10-28
PodDisruptionBudget blocks master disruption


ckotzbauer:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** e.g. registry.opensource.zalan.do/acid/postgres-operator:v1.6.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal Kubernetes
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report

Hi,
I came accross this situation yesterday while preparing a Kubernetes-Cluster-Update. We have this operator deployed, which runs a three-pod-postgres-cluster. When I'm correct there should be always only one master-postgres-pod which holds the `spilo-role: master` label.
The operator creates always a PodDisruptionBudget (even when this is disabled via helm-chart). When it is enabled, the `minAvailable` setting is set to `1`. From outside perspective this seems fine, but the PDBs-selector settings are using the `spilo-role: master` label where always only one pod is running, so no disruptions are allowed at all. From my point of view, there's no problem to drain the master-pod as one of the replicas will immediately take over, so the PDB should not include the `spilo-role: master` role.

Is this assumption correct or did I miss something? 😄 
Thanks!

FxKu:
If you want to rotate nodes you can use the node_readiness_label the operator is watching. If it's removed a failover to one of the replicas will happen.

ckotzbauer:
> If you want to rotate nodes you can use the `node_readiness_label` the operator is watching. If it's removed a failover to one of the replicas will happen.

But this does not solve the problem I described above. I don't want to remove labels from pods, I want to drain a node without getting errors, cause of a wrong configured pdb...

ckotzbauer:
@FxKu I could do a PR to remove the `spilo-role: master` label from the PDB. Would that be okay?

Jan-M:
@ckotzbauer If you use the node readiness label as Felix suggest, then the postgres operator should terminate the master pod and migration works.

Please do not remove the master label from the pdb. What we can look at making this in general configurable. for this you can propose a PR. E.g. no pdb for single pod clusters.

ckotzbauer:
@Jan-M But the `kubectl drain` cannot handle the node-readiness label. From the `kubectl` perspective the pdb can't be "resolved" with its current labels.

Kuzbekov:
[...]

PR #2364, opened 2023-06-26
Make PodDisruptionBudget master label selector optional

b1zzu:
Fix: https://github.com/zalando/postgres-operator/issues/1663

Introduce the `pdb_master_label_selector` config that by default is enabled but if disabled will not add the `spilo-role: master` match label to the cluster PodDisruptionBudget.


b1zzu:
Hi,
I'm open to feedback for this PR, also I didn't know if they helm charts and manifests are autogenerated or if I have to take care of it

elliottneilclark:
This is awesome and would help us to use PDBs with Postgres.
(bumping to the top of the review pile?)

yoshihikoueno:
@FxKu @Jan-M 
Hi, I found this MR very useful and I'm sure many users out there also find the same way.  
Could you have time to take a look at it?

pkg/util/config/config.go
```diff
@@ -219,6 +219,7 @@ type Config struct {
 	ReplicaDNSNameFormat                     StringTemplate    `name:"replica_dns_name_format" default:"{cluster}-repl.{namespace}.{hostedzone}"`
 	ReplicaLegacyDNSNameFormat               StringTemplate    `name:"replica_legacy_dns_name_format" default:"{cluster}-repl.{team}.{hostedzone}"`
 	PDBNameFormat                            StringTemplate    `name:"pdb_name_format" default:"postgres-{cluster}-pdb"`
+	PDBMasterLabelSelector                   *bool             `name:"pdb_master_label_selector" default:"true"`
```
Jan-M, line 222: In line with the others why not call it enable?

Jan-M:
@FxKu sounds like something useful to many people, tiny comment and then looks like we can tackle it.

I think it is important though for readers to understand that with this, you hand of to your infrastructure the responsibility to migrate masters only in healthy situations.

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -1420,7 +1423,7 @@ var OperatorConfigCRDResourceValidation = apiextv1.CustomResourceValidation{
 										},
 									},
 								},
-							},
+              },
```
FxKu, line 1426:
```suggestion
},
```

FxKu, line 1426:
```suggestion
							},
```

FxKu:
:+1:

idanovinda:
👍

PR #2374, opened 2023-07-17
feat(backup): Add timezone parameter for logical backup cronjob

thcdrt:
Hello,

I would like to add the timezone parameter inside the logical backup cronjob.

WDYT ?

Thomas


Jan-M:
Might be good to default this to utc or mention what the default is?

PR #2382, opened 2023-07-26
Fix the wrong default value description in cluster manifest

yoshihikoueno:
In `Prepared Databases` section of the cluster manifest documentation,  
it says `defaultRoles` and `defaultUsers` are **both disabled** by default for prepared schemas.

But `defaultRoles` is actually enabled by default, thus this MR fixes it.


FxKu:
:+1:

jopadi:
👍

FxKu:
Thanks @yoshihikoueno for the docs fix.

PR #2389, opened 2023-08-08
feat: Add pod_service_account_annotations

dennislapchenko:
Added `pod_service_account_annotations`.
This will allow avoiding custom creation of service account with gcp/aws annotations for wal needs.
Since by default service account is created, makes sense to allow passing annotations as well.

Also tested locally via `./run_operator_locally.sh`


oopjot:
Hi, are there any plans regarding this feature? This would be very helpful for IRSA users. Thank you.

dennislapchenko:
> Hi, are there any plans regarding this feature? This would be very helpful for IRSA users. Thank you.

There actually exists a way how to already add this, but its less than ideal when you only need an annotation..
So hopefully this PR still gets merged.

```
configKubernetes:
  pod_service_account_definition: |
    apiVersion: v1
    kind: ServiceAccount
    metadata:
      annotations:
        eks.amazonaws.com/role-arn: <role_arn>
      name: postgres-operator-pod
      namespace: postgres-operator
```

namespace seems to be irrelevant too, since in required namespaces the SA is still created correctly

oopjot:
@dennislapchenko I must've missed this field, thank you so much!

psolarcz:
@dennislapchenko Can we add a unique role per cluster? Currently I'm able to add same annotation for all SAs created by the operator.

dennislapchenko:
> @dennislapchenko Can we add a unique role per cluster? Currently I'm able to add same annotation for all SAs created by the operator.

I am not fully aware of the inner workings, it would make sense to be able to overwrite this from each cluster, but it would most likely require some heavy code alterations.

Issue #2358, opened 2023-06-21
LOGICAL_BACKUP_S3_BUCKET is used in the path for Azure uploads, azure does not require the bucket name on the command line as a prefix.


EraYaN:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/logical-backup:v1.10.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AKS
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Enhancement

It's this line: https://github.com/zalando/postgres-operator/blob/master/docker/logical-backup/dump.sh#L48

I wonder why that variable is there. It results in "my-bucket-url" if it's not set, which seems undesirable.

EDIT seems like setting it to "" will result in a no name folder, also not great.

Issue #2337, opened 2023-05-24
Using pg_dump as logical backup rather than pg_dumpall


nihaldivyam:
- **Which image of the operator are you using?** 
-- registry.opensource.zalan.do/acid/postgres-operator:v1.9.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?**
-- Bare Metal K8s
- **Are you running Postgres Operator in production?** 
-- Yes
- **Type of issue?** 
-- Feature request

Presently, the PostgreSQL Operator allows for logical backups using the `pg_dumpall` utility. However, during the restoration process, it restores all databases rather than the specific one we desire. This becomes problematic, especially on a fresh Postgres cluster, as it includes the restoration of the `postgres` database, resulting in the overwriting of all passwords. Consequently, when other pods access the updated secret, authentication issues arise due to the continued usage of outdated passwords.

I suggest enhancing the PostgreSQL operator by transitioning to the use of `pg_dump` for logical backups instead of  `pg_dumpall` . This modification would yield a more efficient and user-friendly operator, enabling selective restoration of individual databases while avoiding the unintended modification of critical system databases.

EraYaN:
We just ran into this during a data restore event and we are going to build a solution for logical backups in the directory form and per database. That should speed up restores and let us restore specific items and in parallel.

PR #2392, opened 2023-08-11
Initial azure version for per database backups

EraYaN:
Is upstream interested in something like this?

This will only do it for Azure (for now).

It runs dumpall for the globals and pg_dump for each database (in parallel mode). This makes the resulting backups way more usable.

This would be related to: #2337 #2358

Would it be a good idea to port the script to something like python?


raviranjanelastisys:
We would like to have this feature enabled , any update ?

raviranjanelastisys (approved):

FxKu:
Sorry for the late comment, but this change looks quiet hacky to me. I think, you could just build your own logical backup image and then reference it in the configuration. I will close this PR.

Issue #1373, opened 2021-02-19
[UI] Helm chart feature request: Watch multi namespaces and parameterize teams


caiconkhicon:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator-ui:v1.6.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K8s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Feature request

I want to use Helm to deploy postgres-operator-ui (https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator-ui). However, it lacks of two features:
- Multi namespaces: my databases are deployed in many namespaces. ~With the current Helm chart, it is possible to watch one namespace only~. It is possible to watch all namespaces with `targetNamespace: *`. Please create a documentation for the `postgres-operator-ui` Helm chart to mention this.
- Parameterize team names: Team names are specified as a list. However, it is fixed with one item named `acid`. I have tested changing it manually and it worked. So please implement it. (https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator-ui/templates/deployment.yaml#L54-L58)

caiconkhicon:
I created https://github.com/zalando/postgres-operator/pull/1375 to add the TEAMS feature

yfoelling:
@caiconkhicon you can watch all namespaces by setting it to:
```  targetNamespace: "*"```

i found that out by trial-error, since i also did not find it in the documentation

FxKu:
fyi we have a reference doc for all config options where this is explained: https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md#kubernetes-resources as well as in our [admin](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#specify-the-namespace-to-watch) docs.

caiconkhicon:
@FxKu : I don't think that it's exactly correct. The documentation is for `postgres-operator`, not for the UI. The mentioned parameter is also `watched_namespace`, not `targetNamespace` for the UI.

FxKu:
ah yes, true for UI there's not much documentation. But yes, it works with `targetNamespace: "*"` indeed. Not sure, if this should be the default.

caiconkhicon:
[...]

PR #2412, opened 2023-09-10
Document that `*` can be used for UI `TARGET_NAMESPACE` option

MisguidedEmails:
Similar to `WATCHED_NAMESPACE` for the operator itself, `*` is a valid option for the UI. Without it clusters in multiple namespaces - rather, those not in the `TARGET_NAMESPACE` - will not show in the UI. Which can be confusing for users.

Mentioned in: https://github.com/zalando/postgres-operator/issues/1373


FxKu:
:+1:

idanovinda:
👍

PR #2415, opened 2023-09-12
on PVC when_scaled:retain check pod count, too

FxKu:
With #2343 merged the e2e pipeline is not passing anymore. At the moment, I suspect that scaling down the cluster to one instance but only checking PVC count happens very quickly so that we run without replicas into the subsequent e2e test which leads to:

```
======================================================================
FAIL: test_resource_generation (test_e2e.EndToEndTestCase)
Lower resource limits below configured minimum and let operator fix it.
  File "/usr/local/lib/python3.8/dist-packages/timeout_decorator/timeout_decorator.py", line 82, in new_function
  File "/tests/test_e2e.py", line 1275, in test_resource_generation
    self.assertNotEqual(replica_nodes, [])
AssertionError: [] == []
```


FxKu:
:+1:

PR #2418, opened 2023-09-14
ppc64le multiarch support

sumitd2:
*The multiarch image build has been performed successfully (locally) in an Intel Ubuntu 22.04 container.
*A recent version of the image has been tested on a production ppc64le Openshift cluster.

### DIFF
--- a/.github/workflows/publish_ghcr_image.yaml
+++ b/.github/workflows/publish_ghcr_image.yaml
@@ -53,4 +53,4 @@ jobs:
           push: true
           build-args: BASE_IMAGE=alpine:3.15
           tags: "${{ steps.image.outputs.NAME }}"
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64,linux/arm64,linux/ppc64le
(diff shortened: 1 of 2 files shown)


RajeshJeyapaul (commented):
Why the check for "ppc64el" at line 6 ? thanks

sumitd2:
> Why the check for "ppc64el" at line 6 ? thanks

Because on line 19, the link 
https://storage.googleapis.com/golang/go1.19.8.linux-${arch}.tar.gz
should be
https://storage.googleapis.com/golang/go1.19.8.linux-**ppc64le**.tar.gz

RajeshJeyapaul:
> > Why the check for "ppc64el" at line 6 ? thanks
> 
> Because on line 19, the link `https://storage.googleapis.com/golang/go1.19.8.linux-${arch}.tar.gz` should be `https://storage.googleapis.com/golang/go1.19.8.linux-ppc64le.tar.gz`

ok..  used  to tag the filename with architecture should be ok... I understand that  "ppcel" and "ppcle" both indicates the little endian but some flavours still refer it as "el" (Debian I guess)..Just to ensure that it does not break anywhere if we do similar check on the flavour which still uses "el"..thanks

PR #2421, opened 2023-09-15
add support for recovery section in event streams

FxKu:
This PR adds support for Dead Letter Recovery in event streams. So far, only Nakadi is supported as recovery sink (like with data sink). Recovery can be enabled for an entire stream object (`enableRecovery`) or for each stream table individually by specifying an `recoveryEventType`. If recovery is enabled, but no recovery type is specified the generated stream CRD will use an event type with the `-dead-letter-queue` suffix. The recovery event type must already exist in Nakadi. If the name is different from the operator default (event type name + suffix) one has to specify it with recoveryEventType`.


FxKu:
:+1:

PR #2424, opened 2023-09-19
Update prepared database section section with secret name info

spreeker:
add a little note about  underscores / '_'  in the manifest will be dash '-' in the k8s secrets.  

I wasted a bit long time looking for the correct user names..


FxKu:
:+1:

jopadi:
👍

Issue #1838, opened 2022-04-06
dump.sh and k8sres.go utilize LOGICAL_BACKUP_S3_BUCKET, even on GCP


rocket357:
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.7.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** GCP
- **Are you running Postgres Operator in production?** not yet (for GCP)
- **Type of issue?** Documentation (variable name confusion)

https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L2144

Generated logical backup pod config uses "LOGICAL_BACKUP_S3_BUCKET" even on GCP (also: https://github.com/zalando/postgres-operator/blob/master/docker/logical-backup/dump.sh#L113).  This seems confusing, as I would expect it to utilize LOGICAL_BACKUP_GS_BUCKET (following along with WALE_GS_PREFIX/WALG_GS_PREFIX in the spilo project).

It's simple enough to use LOGICAL_BACKUP_S3_BUCKET for now, but it seems it'd make more sense to migrate that to LOGICAL_BACKUP_GS_BUCKET at some point.

FxKu:
Hm, sounds indeed like someone was too lazy back then implementing a correct set of environment variables. Feel free, to publish a PR where `generateLogicalBackupPodEnvVars` method reflects the config better - like `generateSpiloPodEnvVars` does it.

rocket357:
Working on a PR.

rocket357:
Apologies for the delays.  I see Azure has been handled separately, so this issue and referring PR will now just focus on GCS logical backup environment variables.

Lxrdknows77:
@rocket357 @FxKu  why its implemented in that way that all logical backup env variables are fetched from global operator Config, so that force to use same s3 for logical backups for all clusters. Just wondering why there is no option to fetch these variables from env variables of postgresql manifest itself (as its done for backup s3 bucket).

rocket357:
@Lxrdknows77 Ahh, good point.  My bad on implementing in the global operator config.  Your concerns are noted.

That said, my company no longer has a use-case for logical backups, so we'll be disabling them in our configs.  I've closed the PR for this since the implementation is bad and I'm not going to be working on a replacement implementation.

PR #2429, opened 2023-09-21
Update logical backup environment variables for GCS

rocket357:
Per #1838 break GCS logical backup bucket environment variable into its own variable "LOGICAL_BACKUP_GS_BUCKET" (currently using LOGICAL_BACKUP_S3_BUCKET).

While here, update dump.sh to utilize the new GCS logical backup environment variable and document where required.

I looked at cloning aws_delete_outdated in dump.sh as well, but given gsutil doesn't currently have a mechanism for easily listing files older than a given cutoff date, I felt it better to defer on this.  Suggestions welcome for implementing gcs_delete_outdated in dump.sh if there is sufficient interest.

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -503,6 +503,8 @@ spec:
                   logical_backup_docker_image:
                     type: string
                     default: "registry.opensource.zalan.do/acid/logical-backup:v1.10.1"
+                  logical_backup_gs_bucket:
+                    type: string
                   logical_backup_google_application_credentials:
                     type: string
                   logical_backup_job_prefix:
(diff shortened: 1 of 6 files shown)


Jan-M:
Somehow we end up having lots of different configs for "buckets". With this change we might also break user config using S3 bucket to configure there GCS bucket?

rocket357:
> Somehow we end up having lots of different configs for "buckets". With this change we might also break user config using S3 bucket to configure there GCS bucket?

It does seem to me that having a "LOGICAL_BACKUP_BUCKET_NAME" and "LOGICAL_BACKUP_BUCKET_PROVIDER" type of setup would be more efficient.

That said, my company has moved away from using logical backups as our use case for logical backups is no longer required.

PR #2445, opened 2023-10-12
improve additional teams lookup

FxKu:
We came across a team mapping resource where 11 teams where cross referencing one another. Although the lookup of additional teams is already and [exclusion filter](https://github.com/zalando/postgres-operator/blob/dc872abb51efb87df982bf74f636a5a3ff10d0e0/pkg/teams/postgres_team.go#L69) it seemed that there were still to many lookups performed bringing the operator to a halt.

```
"DeltaFIFO Pop Process" ...,Depth:946,Reason:slow event handlers blocking the queue
```

The PR changes how `fetchAdditionalTeams` must be called. The team for which the lookup is done must already be added to the `exclude` filter so that additional teams will be added to the filter as well before doing a recursive call. Transitive teams will be added to the exclude filter, too, to avoid unnecessary look ups in case they are also an additional team to the original team.

I counted the lookups for [one of the unit tests](https://github.com/zalando/postgres-operator/blob/dc872abb51efb87df982bf74f636a5a3ff10d0e0/pkg/teams/postgres_team_test.go#L126) and the number of loops was reduced from 11 to 5 with only little cross-referecing.

### DIFF
--- a/pkg/teams/postgres_team.go
+++ b/pkg/teams/postgres_team.go
@@ -1,6 +1,9 @@
 package teams
 
 import (
+	"fmt"
+	"strings"
+
 	acidv1 "github.com/zalando/postgres-operator/pkg/apis/acid.zalan.do/v1"
 	"github.com/zalando/postgres-operator/pkg/util"
 )
@@ -59,6 +62,7 @@ func fetchTeams(teamset *map[string]struct{}, set teamHashSet) {
 func (ptm *PostgresTeamMap) fetchAdditionalTeams(team string, superuserTeams bool, transitive bool, exclude []string) []string {
 
 	var teams []string
+	fmt.Println("fetching for team " + team + " excluded teams: " + strings.Join(exclude, ", "))
 
 	if superuserTeams {
 		teams = (*ptm)[team].AdditionalSuperuserTeams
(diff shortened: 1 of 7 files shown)


pkg/teams/postgres_team.go
```diff
@@ -66,13 +66,20 @@ func (ptm *PostgresTeamMap) fetchAdditionalTeams(team string, superuserTeams boo
 		teams = (*ptm)[team].AdditionalTeams
 	}
 	if transitive {
-		exclude = append(exclude, team)
 		for _, additionalTeam := range teams {
 			if !(util.SliceContains(exclude, additionalTeam)) {
+				// remember to no check team and additionalTeam again
```
Jan-M, line 71: `not`

pkg/teams/postgres_team.go
```diff
@@ -66,13 +66,20 @@ func (ptm *PostgresTeamMap) fetchAdditionalTeams(team string, superuserTeams boo
...
+				exclude = append(exclude, additionalTeam)
 				transitiveTeams := (*ptm).fetchAdditionalTeams(additionalTeam, superuserTeams, transitive, exclude)
 				for _, transitiveTeam := range transitiveTeams {
-					if !(util.SliceContains(exclude, transitiveTeam)) && !(util.SliceContains(teams, transitiveTeam)) {
-						teams = append(teams, transitiveTeam)
+					if !(util.SliceContains(exclude, transitiveTeam)) {
+						// remember to no check transitive team again in case
```
Jan-M, line 76: `not`

jopadi:
👍

Issue #2448, opened 2023-10-17
Feature Request: Support Running Sidecar With a Command


tabbyl21:
Hi, I would like to support running a sidecar using a command. This is useful in the case that the container being run as a sidecar does not have an entry point and needs to be run explicitly using a command.

Here are my answers to the requested questions:
- I am using the latest operator image. 
- I am running Zalando on GKE and kind as of now. 
- I am not running it in production presently.
- I am using the latest operator image. 
- This is a feature request

I have run the change on my own fork, it is a two liner change. :)

tabbyl21:
PR with required changes: https://github.com/zalando/postgres-operator/pull/2449

FxKu:
#2449 has been merged

PR #2449, opened 2023-10-17
Feat: Support Running Sidecar with a Command.

tabbyl21:
This PR addresses issue #2448 . Some containers may not have entry points, if this is the case they would need to be run using a command. This change extends the definition of sidecar so that there is an optional command field. If the field is present then the container will be run using that command. This is a two line change that is fully backward compatible.


tabbyl21:
Hi @Jan-M @jopadi ,

I am wondering what I need to do to get this PR accepted into your project! I believe I have followed the guidelines from the ReadME to the extent its applicable! Please let me know if there is anything else I need to add!

Thank you for your time,
Tabatha

AleksandrPaul:
Hi @Jan-M , @sdudoladov , @FxKu , @idanovinda , @hughcapet, @jopadi 

Could you please review this PR?

sgielen:
Hi @FxKu could you please review this PR? It is similar to a PR you recently reviewed. Thanks

kamenskiyyyy:
Any updates?

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -220,6 +220,7 @@ type Sidecar struct {
 	DockerImage string             `json:"image,omitempty"`
 	Ports       []v1.ContainerPort `json:"ports,omitempty"`
 	Env         []v1.EnvVar        `json:"env,omitempty"`
+	Command     []string           `json:"command,omitempty" protobuf:"bytes,3,rep,name=command"`
```
FxKu, line 223: why is `protobuf:"bytes,3,rep,name=command"` necessary? We have it nowhere else. Is it similar to filter inputs using pattern in the CRD schema?

FxKu:
@tabbyl21 sorry for not reacting for a looong time. We are usually a bit reluctant to changes that offer more config mean and adding complexity. After taking a closer look I get now why this change is needed. Being able to run any command via sidecar just sounded too scary to me first (well, maybe because it is). If you could answer my one question we could merge it then.

Try to run codegen (./hack/update-codegen.sh) to see if something changes.

icc-garciaju:
One question here.
Why only "command" and not include "args" too?

sgielen:
> One question here. Why only "command" and not include "args" too?

It is a []string so it includes the args, too

pkg/apis/acid.zalan.do/v1/postgresql_type.go
tabbyl21: apologies just seeing this comment. I can remove that part

tabbyl21:
@FxKu apologies didn't see your repsonses! I have addressed your comment! Let me know what changes need to be made to merge this in!

hughcapet:
👍

FxKu:
:+1:

PR #2462, opened 2023-10-27
Add Pod Service Account Annotation update option

### LINKED ISSUE #2459: AWS IRSA - Service Account Annotation
Currently I'm using [pod_service_account_definition](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values.yaml#L186) to add annotation to the role:

```
configKubernetes:
  pod_service_account_definition: '{"apiVersion":"v1","kind":"ServiceAccount","metadata":{"annotations":{"eks.amazonaws.com/role-arn":"arn:aws:iam::ACCID:role/postgres-role"}}}'
```

With this setup I'm unable to use unique role per postgres cluster - it's needed to set proper permissions on the S3 bucket.
Is it possible to generate a unique role per cluster?

psolarcz:
Adding option to specify list of annotation keys, for example: `eks.amazonaws.com/role-arn` - that will be appended with the cluster name. It will allow to use unique IAM Role per cluster.

To fix: https://github.com/zalando/postgres-operator/issues/2459

### DIFF
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -128,6 +128,7 @@ data:
   # pod_priority_class_name: "postgres-pod-priority"
   pod_role_label: spilo-role
   # pod_service_account_definition: ""
+  # pod_service_account_annotation_per_cluster: "eks.amazonaws.com/role-arn"
   pod_service_account_name: "postgres-pod"
   # pod_service_account_role_binding_definition: ""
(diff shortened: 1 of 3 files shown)


Art3mK:
that would be really nice 👍 Not sure if appending cluster name is a good idea, I'd like to just pass role ARN via manifest to the SA.

PR #2471, opened 2023-11-17
Rancher integration

schmidax:
Revision of the Helm-Charts for use in the Rancher dashboard as an app.
grafik
grafik
grafik
grafik


FxKu:
image

Wat?! We need a lot more explanation here than just screenshots if you want us to review this PR. Why does rancher integration touch 246 files? What are things that are currently missing? No need to write a whole essay but I expect to see a short summary of what you've changed. The diff is just too big and scary.

schmidax:
The 246 changes actually only have a small background. Your procedure with the chart versions was to save the old versions as tar files. In order to be able to install the older versions of helm (and Rancher in particular), a separate folder must exist for each version containing the templates for the appropriate version.

For Rancher directly, only one extra file per chart is required. The questions.yaml, which prepares the variables from the values.yaml for Rancher in a readable format. I have not yet built the README for Rancher.

I have only done the Rancher preparation for version 1.10.0 and 1.10.1. I unpacked the tar-files into the version folders for all versions. But this is just a suggestion from me. You can only do this for 1.10.0 and higher...

schmidax:
Hi there,

I have now understood how to deal with helm charts. This index.yaml was not quite clear to me. Now I have added the integration into the existing construct (with version 1.11.0) Sorry for this confusion.

PR #2472, opened 2023-11-17
Standby cluster promotion by changing manifest

FxKu:
Continuation of #1035 

Config update must happen after rolling update that changes STANDBY_* env vars. Added a unit test and updated documentation.


idanovinda:
👍

PR #2474, opened 2023-11-19
Add topologySpreadConstraints configuration.

laiminhtrung1997:
Dear all,

I think we should configure topologySpreadConstraints to pod spec so these pods can spread zones for Highly Availability.

Could someone review it, please? Thank you very much.

Best regards.


laiminhtrung1997:
I moved to PR #2530 for re-coding.

PR #2480, opened 2023-11-23
chore: remove refs to deprecated io/ioutil

testwill:
ioutil. -> os.


FxKu:
Thanks @testwill. Can you replace all occurrences we have in our code? We have some unit test files:
- patroni_test.go
- pod_test.go 
- sync_test.go

FxKu:
@testwill you're going to reflect my review or not? Then I would close it and do the switch everywhere in a separate PR.

PR #2489, opened 2023-11-29
let isSystemUsername check all system users

FxKu:
When one defines a `fes_user` in the manifest and has also `streams` defined, the database password of this user gets overridden. The secret, however, is not touched. Therefore, removing the user from the manifest immediately fixes the issue. 

I found that when initializing robot / manifest users the check for name collisions with already defined system users is only done for the configured `SuperUsername` and `ReplicationUsername` - not for e.g. pooler of stream user. I changed it now to a loop over all system users that are already initialized at this point (not in the unit test though). Because, if this check does not happen, somewhere done the line the database role picks up a new [random password](https://github.com/zalando/postgres-operator/blob/96077c47d64db54bdf60d7e1eefeea3dee6022b5/pkg/cluster/cluster.go#L1325). `fes_user` will be part of c.pgUsers and override the existing sytem user with the same name in the list of `newUsers` to sync [here](https://github.com/zalando/postgres-operator/blob/96077c47d64db54bdf60d7e1eefeea3dee6022b5/pkg/cluster/sync.go#L1041) and [here](https://github.com/zalando/postgres-operator/blob/master/pkg/util/users/users.go#L62). Interestingly, I find no ALTER ROLE statements in the Postgres logs (probably because we don't want to log passwords).

Btw: The same bug affects the connection pooler role as well.

### DIFF
--- a/pkg/cluster/cluster_test.go
+++ b/pkg/cluster/cluster_test.go
@@ -164,6 +164,7 @@ func TestInitRobotUsers(t *testing.T) {
 			err:           nil,
 		},
 	}
+	cl.initSystemUsers()
 	for _, tt := range tests {
 		cl.Spec.Users = tt.manifestUsers
 		cl.pgUsers = tt.infraRoles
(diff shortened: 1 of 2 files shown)


jopadi:
👍

FxKu:
:+1:

PR #2490, opened 2023-11-30
remove Users section from Patroni Bootstrap

FxKu:
Patroni is about to remove the users creation functionality. (https://github.com/zalando/patroni/blob/master/docs/releases.rst#version-320).

Since the operator is relying on Spilo to create the Pam role via env variable [HUMAN_ROLE](https://github.com/zalando/postgres-operator/blob/96077c47d64db54bdf60d7e1eefeea3dee6022b5/pkg/cluster/k8sres.go#L951) the operator is not affected by this change.

### DIFF
--- a/pkg/cluster/cluster_test.go
+++ b/pkg/cluster/cluster_test.go
@@ -26,8 +26,8 @@ import (
 const (
 	superUserName       = "postgres"
 	replicationUserName = "standby"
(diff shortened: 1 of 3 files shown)


hughcapet:
👍

FxKu:
:+1:

idanovinda:
👍

Issue #2492, opened 2023-12-05
array index out of range


seekingua:
when the desired statefulset has different numbers of volume claim template with current cluster,  will be panic because of index out of range

code line:
`https://github.com/zalando/postgres-operator/blob/96077c47d64db54bdf60d7e1eefeea3dee6022b5/pkg/cluster/cluster.go#L483C3-L483C3`

FxKu:
Merged your PR.

PR #2493, opened 2023-12-05
Fix VolumeClaimTemplates index out of range problem

seekingua:
- when the desired statefulset has different numbers of volume claim template with current cluster,  will be panic because of index out of range
- when got different numbers of volume claim template, maybe no need to compare details, after all, it need to be replaced already

issue: [[https://github.com/zalando/postgres-operator/issues/2492](https://github.com/zalando/postgres-operator/issues/2492)]


idanovinda:
👍

FxKu:
:+1:

FxKu:
Thanks @seekingua for this bugfix :)

Issue #2441, opened 2023-10-10
no option to use an existing priorityClassName for posgres Pods or make podPriorityClassName value configurable 


antonmatsiuk:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?**: registry.opensource.zalan.do/acid/postgres-operator:v1.10.1
- **Where do you run it**: cloud (AWS, Azure)
- **Are you running Postgres Operator in production?**: yes
- **Type of issue?**: [Bug report]

In the current implementation there is no option to specify an existing `priorityClassName` for Postgres Cluster pods. The actual value for `priorityClassName` is set only for [postgres-operator](https://github.com/zalando/postgres-operator/blob/dc872abb51efb87df982bf74f636a5a3ff10d0e0/charts/postgres-operator/templates/deployment.yaml#L79) deployment. 

In contrast, `podPriorityClassName`  parameter designed for PostgreSQL cluster disrespects existing priority classes in a cluster and [creates a new priority class](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/templates/postgres-pod-priority-class.yaml) with a [static priority value](https://github.com/zalando/postgres-operator/blob/dc872abb51efb87df982bf74f636a5a3ff10d0e0/charts/postgres-operator/templates/postgres-pod-priority-class.yaml#L15). This makes using the `podPriorityClassName` parameter unusable in production clusters where priority classes are created and managed externally. If an existing priority class is specified under `podPriorityClassName` (`podPriorityClassName`:`critical` in the example below), Helm Chart throws the following error: 
```
Helm upgrade failed: error while running post render on files: may not
        add resource with an already registered id:
        PriorityClass.v1.scheduling.k8s.io/critical.[noNs]
``` 

**Solution**: add additional `create` boolean flag to [pod-priority-class](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/templates/postgres-pod-priority-class.yaml#L15) to ignore its creation and to allow usage of existing external priority classes

PR #2499, opened 2023-12-18
Feat: add flags for podPriorityClassName to support external priorityclasses

capuche2412:
Hello,

To solve this issue https://github.com/zalando/postgres-operator/issues/2441, I present a possible solution by adding flags to create the PriorityClass (or not) and to set the priority. In case of a use of an external priorityclass, only the name should be set.

### DIFF
--- a/charts/postgres-operator/templates/configmap.yaml
+++ b/charts/postgres-operator/templates/configmap.yaml
@@ -10,9 +10,9 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Release.Name }}
 data:
-  {{- if .Values.podPriorityClassName }}
-  pod_priority_class_name: {{ .Values.podPriorityClassName }}
-  {{- end }}
+{{- if or .Values.podPriorityClassName.create .Values.podPriorityClassName.name }}
+  pod_priority_class_name: {{ include "postgres-pod.priorityClassName" . }}
+{{- end }}
   pod_service_account_name: {{ include "postgres-pod.serviceAccountName" . }}
 {{- include "flattenValuesForConfigMap" .Values.configGeneral | indent 2 }}
 {{- include "flattenValuesForConfigMap" .Values.configUsers | indent 2 }}
--- a/charts/postgres-operator/templates/postgres-pod-priority-class.yaml
+++ b/charts/postgres-operator/templates/postgres-pod-priority-class.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.podPriorityClassName }}
+{{- if .Values.podPriorityClassName.create }}
 apiVersion: scheduling.k8s.io/v1
 description: 'Use only for databases controlled by Postgres operator'
 kind: PriorityClass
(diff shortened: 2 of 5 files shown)


charts/postgres-operator/values.yaml
```diff
@@ -462,7 +462,10 @@ podServiceAccount:
 priorityClassName: ""
 
 # priority class for database pods
-podPriorityClassName: ""
+podPriorityClassName:
+  create: true
```
FxKu, line 466: Lets copy the behavior of the service account. If create is true a priority class should get created and if name is empty then we use the default `default (printf "%s-%v" (include "postgres-operator.fullname" .) "pod")` (see _helpers.tpl file)

charts/postgres-operator/templates/postgres-pod-priority-class.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.podPriorityClassName }}
+{{- if and .Values.podPriorityClassName.create .Values.podPriorityClassName.name }}
```
FxKu, line 1: .create should be the only condition here (see my other comment)

capuche2412:
I notice another scenario that is not being taken into account. If someone explicitly does not want to set a priorityClassName, it is not possible to do so here. A priorityClass will still be created. This could be perceived as unnecessary boilerplate, especially in a cluster that does not utilize custom priorityClasses at all.
Should an option like `enabled: true` be added to address this?

FxKu:
I don get your last comment. The default should be false like before. The priority class should not get created if somebody does not want to use it. You have to enable it manually in the values.yaml file. (It's funny after all those years I still see no use in helm charts haha)

charts/postgres-operator/templates/configmap.yaml
```diff
@@ -10,9 +10,7 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Release.Name }}
 data:
-  {{- if .Values.podPriorityClassName }}
-  pod_priority_class_name: {{ .Values.podPriorityClassName }}
-  {{- end }}
+  pod_priority_class_name: {{ include "postgres-pod.priorityClassName" . }}
```
capuche2412, line 13: My comment was not very clear indeed. I'll try to be clearer and more explicit.

Previously, before the changes, if `.Values.podPriorityClassName` was not set, then there would be no  `pod_priority_class_name` field in the configmap. Hence, the pod's PriorityClass is set to the default one. 
So if someone wants to set the pod's PriorityClass to the default one, they would leave the field  `.Values.podPriorityClassName` empty.

Now, with the changes, the field  `pod_priority_class_name` will always be in the configmap. It will be either `"postgres-operator.fullname"-pod` or `.Values.podPriorityClassName.name`.
Maybe, there is a case in which someone does not want to set a `pod_priority_class_name` at all to set the PriorityClass to the default one.

FxKu: ah now I see your change. Why not leave the if and change to {{- if .Values.podPriorityClassName.create }} ?

capuche2412: I think it is better to put `For .Values.podPriorityClassName.create .Values.podPriorityClassName.name }}` because it covers all the cases :
1) `create: false` & `name: ""` : the pod PriorityClass is the default one
2) `create: true` & `name: ""` : the pod PriorityClass is created with the default name
3) `create: false` & `name: "my-custom-class"` : the pod PriorityClass is set to `my-custom-class`
4) `create: true` & `name: "my-custom-class"` : the pod PriorityClass is created with the name `my-custom-class`

With only  ` {{ if .Values.podPriorityClassName.create }}`, the case 3 (use of an external priority class) is not possible and it is the reason why I have open this PR

FxKu: Ok got it now. What is the default name? `default`? Can you document the behavior (what you wrote here) in the values.yaml file?

capuche2412: Sure !

FxKu:
:+1:

jopadi:
👍

PR #2456, opened 2023-10-21
Populating crd labels and annotations in logical backup job pod manifest

andrejshapal:
Hello, 

Fixing https://github.com/zalando/postgres-operator/issues/1834

1. Making sure InheritedLabels set in operator settings / cluster settings are added to backup job pod manifest.
2. Making sure InheritedAnnotations set in operator settings / cluster setting are added to backup job pod manifest.

The goal is to see backup job resources in ArgoCD.

PR #2509, opened 2024-01-09
add unit test and documentation for finalizers

FxKu:
#941 added support for using finalizers. However, the PR lacked documentation and a unit test. We decided to merge it anyway due to the age of the PR in favor for having a quick follow-up PR which adds it.

The unit test became larger than anticipated:

- testing the whole Create() process of cluster.go
- All finalizer functions were turned private as they are not accessed from anywhere else in the code.
- Finalizers will not only added on Create and removed on Delerte but also managed sync
- The introduced JSON patch dependency was removed in favor for marshaling with Go like we do with the status
- The patch/update of Postgresql resource was moved to it's own function in k8sutil
- RemoveString function moved into util package and added unit test
- log messages starting with capital letter were lower cased

This PR also remove a binary debug file which was merged with #2456, unfortunately.


jopadi:
👍

FxKu:
:+1:

PR #2311, opened 2023-05-04
Add hugepages 2Mi and 1Gi fields to ResourceDescription and pass them to the statefulset

silenium-dev:
This is a simple approach to these issues and just allows hugepages requests and limits to be specified in the CR for postgres container and eventual sidecars.
At the moment, it is not possible to use the operator on clusters with enabled hugepages support, because postgres detects and tries to use hugepages, but fails to allocate as the limit on containers defaults to zero. This PR allows specifying a little amount of hugepages to prevent postgres from crashing.

Closes #1788 and closes #1549

PR #2512, opened 2024-01-11
hugepages empty on default and updated date in codegen files

FxKu:
For #2509 I need to implement an update call on the Postgres resource. Turns out this fails if hugepages (introduced just recently with #2311) are not specified in the manifest. For CPU and memory we assign defaults if they are not specified but hugepages remain an emtpy string which [violates](https://github.com/zalando/postgres-operator/blob/743aade45f2dca0b4246b6734ff5ef253aac41c1/manifests/postgresql.crd.yaml#L444) the CRD on update call.

This PR turns the hugepages fields into pointers so they are unset during updates against the cluster manifest.

FYI: I have re-run code generation which updated the year in many files.


FxKu:
:+1:

idanovinda:
👍

PR #2521, opened 2024-01-24
adding finalizer option in all manifests and rework its docs

FxKu:
The `enable_finalizers` option added in #941 was still missing in the example manifests. I also felt I had to add two more sentences to its docs.


jopadi:
👍

FxKu:
:+1:

PR #2524, opened 2024-01-29
allow empty resources when defaults are empty

### LINKED ISSUE #2057: Allow setting null resource requests / limits
> We've removed the `required` flag for `resources` in the CRD schema starting from `v1.8.0`.

_Originally posted by @FxKu in https://github.com/zalando/postgres-operator/issues/2028#issuecomment-1234071399_

This isn't enough to not have any CPU/memory limits for containers. If the value isn't set in the CRD then the configured default or minimum value takes effect. If those are set to 0 then Kubernetes can't create pods because the limit would be lower than the requests.

We need some way of _not setting_ a CPU limit at all in order to avoid unnecessary throttling. Please see https://home.robusta.dev/blog/stop-using-cpu-limits/ for a discussion about this.

FxKu:
closes #2057

This PR removes the hard-coded internal defaults for the (default) resource config options for Spilo and pooler pods allowing for not setting resources at all while relying on the K8s environment to handle it.

This means deployments lacking a resource section and a config map with no defaults defined are likely to fail because memory/cpu request (and memory limits) are missing.

The PR also adds a safeguard function `matchLimitsWithRequestsIfSmaller` for situations where limits are smaller than the actual requests, e.g. if no limit is specified in the manifest and the operator falls back on a default value which is still lower than the request.


idanovinda:
👍

FxKu:
:+1:

antverpp:
This is not working. I updated my operator to 1.11.0. I remove setting the cpu limits, but still receive “1” on limits. The same if I set “0” - it still makes “1”

ktalo:
> This is not working. I updated my operator to 1.11.0. I remove setting the cpu limits, but still receive “1” on limits. The same if I set “0” - it still makes “1”

If you have installed using the helm chart, the defaults are still set in https://github.com/zalando/postgres-operator/blob/7c7aa969350bef67d0283d93589499ae3d114edb/charts/postgres-operator/values.yaml#L233 and thus end up in your operator's config. You need to update your operator using a values that unset those (i.e.. set to null).

marcofunstage:
Hi @ktalo,
I set the following configuration, but doesn't work; setting empty value, restarting postgres or the operator pods didn't help either
```
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: postgres-operator
  namespace: postgres-operator
spec:
  interval: 5m
  chart:
    spec:
      chart: postgres-operator
      version: '1.12.0'
      sourceRef:
        kind: HelmRepository
        name: zalando-postgres-operator
        namespace: flux-system
      interval: 60m
  values:
    configKubernetes:
      enable_cross_namespace_secret: true
    configMajorVersionUpgrade:
      major_version_upgrade_mode: "manual"
    configPostgresPodResources:
      default_cpu_limit: "0"
```

ktalo:
If you still have old CRDs present, the defaults are there, too. The CRDs won't get updated with helm upgrade. The default limits are in CRD operatorconfigurations.acid.zalan.do and that needs to be updated too.

PR #2528, opened 2024-01-31
allow users to opt out from globally enabled secret rotation

### LINKED ISSUE #1994: Disable password rotation for single cluster

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.8.2
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare metal RKE
- **Type of issue?** [Bug report, question, feature request, etc.] Question

Hello,

Is it possible to disable password rotation for a specific cluster? My use case is that I have an application ([Drone](https://github.com/harness/drone)) that does not use a specific secret for the database credentials. The credentials are specified directly in the connection URI. For example :
```
 postgresql://user:password@acid-drone:5432/drone?some_options
```
So when the operator will rotate the password this connection URI will not be updated.

Thanks in advance for your help.

FxKu:
closes #1994

Introducing another slice to define users that should be ignored from secret rotation when it's globally enabled.

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -612,6 +612,11 @@ spec:
                     - SUPERUSER
                     - nosuperuser
                     - NOSUPERUSER
+              usersIgnoringSecretRotation:
+                type: array
+                nullable: true
+                items:
+                  type: string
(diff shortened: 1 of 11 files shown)


FxKu:
:+1:

jopadi:
👍

PR #2535, opened 2024-02-09
patch event stream CRD instead of update

FxKu:
If streams were already used before the introduction of recovery in #2421 the operator will detect a diff because it sets recovery to `type: None` when it's not enabled. Using the `Update` endpoint results in the following error:

```
could not sync cluster: could not sync streams: failed updating event stream my-stream: FabricEventStream.zalando.org "my-stream" is invalid: spec.eventStreams[0].recovery.sink: Invalid value: "null": spec.eventStreams[0].recovery.sink in body must be of type object: "null"
```

The operator does not specify the sink when `type: None` but it's part of the [recovery struct](https://github.com/zalando/postgres-operator/blob/29ea863faf115396a3ca3184d3aa813f0a349030/pkg/apis/zalando.org/v1/fabriceventstream.go#L56) and will be `nil` (because of pointer type). The `Update` endpoint is replacing the K8s resource with the stream that has the `nil` field. If we would use the Patch endpoint, however, only the fields with values are applied which is what `updateStreams` was always meant to do.

### DIFF
--- a/pkg/cluster/streams.go
+++ b/pkg/cluster/streams.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"reflect"
 	"sort"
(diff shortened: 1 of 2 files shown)


FxKu:
:+1:

macedigital:
:+1:

PR #2538, opened 2024-02-15
do not block delete because of emtpy child resources

FxKu:
The finalizer will not get removed if there are any errors during the delete. But, it can happen on cluster creation that child resources never got created and, thus, are empty like c.StatefulSet, c.PodDisruptionBudget etc. If the operator considers this as an error it would never allow the removal of finalizers.

Instead, in all the delete function the operator should only log the fact that a child resource is missing and return with nil. This way finalizer removal and manifest deletion can happen.


idanovinda:
👍

FxKu:
:+1:

PR #2543, opened 2024-02-16
bump go version and update dependencies

FxKu:
- updating go from 1.19 to 1.21 (not yet to 1.22)
- updating K8s libs to 1.28. 1.29+ has a breaking change with new `[VolumeResourceRequirement](https://pkg.go.dev/k8s.io/api/core/v1#VolumeResourceRequirements)` setting for PVCs
- updating other dependencies in go.mod
- reflecting a change in go cache library to have extra bool field in PopProcessFunc type


idanovinda:
👍

PR #2546, opened 2024-02-19
update spec when updating status

FxKu:
[Four years ago](https://github.com/zalando/postgres-operator/commit/0c6655a22dffce9c2388e6ca6a69bc29c26e7c6b) I moved the `setStatus` code to the `k8sutil` package in order to update the status also in the `controller` scope and not only inside `cluster`. For this I removed the line `c.setSpec(newspec)` and left it to the caller to update Spec struct.

However, it was not reflected by all callers which leads to the situation that further `Update` calls during DELETE events (like removing the finalizer) can be blocked because
1. the `resourceVersion` is already higher that it what is stored in c.Spec, in case the status changed (e.g. Running -> UpdateFailed).
2. [Delete()](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L1067) func is not calling c.setSpec unlike the [Update()](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L838) and [Create()](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/cluster.go#L268) because without finalizer the Postgresql resource is already gone.

Instead of implementing extra logic to Delete() this PR attempts to bring back the behavior from 4 years ago to update the spec when updating the status.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -254,11 +254,16 @@ func (c *Cluster) Create() (err error) {
 	)
 
 	defer func() {
+		var pgUpdatedStatus *acidv1.Postgresql
 		if err == nil {
-			c.KubeClient.SetPostgresCRDStatus(c.clusterName(), acidv1.ClusterStatusRunning) //TODO: are you sure it's running?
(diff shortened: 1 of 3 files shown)


jopadi:
👍

FxKu:
:+1:

PR #2549, opened 2024-02-21
change username in secret when switching rotation mode

FxKu:
You may want certain users to switch to in-place rotation, but the operator does not set the username when it branches to in-place rotation in the go code. Thus, the username would always stay the same. It needs an extra line `secret.Data["username"] = []byte(secretUsername)`. Another workaround could be to first opt out with the new option introduced with #2528 and add the user to in-place rotation list afterwards.

But, the operator could also be smart enough to detect when the desired username is not a rotation user and allow for immediate rotation. This is what this PR introduces.

It can also work the other way around when switching from in-place rotation to rotation with extra users, like:
```
	if len(currentUsername) != len(expectedUsername) {
		rotationModeChanged = true
```

Unlike before, this would kick in any initial rotation immediately an not after the rotation period. However, it makes the current unit tests impossible. A rotation can produce a list of retention user which the operator tries to drop from the database. Database connections will fail in the unit test. So in the scenario switching from in-place to users rotation the user will need to wait for the next rotation data which is stored in secret - like it before.

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -942,6 +942,8 @@ func (c *Cluster) rotatePasswordInSecret(
 		err                 error
 		nextRotationDate    time.Time
 		nextRotationDateStr string
+		expectedUsername    string
+		rotationModeChanged bool
 		updateSecretMsg     string
 	)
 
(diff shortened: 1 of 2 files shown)


jopadi:
👍

FxKu:
:+1:

PR #2555, opened 2024-02-23
arm-compatible pooler image

FxKu:
New pooler version with arm support. It lives in a new registry which allows for multi-arch images.

This PR also updates the operator debug dockerfile and base image for ui to use the latest alpine image.


Charliefrodriguez:
Is there any time table on when this will be merged in?

pat-s:
I also just noticed that the arm64 variant is missing here and had to apply a manual nodeSelector patch.

Besides, looking at the diff, alpine 3.15 is still used to build the current image? 👀 
Please update all build components (including the ubuntu one?).

micbar:
How can we push this to get merged? Arm support is still not possible with this operator as of today.

FxKu:
will use this PR as a follow up to #3077

FxKu:
turned out everything could be handled in #3077

Issue #2552, opened 2024-02-22
synchronous_node_count is not updated in cluster settings


cstohr1:
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.10.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** AWS EKS
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report

https://github.com/zalando/postgres-operator/pull/1484/ introduced support for `synchronous_node_count` in the CRD, however I dont see setting this to >1 actually being reflected in the cluster. I deployed the latest postgres-operator image and this CR using `ghcr.io/zalando/spilo-15:3.0-p1`:

```
apiVersion: acid.zalan.do/v1
kind: postgresql
metadata:
  name: postgres-cluster
spec:
  allowedSourceRanges:
    - 127.0.0.1/32
  databases:
    accounts: accounts
  dockerImage: ghcr.io/zalando/spilo-15:3.0-p1
  enableConnectionPooler: false
  enableLogicalBackup: false
  enableMasterLoadBalancer: false
  enableMasterPoolerLoadBalancer: false
  enableReplicaConnectionPooler: false
  enableReplicaLoadBalancer: false
  enableReplicaPoolerLoadBalancer: false
  enableShmVolume: true
  initContainers:
    - command:
        - /bin/date
      image: busybox:1.35
      name: date
  numberOfInstances: 3
  patroni:
    initdb:
      data-checksums: 'true'
      encoding: UTF8
      locale: en_US.UTF-8
    loop_wait: 10
    maximum_lag_on_failover: 33554432
    retry_timeout: 10
    synchronous_mode: true
    synchronous_mode_strict: true
    synchronous_node_count: 2
    ttl: 30
  postgresql:
    parameters:
      archive_timeout: '300'
      autovacuum_vacuum_cost_delay: 10ms
      autovacuum_vacuum_cost_limit: '10000'
      log_filename: postgresql-%M
      log_rotation_age: '1'
      log_statement: none
      max_connections: '200'
      shared_buffers: 1024MB
      synchronous_commit: remote_apply
      synchronous_standby_names: '*'
    version: '11'
  resources:
    limits:
      cpu: 3000m
      memory: 4Gi
    requests:
      cpu: 1000m
      memory: 4Gi
  teamId: knime
  users:
    accounts:
      - createdb
  volume:
    size: 30Gi
```

[...]

PR #2558, opened 2024-02-29
Fix updating SynchronousNodeCount (#2552)

cstohr1:
CRD support was added previously, but comparing desired SynchronousNodeCount to the actual patroni config was missing


FxKu:
:+1:

idanovinda:
👍

PR #2574, opened 2024-03-12
return create and sync error, not setStatus error

FxKu:
Yet another follow-up to adding finalizers on existing clusters. If you cancel the kubectl delete command the controller does not find the deletionTimestamp and keeps on syncing the cluster. Only after restarting the operator by removing the pod the timestamp gets picked up.

I wanted to repeat the experiment but the cluster recreation failed, but I could not see why. With the changes made in #2546 I'm no longer returning the actual error on cluster creation but overriding it with the error returned from setting the status. This is wrong and is fixed with this PR.


idanovinda:
👍

PR #2578, opened 2024-03-14
fix: db row close error be ignored

drivebyer:
We should wrap error like:
https://github.com/zalando/postgres-operator/blob/a5663da64f7a94af8dde395852cb2cf309ae27bc/pkg/cluster/database.go#L537-L545


FxKu:
:+1:

hughcapet:
👍

FxKu:
Thanks @drivebyer for your contribution!

PR #2579, opened 2024-03-14
add option to prevent PVC removal on cluster deletion

FxKu:
closes #2567
addresses #1074

A long wished for feature to stop operator from deleting PVCs. To avoid confusion with the new PVC retention policy feature added in #2343 for which I added the missing documentation.

Issue #1736, opened 2022-01-09
Add secretAnnotations to postgresql cluster manifest


soosap:
Currently only `serviceAnnotations` and `podAnnotations` are supported in the **postgresql** cluster manifest. Annotations are used for many things. I am using helm to manage my Kubernetes workloads. I also maintain a custom postgres helm chart tailored to my company needs.

Creating a postgres cluster works like a charm. The problem arises when I delete a cluster. It also wipes out all the associated secrets. One way to protect from `helm uninstall` wiping out the secrets is by providing the following annotation to all created secrets: `"helm.sh/resource-policy": keep`.

Unfortunately, as of today there is no `secretAnnotations` supported. I would like to request that feature as it greatly improves the integration of these crds into a helm-managed Kubernetes cluster.

FxKu:
What's the reason for keeping secrets of a deleted Postgres cluster? For restoring purposes? In that case, the operator would create new secrets and update the passwords in the DB. Maybeeee, there can be a flag if secrets should be removed or not - I'm uncertain about yet another annotations option.

soosap:
Hi Felix, I was actually fighting another problem https://github.com/zalando/postgres-operator/issues/1739 to get point in time recovery to work. I assumed that wiping out secrets and recreating them has some effect on backups being found by the restore script.

Thanks for clarifying the behaviour of secret management via the operator. Since the operator creates new secrets and updates them in the db accordingly, I do no longer have a need to keep my secrets around after deleting a cluster. Thus, being able to set annotations specifically to secrets is no longer necessary.

dmotte:
@FxKu @excavador FYI I'm working on a fix for this, because #2579 doesn't make much sense if secrets are deleted. I will submit my PR shortly

embik:
While this issue took a different turn due to the specific use case, is the original purpose (adding annotations to secrets) fulfilled? It stills seems to be an open issue, maybe it could be reopened? In my case, I would like for secrets to be synced to a different namespace with e.g. https://github.com/emberstack/kubernetes-reflector, but I don't see a way to add annotations to generated credentials secrets (apart from doing it manually).

PR #2582, opened 2024-03-15
Secrets deletion config

dmotte:
This PR introduces a configuration option to provide a way to disable the **deletion** of Kubernetes **secrets**: `enable_secrets_deletion` (default `true`).

Ratio: see https://github.com/zalando/postgres-operator/issues/1736#issuecomment-1999567249

I think many people would love this feature. This also relates to #2579, which otherwise wouldn't make much sense, because attaching already-existing volumes to a brand new cluster doesn't work due to different secrets.

Inspired by #2579 (huge thanks to @FxKu)

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -2110,7 +2110,7 @@ class EndToEndTestCase(unittest.TestCase):
             self.eventuallyEqual(lambda: k8s.count_statefulsets_with_label(cluster_label), 0, "Statefulset not deleted")
             self.eventuallyEqual(lambda: k8s.count_deployments_with_label(cluster_label), 0, "Deployments not deleted")
             self.eventuallyEqual(lambda: k8s.count_pdbs_with_label(cluster_label), 0, "Pod disruption budget not deleted")
-            self.eventuallyEqual(lambda: k8s.count_secrets_with_label(cluster_label), 3, "Secrets were deleted although disabled in config")
+            self.eventuallyEqual(lambda: k8s.count_secrets_with_label(cluster_label), 7, "Secrets were deleted although disabled in config")
             self.eventuallyEqual(lambda: k8s.count_pvcs_with_label(cluster_label), 3, "PVCs were deleted although disabled in config")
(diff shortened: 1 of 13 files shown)


e2e/tests/test_e2e.py
```diff
@@ -2109,7 +2110,7 @@ def test_zz_cluster_deletion(self):
             self.eventuallyEqual(lambda: k8s.count_statefulsets_with_label(cluster_label), 0, "Statefulset not deleted")
             self.eventuallyEqual(lambda: k8s.count_deployments_with_label(cluster_label), 0, "Deployments not deleted")
             self.eventuallyEqual(lambda: k8s.count_pdbs_with_label(cluster_label), 0, "Pod disruption budget not deleted")
-            self.eventuallyEqual(lambda: k8s.count_secrets_with_label(cluster_label), 0, "Secrets not deleted")
+            self.eventuallyEqual(lambda: k8s.count_secrets_with_label(cluster_label), 3, "Secrets were deleted although disabled in config")
```
FxKu, line 2113:
```suggestion
            self.eventuallyEqual(lambda: k8s.count_secrets_with_label(cluster_label), 7, "Secrets were deleted although disabled in config")
```

There are more secrets created during e2e tests.

FxKu:
:+1:

idanovinda:
👍

FxKu:
Thanks @dmotte

Yingrjimsch:
@dmotte as far as I've understood this configuration is per operator. Is there a way to let the postgres cluster decide if it should keep the secret?

dmotte:
@Yingrjimsch I think it can be implemented by adding a config option to the cluster manifest. To be honest I don't need it, but, if you want to give it a try, feel free to add me as a reviewer :)

Yingrjimsch:
@dmotte I've added the functionallity [here](https://github.com/zalando/postgres-operator/pull/2750) and mentioned you in the pull request, it is fairly simple but effective for our use case. If you have any suggestions just let me know :)

PR #2584, opened 2024-03-15
Fix duplicate key issue when using configmap config target

ThoreKr:
When using the `configTarget: "ConfigMap"` option, the rendered configmap fails to install due to 
```
  line 52: mapping key "persistent_volume_claim_retention_policy" already defined at line 51
```

This is caused by the flatten helper adding the config option too early causing duplicate keys like these to be created:

```
  persistent_volume_claim_retention_policy: "when_deleted:retain"
  persistent_volume_claim_retention_policy: "when_deleted:retain,when_scaled:retain"
``` 

This pr fixes this issue.


FxKu:
:+1:

idanovinda:
👍

FxKu:
Thanks @ThoreKr . I will update the helm chart shortly after.

PR #2608, opened 2024-04-14
Apply template on all keys of operatorconfiguration

LordGaav:
This allows one to specify go templates in every setting applied to the OperatorConfiguration. This in turn allows me to do something like this:

```yaml
global:
    s3:
        endpoint: https://s3.example.com
        access_key_id: xxx
        secret_access_key: xxx
        bucket: xxx
postgres-operator:
    configLogicalBackup:
        logical_backup_s3_endpoint: '{{ .Values.global.s3.endpoint }}'
        logical_backup_s3_access_key_id: '{{ .Values.global.s3.access_key_id }}'
        logical_backup_s3_secret_access_key: '{{ .Values.global.s3.secret_access_key }}'
        logical_backup_s3_bucket: '{{ .Values.global.s3.bucket }}'
some-other-chart:
    some-other-key: '{{ .Values.global.s3.bucket }}'
```

I applied `tpl` to all keys, because even though I only required logical backup for my use case, I see no reason to not add this to all keys.

### DIFF
--- a/charts/postgres-operator/templates/operatorconfiguration.yaml
+++ b/charts/postgres-operator/templates/operatorconfiguration.yaml
@@ -14,7 +14,7 @@ configuration:
   users:
 {{ tpl (toYaml .Values.configUsers) . | indent 4 }}
   major_version_upgrade:
-{{ toYaml .Values.configMajorVersionUpgrade | indent 4 }}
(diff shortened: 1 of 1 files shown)


FxKu:
:+1:

idanovinda:
👍

FxKu:
Thanks @LordGaav for your contribution.

PR #2621, opened 2024-04-25
add logical backup retention as manifest option

FxKu:
So far, it's only possible to configure the retention globally, but different clusters might have different requirements. As usual the manifest setting will take precedence over the global config. I changed the env var generation slightly to only add variables needed for a given LogicalBackupProvider. A new unit test is covering the code changed/added with this PR.


idanovinda:
👍

FxKu:
:+1:

Issue #2464, opened 2023-10-30
No option SubPathExpr for additionalVolumes


smutel:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?**

registry.opensource.zalan.do/acid/postgres-operator:v1.10.0

- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** [AWS K8s | GCP ... | Bare Metal K8s]

Kubernetes on private vms

- **Are you running Postgres Operator in production?** [yes | no]

no

- **Type of issue?** [Bug report, question, feature request, etc.]

Feature request

PR #2638, opened 2024-05-24
fix unit test for new subPathExpr feature

FxKu:
With #2464 the unit test were not implemented correctly. This PR fixes it and also introduces the isSubPathExpr flag for the spilo volume in case somebody might need this. I did some alphabetical re-ordering in the CRDs.


FxKu:
:+1:

idanovinda:
👍

Issue #2644, opened 2024-05-30
infrastructure_role_secrets field doesn't work as expected.


hemakshis:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?**  
ghcr.io/zalando/postgres-operator:v1.11.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** 
AWS K8s 
- **Are you running Postgres Operator in production?** 
Currently no, but will soon go GA
- **Type of issue?** [Bug report, question, feature request, etc.] 
bug

I am trying to create two infrastructure roles with different access, Developer (`batman` - read only) and On-call user (`ironman` - superuser). Every one should be able to read the Developer k8s secret fetch the password and use it. And only the person who is On-call can read the k8s secret for that (this we can manage by k8s RBAC).

The issue is since I can only provide only one secret as part of the `infrastructure_roles_secret_name`, I will have to put the other one in `infrastructure_roles_secrets` but whenever I am trying the following configuration -

Also please note we are using CRD approach.

```yaml
# read only secret
apiVersion: v1
data:
  # batman
  user1: YmF0bWFu
  # justice
  password1: anVzdGljZQ== 
  # pg_read_all_data
  inrole1: cGdfcmVhZF9hbGxfZGF0YQ==
kind: Secret
metadata:
  name: postgresql-infrastructure-roles
  namespace: postgres-operator
type: Opaque
---
apiVersion: v1
data:
  # ironman: marvel
  ironman: bWFydmVs
kind: Secret
metadata:
  name: postgresql-infrastructure-roles-oncall
  namespace: postgres-operator
type: Opaque
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: postgresql-infrastructure-roles-oncall
  namespace: postgres-operator
data:
  ironman: |
    inrole: [pg_write_all_data]
    user_flags:
      - login
      - superuser
---
```

In my postgres-operator helm release values.yaml -
```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: postgres-operator
---
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: HelmRepository
metadata:
  name: postgres-operator-charts
  namespace: postgres-operator
spec:
  interval: 30m
  url: https://opensource.zalando.com/postgres-operator/charts/postgres-operator
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: postgres-operator
[...]

PR #2645, opened 2024-05-30
Fix (#2644) - Add json tags to InfrastructureRole struct

hemakshis:
Fixes: #2644

### DIFF
--- a/pkg/util/config/config.go
+++ b/pkg/util/config/config.go
@@ -70,17 +70,17 @@ type Resources struct {
 type InfrastructureRole struct {
 	// Name of a secret which describes the role, and optionally name of a
 	// configmap with an extra information
-	SecretName spec.NamespacedName `json:"secretname,omitempty"`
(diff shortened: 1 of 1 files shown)


FxKu:
Thanks. All the other mappings in this config file use `name:...` instead of `json:...`. Can you test if it also works with name?

FxKu:
I feel we could also more tests and docs on this topic. But for now we can merge as is to include it into today's release.

hemakshis:
It is working after adding `name` tag as well, do you want me to revert it? 😅

hemakshis:
> I feel we could also more tests and docs on this topic. But for now we can merge as is to include it into today's release.

Yes, docs are very confusing and each doc file gives some extra information which is not present in another. It would be good to make it consistent for all the ways we can provide these infra roles - CRD, ConfigMap and then both Only secrets & both Secret+ConfigMap.

FxKu:
:+1:

idanovinda:
👍

PR #2646, opened 2024-05-31
[UI] use only one logger adapter and update Dockerfile

FxKu:
This PR fixes a long standing issue with UI logging stack traces for when logging was called from code. The reason seems to be multiple loggers are in action here - imported from dependencies vs or own getLogger instances. By using a central logger adapter which deactivates existing loggers in its config (when imported after them) we can get rid of these stack traces.

Moreover, the Dockerfile is now using a new base image which makes the setup more straight forward.

Another minor fix removes a redundant info block from the status page.


idanovinda:
👍

idanovinda:
👍

PR #2648, opened 2024-06-01
Comment out options with default values

pschichtel:
these cause a diff in CI tools like argocd, because the default values will not be included in the yaml output by kubernetes, but they will be rendered by helm.

A workaround is to set these values to null, which will use the default value of the controller.

### DIFF
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -130,7 +130,7 @@ configKubernetes:
   # enables initContainers to run actions before Spilo is started
   enable_init_containers: true
   # toggles if operator should delete secrets on cluster deletion
-  enable_secrets_deletion: true
+  # enable_secrets_deletion: true
   # toggles if operator should delete PVCs on cluster deletion
   enable_persistent_volume_claim_deletion: true
   # toggles pod anti affinity on the Postgres pods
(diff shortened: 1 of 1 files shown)


pschichtel:
This is btw a new issue since the 1.12 release.

FxKu:
I don't get it. What is different compared to previous releases? With nearly every release we introduce new config options also to the helm chart. So yeah, there is a diff. Why is it problematic?

pschichtel:
I just upgraded the operator in another cluster with argocd and the same thing happened.

FxKu:
I still don't really understand what that "same thing" is - the diff in CI tools. How does it look like?

I've checked if we missed some default values somewhere, but no, it's implemented like other options e.g. `enable_persistent_volume_claim_deletion`.

pschichtel:
This is the config I had with 1.11.*:

```yaml
operator:
  configKubernetes:
    enable_cross_namespace_secret: true
    inherited_labels:
    - velero.io/exclude-from-backup
  configTeamsApi:
    enable_team_superuser: true
    enable_teams_api: false
```

This is the config I now use with 1.12.* that avoids this issue:

```yaml
operator:
  configKubernetes:
    enable_cross_namespace_secret: true
    enable_secrets_deletion: null
    inherited_labels:
    - velero.io/exclude-from-backup
  configTeamsApi:
    enable_team_superuser: true
    enable_teams_api: false
  configLogicalBackup:
    logical_backup_s3_bucket_prefix: null
```

I'll try reverting the commit tomorrow.

FxKu:
Can it be that you haven't manually updated the CRD so that these new options are unknown? I remember that one had to update CRD manually when using helm.

pschichtel:
That's what I though as well and why I immediately checked the CRD definition in the cluster, which contained the options. Weirdly, after I reverted the values.yaml changes in my cluster the to the 1.11.* version from above, the issue doesn't seem to reappear.

pschichtel:
It seems this was a known diff issue with argo cd, sorry for this.

PR #2655, opened 2024-06-06
Enable build image when push to master

idanovinda:
Currently, our operator only build image when there is tag release. It should also build image when on merge to master.


idanovinda:
👍

Issue #2642, opened 2024-05-29
inherited_annotations / inherited_labels do not update PVCs / PDBs / secrets with labels/annotations if they were created before applying the config


mightymiracleman:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** e.g. ghcr.io/zalando/postgres-operator:v1.11.0
   1.10
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** [AWS K8s | GCP ... | Bare Metal K8s]
  AKS
- **Are you running Postgres Operator in production?** [yes | no]
  No, but soon, we have been running it in our dev envs for quite some time.
- **Type of issue?** [Bug report, question, feature request, etc.]
  question / possible bug?

Some general remarks when posting a bug report:
- Please, check the operator, pod (Patroni) and postgresql logs first. When copy-pasting many log lines please do it in a separate GitHub gist together with your Postgres CRD and configuration manifest.
- If you feel this issue might be more related to the [Spilo](https://github.com/zalando/spilo/issues) docker image or [Patroni](https://github.com/zalando/patroni/issues), consider opening issues in the respective repos.

I have tried this with both `inherited_annotations` and `inherited_labels`, when creating a new postgresql cluster, it works as expected, the labels / annotations are applied to the pvc / pdb / secrets as expected.  However, if there is an existing postgresql cluster (created before the `inherited_annotations` / `inherited_labels` were set) the pvc / pdb / secret does not get updated with the labels /annotations.   The pooler, db and services seem to get updated with the values.

I should also mention, I tried this with annotations and validated that the values weren't being propagated to pvc/pdb/secrets; I also tried with labels but only validated that values weren't set on the pvc in that case (I'm assuming that the other objects were not tagged in the same manner as annotations).

Is this a bug?  

Thank you!

FxKu:
@hughcapet is currently working on exactly this :smile:

mightymiracleman:
thanks!

PR #2657, opened 2024-06-10
Improve inherited annotations

hughcapet:
- Annotate PVC on Sync/Update, not only change PVC template
- Don't rotate pods when only annotations changed
- Annotate Logical Backup's and Pooler's pods
- Annotate PDB, Endpoints created by the Operator, Secrets, Logical Backup jobs

Inherited annotations are only added/updated, not removed

addresses #2642


pkg/cluster/cluster.go
```diff
@@ -513,9 +519,9 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 				reasons = append(reasons, fmt.Sprintf("new statefulset's name for volume %d does not match the current one", i))
 				continue
 			}
-			if !reflect.DeepEqual(c.Statefulset.Spec.VolumeClaimTemplates[i].Annotations, statefulSet.Spec.VolumeClaimTemplates[i].Annotations) {
+			if changed, reason := c.compareAnnotations(c.Statefulset.Spec.VolumeClaimTemplates[i].Annotations, statefulSet.Spec.VolumeClaimTemplates[i].Annotations); changed {
 				needsReplace = true
-				reasons = append(reasons, fmt.Sprintf("new statefulset's annotations for volume %q does not match the current one", name))
+				reasons = append(reasons, fmt.Sprintf("new statefulset's annotations for volume %q does not match the current one: ", name)+reason)
```
FxKu, line 524:
```suggestion
				reasons = append(reasons, fmt.Sprintf("new statefulset's annotations for volume %q does not match the current one: %s", name, reason))
```

pkg/cluster/cluster.go
```diff
@@ -818,6 +826,17 @@ func (c *Cluster) compareLogicalBackupJob(cur, new *batchv1.CronJob) (match bool
 	return true, ""
 }
 
+func (c *Cluster) ComparePodDisruptionBudget(cur, new *apipolicyv1.PodDisruptionBudget) (bool, string) {
```
FxKu, line 829:
```suggestion
func (c *Cluster) comparePodDisruptionBudget(cur, new *apipolicyv1.PodDisruptionBudget) (bool, string) {
```

we don't need to export it for other packages I think

e2e/tests/test_e2e.py
```diff
@@ -909,25 +909,35 @@ def test_ignored_annotations(self):
...
             old_sts_creation_timestamp = sts.metadata.creation_timestamp
             k8s.api.apps_v1.patch_namespaced_stateful_set(sts.metadata.name, sts.metadata.namespace, annotation_patch)
-            svc = k8s.api.core_v1.read_namespaced_service('acid-minimal-cluster', 'default')
             old_svc_creation_timestamp = svc.metadata.creation_timestamp
             k8s.api.core_v1.patch_namespaced_service(svc.metadata.name, svc.metadata.namespace, annotation_patch)
 
             patch_config_ignored_annotations = {
                 "data": {
-                    "ignored_annotations": "k8s-status",
+                    "ignored_annotations": "k8s-status, foo",
```
FxKu, line 940: this other annotation is just to test with two values?

pkg/cluster/sync.go
```diff
@@ -309,7 +317,7 @@ func (c *Cluster) syncPodDisruptionBudget(isUpdate bool) error {
 	return nil
 }
 
-func (c *Cluster) syncStatefulSet() error {
+func (c *Cluster) syncStatefulSet(force bool) error {
```
FxKu, line 320: is there way we move all checks done outside to set the value of `force` inside syncStatefulSet? It's called in so many places and unit test and I find the extra argument kinda ugly.


In general, I like that we call syncStatefulSet without extra conditions and resolve it inside the function.

pkg/cluster/volumes.go
```diff
@@ -44,16 +44,20 @@ func (c *Cluster) syncVolumes() error {
...
 
 		// resize pvc to adjust filesystem size until better K8s support
-		if err = c.syncVolumeClaims(); err != nil {
+		if err = c.syncVolumeClaims(false); err != nil {
 			err = fmt.Errorf("could not sync persistent volume claims: %v", err)
 			return err
 		}
 	} else if c.OpConfig.StorageResizeMode == "pvc" {
-		if err = c.syncVolumeClaims(); err != nil {
+		if err = c.syncVolumeClaims(false); err != nil {
```
FxKu, line 52: actually I wonder if we should not call c.syncVolumneClaims everytime without all the ifs here but do it inside the function. But I see that for ebs and off mode we need some extra pre and post treatment here.

pkg/cluster/volumes.go
```diff
@@ -184,21 +191,53 @@ func (c *Cluster) populateVolumeMetaData() error {
 }
 
 // syncVolumeClaims reads all persistent volume claims and checks that their size matches the one declared in the statefulset.
-func (c *Cluster) syncVolumeClaims() error {
+func (c *Cluster) syncVolumeClaims(noResize bool) error {
```
FxKu, line 194: I tend to call these negative variable ignoreSomething. I my head it's easier to process the following code, e.g. I need a few seconds more on !noResize

e2e/tests/test_e2e.py
hughcapet: oh. i didn't check that pod is deleted unconditionally in the `update_config()`. so no need to introduce changes, I guess

idanovinda (approved):
LGTM, had a few questions and confirmed them with Polina 👍

pkg/cluster/volumes.go
```diff
@@ -187,18 +181,59 @@ func (c *Cluster) populateVolumeMetaData() error {
...
 	c.setProcessName("syncing volume claims")
 
-	needsResizing, err := c.volumeClaimsNeedResizing(c.Spec.Volume)
+	ignoreResize := false
+
+	if c.OpConfig.StorageResizeMode == "off" || c.OpConfig.StorageResizeMode == "ebs" {
+		ignoreResize = true
+		if c.OpConfig.StorageResizeMode == "off" {
+			c.logger.Infof("Storage resize is disabled (storage_resize_mode is off). Skipping volume size sync.")
+		}
```
FxKu, line 190:
```suggestion
		c.logger.Debugf("Storage resize mode is set to %q. Skipping volume size sync of PVCs.", c.OpConfig.StorageResizeMode)
		
```

pkg/cluster/sync.go
```diff
@@ -1379,6 +1413,16 @@ func (c *Cluster) syncLogicalBackupJob() error {
...
+		if changed, _ := c.compareAnnotations(job.Annotations, desiredJob.Annotations); changed {
+			patchData, err := metaAnnotationsPatch(desiredJob.Annotations)
+			if err != nil {
+				return fmt.Errorf("could not form patch for the logical backup job %q: %v", jobName, err)
+			}
+			_, err = c.KubeClient.CronJobs(c.Namespace).Patch(context.TODO(), jobName, types.MergePatchType, []byte(patchData), metav1.PatchOptions{})
+			if err != nil {
+				return fmt.Errorf("could not patch annotations of the logical backup job %q: %v", jobName, err)
+			}
+		}
```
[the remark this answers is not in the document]
hughcapet: there it is done for pods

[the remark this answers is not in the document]
FxKu: got it

FxKu (approved):

FxKu:
:+1:

hughcapet:
👍

PR #2677, opened 2024-06-27
Add support for EBS CSI Driver

### LINKED ISSUE #1946: Helm chart storage_resize_mode mixed does not work
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.8.2
- **Type of issue?** question

Hi! I wanted to tune gp3 persistent volume iops and throughput.

Documentation at https://postgres-operator.readthedocs.io/en/latest/reference/operator_parameters/#kubernetes-resources
states that there is:
storage_resize_mode parameter - mixed :operator uses AWS API to adjust size, throughput, and IOPS, and calls pvc change for file system resize Default is "pvc".

When i add this parameter in postgres-operator 1.8.2 helm chart values i get error that such parameter is unsupported. 

Checked 1.8.2 helm chart crd manifest and such option "mixed" is missing - https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/crds/operatorconfigurations.yaml#L320

Is iops/throughput tuning feature supported or not?

cosimomeli:
Add support for the recent EBS CSI Driver using the `CSI` field in the PVC to set the provisioner and the volume-id.
I'm keeping the retro compatibility with the old field.


.gitignore
```diff
@@ -104,3 +104,5 @@ e2e/tls
 mocks
 
 ui/.npm/
+
+.DS_Store
```
FxKu, line 108: what's this change about?

cosimomeli: To avoid committing the annoying MacOS Finder metadata files

FxKu:
I see that you created unti test into their own file and not in volume_test.go. This is fine for me. That file is already big enough. But can you check why some of the tests are failing now with `Unexpected call to *mocks.MockVolumeResizer.GetProviderVolumeID ...`?

pkg/cluster/volumes.go
```diff
@@ -151,7 +151,7 @@ func (c *Cluster) populateVolumeMetaData() error {
 	volumeIds := []string{}
 	var volumeID string
 	for _, pv := range pvs {
-		volumeID, err = c.VolumeResizer.ExtractVolumeID(pv.Spec.AWSElasticBlockStore.VolumeID)
+		volumeID, err = c.VolumeResizer.GetProviderVolumeID(pv)
```
FxKu, line 154: calling GetProviderVolumeID is fine with we since it will ultimately also call ExtractVolumeID. However, for some unit test you have to add the mocked reponses of the resizer. Before [these lines](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/volumes_test.go#L216) you would need to `resizer.EXPECT().GetProviderVolumeID(gomock.Eq(...`

cosimomeli: I just pushed a fix to this. Let me know if it's ok. I'm pretty new to Go and I have been away from typed languages for a long time, so I could miss some tricks and practices.

cosimomeli:
The e2e tests passed locally. Is there a known flakiness?

FxKu:
:+1:

idanovinda:
👍

PR #2686, opened 2024-07-04
improve logical backup comparison unit test and improve container sync

FxKu:
The current operator has a small bug when syncing logical backup cronjobs because it considers containers with empty and nil volumeMounts to be different. We should probably check for other places where we have this false assumption of how DeepEqual works. But for now, this PR will only improve upon syncing additionalVolumes for containers.


FxKu:
:+1:

FxKu:
:+1:

hughcapet:
👍

PR #2691, opened 2024-07-12
Add support for ephemeral PostgresQL instances

### LINKED ISSUE #618: Allow creation of ephemeral clusters
For running tests and for local development, it would be useful if it were possible to run the operator with ephemeral storage, such as emptyDir.

I would swear that this was possible with earlier versions of the operator, but I can't find any way to do so now.

nicholascioli:
This PR adds support for using ephemeral volumes as a data backing for PostgresQL instances.

Fixes #618


PavelSorokin:
+1

PR #2684, opened 2024-07-02
Enable slot and publication deletion when stream application is removed

idanovinda:
Follow up on https://github.com/zalando/postgres-operator/pull/2563

The current behaviour of our operator for the stream section allows us to create FES resources. During the creation, the postgres-operator will create FES resources, publications, and replication slots. However, during deletion (when user drop the stream section from the manifest), operator only removes the FES resources, while publications and replications are not deleted. This results in need for manual deletion by someone who has administrator roles.

Changes:
* refactor `syncPublication()` to allow sync publication (CRUD) per database 
* delete replication slot based on the publication it is deleted
* update `createOrUpdateStream()` to allow resource deletion when removed from manifest
* add minimal FES CRD to enable FES resources creation for E2E test
* adding E2E test - create and delete FES resources

PR #2704, opened 2024-07-29
stream: slot and FES should not be created if the publication creation fails

idanovinda:
This pull request addresses two related issues regarding the stream section during publication and slot creation:

1. Conditional FES Resource Creation:
    The FES (Fabric Event Stream) resource will no longer be created if the logical slot does not exist. Previously, there was a bug causing the FES resource to be created regardless of the existence of the logical slot. This was not the expected behavior because creating the FES resource would trigger the FES operator to create publications and slots if they were missing. If publication and slot creation were not possible (e.g., due to missing tables), it would result in failures.

2. Logical Slot Creation/Deletion Based on Publication Modification:
    The logical slot will now only be created or deleted if the publication modification is successful. Following the changes in #2684, a logical slot was being created regardless of whether the publication creation was successful. This change ensures that logical slot creation is consistent with the success of publication modification, preventing potential issues related to dangling slots that accumulate WAL (Write-Ahead Logging) entries, which could result in risk to the cluster.

### DIFF
--- a/pkg/cluster/streams_test.go
+++ b/pkg/cluster/streams_test.go
@@ -41,6 +41,10 @@ var (
 	fesUser     string = fmt.Sprintf("%s%s", constants.EventStreamSourceSlotPrefix, constants.UserRoleNameSuffix)
 	slotName    string = fmt.Sprintf("%s_%s_%s", constants.EventStreamSourceSlotPrefix, dbName, strings.Replace(appId, "-", "_", -1))
 
+	fakeCreatedSlots map[string]map[string]string = map[string]map[string]string{
+		slotName: {},
+	}
+
 	pg = acidv1.Postgresql{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Postgresql",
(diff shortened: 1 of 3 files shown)


pkg/cluster/streams_test.go
```diff
@@ -41,6 +41,10 @@ var (
 	fesUser     string = fmt.Sprintf("%s%s", constants.EventStreamSourceSlotPrefix, constants.UserRoleNameSuffix)
 	slotName    string = fmt.Sprintf("%s_%s_%s", constants.EventStreamSourceSlotPrefix, dbName, strings.Replace(appId, "-", "_", -1))
 
+	fakeCreatedSlots map[string]map[string]string = map[string]map[string]string{
+		slotName: {},
```
FxKu, line 45:
```suggestion
		slotName: getSlotName(dbName, appId),
```

FxKu:
:+1:

idanovinda:
👍

PR #2709, opened 2024-07-31
Fix synchronous_mode/synchronous_mode_strict ignored when set to false

drivebyer:
In my use case, I import the package from https://github.com/zalando/postgres-operator/tree/master/pkg/apis/acid.zalan.do/v1 into my Golang project. In my project, I use Golang to populate the PostgreSQL resource, and then use client-go to create the resource. However, when updating the PostgreSQL resource, like setting spec.patroni.synchronousMode to false, it gets ignored after marshaling to a string because of the omitempty tag. In my opinion, these fields should not have the omitempty tag.


drivebyer:
PTAL @FxKu

FxKu:
hm - this is contrary to #1223. Always thought omitempty only omits the field when it's not specified, but not on false. But anyway, why is it important to list, when the mode is disabled. With a simple coalesce logic you should be able to check for nil and insert whatever you want.

PR #2710, opened 2024-07-31
Implement per-cluster maintenance window for Postgres automatic upgrade

idanovinda:
This feature introduces the ability for users to set a maintenance window for major PostgreSQL version upgrades when automatic upgrade is enabled. This ensures that major upgrades occur during specified times, minimising disruption to service, and improving management of critical database updates.

### DIFF
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -3,6 +3,7 @@ package cluster
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/zalando/postgres-operator/pkg/spec"
 	"github.com/zalando/postgres-operator/pkg/util"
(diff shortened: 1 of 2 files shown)


pkg/cluster/majorversionupgrade.go
```diff
@@ -55,6 +56,27 @@ func (c *Cluster) isUpgradeAllowedForTeam(owningTeam string) bool {
 	return util.SliceContains(allowedTeams, owningTeam)
 }
 
+func (c *Cluster) isInMainternanceWindow() bool {
```
hughcapet, line 59: can we also have unit test for this function? 🥺  (e.g. ,the input scenarios you've tested manually)

hughcapet (approved):

FxKu:
:+1:

idanovinda:
👍

PR #2714, opened 2024-08-02
Improving Postgres CR Status with Additional Details

### LINKED ISSUE #2669: Improving Postgres CR Status with Additional Details
- **Which image of the operator are you using?** e.g. ghcr.io/zalando/postgres-operator:v1.12.2
- **Type of issue?**   feature request

Hi,
Currently the status field of the `postgresql` CR looks something like:

```
status:
  PostgresClusterStatus: Running
```

I'd like to propose following changes to the status:
- **Number Of Instances** can be added to the status to indicate the number of running pods associated with the Postgres CR 
- **Label Selector**  can be included in the status to define the label criteria applied to pods linked with the Postgres CR and this is also required for scale subresource.
- **Observed generation:** It would be useful in identifying if the operator has made any changes to the CR after the CR spec has been updated.
- **LastTransitionTime**  can also be added to to provide insights into the state and lifecycle of resources.

post these changes, the status field of the `postgresql` CR would look something like this:
```
status:
  labelSelector: cluster-name=postgres-cluster-pvc-5
  numberOfInstances: 3
  conditions:
  - lastTransitionTime: "2024-06-03T10:18:04Z"
    status: "True"
    type: postgresClusterStatusRunning
  observedGeneration: 2
```
Along with this, I am proposing to add scale subresource in the CRD of `postgresql` as well, which would use some of the fields proposed above, and then this scale subresource can be further used by external systems to scale or monitor the CR. This scale subresource would look something like this:
```
 subresources:
   status: {}
   scale:
       specReplicasPath: .spec.numberOfInstances  
       statusReplicasPath: .status.numberOfInstances  
       labelSelectorPath: .status.labelSelector
```

RavinaChidambaram:
Fixes https://github.com/zalando/postgres-operator/issues/2669

Following things have been added:
- Added `observedGeneration`
- Added `conditions` of type `ReconciliationSuccessful` and `Ready`
- Added `numberOfInstances` and `labelSelector` for scale subresource
- Added scale subresource configuration in the CRD 
- `postgresClusterStatus` is still part of the CRD

Detailed example flow:
Conditions are updated everytime there is a change in `PostgresClusterStatus`. And if the cluster is running successfully, the status looks like:
```
status:
  conditions:
  - lastTransitionTime: "2024-08-02T12:33:34Z"
    status: "True"
    type: ReconciliationSuccessful
  - lastTransitionTime: "2024-08-02T12:33:34Z"
    status: "True"
    type: Ready
  labelSelector: cluster-name=acid-minimal-cluster
  numberOfInstances: 3
  observedGeneration: 4
  postgresClusterStatus: Running
```
If there is a transition from Running state to a failed state, type `ReconciliationSuccessful` will be false. 
When a failed case (Create failed,Update failed, Sync failed)  is encountered, the error message will be printed in the `message` in `conditions`
```
status:
  conditions:
  - lastTransitionTime: "2024-08-02T13:05:38Z"
    message: 'could not create master endpoint: could not create master endpoint:
      endpoints "acid-minimal-cluster-test-1" already exists'
    reason: CreateFailed
    status: "False"
    type: ReconciliationSuccessful
  - lastTransitionTime: "2024-08-02T13:05:38Z"
    status: "False"
    type: Ready
  labelSelector: cluster-name=acid-minimal-cluster
  numberOfInstances: 0
  postgresClusterStatus: CreateFailed
```


RavinaChidambaram:
Hi @FxKu , Is there any update on this PR?

FxKu:
I had to concentrate on the 1.13.0 release. But we will definitely include it in the next release in autumn. This is a great addition. I have to focus on some internal projects for the next weeks, but still, I hope to find some time in between to review.

FxKu:
@RavinaChidambaram can you fix the failing unit test so e2e pipeline can run

pkg/apis/acid.zalan.do/v1/crds.go
```diff
@@ -25,6 +25,12 @@ const (
 	OperatorConfigCRDResourceShort  = "opconfig"
 )
 
+var (
```
FxKu, line 28: are these not `const`

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -225,9 +226,48 @@ type Sidecar struct {
...
+	equality.Semantic.AddFunc(
+		// Always treat VolatileTime fields as equivalent.
+		func(VolatileTime, VolatileTime) bool {
+			return true
+		},
+	)
+}
+
+// Condition contains the conditions of the PostgreSQL cluster
+type Condition struct {
```
FxKu, line 256: Can this be changed to PostgresqlCondition? This aligns with other K8s resources.

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -225,9 +226,48 @@ type Sidecar struct {
 // UserFlags defines flags (such as superuser, nologin) that could be assigned to individual users
 type UserFlags []string
 
+type Conditions []Condition
+
+type ConditionType string
```
FxKu, line 231:
```suggestion
type PostgresqlConditionType string
```

pkg/util/k8sutil/k8sutil.go
```diff
@@ -192,10 +193,16 @@ func NewFromConfig(cfg *rest.Config) (KubernetesClient, error) {
 }
 
 // SetPostgresCRDStatus of Postgres cluster
-func (client *KubernetesClient) SetPostgresCRDStatus(clusterName spec.NamespacedName, status string) (*apiacidv1.Postgresql, error) {
+func (client *KubernetesClient) SetPostgresCRDStatus(clusterName spec.NamespacedName, status string, numberOfInstances int32, labelSelector string, observedGeneration int64, existingConditions apiacidv1.Conditions, message string) (*apiacidv1.Postgresql, error) {
```
FxKu, line 196: Instead of extending the list of arguments we could also pass a PostgresStatus struct, right?

FxKu:
As half of my comments target the scale subresource, I would suggest to remove it from here and create a separate PR for it. Then you could concentrate only on the status and we get it merged quicker.

Have you tried to run code-generation. I think, the generated code has to get updated with the new status structure.

pkg/apis/acid.zalan.do/v1/crds.go
RavinaChidambaram: Changed to `const`

pkg/apis/acid.zalan.do/v1/postgresql_type.go
RavinaChidambaram: Yeah, changed to `PostgresqlCondition`

pkg/apis/acid.zalan.do/v1/postgresql_type.go
```diff
@@ -225,9 +226,48 @@ type Sidecar struct {
...
+			return true
+		},
+	)
+}
+
+// Condition contains the conditions of the PostgreSQL cluster
+type Condition struct {
+	Type               ConditionType      `json:"type" description:"type of status condition"`
+	Status             v1.ConditionStatus `json:"status" description:"status of the condition, one of True, False, Unknown"`
+	LastTransitionTime VolatileTime       `json:"lastTransitionTime,omitempty" description:"last time the condition transit from one status to another"`
```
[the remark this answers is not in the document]
RavinaChidambaram: Mostly to avoid semantic check inequality with `metav1.time`, with a wrapper around `metav1.time` there won't be a semantic inequality even if the times are different but everything else in the CR is same, this could be handy while writing unit tests or e2e tests I think

pkg/cluster/cluster.go
```diff
@@ -255,16 +255,20 @@ func (c *Cluster) Create() (err error) {
 		ss             *appsv1.StatefulSet
 	)
 
+	//Even though its possible to propogate other CR labels to the pods, picking the default label here since its propogated to all the pods by default. But this means that in order for the scale subresource to work properly, user must set the "cluster-name" key in their CRs with value matching the CR name.
+	labelstring := fmt.Sprintf("%s=%s", "cluster-name", c.Postgresql.ObjectMeta.Labels["cluster-name"]) //TODO: make this configurable.
```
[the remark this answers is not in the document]
RavinaChidambaram: Have changed it, using `ClusterNameLabel`

pkg/util/k8sutil/k8sutil.go
RavinaChidambaram: Yeah, made modifications to use the PostgresStatus struct

RavinaChidambaram:
I have made changes to fix the unit tests and successfully ran code-generation, unit tests and e2e test downstream.
Kindly review the changes and if you still think it's necessary to create a new branch for scale subresource addition, I will proceed with it.

RavinaChidambaram:
Hi @FxKu , any updates?

FxKu:
@RavinaChidambaram sorry for not coming back sooner. I see that you did not reflect all my comments. I also do wonder if we have to change the code logic and move the SetPostgresCRDStatus into another package so that the signature can remain the same `(clusterName spec.NamespacedName, status string)` and we will the other fields inside this function? Because, right now we have so many code lines changed and as far as I can see:
- NumberOfInstances
- LabelSelector
- ObservedGeneration
- Conditions

would always be constructed in the same way. WDYT?

pkg/apis/acid.zalan.do/v1/postgresql_type.go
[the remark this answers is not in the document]
RavinaChidambaram: @FxKu , Is this okay?

RavinaChidambaram:
I’ve made all the changes that you mentioned, except for the metav1.time.
Regarding keeping the `SetPostgresCRDStatus` signature unchanged, I agree - we can move the construction of fields like `NumberOfInstances`, `LabelSelector`, `ObservedGeneration`, and `Conditions` into the function itself.

RavinaChidambaram:
Hi @FxKu , 
As suggested, the construction of the status fields has been moved into the function itself.
But this required adding two new parameters to the function `SetPostgresCRDStatus` - `clusterNameLabel` and `message`.

FxKu:
superceded by #3006

PR #2723, opened 2024-08-12
extend inherited annotations unit test to include logical backup cron job

FxKu:
Required one change in the update logic because before the operator only synced on schedule changes during UPDATE events. Now it syncs always when logical backups are enabled.


hughcapet:
👍

FxKu:
:+1:

PR #2727, opened 2024-08-13
Implement major upgrade result annotations

idanovinda:
In order to prevent Operator endlessly retrying on failing major version upgrade and also to be able to easily identify and be alerted on unsuccessful automated attempts, we need to implement a set of "Failed upgrade" annotations assigned to postgresql resource manifest.

Postgres-operator should annotate postgresql resource with one of the following values based on the return code of the upgrade script:
```
last-major-upgrade-success: <timestamp>
last-major-upgrade-failure: <timestamp>
```

Operator should:
* not execute the upgrade script if `running PG version < manifest PG version` but `last-major-upgrade-failure` is set
* remove the failure annotation if the `running PG version >= manifest PG version`
* set `last-major-upgrade-failure` when upgrade is failed
* set `last-major-upgrade-success` when upgrade is success

closes #2685


pkg/cluster/majorversionupgrade.go
```diff
@@ -54,6 +58,58 @@ func (c *Cluster) isUpgradeAllowedForTeam(owningTeam string) bool {
...
 
+func (c *Cluster) PatchAnnotations(patchBytesAnnotation []byte) error {
+	resourceName := c.Name
+	_, err := c.KubeClient.Postgresqls(c.Namespace).Patch(context.Background(), resourceName, types.MergePatchType, patchBytesAnnotation, metav1.PatchOptions{})
+	if err != nil {
+		c.logger.Errorf("failed to patch annotations: %v", err)
+		return err
+	}
+	return nil
+}
```
FxKu, line 69: why not use metaAnnotationsPatch function?

pkg/cluster/majorversionupgrade.go
```diff
@@ -54,6 +58,58 @@ func (c *Cluster) isUpgradeAllowedForTeam(owningTeam string) bool {
...
+
+func (c *Cluster) annotatePostgresResource(isSuccess bool) error {
+	annotations := make(map[string]string)
+	currentTime := metav1.Now().Format("2006-01-02T15:04:05Z")
+	if isSuccess {
+		annotations["last-major-upgrade-success"] = currentTime
+	} else {
+		annotations["last-major-upgrade-failure"] = currentTime
+	}
```
FxKu, line 78: lets define constants for these annotation keys

pkg/cluster/majorversionupgrade.go
```diff
@@ -54,6 +58,58 @@ func (c *Cluster) isUpgradeAllowedForTeam(owningTeam string) bool {
...
+		c.logger.Errorf("failed to marshal annotations: %v", err)
+		return err
+	}
+
+	err = c.PatchAnnotations(patchBytes)
+	if err != nil {
+		c.logger.Errorf("failed to patch annotations to Postgres resource: %v", err)
+		return err
+	}
+	return nil
```
FxKu, line 96: so whenever there was a fail the `last-major-upgrade-failure` annotation will be present in the manifest and not touched again until it fails another time. Do we want to have it like this, or should it be set to nil?

pkg/cluster/majorversionupgrade.go
```diff
@@ -67,12 +123,23 @@ func (c *Cluster) majorVersionUpgrade() error {
...
 	if c.currentMajorVersion >= desiredVersion {
+		if lastMajorUpgradeAnnotations["last-major-upgrade-success"] == "" {
+			c.annotatePostgresResource(isUpgradeSuccess)
+			c.logger.Info("update last major upgrade success annotation to current timestamp as it was not set")
+		}
 		c.logger.Infof("cluster version up to date. current: %d, min desired: %d", c.currentMajorVersion, desiredVersion)
 		return nil
 	}
 
+	if lastMajorUpgradeAnnotations["last-major-upgrade-failure"] != "" {
```
FxKu, line 138: following up my comment above, if the failure annotations stays in the manifest, we have to compare the time with the last success. If failure is older than success major version we can proceed and should not exit here.

pkg/cluster/majorversionupgrade.go
```diff
@@ -54,6 +58,58 @@ func (c *Cluster) isUpgradeAllowedForTeam(owningTeam string) bool {
...
+	for _, annotationKey := range upgradeAnnotationKeys {
+		if val, exists := annotations[annotationKey]; exists {
+			upgradeAnnotationValues[annotationKey] = val
+		} else {
+			upgradeAnnotationValues[annotationKey] = ""
+		}
+	}
+	return upgradeAnnotationValues
+}
+
```
FxKu, line 112: do we really need this function? Can someone not just do `if _, exists := c.Spec.ObjectMeta.Annotations[annotationKey]; if exists` wherever this needs to be checked?

e2e/tests/test_e2e.py
```diff
@@ -1209,68 +1215,108 @@ def check_version():
...
-            "acid.zalan.do", "v1", "default", "postgresqls", "acid-upgrade-test", pg_patch_version_15)
+            "acid.zalan.do", "v1", "default", "postgresqls", "acid-upgrade-test", pg_patch_version_14)
         self.eventuallyEqual(lambda: k8s.get_operator_state(), {"0": "idle"}, "Operator does not get in sync")
 
-        # should have finish failover
         k8s.wait_for_pod_failover(master_nodes, 'spilo-role=master,' + cluster_label)
         k8s.wait_for_pod_start('spilo-role=master,' + cluster_label)
         k8s.wait_for_pod_start('spilo-role=replica,' + cluster_label)
-        self.eventuallyEqual(check_version, 14, "Version should not be upgraded")
+        self.eventuallyEqual(check_version, 13, "Version should not be upgraded")
```
hughcapet, line 1258: let's also check failure annotation is not set in this case

e2e/tests/test_e2e.py
```diff
@@ -1209,68 +1215,108 @@ def check_version():
...
+        k8s.wait_for_pod_start('spilo-role=master,' + cluster_label)
+        k8s.wait_for_pod_start('spilo-role=replica,' + cluster_label)
+        self.eventuallyEqual(check_version, 15, "Version should be upgraded from 13 to 15")
+
+        # check if annotation for last upgrade's success is updated after second upgrade
+        second_annotations = get_annotations()
+        self.assertIsNotNone(second_annotations.get("last-major-upgrade-success"), "Annotation for last upgrade's success is not set")
+        self.assertNotEqual(annotations.get("last-major-upgrade-success"), second_annotations.get("last-major-upgrade-success"), "Annotation for last upgrade's success is not updated")
+
+        # test annotation with failed upgrade annotation
```
hughcapet, line 1287:
```suggestion
        # test update with failed upgrade annotation
```

hughcapet (approved):

FxKu (approved):

Issue #2716, opened 2024-08-03
Logical backup secret results in duplicate environment keys


nickmansrob:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** Helm: postgres-operator:v1.12.2
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K3s
- **Are you running Postgres Operator in production?** almost
- **Type of issue?** [Bug report, question, feature request, etc.] Bug report

When generating the cronjob for the logical backup when using Azure, the account key gets fetched form the `OperatorConfiguration`, even when it's empty. This happens [here](https://github.com/zalando/postgres-operator/blob/7c7aa969350bef67d0283d93589499ae3d114edb/pkg/cluster/k8sres.go#L2492-L2495). I pass it through a SOPS encrypted secret, and this results in a duplicate environment entry in the cronjob with following kubernetes warning:
```
Warning: spec.template.spec.containers[0].env[16]: hides previous definition of "LOGICAL_BACKUP_AZURE_STORAGE_ACCOUNT_KEY"
```
Everything still works so this is a low priority bug.

A check should be added for Azure and GCP in the same way this is done for AWS.

I can fix this in a PR if needed and if I find the time.

FxKu:
A PR to fix it is more than welcome :+1:

FxKu:
Fix has been merged

PR #2732, opened 2024-08-19
fix: add secret only when not in secret file

nickmansrob:
This PR fixes issue #2716 by adding logic to make sure a ENV variable which is specified in a Secret is not added as an empty ENV variable too, resulting in a duplicate key (one from the Secret, one as the default value in OperatorConfiguration which is "").


pkg/cluster/k8sres.go
```diff
@@ -2481,7 +2481,9 @@ func (c *Cluster) generateLogicalBackupPodEnvVars() []v1.EnvVar {
 		}
 
 	case "gcs":
-		envVars = append(envVars, v1.EnvVar{Name: "LOGICAL_BACKUP_GOOGLE_APPLICATION_CREDENTIALS", Value: c.OpConfig.LogicalBackup.LogicalBackupGoogleApplicationCredentials})
+        if c.OpConfig.LogicalBackup.LogicalBackupGoogleApplicationCredentials != "" {
+          envVars = append(envVars, v1.EnvVar{Name: "LOGICAL_BACKUP_GOOGLE_APPLICATION_CREDENTIALS", Value: c.OpConfig.LogicalBackup.LogicalBackupGoogleApplicationCredentials})
+        }
```
FxKu, line 2486:
```suggestion
		if c.OpConfig.LogicalBackup.LogicalBackupGoogleApplicationCredentials != "" {
			envVars = append(envVars, v1.EnvVar{Name: "LOGICAL_BACKUP_GOOGLE_APPLICATION_CREDENTIALS", Value: c.OpConfig.LogicalBackup.LogicalBackupGoogleApplicationCredentials})
		}
```

pkg/cluster/k8sres.go
```diff
@@ -2492,11 +2494,11 @@ func (c *Cluster) generateLogicalBackupPodEnvVars() []v1.EnvVar {
...
 				Value: c.OpConfig.LogicalBackup.LogicalBackupAzureStorageContainer,
-			},
-			{
-				Name:  "LOGICAL_BACKUP_AZURE_STORAGE_ACCOUNT_KEY",
-				Value: c.OpConfig.LogicalBackup.LogicalBackupAzureStorageAccountKey,
 			}}...)
+
+		if c.OpConfig.LogicalBackup.LogicalBackupAzureStorageAccountKey != "" {
+			envVars = append(envVars, v1.EnvVar{Name: "LOGICAL_BACKUP_AZURE_STORAGE_ACCOUNT_KEY", Value: c.OpConfig.LogicalBackup.LogicalBackupAzureStorageAccountKey})
+		}
```
FxKu, line 2501:
```suggestion
		if c.OpConfig.LogicalBackup.LogicalBackupAzureStorageAccountKey != "" {
			envVars = append(envVars, v1.EnvVar{Name: "LOGICAL_BACKUP_AZURE_STORAGE_ACCOUNT_KEY", Value: c.OpConfig.LogicalBackup.LogicalBackupAzureStorageAccountKey})
		}
```

FxKu:
:+1:

pkg/cluster/k8sres.go
nickmansrob: Seems my formatter did not catch that one :thinking:

PR #2713, opened 2024-08-02
sync all resources to cluster fields

FxKu:
While developing owner references support I realized once again that we store many child resources as cluster objects so we can access them even after the manifest is gone. But not for all resources, which might cause issues when deleting them because c.Namespace field might be empty:

```
time="2024-08-01T12:32:33Z" level=debug msg="removing leftover Patroni objects (endpoints / services and configmaps)" cluster-name=default/acid-clone-cluster pkg=cluster worker=2
time="2024-08-01T12:32:33Z" level=warning msg="could not fetch service \"/\": an empty namespace may not be set when a resource name is provided" cluster-name=default/acid-clone-cluster pkg=cluster worker=2
```

Fixing it for missing resources such as logical backup cron job and streams is easy. But it gets a bit tricky with objects created by Patroni, namely config service/endpoint and optionally configmaps. This PR introduces extra sync treatment for these, but only mapping to cluster fields. No diffs, no updates. This should allow us to loop over them on deletion.

Some more minor changes:
- we don't need to init the clusters kubeResources with nil
- bugfix: patched secrets and endpoints are now synced in cluster fields
- Patroni endpoints are treated separately to master/replica endpoints (only config service reused existing code, hence the extra PostgresRole)
- have only one function to determine the service/endpoint name
- small refactoring of code when services/endpoints are already deleted
- deleteSecret does not require passing the secret, we fetch it from cluster.Secrets map via uid
- if second parameter from a range loop is not needed we don't need to write an underscore (idx, _ := range -> idx := range)

The streams feature saw some bigger refactoring:
- createOrUpdateStreams was turned into syncStream for a single stream CRD which is more in line with the rest of the code
- loop over appIds is now done in syncStreams as well as the deletion of streams that got removed from the manifest (and, thus, missing in appIds list)
- renamed gatherApplicationIds function to more suitable name getDistinctApplicationIds
- deleteStream does not require passing stream, we fetch it from cluster.Streams map via appId
- the recently added check for existing slots (#2704) to allow FES creation is now comparing desired databaseSlots with actual slotsToSync. The previous comparison was not complete: appId loop with index lookup on c.Spec.Streams, which can be bigger, also with a different order.

fixes #2719

PR #2733, opened 2024-08-19
fix stream duplication on operator restart

FxKu:
With the refactoring in #2713 I removed one important part of syncing the streams: Fetching the current state from K8s API. Instead I looped over the new cluster struct field Streams which is always empty on start up. This will produce a new stream resource on ever operator pod start. Therefore, I added a delete logic to cleanup the ill-created duplicates.

The unit tests were extended to test syncing and removal as well as owner references update.


FxKu:
:+1:

PR #2735, opened 2024-08-20
Fix resource constraints

dmotte:
The documentation about `postgres_pod_resources` says that you can disable the defaults by setting them to zero **OR empty string**:

> Empty string or `0` disables the default.

Source: https://github.com/zalando/postgres-operator/blob/2582b934bfb5bfaba9e5d7116ef270e507c5d6a2/docs/reference/operator_parameters.md#L570-L604

But it's not currently possible to set them to empty string, due to the **regular expressions** in `charts/postgres-operator/crds/operatorconfigurations.yaml`.

Also, the `enforceMinResourceLimits` function currently supports only the empty string case, which is impossible to use due to the aforementioned regexes.

This PR aims to fix these two issues 🙂

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -157,7 +157,11 @@ func (c *Cluster) enforceMinResourceLimits(resources *v1.ResourceRequirements) e
 	// setting limits too low can cause unnecessary evictions / OOM kills
 	cpuLimit := resources.Limits[v1.ResourceCPU]
 	minCPULimit := c.OpConfig.MinCPULimit
-	if minCPULimit != "" {
+	minCPU, err := resource.ParseQuantity(minCPULimit)
+	if err != nil {
+		return fmt.Errorf("could not parse min CPU limit %s: %v", minCPULimit, err)
+	}
+	if !minCPU.IsZero() {
 		isSmaller, err = util.IsSmallerQuantity(cpuLimit.String(), minCPULimit)
 		if err != nil {
 			return fmt.Errorf("could not compare defined CPU limit %s for %q container with configured minimum value %s: %v",
@@ -168,13 +172,17 @@ func (c *Cluster) enforceMinResourceLimits(resources *v1.ResourceRequirements) e
 				cpuLimit.String(), constants.PostgresContainerName, minCPULimit)
 			c.logger.Warningf(msg)
 			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeWarning, "ResourceLimits", msg)
-			resources.Limits[v1.ResourceCPU], _ = resource.ParseQuantity(minCPULimit)
+			resources.Limits[v1.ResourceCPU] = minCPU
 		}
 	}
 
 	memoryLimit := resources.Limits[v1.ResourceMemory]
 	minMemoryLimit := c.OpConfig.MinMemoryLimit
-	if minMemoryLimit != "" {
+	minMemory, err := resource.ParseQuantity(minMemoryLimit)
+	if err != nil {
+		return fmt.Errorf("could not parse min memory limit %s: %v", minMemoryLimit, err)
+	}
+	if !minMemory.IsZero() {
 		isSmaller, err = util.IsSmallerQuantity(memoryLimit.String(), minMemoryLimit)
 		if err != nil {
 			return fmt.Errorf("could not compare defined memory limit %s for %q container with configured minimum value %s: %v",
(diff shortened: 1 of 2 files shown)


FxKu:
@dmotte thanks! Could you provide also unit tests that cover this "new" behavior so I could also try them on the main branch and see what doesn't work? This would be a great improvement to this PR.

dmotte:
@FxKu done :slightly_smiling_face: I hope it's ok

pkg/cluster/k8sres_test.go
```diff
@@ -2993,6 +2993,44 @@ func TestGenerateResourceRequirements(t *testing.T) {
...
+					ClusterLabels:        map[string]string{"application": "spilo"},
+					ClusterNameLabel:     clusterNameLabel,
+					DefaultCPURequest:    "",
+					DefaultCPULimit:      "",
+					MaxCPURequest:        "",
+					MinCPULimit:          "",
+					DefaultMemoryRequest: "",
+					DefaultMemoryLimit:   "",
+					MaxMemoryRequest:     "",
+					MinMemoryLimit:       "",
```
FxKu, line 3009: You can change all these to "0", because "" is the internal default for string type in Golang. It is already tested in the subTest `test generation of resources when default is not defined`.

pkg/cluster/k8sres_test.go
```diff
@@ -2993,6 +2993,44 @@ func TestGenerateResourceRequirements(t *testing.T) {
...
+						ResourceLimits: acidv1.ResourceDescription{CPU: k8sutil.StringToPointer("5m"), Memory: k8sutil.StringToPointer("5Mi")},
+					},
+					TeamID: "acid",
+					Volume: acidv1.Volume{
+						Size: "1G",
+					},
+				},
+			},
+			expectedResources: acidv1.Resources{
+				ResourceLimits: acidv1.ResourceDescription{CPU: k8sutil.StringToPointer("5m"), Memory: k8sutil.StringToPointer("5Mi")},
```
FxKu, line 3031: So in the end the purpose of the test would be to see if no requests are defined even when minCpiLimit is set to "0".

pkg/cluster/k8sres.go
```diff
@@ -150,42 +150,53 @@ func makeLogicalBackupResources(config *config.Config) acidv1.Resources {
...
-				cpuLimit.String(), constants.PostgresContainerName, minCPULimit, err)
+			return fmt.Errorf("could not parse min CPU limit %s: %v", minCPULimit, err)
 		}
-		if isSmaller {
-			msg = fmt.Sprintf("defined CPU limit %s for %q container is below required minimum %s and will be increased",
-				cpuLimit.String(), constants.PostgresContainerName, minCPULimit)
-			c.logger.Warningf(msg)
-			c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeWarning, "ResourceLimits", msg)
-			resources.Limits[v1.ResourceCPU], _ = resource.ParseQuantity(minCPULimit)
+		if !minCPU.IsZero() {
```
FxKu, line 164: this extra code is not required. If c.OpConfig.MinCPULimit is not defined it will be "". Even when it's "0" and the comparison code after isSmaller is executed if won't get enforced even when no limits are specified.

pkg/cluster/k8sres.go
```diff
@@ -150,42 +150,53 @@ func makeLogicalBackupResources(config *config.Config) acidv1.Resources {
...
+				return fmt.Errorf("could not compare defined memory limit %s for %q container with configured minimum value %s: %v",
+					memoryLimit.String(), constants.PostgresContainerName, minMemoryLimit, err)
+			}
+			if isSmaller {
+				msg = fmt.Sprintf("defined memory limit %s for %q container is below required minimum %s and will be increased",
+					memoryLimit.String(), constants.PostgresContainerName, minMemoryLimit)
+				c.logger.Warningf(msg)
+				c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeWarning, "ResourceLimits", msg)
+				resources.Limits[v1.ResourceMemory] = minMemory
+			}
```
FxKu, line 199: same here. All these changes are not needed. The code works as expected imo.

FxKu:
Seems to me that only the schema validation needs to change. It's still good though to add a "0" example to the unit test.

dmotte:
@FxKu yeah, I think I got your point and you are right :)

Just to double check, correct me if I'm wrong: `util.IsSmallerQuantity(somePositiveQuantity, "0")` returns `false`, so `isSmaller` ends up being `false` and the limit is not enforced.

Thank you! Fixed in 24df591

FxKu:
Thanks. Can you also update these files with the new validation pattern: 
- [CRD Manifest example](https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql.crd.yaml)
- [CRD go representation](https://github.com/zalando/postgres-operator/blob/master/pkg/apis/acid.zalan.do/v1/crds.go)

dmotte:
@FxKu I think you meant `manifests/operatorconfiguration.crd.yaml` instead of `manifests/postgresql.crd.yaml`, so I updated that one instead. Done in 107fe9b :wink: let me know if it's OK or you have other requests

FxKu:
:+1:

idanovinda:
👍

PR #2737, opened 2024-08-21
remove streams delete and extend unit tests

FxKu:
#2733 fixed a bug with stream duplicates created on every operator restart. To facilitate the cleanup the operator was deleting every extra stream it found based on the filter labels.

In general, the operator should be careful with what it deletes. Therefore this PR removes the delete call again assuming that duplicates should not exist only if somebody adds them manually with the same labels which is unlikely to happen. Sorting is not supported in go client so we would not be able to tell which stream gets synced in the cluster struct (same for deletions).

I have split the update streams unit test to have a dedicated test only for removal.

### DIFF
--- a/pkg/cluster/streams.go
+++ b/pkg/cluster/streams.go
@@ -451,15 +451,6 @@ func (c *Cluster) syncStream(appId string) error {
 		if stream.Spec.ApplicationId != appId {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


idanovinda:
👍

FxKu:
:+1:

PR #2738, opened 2024-08-23
align sync and update logs

FxKu:
This PR started with a small fix of the logs for syncing the config service (endpoint vs service). There was no debug message for syncing it. While looking at the operator logs in UI I found a few more log lines which were unnecessarily using the info log level instead of debug.

I went through the code and found also debug lines which could in fact be info log lines because they happen during cluster updates. In general, syncing should only be on debug level, actual changes on info level (at least the lines explaining the diff and reporting successful update - intermediate steps can still be on debug level)

After all, most changes in this PR is changing from Debugf (and Debugln) to Debug function when no formatting is needed.


macedigital (approved):
Looks good to me

macedigital:
:+1:

FxKu:
👍

PR #2740, opened 2024-08-23
update K8s version and reflect necessary changes

FxKu:
I intentionally did not change it for v1.13.0 because I'm not sure if the operator would then stop working for K8s 1.28- any longer.


macedigital (approved):
Looks good to me. I wouldn't expect the changes to be backward necessarily incompatible with 1.28.x, but highly recommend to mention this as potentially breaking change.

Wouldn't it also be possible to spin up a 1.28.x minikube cluster to check for compatibility ?

macedigital:
:+1:

FxKu:
I tested with minikube 1.28 and I could spin up clusters. Therefore, we can merge it.

FxKu:
:+1:

PR #2743, opened 2024-08-28
remove zappr

FxKu:
We're migrating to ComPR and zappr will be EOL soon. Good bye [Zappr](https://zappr.opensource.zalan.do/login) and thank you for all those years of guaranteeing 4-eye approval.


macedigital:
Hi, I think ComPR requires a `.zappr.yaml` file to be present still. What could be removed is the `approvals` configuration (basically leaving the team / type in the file)

PR #2750, opened 2024-09-05
fix(2582): added cluster level delete secrets config

Yingrjimsch:
Here [#2582 ](https://github.com/zalando/postgres-operator/pull/2582) it has been discussed to add a config for deleting secrets globally on operator level. This is very nice and helps greatly.

In my case I have a setup where my operator contains the config `delete_annotation_name_key: delete-clustername` and as a safety config `enable_secrets_deletion: false` which is needed if I want to clear a postgres cluster completely but still need the secrets. This works like a charm but I have temporary walg cluster clones which can be applied to get some backups. This temporary walg clones should still delete their secrets post deletion therefore a config is needed on cluster level.
I implemented the config `enable_secrets_deletion_key` in a similar way as `delete_annotation_name_key` works. This allows me to override the operator wide `enable_secrets_deletion` configuratin per postgresql cluster. If it is not set, the operator decides whether the secrets should be deleted, if it is set to `true` the secrets are deleted nevertheless and if it is set to `false` the secrets will not be deleted, allowing for a more fine granular configuration.

I have checked several possible scenarios with the following results:
| `enable_secrets_deletion` | `delete_annotation_name_key` | `delete-clustername` | `enable-secrets-deletion` |  Keep Secrets? |
| -------- | ------- | ------- | ------- | ------- | 
| - | - | - | - | NO |
| true | - | - | - | NO | 
| false | - | - | - | YES | 
| - | delete-clustername | acid-minimal-cluster | - | NO | 
| - | delete-clustername | - | - | YES | 
| true | delete-clustername | - | - | YES | 
| false | delete-clustername | - | - | YES |
| true | delete-clustername | - | true | YES | 
| false | delete-clustername | - | false | YES |
| true | delete-clustername | acid-minimal-cluster | - | NO | 
| false | delete-clustername | acid-minimal-cluster | - | YES |
| true | delete-clustername | acid-minimal-cluster | true | NO | 
| false | delete-clustername | acid-minimal-cluster | false | YES |

@dmotte can you review this please?

### DIFF
--- a/pkg/util/config/config.go
+++ b/pkg/util/config/config.go
@@ -66,6 +66,7 @@ type Resources struct {
 	MaxInstances                      int32  `name:"max_instances" default:"-1"`
 	MinInstances                      int32  `name:"min_instances" default:"-1"`
 	IgnoreInstanceLimitsAnnotationKey string `name:"ignore_instance_limits_annotation_key"`
+	EnableSecretsDeletionKey          string `name:"enable_secrets_deletion_key"`
 }
 
 type InfrastructureRole struct {
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -229,6 +229,8 @@ spec:
                   enable_secrets_deletion:
                     type: boolean
                     default: true
+                  enable_secrets_deletion_key:
+                    type: string
                   enable_sidecars:
                     type: boolean
                     default: true
--- a/charts/postgres-operator/values.yaml
+++ b/charts/postgres-operator/values.yaml
@@ -141,6 +141,9 @@ configKubernetes:
   enable_readiness_probe: false
   # toggles if operator should delete secrets on cluster deletion
   enable_secrets_deletion: true
+  # key name for annotation that overrides enable_secrets_deletion on cluster level
+  # enable_secrets_deletion_key: "enable-secrets-deletion"
+
   # enables sidecar containers to run alongside Spilo in the same pod
   enable_sidecars: true
 
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -365,6 +365,9 @@ configuration they are grouped under the `kubernetes` key.
   By default, the operator deletes secrets when removing the Postgres cluster
   manifest. To keep secrets, set this option to `false`. The default is `true`.
 
+* **enable_secrets_deletion_key**
+  By default, the `enable_secrets_deletion` decides on the deletion of secrets for the entire operator. To overwrite `enable_secrets_deletion` this property can be set and an annotation on cluster level can be added with the values: delete secrets `true` or `false`.
+
 * **enable_persistent_volume_claim_deletion**
   By default, the operator deletes PersistentVolumeClaims when removing the
   Postgres cluster manifest, no matter if `persistent_volume_claim_retention_policy`
--- a/manifests/configmap.yaml
+++ b/manifests/configmap.yaml
@@ -61,6 +61,7 @@ data:
   enable_replica_load_balancer: "false"
   enable_replica_pooler_load_balancer: "false"
   enable_secrets_deletion: "true"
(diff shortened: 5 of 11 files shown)


2
dmotte, line 1: I guess this file has been committed by accident :)

dmotte (commented):
In general it looks good to me @Yingrjimsch thank you so much for taking the time to do this!

Maybe it's also worth to document this new config option, WDYT? I think it should go into the `docs/reference/cluster_manifest.md` file.

Last thing: I think the `pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go` file needs to be regenerated after your changes. Could you please regenerate and commit it? Thanks!

2
Yingrjimsch: @dmotte thanks for reviewing. I added the documentation in `docs/reference/operator_parameters.md` because there all the other properties `delete_annotation_name_key` and your `enable_secret_deletion` are placed. 
Can you tell me how to regenerate the `pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go`? I tried by running `hack/update-codegen.sh` but this did not update anything.

@FxKu I saw you have approved the request from @dmotte is it possible for you to review this as well? (Shouldn't take too much time ;)

dmotte: > Can you tell me how to regenerate the `pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go` ?

@Yingrjimsch honestly I'm not sure, but I think you just have to put this piece of code

```go
	if in.EnableSecretsDeletionKey != nil {
		in, out := &in.EnableSecretsDeletionKey, &out.EnableSecretsDeletionKey
		*out = new(string)
		**out = **in
	}
```

right after the `if in.EnableSecretsDeletion != nil { ... }` block, i.e. right here:

https://github.com/Yingrjimsch/postgres-operator/blob/6fdf2c92fa71e1bf3d99deff35401fd93d736086/pkg/apis/acid.zalan.do/v1/zz_generated.deepcopy.go#L289-L290

I suggest you to simply do that for now, and then let's wait for other reviews :wink:

Yingrjimsch:
@FxKu could you please review this "small" fix? I would really appreciate it 😄

2
dmotte: > I suggest you to simply do that for now

@Yingrjimsch I see that you reacted with a "thumbs up" but I cannot see the fix yet :) let me know if you need further explanation

Yingrjimsch: > > I suggest you to simply do that for now
> 
> @Yingrjimsch I see that you reacted with a "thumbs up" but I cannot see the fix yet :) let me know if you need further explanation

@dmotte I've tried it with your suggested codeblock. The result is, that the build fails. I saw that the `DeleteAnnotationNameKey` is also missing in the zz_generated.deepcopy.go and due to the fact that the `hack/update-codegen.sh` has not done anything I thought I will leave it for now. If you know why it is not able to build I'm gladly adding it.

dmotte: @Yingrjimsch ok, got it. In this case yeah maybe you're right and that's not needed. Thank you for the explanation :+1:

dmotte (approved):

pkg/util/config/config.go
```diff
@@ -66,6 +66,7 @@ type Resources struct {
 	MaxInstances                      int32  `name:"max_instances" default:"-1"`
 	MinInstances                      int32  `name:"min_instances" default:"-1"`
 	IgnoreInstanceLimitsAnnotationKey string `name:"ignore_instance_limits_annotation_key"`
+	EnableSecretsDeletionKey          string `name:"enable_secrets_deletion_key"`
```
FxKu, line 69: Can you rename the to `IgnoreSecretDeletionAnnotationKey`?

FxKu:
I haven't checked yet, if the option works vice versa, but this would be my preferred implementation. So that you can have:
a. global config to drop secrets on cluster deletion, keep for single clusters which use the ignore annotation == "true"
b. global config to keep secrets on cluster deletion, drop for single clusters which use the ignore annotation == "true"

PR #2755, opened 2024-09-09
Bugfix: skip upgrade when failed annotation should happen after recheck version

idanovinda:
When operator (re)-start, `c.currentMajorVersion` will be 0. This value will be updated only if we execute `majorVersionUpgrade` until the part where we get the current version from the Patroni. In #2727, we skipped the part where we do update this value when a cluster have `last-major-upgrade-failure` annotation. Hence, this cluster will never update its `c.currentMajorVersion` after operator restarts and will always be 0. Our e2e test couldn't catch this as we test the `last-major-upgrade-failure` after doing some upgrades. 

**Solution**

We should only skip the upgrade of the failure upgrade only after we make sure that the recheck is done.

### DIFF
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -127,11 +127,6 @@ func (c *Cluster) majorVersionUpgrade() error {
 		return nil
 	}
 
-	if _, exists := c.ObjectMeta.Annotations[majorVersionUpgradeFailureAnnotation]; exists {
-		c.logger.Infof("last major upgrade failed, skipping upgrade")
-		return nil
-	}
-
 	if !isInMainternanceWindow(c.Spec.MaintenanceWindows) {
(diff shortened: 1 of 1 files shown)


idanovinda:
👍

pkg/cluster/majorversionupgrade.go
```diff
@@ -162,10 +157,19 @@ func (c *Cluster) majorVersionUpgrade() error {
 
 	// Recheck version with newest data from Patroni
 	if c.currentMajorVersion >= desiredVersion {
+		if _, exists := c.ObjectMeta.Annotations[majorVersionUpgradeFailureAnnotation]; exists { // if failure annotation exists, remove it
```
idanovinda, line 160:
```suggestion
		if _, exists := c.ObjectMeta.Annotations[majorVersionUpgradeFailureAnnotation]; exists {
```

FxKu:
:+1:

PR #2772, opened 2024-10-11
Add major upgrade prechecks

hughcapet:
Don't fail major upgrade (don't set annotation) if replica(s) are not (yet) streaming or replication lag is too high


FxKu:
:+1:

hughcapet:
👍

Issue #2745, opened 2024-08-30
could not create user \"app-new_public_owner\": pq: syntax error at or near \"-\"'


hvp1991:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** e.g. ghcr.io/zalando/postgres-operator:v1.11.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare
- **Are you running Postgres Operator in production?** [yes | no] No
- **Type of issue?** [Bug report, question, feature request, etc.], Bug Report / Question

We are using zalando operator, However when we try to create database with special character in name (to be specific hyphen `-`), we do notice some errors in operator logs! In this example you can find db name is `app-new`

I do see credentials are getting created in zalando , However while syncing roles it gives specifically following error while creating role, The strange part is `app-new_owner` **role** and `app-new_owner_user` **user** got created  but **fails** with **reader**, **writer** roles and users!

As per the code base special characters are allowed since queries are used with special character..!!

**Error**:
```
[...]

PR #2782, opened 2024-10-17
quote schema names in case they use special characters

FxKu:
fixes error: `syntax error at or near "-"', 'could not set custom user "dbname_schema-with-dashes_owner" parameters` 

addresses #2745


FxKu:
:+1:

idanovinda:
👍

PR #2783, opened 2024-10-17
fix(postgres-operator/deployment): Set 'nindent' to 8 for 'extraEnvs'

Mart-Kuc:
This PR fixes a template error in `postgres-operator/deployment`.
If you want to use `extraEnvs` in `values.yaml`, the current indentation for `extraEnvs` causes the resulting `deployment.yaml` to break due to incorrect indentation.

The nindent value has been updated from `12` to `8`.

**Result with the old values.yaml (nindent 12):**

Old-template

**Result with the new nindent set to 8:**
New-template

Bumped Chart.yaml version to 1.13.1 but I'm not sure if it's correct

### DIFF
--- a/charts/postgres-operator/Chart.yaml
+++ b/charts/postgres-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: postgres-operator
-version: 1.13.0
+version: 1.14.0
 appVersion: 1.13.0
 home: https://github.com/zalando/postgres-operator
 description: Postgres Operator creates and manages PostgreSQL clusters running in Kubernetes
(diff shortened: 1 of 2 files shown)


daleckystepan:
That would be nice. We also hit that.

charts/postgres-operator/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: postgres-operator
-version: 1.13.0
```
FxKu, line 3: this will be bumped to 1.14.0. We have no branching in place to create a bugfix release for 1.13.0. Too many feature already merged to main trunk.

FxKu:
Thanks for providing the fix. Could you please keep the version on the Chart.yaml? (see my comment)

charts/postgres-operator/Chart.yaml
FxKu: I wanted to keep it at 1.13.0 for now, but ok, we can also set to 1.14.0.

FxKu:
:+1:

hughcapet (approved):

PR #2785, opened 2024-10-18
add cluster field for PVCs

FxKu:
PVCs are the only resource which is not covered by OwnerReferences. We have observed that on cluster deletion operator can report that all PVCs have already been deleted, although they still exist. We believe that there might be a race condition with cluster fields such as Name and Namespace getting unset after cascaded deletion via OwnerReferences resulting in an empty list when [fetching PVCs to delete](https://github.com/zalando/postgres-operator/blob/002d0f94a1463c15500dbfa77ea42c48e5790e6b/pkg/cluster/volumes.go#L266) (labelSet function [using c.Name](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/util.go#L479) for filtering). PVC management by StatefulSet is set to Retain in our case.

This PR present an idea which we use for other resources as well: Storing resource information in a cluster field and loop over this map on deletion. However, unlike resources like Secrets and Services, PVCs are created by K8s via StatefulSet template. So we rely on a [syncVolumeClaims](https://github.com/zalando/postgres-operator/blob/002d0f94a1463c15500dbfa77ea42c48e5790e6b/pkg/cluster/volumes.go#L47) call to have the PVCs info stored in the cluster struct. If this sync does not happen for some reason, the PVCs will not get deleted. I don't think an extra call of listPersisntentVolumeClaims is needed as c.Name would mostly likely be empty as well in this scenario.

I found that syncVolumes is not called in Create() event. Adding it there would greatly reduce the chance that PVCs are not synced before a potential deletion.

### DIFF
--- a/pkg/cluster/volumes.go
+++ b/pkg/cluster/volumes.go
@@ -185,7 +185,7 @@ func (c *Cluster) syncVolumeClaims() error {
 
 	if c.OpConfig.StorageResizeMode == "off" || c.OpConfig.StorageResizeMode == "ebs" {
 		ignoreResize = true
-		c.logger.Debugf("Storage resize mode is set to %q. Skipping volume size sync of PersistentVolumeClaims.", c.OpConfig.StorageResizeMode)
+		c.logger.Debugf("Storage resize mode is set to %q. Skipping volume size sync of persistent volume claims.", c.OpConfig.StorageResizeMode)
 	}
 
 	newSize, err := resource.ParseQuantity(c.Spec.Volume.Size)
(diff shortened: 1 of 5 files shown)


pkg/cluster/volumes.go
```diff
@@ -268,28 +272,40 @@ func (c *Cluster) listPersistentVolumeClaims() ([]v1.PersistentVolumeClaim, erro
...
+		c.logger.Warningf("could not delete all persistent volume claims: %v", strings.Join(errors, `', '`))
+	}
+
+	return nil
+}
+
+func (c *Cluster) deletePersistentVolumeClaim(uid types.UID) error {
+	c.setProcessName("deleting persistent volume claim")
+	pvc := c.VolumeClaims[uid]
+	c.logger.Debugf("deleting secret %q", pvc.Name)
```
hughcapet, line 300: secret?

FxKu: good catch

hughcapet:
👍

FxKu:
:+1:

PR #2802, opened 2024-11-19
Feat: adding configurable parameter for custom role label in multiple resources

idanovinda:
New Spilo with Patroni 4.x integration: https://github.com/zalando/spilo/pull/1050

This PR enhances operator to support Patroni 4.x while retaining compatibility with existing clusters by making the new role label configurable. The role terminology has changed from "master" to "primary". By default, Spilo will use the same value as before, which is `master`. But to allow flexibility in label configurations, we need to ensure that the role label are configurable.

This is achieved by adding configurable parameters:
* `pod_leader_label_value`: default is set to "master"

By setting this new configurable parameters to other value than default, it means that existing clusters need to be migrated to have the new role value. Migration hints from Patroni documentation: https://patroni.readthedocs.io/en/latest/kubernetes.html#customize-role-label


pkg/cluster/sync.go
```diff
@@ -340,7 +340,7 @@ func (c *Cluster) syncPatroniService() error {
 }
 
 func (c *Cluster) syncServices() error {
-	for _, role := range []PostgresRole{Master, Replica} {
+	for _, role := range []PostgresRole{c.masterRole(), c.replicaRole()} {
```
hughcapet, line 343: another thing we need to think about is the transition period: e.g., when "master" is changed to something custom.
this way, we probably need to delete the old objects on sync and also remove those from the Cluster struct

otherwise
1. master -> meow change
2. sync event
3. creates service for the "meow" role (now there are both master and meow services)
4. c.Services[role] = createdSvc (where role == "meow" but we also already have c.Services["master"])

hughcapet: and of course test the transition period at list manually:)

hughcapet:
The PR is still WIP -- migration procedure needs to be implemented. See https://patroni.readthedocs.io/en/latest/kubernetes.html#customize-role-label

idanovinda:
close as it's not relevant anymore

PR #2803, opened 2024-11-19
do not use extra labels to list stream CRDs

FxKu:
Using extra labels for searching stream CRs leads to the creation of new resources when, for example, the teamId label changes. This should not happen. Only use cluster labels for search like we do in other list API calls. Unit test have to reflect the behavior, too.

The stream resource, on the other hand, should get all labels and we have to include a label comparison to the compareStream function. The corresponding unit test has been extended.


idanovinda:
👍

idanovinda:
👍

PR #2810, opened 2024-12-04
Extend MaintenanceWindows parameter usage

hughcapet:
Consider maintenance window when migrating master pods and replacing pods (rolling update)

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -209,7 +209,7 @@ path are ignored for the diff. They will be applied through Patroni's rest api
 interface, following a restart of all instances.
 
 Rolling update is postponed until the next maintenance window if any is defined
-under the `maintenanceWindows` cluster configuration parameter.
+under the `maintenanceWindows` cluster manifest parameter.
 
 The operator also support lazy updates of the Spilo image. In this case the
 StatefulSet is only updated, but no rolling update follows. This feature saves
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -208,6 +208,9 @@ Note that, changes in `SPILO_CONFIGURATION` env variable under `bootstrap.dcs`
 path are ignored for the diff. They will be applied through Patroni's rest api
 interface, following a restart of all instances.
 
+Rolling update is postponed until the next maintenance window if any is defined
+under the `maintenanceWindows` cluster configuration parameter.
+
 The operator also support lazy updates of the Spilo image. In this case the
 StatefulSet is only updated, but no rolling update follows. This feature saves
 you a switchover - and hence downtime - when you know pods are re-started later
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1736,6 +1736,12 @@ func (c *Cluster) Switchover(curMaster *v1.Pod, candidate spec.NamespacedName) e
 
 	var err error
 	c.logger.Debugf("switching over from %q to %q", curMaster.Name, candidate)
+
+	if !isInMaintenanceWindow(c.Spec.MaintenanceWindows) {
+		c.logger.Infof("skipping switchover, not in maintenance window")
+		return nil
+	}
+
 	c.eventRecorder.Eventf(c.GetReference(), v1.EventTypeNormal, "Switchover", "Switching over from %q to %q", curMaster.Name, candidate)
 	stopCh := make(chan struct{})
 	ch := c.registerPodSubscriber(candidate)
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -1741,18 +1741,15 @@ func (c *Cluster) GetSwitchoverSchedule() string {
 		if window.Everyday {
 			if now.After(possibleSwitchover) {
 				// we are already past the time for today, try tomorrow
-				day := now.AddDate(0, 0, 1)
-				possibleSwitchover = time.Date(day.Year(), day.Month(), day.Day(), window.StartTime.Hour(), window.StartTime.Minute(), 0, 0, time.UTC)
+				possibleSwitchover = possibleSwitchover.AddDate(0, 0, 1)
 			}
 		} else {
 			if now.Weekday() != window.Weekday {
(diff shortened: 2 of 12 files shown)


pkg/cluster/util.go
```diff
@@ -663,7 +663,7 @@ func parseResourceRequirements(resourcesRequirement v1.ResourceRequirements) (ac
 	return resources, nil
 }
 
-func isInMainternanceWindow(specMaintenanceWindows []acidv1.MaintenanceWindow) bool {
+func IsInMainternanceWindow(specMaintenanceWindows []acidv1.MaintenanceWindow) bool {
```
idanovinda, line 666: let's also fix the typo as well: `IsInMainternanceWindow` -> `IsInMaintenanceWindow`

pkg/cluster/majorversionupgrade.go
```diff
@@ -128,7 +128,7 @@ func (c *Cluster) majorVersionUpgrade() error {
 		return nil
 	}
 
-	if !isInMainternanceWindow(c.Spec.MaintenanceWindows) {
+	if !IsInMainternanceWindow(c.Spec.MaintenanceWindows) {
```
hughcapet, line 131:
```suggestion
	if !IsInMaintenanceWindow(c.Spec.MaintenanceWindows) {
```

pkg/cluster/util.go
hughcapet, line 666:
```suggestion
func IsInMaintenanceWindow(specMaintenanceWindows []acidv1.MaintenanceWindow) bool {
```

idanovinda:
👍

pkg/cluster/sync.go
```diff
@@ -658,7 +658,9 @@ func (c *Cluster) syncStatefulSet() error {
 	// if we get here we also need to re-create the pods (either leftovers from the old
 	// statefulset or those that got their configuration from the outdated statefulset)
 	if len(podsToRecreate) > 0 {
-		if isSafeToRecreatePods {
+		if !isInMaintenanceWindow(c.Spec.MaintenanceWindows) {
+			c.logger.Infof("postpone pod recreation - not in maintenance window")
+		} else if isSafeToRecreatePods {
```
FxKu, line 663: just a refactoring question - maybe !isInMaintenanceWindow could be it's own if a few lines above to set isSafeToRecreatePods to false. We could collect the reason from all the blocks that set isSafeToRecreatePods = false and then we log here: c.logger.Warnf("postpone pod recreation - reason: %s", strings.Joins(postponeReasons))

pkg/cluster/cluster.go
```diff
@@ -1733,16 +1733,56 @@ func (c *Cluster) GetStatus() *ClusterStatus {
...
+	if !isInMaintenanceWindow(c.Spec.MaintenanceWindows) {
+		c.logger.Infof("postponing switchover, not in maintenance window")
+
+		var possibleSwitchover, schedule time.Time
+
+		now := time.Now().UTC()
+		for _, window := range c.Spec.MaintenanceWindows {
+			if window.Everyday {
+				possibleSwitchover = time.Date(now.Year(), now.Month(), now.Day(), window.StartTime.Hour(), window.StartTime.Minute(), 0, 0, time.UTC)
+				if now.After(possibleSwitchover) {
```
FxKu, line 1748:
```suggestion
			possibleSwitchover = time.Date(now.Year(), now.Month(), now.Day(), window.StartTime.Hour(), window.StartTime.Minute(), 0, 0, time.UTC)
			if window.Everyday && now.After(possibleSwitchover) {
```

pkg/cluster/cluster.go
```diff
@@ -1733,16 +1733,56 @@ func (c *Cluster) GetStatus() *ClusterStatus {
...
+					day := now.AddDate(0, 0, 1)
+					possibleSwitchover = time.Date(day.Year(), day.Month(), day.Day(), window.StartTime.Hour(), window.StartTime.Minute(), 0, 0, time.UTC)
+				}
+			} else {
+				timeToday := time.Date(now.Year(), now.Month(), now.Day(), window.StartTime.Hour(), window.StartTime.Minute(), 0, 0, time.UTC)
+				// is it still possible today?
+				if now.Weekday() == window.Weekday {
+					if now.Before(timeToday) {
+						possibleSwitchover = timeToday
+					}
```
FxKu, line 1759: we only need to act on `if now.Weekday() != window.Weekday`, no?

FxKu:
We have to update [reference docs](https://github.com/zalando/postgres-operator/blob/master/docs/reference/cluster_manifest.md#top-level-parameters) and also sections that talk about rolling updates like in [admin docs](https://github.com/zalando/postgres-operator/blob/master/docs/administrator.md#understanding-rolling-update-of-spilo-pods). About user docs I'm not sure. So far they do not mention maintenance windows at all.

pkg/cluster/cluster.go
hughcapet: why? what if it is 10am while the window is at 20:00-22:00 today

pkg/cluster/cluster.go
hughcapet: this way "else" path's logic with Weekday attribute will be executed for "window is everyday but is after the current time.

hughcapet: ok. but I started to run tests and now see the general problem here...

docs/administrator.md
```diff
@@ -208,6 +208,9 @@ Note that, changes in `SPILO_CONFIGURATION` env variable under `bootstrap.dcs`
 path are ignored for the diff. They will be applied through Patroni's rest api
 interface, following a restart of all instances.
 
+Rolling update is postponed until the next maintenance window if any is defined
+under the `maintenanceWindows` cluster configuration parameter.
```
FxKu, line 212:
```suggestion
under the `maintenanceWindows` cluster manifest parameter.
```

FxKu:
:+1:

hughcapet:
👍

PR #2814, opened 2024-12-11
Fix deployment extraEnvs indentation in operator chart

DaDummy:
Fix bad operator extraEvns indentation by matching the statement to how other lists are expanded in the deployment template.

The previous implementation caused a yaml parsing error due to the expanded extraEnvs list being indented by more than 8 spaces while the other list entries from the template just have 8 spaces indentation.


charts/postgres-operator/templates/deployment.yaml
```diff
@@ -54,7 +54,7 @@ spec:
           value: {{ template "postgres-operator.controllerID" . }}
       {{- end }}
       {{- if .Values.extraEnvs }}
-      {{- .Values.extraEnvs | toYaml | nindent 8 }}
+{{ toYaml .Values.extraEnvs | nindent 8 }}
```
FxKu, line 57: why is it nindent and not just indent?

DaDummy: I missed that the comparable lines in the same file all used indent. I've adjusted the line now to use indent, too.

FxKu:
:+1:

hughcapet:
👍

PR #2815, opened 2024-12-11
allow hyphens in database name

JeffResc:
Allow for hyphens in the database name, they are currently supported by Postgres, but not this operator

Current behavior:
```
time="2024-12-10T22:58:31Z" level=info msg="database \"my-database\" has invalid name" cluster-name=postgres/pg-cluster pkg=cluster
```


FxKu:
@JeffResc thanks for your PR. I would prefer to make the database regex configurable like it was attempted in #2139. Maybe you can pick up the work done there?

PR #2820, opened 2024-12-17
fix comparing stream annotations and improve unit test

FxKu:
The annotation comparison for streams was not ready yet, for comparing new annotations. Only the current ones were copied to `desiredAnnotations`. This was not caught because unit test compared either different streams with different specs, and we did not compare the full reason string. By extending the latter, tests started failing.

After fixing the annotation comparison the stream generator function started failing because the current code did not set the resource annotation when they were missing before. Only when a value could be found for comparison (which is impossible when setting the annotation does not work, of course :zany_face: )

I refactored the code and created a new pass by reference function `setResourceAnnotation` so that the FES generation code remains readable.

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -2204,6 +2204,8 @@ class EndToEndTestCase(unittest.TestCase):
                         {
                             "applicationId": "test-app",
                             "batchSize": 100,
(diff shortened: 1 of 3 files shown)


idanovinda:
👍

FxKu:
:+1:

PR #2822, opened 2024-12-19
fix flaky stream comparison unit test of retruned errors

FxKu:
in #2820 the stream comparison unit test got greatly improved and we added a comparison of expected versus returned reason. It turns out that the order or reported added or mismatching annotations is unpredictable. When we remove the memory setting from the unit test, the chances for differences should be minimized (maybe even impossible)


idanovinda:
👍

FxKu:
:+1:

PR #2826, opened 2024-12-23
Ensure podAnnotations are removed from pods if reset in the config

hughcapet:
fix use case when an annotation is removed from the `podAnnotations` setting  (should be reflected for the running Pods and Pod templates)

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -430,7 +430,7 @@ func (c *Cluster) Create() (err error) {
 	return nil
 }
 
-func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compareStatefulsetResult {
+func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet, deltePodAnnotations *[]string) *compareStatefulsetResult {
 	reasons := make([]string, 0)
 	var match, needsRollUpdate, needsReplace bool
 
@@ -445,7 +445,7 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 		needsReplace = true
 		reasons = append(reasons, "new statefulset's ownerReferences do not match")
 	}
-	if changed, reason := c.compareAnnotations(c.Statefulset.Annotations, statefulSet.Annotations); changed {
+	if changed, reason := c.compareAnnotations(c.Statefulset.Annotations, statefulSet.Annotations, nil); changed {
 		match = false
 		needsReplace = true
 		reasons = append(reasons, "new statefulset's annotations do not match: "+reason)
@@ -519,7 +519,7 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 		}
 	}
 
-	if changed, reason := c.compareAnnotations(c.Statefulset.Spec.Template.Annotations, statefulSet.Spec.Template.Annotations); changed {
+	if changed, reason := c.compareAnnotations(c.Statefulset.Spec.Template.Annotations, statefulSet.Spec.Template.Annotations, deltePodAnnotations); changed {
 		match = false
 		needsReplace = true
 		reasons = append(reasons, "new statefulset's pod template metadata annotations does not match "+reason)
@@ -541,7 +541,7 @@ func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compa
 				reasons = append(reasons, fmt.Sprintf("new statefulset's name for volume %d does not match the current one", i))
 				continue
 			}
-			if changed, reason := c.compareAnnotations(c.Statefulset.Spec.VolumeClaimTemplates[i].Annotations, statefulSet.Spec.VolumeClaimTemplates[i].Annotations); changed {
+			if changed, reason := c.compareAnnotations(c.Statefulset.Spec.VolumeClaimTemplates[i].Annotations, statefulSet.Spec.VolumeClaimTemplates[i].Annotations, nil); changed {
 				needsReplace = true
 				reasons = append(reasons, fmt.Sprintf("new statefulset's annotations for volume %q do not match the current ones: %s", name, reason))
 			}
@@ -781,7 +781,7 @@ func volumeMountExists(mount v1.VolumeMount, mounts []v1.VolumeMount) bool {
 	return false
 }
 
-func (c *Cluster) compareAnnotations(old, new map[string]string) (bool, string) {
+func (c *Cluster) compareAnnotations(old, new map[string]string, removedList *[]string) (bool, string) {
 	reason := ""
 	ignoredAnnotations := make(map[string]bool)
 	for _, ignore := range c.OpConfig.IgnoredAnnotations {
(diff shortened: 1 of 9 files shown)


idanovinda:
👍

pkg/cluster/connection_pooler.go
```diff
@@ -1037,10 +1037,47 @@ func (c *Cluster) syncConnectionPoolerWorker(oldSpec, newSpec *acidv1.Postgresql
...
+							if anno == ignore {
+								continue
+							}
+						}
+						annotationToRemove += fmt.Sprintf(`"%s":null,`, anno)
+						annotationToRemoveTemplate += fmt.Sprintf(`"%s":null,`, anno)
+					}
+				}
+				annotationToRemove = strings.TrimSuffix(annotationToRemove, ",") + `}}}`
+				annotationToRemoveTemplate = strings.TrimSuffix(annotationToRemoveTemplate, ",") + `}}}}}`
```
FxKu, line 1065: can we not use json marshalling here? This concatenating {{{ + }}} looks a bit "ugly".

pkg/cluster/connection_pooler.go
```diff
@@ -1060,7 +1097,6 @@ func (c *Cluster) syncConnectionPoolerWorker(oldSpec, newSpec *acidv1.Postgresql
 			if err != nil {
 				return syncReason, err
 			}
-			c.ConnectionPooler[role].Deployment = deployment
```
FxKu, line 1063: so we change to update internal deployment only if every sync step works, right? Have to think about consequences it might have

pkg/cluster/connection_pooler.go
```diff
@@ -1069,15 +1105,10 @@ func (c *Cluster) syncConnectionPoolerWorker(oldSpec, newSpec *acidv1.Postgresql
...
 		}
 	}
 
 	// check if pooler pods must be replaced due to secret update
-	listOptions := metav1.ListOptions{
-		LabelSelector: labels.Set(c.connectionPoolerLabels(role, true).MatchLabels).String(),
-	}
-	pods, err = c.listPoolerPods(listOptions)
```
FxKu, line 1080: ToDo for me: Check if we can go with a version of pods we now fetch before changing annotations? Or do we need to fetch them again for the following steps?

pkg/cluster/sync.go
```diff
@@ -573,6 +573,33 @@ func (c *Cluster) syncStatefulSet() error {
...
+				annotationToRemove = strings.TrimSuffix(annotationToRemove, ",") + `}}}`
+				for _, pod := range pods {
+					_, err = c.KubeClient.Pods(c.Namespace).Patch(context.Background(), pod.Name,
+						types.StrategicMergePatchType, []byte(annotationToRemove), metav1.PatchOptions{})
+					if err != nil {
+						c.logger.Errorf("failed to remove annotations from pod %s: %v", pod.Name, err)
+						return err
+					}
+				}
+			}
```
FxKu, line 602: Sorry, another thinking out loud comment: We already introduced patch code for annotations. Could it not be reused?

pkg/cluster/connection_pooler.go
hughcapet: oh, this is actually a redundant change

pkg/cluster/sync.go
hughcapet: you mean for upgrade annotations? there the merge type is different. but maybe we can actually use this one there...

FxKu: After checking out the branch and looking at the code as a whole I think this part should be combined with the lines before. The preceding annotation comparison only lacks the nil setting for removed annotations, doesn't it?

pkg/cluster/sync.go
```diff
@@ -1594,6 +1619,27 @@ func (c *Cluster) syncLogicalBackupJob() error {
...
+					}
+				}
+				annotationToRemoveTemplate = strings.TrimSuffix(annotationToRemoveTemplate, ",") + `}}}}}}}`
+				job, err = c.KubeClient.CronJobs(c.Namespace).Patch(context.TODO(),
+					jobName, types.StrategicMergePatchType, []byte(annotationToRemoveTemplate), metav1.PatchOptions{}, "")
+				if err != nil {
+					c.logger.Errorf("failed to remove annotations from the logical backup job %q pod template: %v", jobName, err)
+					return err
+				}
+			}
```
FxKu, line 1642: This part is still using the string concat patch contruction ... 

Again more patch code is following down the line. Is there no other was to do it. The code is becoming harder to understand with the extra comparison and patch code. It looks redundant with minor differences.

pkg/cluster/sync_test.go
```diff
@@ -142,6 +142,163 @@ func TestSyncStatefulSetsAnnotations(t *testing.T) {
...
+			EnableLogicalBackup:           true,
+			EnableReplicaConnectionPooler: boolToPointer(true),
+			PodAnnotations:                podAnnotations,
+			NumberOfInstances:             2,
+		},
+	}
+
+	var cluster = New(
+		Config{
+			OpConfig: config.Config{
```
FxKu, line 175: You could also include the global setting of podAnnotations: CustomPodAnnotations

pkg/cluster/connection_pooler.go
```diff
@@ -1041,6 +1043,32 @@ func (c *Cluster) syncConnectionPoolerWorker(oldSpec, newSpec *acidv1.Postgresql
...
+				if err != nil {
+					return nil, fmt.Errorf("could not marshal ObjectMeta for %s connection pooler's pod template: %v", role, err)
+				}
+				deployment, err = c.KubeClient.Deployments(c.Namespace).Patch(context.TODO(),
+					deployment.Name, types.StrategicMergePatchType, patch, metav1.PatchOptions{}, "")
+				if err != nil {
+					c.logger.Errorf("failed to remove annotations from %s connection pooler's pod template: %v", role, err)
+					return nil, err
+				}
+			}
```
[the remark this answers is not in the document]
hughcapet: we can not because of the necessity to set `= nil`

pkg/cluster/sync.go
hughcapet: i have forgotten refactor this part regarding the marshalling

pkg/cluster/connection_pooler.go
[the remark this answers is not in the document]
hughcapet: moreover, patching below is about annotations of the deployment itself. this one is about pods (template's metadata)

pkg/cluster/cluster.go
```diff
@@ -430,7 +430,7 @@ func (c *Cluster) Create() (err error) {
 	return nil
 }
 
-func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet) *compareStatefulsetResult {
+func (c *Cluster) compareStatefulSetWith(statefulSet *appsv1.StatefulSet, deltePodAnnotations *[]string) *compareStatefulsetResult {
```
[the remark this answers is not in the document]
FxKu: so that it can be removed as an argument to compareStatefulSetWith

pkg/cluster/cluster.go
```diff
@@ -836,7 +841,7 @@ func (c *Cluster) compareServices(old, new *v1.Service) (bool, string) {
 	return true, ""
 }
 
-func (c *Cluster) compareLogicalBackupJob(cur, new *batchv1.CronJob) (match bool, reason string) {
+func (c *Cluster) compareLogicalBackupJob(cur, new *batchv1.CronJob, deletedPodAnnotations *[]string) (match bool, reason string) {
```
FxKu, line 844: could you create a result struct similar to the one we have for comparing stateful sets? For compareAnnotations I'm fine to keep it as is, but for compareLogicalBackupJob it should not require to many changes.

FxKu:
:+1:

hughcapet:
👍

PR #2829, opened 2024-12-31
Update logical backup docker image

lukasredev:
Update logical backup docker image frm `13.0` to `14.0` so backups also work with postgres 17


idanovinda:
👍

PR #2842, opened 2025-01-16
Only check maintenance window for upgrade after pg version recheck

hughcapet:
This way we avoid misleading "skipping major version upgrade, not in maintenance window" log line when `c.currentMajorVersion` is not initialized (==0)

### DIFF
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -173,7 +173,7 @@ func (c *Cluster) majorVersionUpgrade() error {
 		c.logger.Infof("recheck cluster version is already up to date. current: %d, min desired: %d", c.currentMajorVersion, desiredVersion)
 		return nil
 	} else if isStandbyCluster {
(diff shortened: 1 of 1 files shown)


pkg/cluster/majorversionupgrade.go
```diff
@@ -175,13 +172,21 @@ func (c *Cluster) majorVersionUpgrade() error {
 		}
 		c.logger.Infof("recheck cluster version is already up to date. current: %d, min desired: %d", c.currentMajorVersion, desiredVersion)
 		return nil
+	} else if isStandbyCluster {
+		c.logger.Errorf("skipping major version upgrade for %s/%s standby cluster. Re-deploy standby cluster with the required Postgres version specified", c.Namespace, c.Name)
```
FxKu, line 176: shouldn't this be changed to Warnf maybe (even if it was also Error before)?

hughcapet, line 176:
```suggestion
		c.logger.Warnf("skipping major version upgrade for %s/%s standby cluster. Re-deploy standby cluster with the required Postgres version specified", c.Namespace, c.Name)
```

FxKu:
:+1:

hughcapet:
👍

PR #2830, opened 2024-12-31
Critical operation PDB

hughcapet:
Create the second PDB to cover Pods with a special "critical operation" label set.

This label is going to be assigned to all pg cluster's Pods by the Operator during a PG major version upgrade, by Patroni during a cluster/replica bootstrap. It can also be set manually or by any other automation tool.

PR #2844, opened 2025-01-17
Protect Pods from disruptions during upgrades

hughcapet:
Set "critical-operation" label to all postgres Pods to protect them with [the second PDB](https://github.com/zalando/postgres-operator/pull/2830) during the major version upgrade

### DIFF
--- a/pkg/cluster/majorversionupgrade.go
+++ b/pkg/cluster/majorversionupgrade.go
@@ -106,18 +106,17 @@ func (c *Cluster) removeFailuresAnnotation() error {
 	return nil
 }
 
-func (c *Cluster) criticalOperationLabel(pods []v1.Pod, remove bool) error {
+func (c *Cluster) criticalOperationLabel(pods []v1.Pod, active bool) error {
 	var action string
 	var metadataReq map[string]map[string]map[string]*string
 
-	if remove {
-		action = "remove"
-		metadataReq = map[string]map[string]map[string]*string{"metadata": {"labels": {"critical-operaton": nil}}}
-
-	} else {
+	if active {
 		action = "assign"
 		val := "true"
-		metadataReq = map[string]map[string]map[string]*string{"metadata": {"labels": {"critical-operaton": &val}}}
+		metadataReq = map[string]map[string]map[string]*string{"metadata": {"labels": {"critical-operation": &val}}}
+	} else {
+		action = "remove"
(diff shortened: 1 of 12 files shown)


pkg/cluster/majorversionupgrade.go
```diff
@@ -106,6 +106,34 @@ func (c *Cluster) removeFailuresAnnotation() error {
...
+	var metadataReq map[string]map[string]map[string]*string
+
+	if remove {
+		action = "remove"
+		metadataReq = map[string]map[string]map[string]*string{"metadata": {"labels": {"critical-operaton": nil}}}
+
+	} else {
+		action = "assign"
+		val := "true"
+		metadataReq = map[string]map[string]map[string]*string{"metadata": {"labels": {"critical-operaton": &val}}}
```
FxKu, line 120:
```suggestion
		metadataReq = map[string]map[string]map[string]*string{"metadata": {"labels": {"critical-operation": nil}}}

	} else {
		action = "assign"
		val := "true"
		metadataReq = map[string]map[string]map[string]*string{"metadata": {"labels": {"critical-operation": &val}}}
```

pkg/cluster/majorversionupgrade.go
```diff
@@ -106,6 +106,34 @@ func (c *Cluster) removeFailuresAnnotation() error {
 	return nil
 }
 
+func (c *Cluster) criticalOperationLabel(pods []v1.Pod, remove bool) error {
```
FxKu, line 109: when I only see the call I would expect `c.criticalOperationLabel(pods, true)` to add the label and `c.criticalOperationLabel(pods, false)` to remove it. One can think of also two wrapper functions like `addCriticalOperationLabel` and `removeCriticalOperationLabel` which call this func here.

FxKu:
Do we only set this during major version upgrades? I thought or main focus was to protect cluster bootstrap phase.

hughcapet:
@FxKu bootstrap can not be efficiency covered by operator. There will be a separate configuration in patroni to set a custom label

pkg/cluster/majorversionupgrade.go
hughcapet: wanted some code deduplication 🥲

FxKu:
:+1:

FxKu:
:+1:

PR #2849, opened 2025-01-21
Partially revert #2810

hughcapet:
Consider MaintenanceWindows only for pod migration

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -1187,7 +1187,7 @@ class EndToEndTestCase(unittest.TestCase):
         Test major version upgrade: with full upgrade, maintenance window, and annotation
         """
         def check_version():
-            p = k8s.patroni_rest("acid-upgrade-test-0", "")
+            p = k8s.patroni_rest("acid-upgrade-test-0", "") or {}
             version = p.get("server_version", 0) // 10000
             return version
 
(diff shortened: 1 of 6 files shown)


e2e/tests/test_e2e.py
```diff
@@ -1187,7 +1187,7 @@ def test_major_version_upgrade(self):
         Test major version upgrade: with full upgrade, maintenance window, and annotation
         """
         def check_version():
-            p = k8s.patroni_rest("acid-upgrade-test-0", "")
+            p = k8s.patroni_rest("acid-upgrade-test-0", "") or {}
```
FxKu, line 1190: or {}?
I don't understand python sometimes

hughcapet: it can return None and then `p.get("server_version", 0)` will break everything with exception

`or {}` returns empty dict instead of None

pkg/cluster/cluster.go
```diff
@@ -1761,35 +1766,39 @@ func (c *Cluster) GetSwitchoverSchedule() string {
 }
 
 // Switchover does a switchover (via Patroni) to a candidate pod
-func (c *Cluster) Switchover(curMaster *v1.Pod, candidate spec.NamespacedName) error {
+func (c *Cluster) Switchover(curMaster *v1.Pod, candidate spec.NamespacedName, inMaintWindow bool) error {
```
FxKu, line 1769:
```suggestion
func (c *Cluster) Switchover(curMaster *v1.Pod, candidate spec.NamespacedName, scheduled bool) error {
```

I think this makes it easier to read that the switchover should be scheduled or not because of the maintenanceWindow. Otherwise this reads to me as Switchover being called during maintenanceWindow = true or false. And interestingly, if the call happens during the window there's no need for schedule and `inMaintWindow` will be false. Counter-intuitive naming.

hughcapet:
👍

FxKu:
:+1:

PR #2863, opened 2025-02-12
fix creating secrets for rotation users

FxKu:
In #2657 a [bug](https://github.com/zalando/postgres-operator/commit/47efca33c9857b5ada5cade4051468170f8ff17a#diff-f4933755bd63c9e160a2068e8ccb4048e8df1b43f7f4544dcbc7ae2650a07478R975) was introduced which produces additional secrets for rotation users which are redundant to the original secret. Before syncing secrets we usually initialize the user list to get the desired state.

On UPDATE events we only needed to sync secrets in case the user setup changed. `initUsers` and `syncSecrets` were called in one if block. The behavior was changed to have to if blocks where secrets can be synced on annotations diff while skipping user initializing. Therefore, the current pgUser list is used which contains the rotation users instead of the original ones.

We better return to the previous behavior incl. the annotation diff check, even though we would probably always init users then because UPDATEs are usually tiggered from deployments changing the deployment-time annotation.

Another quick hack would be to not generate a secret for rotation users (when `Rotated` flag is true), but since rotation users replace their original counterparts in the pgUsers list, we would then not sync the corresponding secrets.

In a first step I extended the e2e test without changing any other code (quick hack only as a comment) to reproduce the bug.

### DIFF
--- a/e2e/tests/test_e2e.py
+++ b/e2e/tests/test_e2e.py
@@ -1814,6 +1814,7 @@ class EndToEndTestCase(unittest.TestCase):
         enable_password_rotation = {
             "data": {
                 "enable_password_rotation": "true",
+                "inherited_annotations": "environment",
                 "password_rotation_interval": "30",
                 "password_rotation_user_retention": "30",  # should be set to 60 
             },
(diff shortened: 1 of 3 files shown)


hughcapet:
👍

FxKu:
:+1:

PR #2912, opened 2025-05-16
Skip creation of OwnerReference if user is in a different namespace

### LINKED ISSUE #2910: Operator fails to create secret in separate namespace when using ownerReferences and preparedDatabases

- **Which image of the operator are you using?** e.g. ghcr.io/zalando/postgres-operator:v1.13.0 
  `ghcr.io/zalando/postgres-operator:v1.14.0`
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** [AWS K8s | GCP ... | Bare Metal K8s]
  `GCP`
- **Are you running Postgres Operator in production?** [yes | no]
  Not yet, currently setting up for initial deployment
- **Type of issue?** [Bug report, question, feature request, etc.]
  Bug report

Some general remarks when posting a bug report:
- Please, check the operator, pod (Patroni) and postgresql logs first. When copy-pasting many log lines please do it in a separate GitHub gist together with your Postgres CRD and configuration manifest.
- If you feel this issue might be more related to the [Spilo](https://github.com/zalando/spilo/issues) docker image or [Patroni](https://github.com/zalando/patroni/issues), consider opening issues in the respective repos.

When enabling ownerReferences, and using `preparedDatabases` in the cluster definition with `defaultUsers:  true`, the operator fails to create the secrets because it attempts to attach a cross-namespace ownerReference.

The problem seems to be that [this check](https://github.com/zalando/postgres-operator/blob/68c4b496365f02afb57b9066492dfa319120622a/pkg/cluster/k8sres.go#L1930) doesn't account for the fact that when using `preparedDatabases` you can set `secretNamespace` to a value that differs from the namespace the cluster is defined in.

We realize we can create users via `users` instead of relying on the default users, but this still seems to be a bug that should be addressed.

mortenlj:
Instead of doing a string compare on the username, check the actual namespace of the user to determine if an owner reference can be created.

I believe this should fix #2910

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -1928,7 +1928,7 @@ func (c *Cluster) generateSingleUserSecret(pgUser spec.PgUser) *v1.Secret {
 
 	// if secret lives in another namespace we cannot set ownerReferences
 	var ownerReferences []metav1.OwnerReference
-	if c.Config.OpConfig.EnableCrossNamespaceSecret && strings.Contains(username, ".") {
+	if c.Config.OpConfig.EnableCrossNamespaceSecret && c.Postgresql.ObjectMeta.Namespace != pgUser.Namespace {
 		ownerReferences = nil
 	} else {
 		ownerReferences = c.ownerReferences()


FxKu:
:+1:

idanovinda:
👍

PR #2234, opened 2023-02-22
Names to container ports for postgresql, patroni and operator

stacksegv:
This PR is about to get prometheus-operator fetch the metrics of the patroni metrics endpoint.
The prometheus operator CR for podmonitor definition of the endpoint only supports port names and no numbers.

So adding names for the container ports should IMHO not interfere with any component, but i surely may be corrected and there are no names for reasons. 
If there shall be any documentation to be extended, i'd go for it - i did not find a good place for it.

The DOCKER_CLI Variable makes it possible to use e.g. nerdctl for container build.

This is my first PR of all time - please be gracious :-D

PR #2934, opened 2025-07-14
Add names for containerPorts

bo0ts:


FxKu:
Duplicate to #2234. We are currently working on deprecating wal-e in Spilo. If this changes of the env variables causes a rolling update on every Postgres cluster, I will also include this change (I mean #2234).

Issue #2859, opened 2025-01-29
Postgres 1.14 logical operator default version is 1.13


arconsis-subash:
Currently I am using Postgres Operator version 1.14.0 and I leave the logical backup version to default so that I won't be interfering with the default version of it later when its upgraded in future. But the problem is,  the default version it takes is 1.13. If I specify the the version of logical backuip to 1.14.0 then it is working fine.

So the question is should I provide the version of logical backup

I am running it in k8s and it is only testing environment.

FxKu:
Hi. Are you using the helm chart? There it is already fixed: https://github.com/zalando/postgres-operator/pull/2829. The only difference between the version is the Postgres 17 support. You can always use the same tag for operator deployment and logical backup cron job. They are build with the same pipeline.

maxlaverse:
Just as a note: the fix is on master but a new Helm chart hasn't been released yet. You need to configure `logical_backup_docker_image` in the meantime, as in the PR from the comment above.

FxKu:
Closing this issue. See my comment in the [PR](https://github.com/zalando/postgres-operator/pull/2940#issuecomment-3669918920)

PR #2940, opened 2025-08-01
Fix: logical_backup_docker_image to v1.14.0 (#2859)

nikkytub:
### 📌 Summary

This PR fixes issue #2859 and updates the Helm chart package to include the correct version of the `logical_backup_docker_image`, aligning it with the operator version `v1.14.0`. This ensures consistency across deployments and avoids potential mismatches between the operator and its backup tooling.

### ✅ What’s Changed

- Re-packaged the Helm chart to reflect the updated `logical_backup_docker_image` version
- Committed the updated `postgres-operator-1.14.0.tgz` chart archive

### 🔍 Why It Matters

- Ensures that users deploying the operator via Helm get the correct backup image version
- Prevents compatibility issues between the operator and logical backup container
- Supports smoother upgrades and version alignment

### 🧪 Validation

- Verified that the chart includes the correct image reference
- Confirmed that the `.tgz` archive was properly updated and committed


mjtrangoni:
It is the same for v1.15.0 now. PTAL.

See the [manifests/operatorconfiguration.crd.yaml](https://github.com/zalando/postgres-operator/blob/master/manifests/operatorconfiguration.crd.yaml#L511) and the [](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/crds/operatorconfigurations.yaml#L513)

Is this something on purpose, related to any issue or just a bug in the release process?

Thanks!

FxKu:
@mjtrangoni thanks for pointing it out. I have also noticed the mistake and fixed it for the v1.15.1 release. I will publish the helm charts later.

Closing this PR then since we are currently not back-patching past releases.

PR #2941, opened 2025-08-04
fix(ui): add missing endpoint_url for a couple of s3 client initializations

rcmorano:
I just mimicked what's done in a few other client initializations and it fixed backup listing while using minio :)


flowftw:
We tested this fix with the latest version of the operator and it still works for both Minio and S3. @FxKu maybe you could have another look at it?

PR #2943, opened 2025-08-18
Ensure compatibility with Spilo after WAL-E removal (support old and new versions)

idanovinda:
Spilo https://github.com/zalando/spilo/pull/1143 removes WAL-E and introduces changes in environment variables.
This PR updates the Postgres Operator to:
- Support both the old Spilo versions (with WAL-E env vars) and the new versions (without WAL-E).
- Ensure smooth upgrades and backward compatibility.


idanovinda:
close, as Spilo supports backward compatibility, we don't need to change anything from operator part

PR #2949, opened 2025-08-28
DOC: Minikube has many drivers now, incl. Docker

remram44:
Minikube with Docker is good, it works with other VM providers too


jopadi:
👍

FxKu:
:+1:

Issue #2946, opened 2025-08-25
Operator deprecation warning: "v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice


monotek:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** ghcr.io/zalando/postgres-operator:v1.14.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** azure aks 1.33.1
- **Are you running Postgres Operator in production?**  yes
- **Type of issue?** Bug report

Seems the operator will have some problems with k8s 1.33+ as the Endpoints api is deprecated and replaced by EndpointSlice.

I get a lot of warnings in the operator logs like:

```
W0825 12:02:43.852245 1 warnings.go:70] v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice
```

FxKu:
Thanks for letting us know. The operator has adopt here for sure. Feel free to provide a PR. Otherwise we might fix it maybe beginning of next year.

FxKu:
Or maybe we tackle it a bit sooner. But likely we cannot use EndpointSlices and have to switch to using configmaps.

schmitch:
@FxKu there is now a new release with this, but I do wonder how do services work with the configmap logic?

in the past a service was created that was backed by endpoints. so I guess if the endpoints are removed it will just use the normal labelSelector is that correct?

Ref:

- https://github.com/zalando/postgres-operator/issues/340
- https://github.com/zalando/postgres-operator/issues/2565

Without Endpoints the Service needs some other way to find the correct service to route?

PR #2961, opened 2025-10-13
extend RBAC in prepatation to switch to configmap-based cluster management

FxKu:
In #2946 it was correctly pointed out that endpoint are marked for deprecation in K8s 1.33. However, it's successor EndpointSlices is not regarded an option for Patroni to manage the database clusters. Our only choice atm is to switch to configmap-based cluster management.

One has to follow these three steps:
1. Scale all clusters to 1 pod
2. Change `kubernetes_use_configmaps: true` in the config
3. Revert 1 and ideally remove endpoints manually.

The PR is only a preparation for future releases. For 1.15. we'd better not change the default behavior. But for 1.16 we should.


jopadi:
👍

FxKu:
:+1:

PR #2971, opened 2025-10-29
Consider container commands in Statefulset diffs

baarde:
#2449 added Command to sidecars.

So container's commands should be compared when deciding whether a Statefulset has been updated.


mikkeloscar:
Would be good to expand the test cases to cover this

mortenlj:
When a sidecar is defined in the OperatorConfiguration, all fields on the container may be specified. One commonly used would be `args`, it would probably be nice to compare those too.

Issue #2966, opened 2025-10-20
Sidecar without image specification causes all PostgreSQL pods to shutdown


ovnozdrach:
- **Which image of the operator are you using?** ghcr.io/zalando/postgres-operator:v1.14.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K8s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Bug report

It's possible to add a sidecar specification without providing any image (by mistake) for that additional container. 
This leads to a cascading failure where all cluster instances are shut down sequentially during sync intervals (30 minutes) until all pods are terminated, resulting in complete PostgreSQL unavailability.

New pods cannot start due to the Kubernetes error: "create Pod ... in StatefulSet ... failed error: Pod ... is invalid: spec.containers[2].image: Required value"

The operator logs show this -
```
{"cluster-name":"dev-mdb-mq-clusters/test-no-image","level":"info","msg":"performing rolling update","pkg":"cluster","time":"2025-10-20T09:10:42Z","worker":2}
{"cluster-name":"dev-mdb-mq-clusters/test-no-image","level":"info","msg":"there are 2 pods in the cluster to recreate","pkg":"cluster","time":"2025-10-20T09:10:42Z","worker":2}
{"cluster-name":"dev-mdb-mq-clusters/test-no-image","level":"debug","msg":"subscribing to pod \"dev-mdb-mq-clusters/test-no-image-1\"","pkg":"cluster","time":"2025-10-20T09:10:42Z","worker":2}
...
{"cluster-name":"dev-mdb-mq-clusters/test-no-image","level":"debug","msg":"unsubscribing from pod \"dev-mdb-mq-clusters/test-no-image-1\" events","pkg":"cluster","time":"2025-10-20T09:20:52Z","worker":2}
{"cluster-name":"dev-mdb-mq-clusters/test-no-image","level":"error","msg":"could not sync statefulsets: could not recreate pods: could not recreate replica pod \"dev-mdb-mq-clusters/test-no-image-1\": pod label wait timeout","pkg":"cluster","time":"2025-10-20T09:20:52Z","worker":2}
{"cluster-name":"dev-mdb-mq-clusters/test-no-image","level":"debug","msg":"syncing roles","pkg":"cluster","time":"2025-10-20T09:20:52Z","worker":2}
[...]

PR #2977, opened 2025-11-06
Fix Sidecar without image specification issue

ovnozdrach:
This is to address issue https://github.com/zalando/postgres-operator/issues/2966

It's possible to add a sidecar specification without providing its image.
This leads to a cascading failure where all cluster instances are shut down sequentially during sync intervals (30 minutes) until all pods are terminated, resulting in complete PostgreSQL unavailability

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
 # global owners
-*        @sdudoladov @Jan-M @FxKu @jopadi @idanovinda @hughcapet
+*        @sdudoladov @Jan-M @FxKu @jopadi @idanovinda @hughcapet @mikkeloscar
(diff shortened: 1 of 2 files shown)


mikkeloscar:
Likely is also mitigated via schema update from: https://github.com/zalando/postgres-operator/pull/3007 but good to have it covered here in code as well.

mikkeloscar:
:+1:

FxKu:
:+1:

ovnozdrach:
Hi @FxKu ! Looks like the change was mistakenely mapped to #2997 in the [Release notes](https://github.com/zalando/postgres-operator/releases/tag/v1.15.1) . Maybe it's worth to correct that?
`- prevent sidecars without image specification to not run into cascading failures (https://github.com/zalando/postgres-operator/pull/2997)`

PR #2983, opened 2025-11-12
Allow user env vars to override operator-generated ones

PavelZaytsev:
This change modifies appendEnvVars() to allow environment variables defined in the PostgreSQL CRD spec.env to override operator-generated environment variables (like SPILO_CONFIGURATION) instead of being silently ignored.

Previously, if an env var already existed in the list, user-provided values were skipped. Now, user values take precedence and replace the operator-generated ones.

This enables users to customize SPILO_CONFIGURATION and other operator-managed env vars through the CRD, which is useful for adding custom Patroni DCS configuration like ignore_slots.

## Testing

Tested with:
- PostgreSQL 17 cluster with custom `SPILO_CONFIGURATION`
- Verified that user-provided env vars properly override operator-generated ones
- Confirmed Patroni configuration contains both operator defaults and user customizations
- Validated that `ignore_slots` configuration works correctly with native PostgreSQL 17 slot synchronization

Example working configuration:
```
kind: postgresql
spec:
  env:
    - name: SPILO_CONFIGURATION
      value: |
        bootstrap:
          dcs:
            ignore_slots:
              - type: logical
```
Result: Patroni config includes both operator settings (ttl, loop_wait, etc.) and user's ignore_slots configuration.

### DIFF
--- a/pkg/cluster/cluster.go
+++ b/pkg/cluster/cluster.go
@@ -716,12 +716,12 @@ func compareSpiloConfiguration(configa, configb string) bool {
 	if err != nil {
 		return false
 	}
-	oa.Bootstrap.DCS = patroniDCS{}
 	err = json.Unmarshal([]byte(configb), &ob)
 	if err != nil {
 		return false
 	}
-	ob.Bootstrap.DCS = patroniDCS{}
+	// Compare the full configuration including DCS section
+	// This allows user-provided DCS settings (like ignore_slots) to trigger pod restarts
 	return reflect.DeepEqual(oa, ob)
 }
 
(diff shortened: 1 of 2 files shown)


mikkeloscar:
Please add a test case covering this change

PavelZaytsev:
@mikkeloscar I have added the test case

mikkeloscar:
@PavelZaytsev FYI: the tests are failing

PavelZaytsev:
After discussion, we're taking a simpler approach by adding ignore_slots as a native operator config option rather than a general env var override mechanism, looks like general override breaks the upgrade

Issue #2787, opened 2024-10-21
`allowedSourceRanges` should support ipv6


viceice:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** ghcr.io/zalando/postgres-operator:v1.13.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K8s
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** feature request

I've a cluster which supports dualstack networking and i like to allow ipv6 access to my postgresql clusters.

To get this working i need to set ipv6 ranges to `allowedSourceRanges` to the postgresql crd.

it should also be possible to set the `ipFamilyPolicy` and `ipFamilies` fields on the service opbjects.

Workaround: Create my own service

FxKu:
I welcome a PR that updates the [pattern](https://github.com/zalando/postgres-operator/blob/c206eb38a80853c7dbc989875565dad938cb82c6/manifests/postgresql.crd.yaml#L109) for `allowedSourceRanges` field to allow all the stuff you've mentioned @viceice :smiley:

ManjunathCV91:
Any plans to include this feature ?
I also want to install PostgreSQL in dual stack openshift cluster.

if any other workaround please suggest.

AlyHKafoury:
@FxKu I would love to work on this, is that ok ?

amk9978:
Any updates on this?
@AlyHKafoury 
@FxKu

AlyHKafoury:
I am interested to work on this, but got no answer if it is ok to start working on it

viceice:
I don't have time to work on this

raphaeltorquat0:
Hi! I've implemented IPv6 support for `allowedSourceRanges` and will be submitting a PR shortly.

**Changes:**
- Updated the regex pattern in all 3 CRD files to accept both IPv4 and IPv6 CIDR notation
- Updated documentation in `docs/reference/cluster_manifest.md`

**Tested formats:**
- IPv4: `192.168.1.0/24`, `0.0.0.0/0`, `127.0.0.1/32`
- IPv6: `fd01::/48`, `::1/128`, `::/0`, `2001:db8::/32`, `fe80::1/64`

All unit tests pass.

FxKu:
[...]

PR #2985, opened 2025-11-13
Extend validation pattern of allowedSourceRanges to allow for IPv6 ranges

exciler:
first part of #2787


mikkeloscar:
I'm not sure if the current e2e setup allows for testing this pattern in a meaningful way, but it would be good to have a test that shows this works for IPv6 and IPv4 as it's hard to reason about such a huge regex. If we can't do it as part of e2e, then at least let's add a unit test which shows the regex works for some cases of IPv4 and IPv6 ranges.

exciler:
Besides that I have manually tested this regex against some valid and invalid IPv4 and IPv6 addresses to, it would of course be very beneficial to have some sort of automated testing for this.

Unfortunately I am not a "Go" expert and not really sure where I should add those tests in this quite large project. Would you mind giving me a hint where to start?

mikkeloscar:
The only place the pattern is defined in go is in `pkg/apis/acid.zalan.do/v1/crds.go` so maybe it makes sense to add it as a `const` and then add a test file `crds_test.go` which has a test case validating it. That would be a unit test which would just show the pattern is correct.

For e2e test, even better, it would be something like adding a test case similar to this one: https://github.com/zalando/postgres-operator/blob/add9912f7b23b80f325739c4633af116e0ca752b/e2e/tests/test_e2e.py#L867-L919 but add `allowedSourceRanges` and check that it succeeds on a valid IPv6/IPv4 and fails on an invalid one. I can't 100% reason if this works in the e2e environment but I believe it does.

jopadi:
the PR will be worked on in this https://github.com/zalando/postgres-operator/pull/3082

Issue #983, opened 2020-05-19
Expose services over NodePort


k0da:
For migration purposes we must expose postgres outside of k8s cluster, our main exposure method is a NodePort. 

It would be nice if cluster CRD will allow to alter service type to a NodePort and optionally specify a value for nodeport number

nutellinoit:
I would also like this option!

Jan-M:
Sounds like a reasonable small PR to do. There is some code already around service creation and config so that would be the place to start. (compare lb vs non lb service)

k0da:
There is a WIP PR https://github.com/zalando/postgres-operator/pull/994 from me

arpitarathi:
Thanks. I am waiting for this option too

worldofgeese:
Can one simply update the ClusterIP service to be of NodePort type while we wait for the PR to approve?

worldofgeese:
I'll also take the risk of drawing @Jan-M's attention back to this one.

Vivida1:
Still no update on this?

darkfessing:
NodePort could be very helpful feature for migration purposes.

lucasl0st:
opened another PR for this: https://github.com/zalando/postgres-operator/pull/2986

PR #2986, opened 2025-11-14
feat: implement service type NodePort

lucasl0st:
This PR adds support for Kubernetes services of type `NodePort`.
I tried my best to keep the structure of the configuration and code the same as for the LoadBalancer service type.

If the configuration for the NodePort is enabled the LoadBalancer configuration is overwritten.
Specific ports to be used on the kubernetes node(s) can be specified, if left empty kubernetes picks a port based on the clusters configuration: https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport

Implements https://github.com/zalando/postgres-operator/issues/983
Probably supersedes https://github.com/zalando/postgres-operator/pull/994

Since this is my first PR at Zalando and this repository, please feel free to let me know if you would like me to make changes, if you want a specific test or if I need to bump a version somewhere. Thank you!


FxKu:
@lucasl0st sorry for the delay reviewing you nice contribution. This looks like the best attempt so far to get NodePort service type support in our operator. I've checked most changes briefly so far and it looks sound.

On general implementation detail we need to think about is what takes precedence when both NodePort and LoadBalancer service type are enabled from the manifest or config? I think this should create an LB, maybe with a hard-coded port if NodePort is specified in the manifest. In the pooler code, for example, I saw that you added NodePort configuration after the LB.

pkg/cluster/resources.go
```diff
@@ -324,7 +324,15 @@ func (c *Cluster) updateService(role PostgresRole, oldService *v1.Service, newSe
...
+			// Kubernetes rejects updates that change type away from LoadBalancer while
+			// loadBalancerSourceRanges is still set; clear it before updating
+			newService.Spec.LoadBalancerSourceRanges = nil
+			newService.ResourceVersion = oldService.ResourceVersion
+			if newServiceType == v1.ServiceTypeClusterIP {
+				newService.Spec.ClusterIP = oldService.Spec.ClusterIP
+			}
+		} else if newServiceType == v1.ServiceTypeClusterIP && newServiceType != oldServiceType {
 			newService.ResourceVersion = oldService.ResourceVersion
 			newService.Spec.ClusterIP = oldService.Spec.ClusterIP
```
FxKu:
```suggestion
		if newServiceType != oldServiceType && oldServiceType == v1.ServiceTypeLoadBalancer {
			// Kubernetes rejects updates that change type away from LoadBalancer while
			// loadBalancerSourceRanges is still set; clear it before updating
			newService.Spec.LoadBalancerSourceRanges = nil
			newService.ResourceVersion = oldService.ResourceVersion
			if newServiceType == v1.ServiceTypeClusterIP {
				newService.Spec.ClusterIP = oldService.Spec.ClusterIP
			}
```

lucasl0st:
> @lucasl0st sorry for the delay reviewing you nice contribution. This looks like the best attempt so far to get NodePort service type support in our operator. I've checked most changes briefly so far and it looks sound.
> 
> On general implementation detail we need to think about is what takes precedence when both NodePort and LoadBalancer service type are enabled from the manifest or config? I think this should create an LB, maybe with a hard-coded port if NodePort is specified in the manifest. In the pooler code, for example, I saw that you added NodePort configuration after the LB.

Hi there!
Sorry for the late reply, I have been trying to find some time to get back into it but haven't been able to.
I see that @jopadi took something over, feel free to do so, I hope you can continue here without me?
Otherwise, I would probably be able to take a look again in a couple of weeks, but I cant provide any guarantee

pkg/cluster/connection_pooler.go
```diff
@@ -570,6 +570,10 @@ func (c *Cluster) generateConnectionPoolerService(connectionPooler *ConnectionPo
 		c.configureLoadBalanceService(&serviceSpec, spec.AllowedSourceRanges)
 	}
 
+	if ok, port := c.shouldCreateNodePortForPoolerService(poolerRole, spec); ok {
+		c.configureNodePortService(&serviceSpec, port)
+	}
+
```
FxKu, line 574: this part should be moved before the LB check so we only call configureLoadBalanceService when shouldCreateNodePortForPoolerService has returned false

pkg/cluster/resources.go
FxKu:
```suggestion
		}
```
can be simplified further

FxKu:
👍

jopadi:
👍

PR #2995, opened 2025-11-19
fix switchover schedule tests

berler:
Previously the tests would fail depending on the local time zone and the time of day the test was being run (even if you set the machine to UTC it would fail if run within an hour of midnight; similarly the tests would fail any time the date component of the local time and UTC was different as the schedule would not match on the right day). Changes the tests to pass in a consistent specific time for `now` so that the tests no longer have different behavior based on when (or where) they are run.


FxKu:
@berler can you reflect Mikkel's comment. Then we can merge the PR.

Edit: Committed the suggestions myself via Gh.

FxKu:
:+1:

idanovinda:
👍

PR #3000, opened 2025-12-01
Modernize code generation

mikkeloscar:
This "modernizes" code generation setup to make it simpler to run and have less implicit dependencies. It does so in the following way:

* Use Go 1.24 tool feature to manage Go tool dependencies (no longer need for `tools.go` hack)
  * Drop the `make tools` directive as it will be implicitly working just when running go.
* Update `hack/update-codegen.sh` to not depend on a bash-script from the code-gen toolchain, instead run  with the pure go implementing such that the tooling can be fully managed in `go.mod`
* Update `k8s.io/code-generator` from `v0.25.9` to `v0.32.9` (This is the same version used for other k8s libraries, and required for the modernization to work).
* Regenerate k8s client code, this has a bunch of changes because of the `k8s.io/code-generator` version update.
  * Add `return` in error cases when updating status in `cluster.go` and `sync.go`. This is needed because the newer version of the generated code doesn't return `nil` for the object on error, so `if pgUpdatedStatus != nil` checks are no longer valid to detect if it was not set. The right thing is to do this check as part of the error checking which is done by `returning` there.
* Drop use of `go mod vendor` which is not needed with the use of `go tool`.
* Drop `make deps` target as it's not needed to be explicitly called.


mikkeloscar:
Closing in favor of https://github.com/zalando/postgres-operator/pull/3003 (from local branch)

PR #3005, opened 2025-12-04
Use UpdateStatus instead of patch

mikkeloscar:
This updates the `postgresql` status update logic to use `UpdateStatus` instead of Patch.

The Purpose of this change is to make it simpler to extend the status field and update it via modifying the struct and allow updates of multiple status fields at a time.

For now there is still only the field: `PostgresClusterStatus` so there is no functional change, but a follow up PR intend to add a new `ObservedGeneration` field.

This also fixes a test case which silently ignored updating the status of a resource which was not in the API already. It does so by registering the postgresql resource in the fake client used for testing.

### DIFF
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -493,13 +493,19 @@ spec:
                     type: string
                   standby_port:
                     type: string
-                oneOf:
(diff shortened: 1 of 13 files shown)


mikkeloscar:
:+1:

FxKu:
👍

mikkeloscar:
:+1:

PR #3007, opened 2025-12-04
Generate postgresql CRD from go structs

mikkeloscar:
Similar to  #3004 this introduces CRD (yaml) generation based on the go structs using [`controler-gen`](https://github.com/kubernetes-sigs/controller-tools)

This enables the generation of the `postgresql` CRD.

The PR has two commits:

* `Sort postgresql.crd.yaml` this just sorts the yaml keys similar to the CRD generation does, but doesn't change the yaml from what is in `master`. This is just done to make the diff cleaner
* `Generate postgresql CRD from go structs` this includes the new YAML generated from the go structs. Look at this commit to see how it's different from `master`.

As this goes from a "manually" defined schema to a generated one there are a few changes:

* The field `patroni.maximum_lag_on_failover` was changed from `float32` to `int64` as it was defined as `integer` in the schema. Related code was changed to cast to `float32`.
* Enum for tolerations are not generated from upstream types, not simple to fix without copying the whole upstream structure.
* Types of the structure `map[string][]string` allowed nullable of the `[]string`. This is not possible to set in kubebuilder so it's missing meaning you can't do `key: null` in those cases, which I anyway think is better to not allow.
* A few constraints are not possible to define via kubebuilder markers. Namely `oneOf` on the `standby` field and specifying `maintenanceWindows` field as a list of strings (as opposed to a list of objects). To handle this I have added a custom script `hack/adjust_postgresql_crd.sh` which injects the right settings via `sed` after the manifest is generated. This is intended to be temporary until those features are supported upstream by `controller-gen`.
* Change the in-process CRD which is used when the operator is enabled to create the CRD resource in the target cluster. Before it was manually defined as a Go struct. Now it's generated at runtime by embedding the generated yaml CRD in the go binary and parse it at runtime (this is in `pkg/apis/acid.zalan.do/v1/crds.go`). This avoids the need to manually adjust the schema in multiple places.

### TODO:

* [x] Fix types for `maintenanceWindows` (this uses a string in the yaml, but a go struct in code)
* [x] Enable full schema generation for big fields which was left out initially to make it simpler to view the diff:
  * [x] sidecars
  * [x] initContainers
  * [x] env
  * [x] `init_containers`
* [x] Handle `oneOf` on the `standby` field
* [x] Validate new schema against existing resources across Zalando clusters to ensure it will continue to work.

Checking the new schema across _all Zalando clusters_ I find only two that fail the validation:

```
cluster: db, resource: godfathers-barzini, namespace: default
Error from server (BadRequest): error when creating "STDIN": postgresql in version "v1" cannot be handled as a postgresql: strict decoding error: unknown field "metadata.clusterName"

cluster: stups-test, resource: zmon-entity-service-load-test-db-cluster, namespace: default
Error from server (BadRequest): error when creating "STDIN": postgresql in version "v1" cannot be handled as a postgresql: strict decoding error: unknown field "spec.sidecars[0].foo"
```

(The last one in `stups-test` was me validating that the check catches issues correctly)


mikkeloscar:
:+1:

FxKu:
:+1:

PR #3011, opened 2025-12-11
bump to v1.15.1

FxKu:
preparing the next release - I kept the 1.15.0 operator chart if somebody is already using it, but the UI chart is dropped, since there's no 1.15.0 ghcr image which the UI chart references.


mikkeloscar:
:+1:

FxKu:
:+1:

PR #3015, opened 2025-12-14
Fix #3014: deletion timestamp handling for clusters with finalizers

### LINKED ISSUE #3014: Delete of postgres resource is ignored when finalizer is used

- **Which image of the operator are you using?** e.g. ghcr.io/zalando/postgres-operator:v1.15.0
- **Type of issue?** Bug Report

It seems that the operator ignores changes to the DeletionTimestamp, only Spec and Annotations are compared, for Deletions DeletionTimestamp is set though, and Delete is never called, because it's actually not yet deleted:

https://github.com/zalando/postgres-operator/blob/master/pkg/controller/postgresql.go#L538

thoro:
When a Postgres cluster has a finalizer, deleting it sets a DeletionTimestamp but doesn't remove the object until the finalizer is cleared. The operator was not properly handling these DeletionTimestamp changes:

1. postgresqlUpdate() was filtering out events where only DeletionTimestamp changed (it only checked Spec and Annotations), causing the delete to never be processed.

2. EventUpdate case in processEvent() didn't check for DeletionTimestamp, so even if the event reached the processor, it would run Update() instead of Delete().

3. removeFinalizer() used a cached object with stale resourceVersion, causing "object has been modified" errors.

Fixes:
- Add explicit DeletionTimestamp check in postgresqlUpdate() to queue the event
- Add DeletionTimestamp check in EventUpdate to call Delete() when set
- Fetch latest object from API before removing finalizer to avoid conflicts


FxKu:
👍

jopadi:
👍

FxKu:
Thanks @thoro for the fix!

PR #3017, opened 2025-12-17
fix docker build for UI and bumped some outdated versions in docs and config

FxKu:
Currently UI build fails and gemini suggested to remove python3-distutils as a hotfix.

I have also found some outdated versions in docs and one in the operator config for logical backup image.


idanovinda:
👍

FxKu:
:+1:

PR #3018, opened 2025-12-18
Do not promote operator image when merge to master

idanovinda:
Current internal build will always promote operator image when merge to master. But this image won't be used until we successfully test it. So image promotion should only happen when smoke test is running successfully.


idanovinda:
👍

FxKu:
:+1:

PR #3019, opened 2025-12-18
Check cluster UID

mikkeloscar:
In situations where an existing Postgresql cluster is deleted and a new cluster, with the same name, is created, then the operator can get "confused" because it still holds the internal state of the previous cluster.

To avoid operating on the old state, check if the UID matches and mark the cluster as not found if it doesn't.


pkg/controller/postgresql.go
```diff
@@ -185,19 +185,25 @@ func (c *Controller) addCluster(lg *logrus.Entry, clusterName spec.NamespacedNam
...
 	} else {
 		clusterName = util.NameFromMeta(event.OldSpec.ObjectMeta)
+		clusterUID = event.OldSpec.UID
 	}
 	lg = lg.WithField("cluster-name", clusterName)
 
 	c.clustersMu.RLock()
 	cl, clusterFound := c.clusters[clusterName]
+	if clusterFound && cl.UID != clusterUID {
+		clusterFound = false
```
FxKu, line 205: I think this will lead to an error during deletion, see https://github.com/zalando/postgres-operator/blob/a3d17c12d27ba63b1497fc993ef566759b828cf9/pkg/controller/postgresql.go#L297

Must be the reason why e2e tests always fail with the cluster deletion test (if they even get there)

PR #3026, opened 2026-01-06
ImagePullSecret for the pooler image

mbuken:
Previously, it was possible to configure a custom pooler image, but images from private registries could not be pulled. This change allows specifying imagePullSecrets for the connection pooler.

I tested this locally in minikube by applying a postgresql CRD with connectionPooler.


NilsGriebner:
Bump. This would be crucial for our deployment.

PR #3030, opened 2026-01-10
add annotation to ignore resources thresholds

FxKu:
Similar to ignoring defined instance limits with a configurable annotation key, this PR introduces the same logic for bypassing globally defined maximums for requests and minimums for limits. An empty key disables the feature. The value of the annotation needs to be a `"true"` string to be effective.


hughcapet:
👍

FxKu:
:+1:

PR #3031, opened 2026-01-13
Fix serving CRD at runtime

mikkeloscar:
This fixes a bug where creating/updating the CRD from the operator doesn't work since #3007

The issue is fixed in the following way

* Import and use the correct yaml library `sigs.k8s.io/yaml` for unmarshalling the crd YAML at runtime. (The main problem)
* Define the temporary CRD yaml location needed for build in the `Makefile` this makes it less likely to not be generated no matter how `make` is run during build.
* Use `Update` instead of `Patch` for updating the CRD at runtime. Not critical but a bit nicer way to manage it.

### DIFF
--- a/manifests/postgresql.crd.yaml
+++ b/manifests/postgresql.crd.yaml
@@ -3523,11 +3523,11 @@ spec:
                     type: object
                   version:
                     enum:
-                    - 13
-                    - 14
-                    - 15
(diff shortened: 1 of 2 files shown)


FxKu:
:+1:

mikkeloscar:
:+1:

PR #3037, opened 2026-01-28
bugfix: fix image naming to use it correctly in e2e test

idanovinda:
This PR fixes an issue where the postgres operator image was always pulled using the `:latest` tag, even when a specific tag was intended.

Previously, docker would resolve the image name in a way that caused it to fall back to `docker.io/library/postgres-operator:$tag`, resulting in latest released image with tag `:latest` being used implicitly. This meant tests always run agains an unintended image version.

This script will no longer automatically pull the `latest` image if a local tag is missing. Users must ensure their local build is up to date.

Changes:
* Explicitly constructs full image with "ghcr.io/zalando" as a prefix
* Failed fast if the tagged image cannot be pulled
* Ensure tests run agains the exact image build from the current code
* [small unrelated change] update run.sh to be mac os friendly

This guarantees we test the current code, not whatever happens to be in the `:latest`.


idanovinda:
👍

hughcapet:
👍

PR #3038, opened 2026-01-28
add support for global maintenance windows

FxKu:
Allow users to define global default maintenance windows for all clusters which are applied when set and unless there are no windows defined in the cluster manifest.


jopadi:
👍

FxKu:
👍

PR #3043, opened 2026-02-11
Remove hardcoded VersionMap from majorversionupgrade

jorsol:
This removes the VersionMap Map of version numbers from the majorversionupgrade.go file with simple math calculation and conversion to int, this makes it more flexible to Postgres versions change.

This is actually blocking us the use of Postgres 18 since while the CRD enum versions can be updated, the Go code would need recompilation.

This is a non-breaking-change, so I hope that this can be released in a 1.15.2 patch version very soon.


FxKu:
👍

hughcapet:
👍

FxKu:
@jorsol thanks for your contribution!

PR #3044, opened 2026-02-17
do not reset secrets of standby clusters

FxKu:
This PR changes two things:

Standby clusters are not only determined by the existence of the `standby` manifest section, but also searching for the `STANDBY_WALE_S3_PREFIX` env var.

The second change deals with password rotation. The current implementation resets secrets if rotation is not allowed. But, when using standby clusters we typically copy the secrets first, before creating the standby via the operator. We require credentials to remain stable. After all, it's a read only cluster, and the operator could not alter password in the database anyway after resetting the username and password in the secret.

I did some refactoring and moved all password rotation bits into a new function so I can easily bypass it when dealing with a standby cluster.


hughcapet:
👍

FxKu:
👍

PR #3045, opened 2026-02-17
try to set infra roles also if one fails

FxKu:
atm one could use the old and new way to define two infrastructure roles. However if fetching one fails, no roles will be set. This PR will add an extra check to set infraRoles when at least one was found


FxKu:
👍

hughcapet:
👍

PR #3047, opened 2026-02-18
Checkin CRD to make go get work

mikkeloscar:
This checks in the file: `pkg/apis/acid.zalan.do/v1/postgresql.crd.yaml` which is generated during the build.

This is required to be able to import the project as a library, otherwise you get an error like this:

```
../../go/pkg/mod/github.com/zalando/postgres-operator@v1.15.2-0.20260218073817-cffa0ee63cce/pkg/apis/acid.zalan.do/v1/crds.go:980:12: pattern postgresql.crd.yaml: no matching files found
```

This was introduced in https://github.com/zalando/postgres-operator/pull/3007 as a way to automatically serve the CRD

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,3 @@ mocks
 ui/.npm/
 
 .DS_Store
-
-# temp build files
-pkg/apis/acid.zalan.do/v1/postgresql.crd.yaml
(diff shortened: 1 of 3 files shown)


FxKu:
👍

mikkeloscar:
:+1:

PR #3054, opened 2026-03-17
Do not set aws-load-balancer-connection-idle-timeout by default

hughcapet:
As this annotation has no effect on NLBs, remove the default annotation setting. Users can still define it (or the NLB alternative) via custom_service_annotations


hughcapet:
👍

FxKu:
👍

PR #3058, opened 2026-03-23
fix: invalid switchover scheduling with default maintenance windows

idanovinda:
Fixes an issue where the operator schedules a switchover with an invalid (zero) timestamp when default maintenance window is applied, causing Patroni to reject the request. 

Ensures a valid scheduled_at value is always set, allowing switchovers to proceed correctly during node decommissioning.

Additional fix: major upgrade e2e test was failing because of wrong architecture being loaded to kind.


FxKu:
👍

hughcapet:
👍

PR #3065, opened 2026-04-06
Avoid rotating pods for PGVERSION change outside of maintenance window

hughcapet:
Do reset of `PGVERSION` during Sync() before the setSpec() call, like in Update() to avoid disruptions caused by `PGVERSION` environment variable mismatch outside of the maintenance windows


FxKu:
👍

hughcapet:
👍

PR #3073, opened 2026-04-15
make pooler labels configurable and change defaults

FxKu:
All resources of one cluster should belong to one application label, not an extra application. On the other hand, there should not be two pods queryable with `application: spilo, spilo-role: master`. Therefore, the role identifier for pooler pods must be changed, too.

As this is a breaking change, we introduce new config options to be able to continue using the old way of labelling, meaning:
- `connection_pooler_labels: application:db-connection-pooler`
- `connection_pooler_role_label: spilo-role`

The new defaults used for these two config options are:
- `connection_pooler_labels: component:db-connection-pooler`
- `connection_pooler_role_label: pooler-role`


FxKu:
Closing this PR as we decided to better not give user the chance to configure the same application name as the database pods because it would break the cluster.

PR #3074, opened 2026-04-16
feature toggle for using maintenance windows

FxKu:
Give operator admins an option to turn on the maintenance windows feature. For example, in staging environments it might not be very useful to delay or block K8s maintenance tasks.

### DIFF
--- a/pkg/cluster/util.go
+++ b/pkg/cluster/util.go
@@ -675,7 +675,9 @@ func isStandbyCluster(spec *acidv1.PostgresSpec) bool {
 }
 
 func (c *Cluster) isInMaintenanceWindow(specMaintenanceWindows []acidv1.MaintenanceWindow) bool {
-	if len(specMaintenanceWindows) == 0 && len(c.OpConfig.MaintenanceWindows) == 0 {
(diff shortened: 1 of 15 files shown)


pkg/cluster/util.go
```diff
@@ -675,7 +675,9 @@ func isStandbyCluster(spec *acidv1.PostgresSpec) bool {
 }
 
 func (c *Cluster) isInMaintenanceWindow(specMaintenanceWindows []acidv1.MaintenanceWindow) bool {
-	if len(specMaintenanceWindows) == 0 && len(c.OpConfig.MaintenanceWindows) == 0 {
+	ignoreMaintenanceWindows := c.OpConfig.EnableMaintenanceWindows == nil || !*c.OpConfig.EnableMaintenanceWindows
```
hughcapet: c.OpConfig.EnableMaintenanceWindows == nil should not it mean "not ignore" if default is to enable?

FxKu: good catch. In this case nil should be treated as true. Changed to != nil &&

hughcapet:
👍

FxKu:
👍

PR #3077, opened 2026-04-17
build multi-arch pooler image

FxKu:
and stop maintaining internal repo.

Our internal pooler was using a few hacks to
- allow connections to admin DB by any use with a defined prefix
- ignoring errors when SET commands are used

We still need to allow our monitoring user served from an infrastructure role secret to be able to connect to the pgbouncer admin DB, but as a read-only `stats_user`. In the passed we only copied a hardcoded auth file into the Docker image. But since the list of infra users can be vary and we don't want to hard code credentials into env variables I'm proposing the following architecture:

1. Set an INFRASTRUCTURE_ROLES env var on the pooler pod with the user names which is passed to the pgbouncer.ini template to be set as `stats_users`.
2. Create a secret per pooler deployment which stores a dynamically generated userlist.txt file.
3. Mount this file into the pooler container for pgBouncer to allow connections from the pooler user as well as infrastructure roles. Could also be extended to more users. But the normal flow would see users connecting via the auth_query (and not auth_file) to the actual database.


pooler/Dockerfile
```diff
@@ -0,0 +1,53 @@
+ARG BASE_IMAGE=alpine:3.19
```
mikkeloscar, line 1: Why not use latest stable version 3.22?

idanovinda:
👍

FxKu:
👍

PR #3083, opened 2026-04-26
Add support for loadBalancerClass in LoadBalancer services

### LINKED ISSUE #2600: Support setting `loadBalancerClass` for loadbalancer services

- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.11.0 (via helm chart)
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K3s

We use different kinds of loadbalancers on our cluster, so we need to set `spec.loadBalancerClass`[^1], so the right loadbalancer provider can set the appropriate values. Currently we can only use annotations via `replicaServiceAnnotations` which issn't enough.
Maybe a generic `spec` field makes sense, so all kind of service overrides can be used.

[^1]: https://kubernetes.io/docs/concepts/services-networking/service/#load-balancer-class

raphaeltorquat0:
## What this PR does

This adds a new configuration option `load_balancer_class` that allows users to specify which load balancer provider should handle the service when running multiple providers in the same cluster.

Fixes #2600

## Changes

- Added `load_balancer_class` field to `LoadBalancerConfiguration` struct
- Added config mapping in `importConfigurationFromCRD()`
- Modified `configureLoadBalanceService()` to set `spec.loadBalancerClass` when configured
- Updated CRD schema with new field
- Updated default configuration manifest
- Added documentation

## Example usage

```yaml
apiVersion: acid.zalan.do/v1
kind: OperatorConfiguration
metadata:
  name: postgres-operator
configuration:
  load_balancer:
    enable_master_load_balancer: true
    load_balancer_class: "service.k8s.aws/nlb"
```

## Type of change

- [x] New feature (non-breaking change which adds functionality)

Note: parts of this implementation were developed with AI assistance.

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -719,6 +719,13 @@ In the CRD-based configuration they are grouped under the `load_balancer` key.
   defines external traffic policy for load
   balancers. Allowed values are `Cluster` (default) and `Local`.
 
+* **load_balancer_class**
+  specifies the load balancer class to use for LoadBalancer services. This is
+  useful when running multiple load balancer providers in the same cluster,
+  allowing you to select which provider should handle the service. For example,
+  `service.k8s.aws/nlb` for AWS Network Load Balancer. When not set, the
+  cluster's default load balancer class is used.
+
 * **master_dns_name_format**
   defines the DNS name string template for the master load balancer cluster.
   The default is `{cluster}.{namespace}.{hostedzone}`, where `{cluster}` is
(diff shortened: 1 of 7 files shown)


raphaeltorquat0:
Hi! Could a maintainer please add the appropriate label to this PR? I believe `minor` would be suitable as this adds a new non-breaking feature. Thank you!

FxKu:
To use NLBs we usually do it via serviceAnnotations - begs the question if we really need another option. Plus, how would you get two LBs? Is the old LB with other class not dropped when you switch the class?

```
  serviceAnnotations:
    service.beta.kubernetes.io/aws-load-balancer-type: nlb
    service.beta.kubernetes.io/aws-load-balancer-internal: "true"
 ```

PR #3087, opened 2026-04-26
feat: allow overriding external-dns hostname annotation

### LINKED ISSUE #2591: Question: Why is it impossible to override external-dns.alpha.kubernetes.io/hostname overriding on cluster-level?
-     registry.opensource.zalan.do/acid/postgres-operator:v1.10.1
-     Kubernetes [Bare Metal K8s]
-     Are you running Postgres Operator in production? [yes]
-     Type of issue? [question]

Hi. I want to use external-dns integration, but unfortunately master_dns_name_format is not suitable for me.  I searched the documentation and found the line: 

> external-dns.alpha.kubernetes.io/hostname with the value defined by the operator configs master_dns_name_format and replica_dns_name_format. This value can't be overwritten. If any changing in its value is needed, it MUST be done changing the DNS format operator config parameters

Why can't overriding be done at masterServiceAnnotations/replicaServiceAnnotations level?

raphaeltorquat0:
## Summary
Allows users to override the `external-dns.alpha.kubernetes.io/hostname` annotation by specifying it in `serviceAnnotations`, `masterServiceAnnotations`, or `replicaServiceAnnotations` in the cluster manifest.

Fixes #2591

## Changes
- **pkg/cluster/k8sres.go**: Modified `generateServiceAnnotations()` to check if the DNS annotation already exists before setting it
- **pkg/cluster/cluster_test.go**: Updated tests to verify user-defined annotations are preserved (changed test case names and expectations)
- **docs/administrator.md**: Updated documentation to reflect that the annotation can now be overwritten

## Before
The operator always overwrote the `external-dns.alpha.kubernetes.io/hostname` annotation with the value from `master_dns_name_format` or `replica_dns_name_format`, ignoring any user-defined value.

## After
If the user has already defined the `external-dns.alpha.kubernetes.io/hostname` annotation in their service annotations, the operator will preserve that value instead of overwriting it.

## Test plan
- [x] Updated existing tests to verify new behavior
- [x] All unit tests pass (`go test ./...`)

---

**Disclosure**: I used AI assistance (Claude) to help implement this change under my review and testing.

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -896,8 +896,9 @@ its services:
 
 - `external-dns.alpha.kubernetes.io/hostname` with the value defined by the
   operator configs `master_dns_name_format` and `replica_dns_name_format`.
-  This value can't be overwritten. If any changing in its value is needed, it
-  MUST be done changing the DNS format operator config parameters; and
+  This value can be overwritten by specifying the annotation in
+  `serviceAnnotations`, `masterServiceAnnotations`, or `replicaServiceAnnotations`
+  in the cluster manifest; and
 
 There are multiple options to specify service annotations that will be merged
 with each other and override in the following order (where latter take
--- a/pkg/cluster/cluster_test.go
+++ b/pkg/cluster/cluster_test.go
@@ -767,16 +767,16 @@ func TestServiceAnnotations(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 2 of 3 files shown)


raphaeltorquat0:
Hi @FxKu, could you please add the `minor` label so the CI checks can proceed? Thanks!

PR #3088, opened 2026-04-27
update standby check in pooler code

FxKu:
fixing the last left over check for standby clusters which ignored the possibility of creating a standby by setting an env variable


FxKu:
👍

idanovinda:
👍

PR #3089, opened 2026-04-29
Provide liveness probe for CRD and config

FxKu:
rebased #1972

With following additions:
- config map option not possible because it does not support structs
- re-run auto-generation of CRD (which also updated the topologySpreadConstraint validation)


FxKu:
👍

jopadi:
👍

PR #3093, opened 2026-05-08
Volume mount length of pooler users

FxKu:
Turns out we should be careful with naming of extra pooler resources like the new secret. Risk for exceeding the name length of 63 characters is was higher with a `-userlist-volume` suffix.

This PR also updates the CRD in the charts folder


FxKu:
👍

jopadi:
👍

PR #3104, opened 2026-06-07
feat(pooler): operator-generated PgBouncer config

mkm29:
# feat(pooler): operator-generated PgBouncer config

## Summary
Adds an **opt-in** mode where the operator renders `pgbouncer.ini` itself and manages it as an operator-owned ConfigMap, instead of relying on the pooler image to generate its own config from environment variables.

## Motivation
The default Spilo/PgBouncer image renders `pgbouncer.ini` at startup, but some images (e.g. the **Chainguard PgBouncer** distroless image) ship no such entrypoint. Supporting them previously required a static ConfigMap plus a mutating webhook. This PR moves that responsibility into the operator and removes the workaround.

## What changes
When `connection_pooler_generate_config: true`, for each pooler role the operator:
- renders `pgbouncer.ini` (pool mode, auth type/query, TLS, sizes) and writes it to a `<pooler>-config` ConfigMap (create/sync/delete in the pooler lifecycle);
- mounts it at `config_path` via `subPath`, and overrides the container `command`/`args` to point PgBouncer at it;
- stamps a config checksum annotation on the pod template so config changes trigger a rollout.

Default is `false` — **no behavior change** for existing clusters.

## New configuration (ConfigMap + `OperatorConfiguration` CRD)
| Parameter | Default | Notes |
|---|---|---|
| `connection_pooler_generate_config` | `false` | master switch |
| `connection_pooler_command` | _(unset)_ | keeps image entrypoint when empty |
| `connection_pooler_args` | `["/etc/pgbouncer/pgbouncer.ini"]` | applied only when generating |
| `connection_pooler_auth_type` | `scram-sha-256` | written into `pgbouncer.ini` |
| `connection_pooler_config_path` | `/etc/pgbouncer/pgbouncer.ini` | mount path |

## Tests
New/extended coverage in `config_test.go`, `connection_pooler_test.go`, and `pgbouncer_config_test.go` (ini rendering incl. TLS, checksum stability, ConfigMap generation, pod-template on/off, and sync create + update-on-drift). Verify by running the tests:

```bash
go generate ./...
go test ./pkg/util/config/ -run TestConnectionPoolerGenerateConfigDefaults -v
go test ./pkg/cluster/ -run 'TestConnectionPoolerSizes|TestPoolerEnv|ConnectionPooler|TestGeneratePgBouncerIni' -v
```

## Also updated
Helm `values.yaml`, `operator_parameters.md`, both `OperatorConfiguration` CRDs, and the default operator-configuration manifest.


cross-du:
Any chance of this PR and #3105 getting a review soon? It would be a big help for my use case.

FxKu:
The PR is too big for the next release which I plan to drop this week, but I like to review and include it after that.

charts/postgres-operator/crds/operatorconfigurations.yaml
```diff
@@ -942,6 +942,25 @@ spec:
...
+                    type: array
+                    items:
+                      type: string
+                  connection_pooler_auth_type:
+                    type: string
+                    default: "scram-sha-256"
+                  connection_pooler_config_path:
+                    type: string
+                    default: "/etc/pgbouncer/pgbouncer.ini"
+              patroni:
```
FxKu, line 963: this diff does not look correct. You only need to add new fields to the pkg.api. package and the CRDs will be generated when running `make` for example

docs/pgbouncer-generated-config.md
```diff
@@ -0,0 +1,133 @@
+<h1>Operator-generated PgBouncer config (Helm)</h1>
```
FxKu, line 1: this isn't specific to helm chart, right? Not sure about a new single docs file. Can you move this section to the administrator.md (global config) and add one for users.md, too (local config)

manifests/acid.zalan.do_operatorconfigurations.yaml
```diff
@@ -0,0 +1,986 @@
+---
```
FxKu, line 1: pls remove this file from the PR

manifests/acid.zalan.do_postgresteams.yaml
```diff
@@ -0,0 +1,84 @@
+---
```
FxKu, line 1: pls remove this file from the PR

manifests/operatorconfiguration.crd.yaml
```diff
@@ -942,6 +942,25 @@ spec:
...
+                    type: array
+                    items:
+                      type: string
+                  connection_pooler_auth_type:
+                    type: string
+                    default: "scram-sha-256"
+                  connection_pooler_config_path:
+                    type: string
+                    default: "/etc/pgbouncer/pgbouncer.ini"
+              patroni:
```
FxKu, line 963: again, looks like the new fields got inserted incorrectly

manifests/postgresql.crd.yaml
```diff
@@ -2850,17 +2847,6 @@ spec:
...
-                - required:
-                  - s3_wal_path
-                - required:
-                  - gs_wal_path
-                - required:
-                  - standby_host
-                not:
-                  required:
-                  - s3_wal_path
-                  - gs_wal_path
```
FxKu, line 2863: do not remove fields in postgresql CRD

manifests/postgresql.crd.yaml-e
```diff
@@ -0,0 +1,4824 @@
+---
```
FxKu, line 1: pls remove this file from the PR

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -347,6 +347,15 @@ type ConnectionPoolerConfiguration struct {
...
 	DefaultMemoryLimit string `json:"connection_pooler_default_memory_limit,omitempty"`
+	// +kubebuilder:default=false
+	GenerateConfig *bool    `json:"connection_pooler_generate_config,omitempty"`
+	Command        []string `json:"connection_pooler_command,omitempty"`
+	// +kubebuilder:default={"/etc/pgbouncer/pgbouncer.ini"}
+	Args []string `json:"connection_pooler_args,omitempty"`
+	// +kubebuilder:default="scram-sha-256"
+	AuthType string `json:"connection_pooler_auth_type,omitempty"`
+	// +kubebuilder:default="/etc/pgbouncer/pgbouncer.ini"
+	ConfigPath string `json:"connection_pooler_config_path,omitempty"`
```
FxKu, line 358: I think, it makes sense to also add these fields to the postgresql_type, too. At least we would need this for some teams to still use pgbouncer with md5 while globally we want scram to be the default

pkg/cluster/connection_pooler.go
```diff
@@ -441,6 +467,33 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
...
+				ConfigMap: &v1.ConfigMapVolumeSource{
+					LocalObjectReference: v1.LocalObjectReference{
+						Name: configVolumeName,
+					},
+				},
+			},
+		})
+		volumeMounts = append(volumeMounts, v1.VolumeMount{
+			Name:      configVolumeName,
+			MountPath: c.OpConfig.ConnectionPooler.ConfigPath,
```
FxKu, line 487: not 100% sure that this variable is always set at this point

pkg/cluster/connection_pooler.go
```diff
@@ -441,6 +467,33 @@ func (c *Cluster) generateConnectionPoolerPodTemplate(role PostgresRole) (
...
+			SubPath:   pgBouncerConfigFileName,
+			ReadOnly:  true,
+		})
+		if len(c.OpConfig.ConnectionPooler.Command) > 0 {
+			poolerContainer.Command = c.OpConfig.ConnectionPooler.Command
+		}
+		poolerContainer.Args = c.OpConfig.ConnectionPooler.Args
```
FxKu, line 494: same here. Can Args be empty? Should we always set it or use an if like with Command

pkg/cluster/pgbouncer_config.go
```diff
@@ -0,0 +1,192 @@
...
+
+// connectionPoolerPodAnnotations returns the pooler pod annotations, adding the
+// config checksum when generated config is enabled so config changes roll pods.
+func (c *Cluster) connectionPoolerPodAnnotations(role PostgresRole) (map[string]string, error) {
+	annotations := c.annotationsSet(c.generatePodAnnotations(&c.Spec))
+	if c.OpConfig.ConnectionPooler.GenerateConfig {
+		checksum, err := c.connectionPoolerConfigChecksum(role)
+		if err != nil {
+			return nil, err
+		}
```
FxKu, line 185: when we maybe just raise a warning when the checksum annotation is missing? I would not break the cluster creation just because an annotation of a pooler could not get generated

mkdocs.yml
```diff
@@ -8,6 +8,7 @@ nav:
   - Postgres Operator UI: 'operator-ui.md'
   - Admin guide: 'administrator.md'
   - User guide: 'user.md'
+  - PgBouncer generated config: 'pgbouncer-generated-config.md'
```
FxKu, line 11: like I wrote above I would not create a dedicated docs file for this config option now but rather include it in admin and user docs

PR #3105, opened 2026-06-08
feat(pooler): make connection pooler security contexts configurable

mkm29:
## Description

Allows operators to override the connection pooler's pod-level and container-level securityContext via the
OperatorConfiguration CRD. Previously these were hardcoded, which prevented running the pooler under hardened images (different pgbouncer UID/GID) or satisfying restricted Pod Security Standards. For example, the Iron Bank image (`registry1.dso.mil/ironbank/opensource/zalando/pgbouncer:1.25.1`) runs as `uid=997(pgbouncer) gid=997(pgbouncer) groups=997(pgbouncer)`.

## Changes

- New config fields `connection_pooler_pod_security_context` and `connection_pooler_security_context` on the
`OperatorConfiguration` CRD (CRD-only; excluded from the ConfigMap decoder, mirroring liveness_probe).
- `connection_pooler.go` now builds the pod/container security contexts via two helpers that use a configured context as
a deep-copied base and fill in defaults only for unset fields.
- Generated deepcopy, CRD schema (`crds.go` + `operatorconfigurations.yaml`), and chart `values.yaml` documentation updated.

## Backward compatibility

Fully backward compatible. When unset, the operator preserves its historical defaults:
- Pod: `RunAsUser`/`RunAsGroup` 100/101 + spilo FSGroup
- Container: `allowPrivilegeEscalation: false`

TestConnectionPoolerDefaultSecurityContext asserts byte-identical output when no override is configured.

## Testing

- New unit tests: default behavior, pod-context override, container-context override
- go build ./..., go vet, and pkg/cluster / pkg/util/config / pkg/controller test suites all pass


mkm29:
I was wondering what the status of this PR is?

PR #3106, opened 2026-06-10
drop deprecated fields from Postgresql CRD

FxKu:
time to finally remove the old load balancer flags after it was marked as deprecated for years. We will temporarily use the useLoadBalancer field to created an internal NLB which we use to migrate clusters between different K8s infras


pkg/cluster/k8sres.go
```diff
@@ -2067,6 +2067,54 @@ func (c *Cluster) generateService(role PostgresRole, spec *acidv1.PostgresSpec)
 	return service
 }
 
+func (c *Cluster) generateMigrationService() *v1.Service {
+	serviceSpec := v1.ServiceSpec{
+		ExternalTrafficPolicy: v1.ServiceExternalTrafficPolicyType(c.OpConfig.ExternalTrafficPolicy),
+		Ports:                 []v1.ServicePort{{Name: "postgresql", Port: pgPort, TargetPort: intstr.IntOrString{IntVal: pgPort}}},
+		Selector:              c.roleLabelsSet(false, Master),
```
mikkeloscar: Does this correctly work during a switchover? My understanding was that partroni is modifying the main service endpoints directly instead of relying on a selector and thus if we rely on a selector here the behavior is potentially slightly different?

pkg/cluster/sync.go
```diff
@@ -456,6 +461,46 @@ func (c *Cluster) syncEndpoint(role PostgresRole) error {
 	return nil
 }
 
+func (c *Cluster) syncMigrationService() error {
```
mikkeloscar: It would make sense to already pass a context here and use that in the relevant calls inside the function.

mikkeloscar:
As a general comment, the migration service is very specific to Zalando EKS migration and Zalando Clusters, it's not reasuable by anyone from opensource and it needs to stay until the migration is complete and possible longer if we make migration a more general feature.

We have to understand and accept this overhead. The alternative solution would be to have the spilo-migrator create the service as part of the migration and bind it to the postgresql resource, that way the Zalando specific migration stuff would be located in one place outside of the postgres-operator.

pkg/cluster/sync.go
FxKu: I think in all our API calls we only use context.TODO() everywhere

pkg/cluster/k8sres.go
FxKu: we tested switchovers and it worked. General fyi, we already run the operator with services that have selectors where Patroni alters configmaps instead of endpoints and it works fine, too

FxKu:
The context of this PR has switched to only drop four fields already marked as deprecated from the Postgresql CRD.

idanovinda (approved):

FxKu:
👍

PR #3108, opened 2026-06-11
Update to Go 1.26.4, build runners and go.mod depedencies

FxKu:
## Golang

- update golang to 1.26.4 and remove old compatibility GO111MODULE=on flags
- new compiler errors popped up with newer go (container loop, Event logging)

## Build process

- define GOARCH in Dockerfile to avoid build issues for ARM
- call go clean -cache -modcache during build
- update to latest github actions and docker runner versions

## Dependencies

- reflect changes in deepcopy-gen
- reflect changes in go-client/cache.FIFO by adding a store next to it because it still offers List(), ListKeys() and GetByKey() functions that we formerly called on FIFO queues and now got removed
- moved to AWS SDK v2 which requires using context objects in AWS API calls, and also changes some variables from int64 to int32


jopadi:
👍

FxKu:
👍

PR #3109, opened 2026-06-12
re-generate CRDs to latest state

FxKu:
build produces a diverged CRD which is why unit test step is failing - despite all test finishing


jopadi:
👍

FxKu:
👍

PR #3110, opened 2026-06-12
update new source ranges validation for kubebuilder

FxKu:
another inconsistency that slipped through and produces a diff between build and code


FxKu:
👍

nctram:
👍

PR #3112, opened 2026-06-16
fix(sync): prevent rolling update when unexpected pods are down

ovnozdrach:
### Description
This is to completely address issue https://github.com/zalando/postgres-operator/issues/2966

### Problem
It's possible to introduce an invalid configuration into a sidecar (e.g., negative container port). Previously, this would trigger a cascading failure. The operator would try to recreate pods one by one during each sync interval (30 minutes), but because the spec was invalid, each new pod would not start. This process would continue until the entire cluster was terminated, resulting in a complete outage.

### Solution
This change introduces a pre-flight check before starting any rolling update. The operator now ensures that the total number of pods is stable and that any pod that is not running is already part of the planned update. If an unexpected pod is down or missing, the operation is postponed, preventing the cascading deletion of healthy instances. This logic is compatible with and enhances the stability checks from #3051.

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -745,6 +745,45 @@ func (c *Cluster) syncStatefulSet() error {
 	// statefulset or those that got their configuration from the outdated statefulset)
 	if len(podsToRecreate) > 0 {
 		if isSafeToRecreatePods {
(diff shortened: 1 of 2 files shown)


ovnozdrach:
@FxKu , hello! can you review this please?

ovnozdrach:
@FxKu , any comment please?

PR #3115, opened 2026-06-24
feat: tag EBS volumes from PostgreSQL CR annotations

tcondeixa:
## Summary

- Adds `ebs_volume_tags_from_annotations` config option mapping CR annotation keys to EBS volume tag names

## Configuration

```yaml
ebs_volume_tags_from_annotations: "application:myorg.org/owning-application,team:myorg.org/team"
```

Test plan

- [ ] Configure ebs_volume_tags_from_annotations in operator config
- [ ] Create a PostgreSQL CR with the mapped annotations
- [ ] Verify EBS volumes are tagged with the correct values after a sync cycle
- [ ] Update annotation value and verify the EBS tag is updated on the next sync
- [ ] Verify no CreateTags calls are made when tags already match

Co-Authored-By: Claude Sonnet 4.6
Signed-off-by: tcondeixa <tscondeixa@gmail.com>


FxKu:
We already have the option inherited_annotations which allows for defining keys which will be inherited from CR. Could we not reuse it?

tcondeixa:
> We already have the option inherited_annotations which allows for defining keys which will be inherited from CR. Could we not reuse it?

The main use-case it to tag the EBS volume with the application tag (optionally the component)
I still need to test and validate but this is a different use-case. From what I understood `inherited_annotations` are for propagate annotations to the pods (maybe other kubernetes resources), this one is about to add tags to the EBS volumes, but I can adjust the name of the config to something else if preferred.
I'm not sure if the owning annotation is mandatory, but we can also just use a `ebs_volume_inherited_labels` and just copy the `application` and `component` to the EBS volume tags.

tcondeixa:
@FxKu I forked this repo and created a PR from there, do I need to create a PR directly to have the cdp pipeline running and an container image to test? I don't think I have permissions in this repo to create a PR directly.

FxKu:
@tcondeixa I don't know why the CDP pipeline was triggered. For external contribution there's always a button for me to approve the pipeline run - or I can retrigger a blocked pipeline in CDP. But in this case here it does not even show up in CDP.

PR #3117, opened 2026-06-25
update docs about CRD generation and copy paste to chart dir, too

FxKu:
Our docs needed a refreshment of how new config options should be added to the codebase. Contributors should re-generate the CRDs which is currently not happening in external PRs. Running make now copies the CRDs to the chart director to to avoid copy-pasting all the time. The CRD diff in the charts dir already show that this is often forgotten.

The `app.kubernetes.io/name` used by the helm chart files is also included into the CRD generation.


mikkeloscar:
:+1:

FxKu:
👍

PR #3119, opened 2026-06-25
Preserve sync quorum in primary PDB without master selector

annielzy:
## Problem

Clusters with `synchronous_mode_strict: true` need master + 1 in-sync replica to accept writes. Setting `pdb_master_label_selector: false` keeps `minAvailable: 1`, which allows up to N-1 simultaneous evictions on an N-node cluster, enough to drop below sync quorum and make the cluster read-only during rolling node drains.

## Change

When `pdb_master_label_selector` is false and `synchronous_mode_strict` is true, the primary PDB now sets:

`minAvailable: synchronous_node_count + 1`

This preserves the write quorum of primary plus required synchronous replicas during voluntary disruptions such as rolling node drains.

Default behavior is unchanged when `pdb_master_label_selector` is enabled/unset or `synchronous_mode_strict `is false.


FxKu:
👍

idanovinda:
👍

FxKu:
👍

PR #3120, opened 2026-06-26
fix array validation for global maintenance window and protected role names

FxKu:
After observing the following error in the e2e test run:
```
stdout: b'', stderr: b'The CustomResourceDefinition "operatorconfigurations.acid.zalan.do" is invalid: spec.validation.openAPIV3Schema.properties[configuration].properties[maintenance_windows].items: Required value: must be specified\n'
```

I figured out that kubebuilder was not generating the `items` section in the CRD validation. This is probably happening because the OpConfig CRD is cross referencing the maintenance window struct from the postgresql CRD and somehow it cannot fully interpret it.

I tried different validation methods but in the end the solution is much simpler. The data type only needs to change to []string - like in the [Config struct](https://github.com/zalando/postgres-operator/blob/master/pkg/util/config/config.go#L183). At the moment the OpConfig setting is [converted](https://github.com/zalando/postgres-operator/blob/master/pkg/controller/operator_config.go#L55) to a string array anyway, meaning we can also simplify the code in some places by changing the type.

#3102 also introduced a bug to set an invalid default value for protected_role_names. 

```
properties[teams_api].properties[protected_role_names].default: Invalid value: "string":  in body must be of type array: "string"
```

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -706,11 +706,11 @@ spec:
                     type: integer
                 type: object
               maintenance_windows:
+                items:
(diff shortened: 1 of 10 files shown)


mikkeloscar:
:+1:

idanovinda:
👍

FxKu:
👍

PR #3121, opened 2026-06-26
switch to metav1.Duration for timeout options

FxKu:
With #3102 moving to an autogenerated OperatorConfiguration and setting default values for timeout field we produced an invalid CRD, because the used time.Duration type requires int64. But we expect numbers with interval letters hence the extra marshalling code.

This could be avoided by using the native metav1.Duration type which turns all timeout fields to type string.


mikkeloscar:
:+1:

FxKu:
👍

PR #3123, opened 2026-06-29
feat: EBS Volumes tags inherit from labels 

tcondeixa:
## Summary

This is another approach to solve the same problem as https://github.com/zalando/postgres-operator/pull/3124

- Adds `ebs_tags_inherit_labels ` config option to propagate labels to EBS volume tags
- Labels from the `ebs_tags_inherit_labels` that are not in the PostgreSQL CR and are in the EBS tags are removed

NOTE: This requires to add `ec2:CreateTags` and `ec2:DeleteTags` to the operator IAM role policy

## Configuration

```yaml
ebs_tags_inherit_labels: "application,team"
```

Test plan

- [x] Configure ebs_tags_inherit_labels in operator config
- [x] Create a PostgreSQL CR with the inherit labels
- [x] Verify EBS volumes are tagged with the correct values after a sync cycle
- [x] Update label value and verify the EBS tag is updated on the next sync
- [x] Verify no CreateTags calls are made when tags already match
- [x] Remove one of the inherit labels from the PostgreSQL CR
- [x] Verify EBS volume don't have the removed label

NOTE: This PR was written in part with the assistance of generative AI


tcondeixa:
As an option we can make the config more generic with the possibility to extend it later (e.g. aws_resources_inherit_labels), so the labels can be extended to other aws resources.

FxKu:
why can we not re-use the keys from `inherited_labels` config option to tag volumes. Is there a specific reason we want to have an extra options for volume tagging?

tcondeixa:
> why can we not re-use the keys from `inherited_labels` config option to tag volumes. Is there a specific reason we want to have an extra options for volume tagging?

The issue is we cannot add the application to the `inherited_labels` and we would need it in the AWS resources (EBS volume and LBs). I also tried with a new config item to be opt-in for OS, and make it more generic than just magically convert our internal annotation to the application in the AWS resources.

pkg/util/teams/teams_test.go
```diff
@@ -1,243 +1,243 @@
...
+		if resp != nil {
+			t.Errorf("response expected to be nil")
+			continue
+		}
+
+		if err.Error() != tc.err.Error() {
+			t.Errorf("expected error: %v, got: %v", tc.err, err)
+		}
+	}
+}
```
FxKu, line 243: why is there a diff on teams_test.go? Can you remove it from this PR?

tcondeixa:
I will go for a few weeks on vacations but I will go back to this PR after. Meanwhile please provide feedback on the approach @mikkeloscar. thanks

PR #3124, opened 2026-06-29
feat: EBS Volume tags from annotations with mapping

tcondeixa:
## Summary

This is another approach to solve the same problem as https://github.com/zalando/postgres-operator/pull/3123, but the other approach seems cleaner and more generic.

- Adds `ebs_volume_tags_from_annotations` config option mapping CR annotation keys to EBS volume tag names

## Configuration

```yaml
ebs_volume_tags_from_annotations: "application:myorg.org/owning-application,team:myorg.org/team"
```

Test plan

- [ ] Configure ebs_volume_tags_from_annotations in operator config
- [ ] Create a PostgreSQL CR with the mapped annotations
- [ ] Verify EBS volumes are tagged with the correct values after a sync cycle
- [ ] Update annotation value and verify the EBS tag is updated on the next sync
- [ ] Verify no CreateTags calls are made when tags already match

NOTE: This PR was written in part with the assistance of generative AI


FxKu:
same comment like in #3123: we do have the `inherited_annotations` options already. We can just use it for tagging. But maybe there's indeed a reason that somebody wants control annotation of K8s resources and tags on EBS volumes ...

PR #3126, opened 2026-06-30
experiment: adjusts to migrate from application to app.kubernetes.io/managed-by label

tcondeixa:
This is an experiment to allow to migrate from the `application` to `app.kubernetes.io/managed-by` label.

Phase 1: cluster_labels: application:spilo,app.kubernetes.io/managed-by:spilo
- needs pod relabeling before replace ( current PR)
- if relabeling partially fails → we abort before deleting the old StatefulSet → pods still match the old application:spilo selector → still fully managed → no split-brain risk
- worst case: nothing changes, operator logs a warning, you retry

Phase 2: cluster_labels: app.kubernetes.io/managed-by:spilo
- Existing selector: {application:spilo, app.kubernetes.io/managed-by:spilo, cluster-name:x}
- Desired selector: {app.kubernetes.io/managed-by:spilo, cluster-name:x}
- MapContains(existing, desired) → true → already works with existing code, no fix needed
- Pods already carry owner:spilo from Phase 1 → new StatefulSet adopts them immediately
- No relabeling step, no risk window

So two-phase concentrates the risk only in Phase 1, and Phase 1's risk is bounded because the old selector still exists on pods if anything goes wrong. Phase 2 is completely safe with zero code changes.

The code change needed is only in Phase 1: relabel pods with the new labels before doing the cascade=orphan replace, and abort if any pod fails to relabel.


pkg/cluster/resources.go
```diff
@@ -211,6 +212,61 @@ func (c *Cluster) updateStatefulSet(newStatefulSet *appsv1.StatefulSet) error {
 }
 
 // replaceStatefulSet deletes an old StatefulSet and creates the new using spec in the PostgreSQL CRD.
+func (c *Cluster) relabelPodsForSelector(oldSelector, newSelector map[string]string) error {
```
FxKu, line 215: comment needs to be moved to matching function

PR #3128, opened 2026-07-13
feat (operator): add support for IRSA (aws resources access)

tcondeixa:
## Changes

Add `irsa_role_arn` config field to enable IRSA (IAM Roles for Service Accounts) as an alternative to kube2iam.

Setting `irsa_role_arn` to a full IAM role ARN annotates the pod ServiceAccount with `eks.amazonaws.com/role-arn`, which causes the EKS mutating webhook to inject an OIDC web identity token volume into new pods. The AWS SDK credential chain prefers this token over the EC2 metadata endpoint (kube2iam), so IRSA takes effect immediately on the next pod rotation.

Zero-downtime migration: running pods are not patched in-place — the kube2iam annotation (`iam.amazonaws.com/role`) drains naturally as pods rotate. Both `kube_iam_role` and `irsa_role_arn` can be set simultaneously during migration. Once all pods have rotated, `kube_iam_role` can be removed. The operator logs migration progress on every reconcile.

Fix `date` to use POSIX format and auto-detect `gsed` when available, so `make` works on macOS.

## Test
- [x] Update operator without `irsa_role_arn` set — existing DBs unaffected, no SA annotation changes
- [x] Update operator with `irsa_role_arn` set and `kube_iam_role` kept — SA annotated immediately, new pods get IRSA token volume, existing pods keep kube2iam until rotated
- [x] Update operator with only `irsa_role_arn` (single deploy, no `kube_iam_role`) — one rotation produces clean pod with IRSA only
- [x] Create new DB with `irsa_role_arn` set — pod starts with IRSA from day one
- [x] Operator logs "IRSA migration complete" after all pods have rotated


FxKu:
can you move syncPodServiceAccount into the sync.go file for now so we have it all in one place? In general, it's a good idea to split the code into single files per K8s resource type, but lets do that in a separate PR.

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
```diff
@@ -252,6 +252,7 @@ type AWSGCPConfiguration struct {
 	WALAZStorageAccount          string `json:"wal_az_storage_account,omitempty"`
 	LogS3Bucket                  string `json:"log_s3_bucket,omitempty"`
 	KubeIAMRole                  string `json:"kube_iam_role,omitempty"`
+	IrsaRoleARN                  string `json:"irsa_role_arn,omitempty"`
```
mikkeloscar:
```suggestion
	IRSARoleARN                  string `json:"irsa_role_arn,omitempty"`
```

Go idiomatic to capitalize initialisms or acronyms: https://google.github.io/styleguide/go/decisions.html#initialisms

Makefile
```diff
@@ -29,16 +29,19 @@ PKG := `go list ./... | grep -v /vendor/`
...
 else
 	DOCKERFILE = Dockerfile
 endif
 
+
 ifeq ($(FRESH),1)
   DEBUG_FRESH=$(shell date +"%H-%M-%S")
 endif
 
+SED := $(shell command -v gsed 2>/dev/null || command -v sed)
```
mikkeloscar, line 43: This and the other SED fix really should be a different PR

pkg/cluster/sync_service_account.go
```diff
@@ -0,0 +1,79 @@
...
+			changed = true
+		}
+	}
+
+	if changed {
+		patchData, err := json.Marshal(patch)
+		if err != nil {
+			return fmt.Errorf("could not marshal service account patch: %v", err)
+		}
+		if _, err = c.KubeClient.ServiceAccounts(c.Namespace).Patch(context.TODO(), sa.Name, types.StrategicMergePatchType, patchData, metav1.PatchOptions{}); err != nil {
```
mikkeloscar: It would be simpler to just use a Get/Update pattern instead of constructing a patch. The difference is the call to Update will fail _only_ if the service account was modified in the meantime, what is unlikely to happen and fine to fail on if it was.

Makefile
tcondeixa: ok, I would remove from this PR and create a separated one.
I will also move the fix on the POSIX format for the date to this separated PR.

pkg/cluster/sync_service_account.go
tcondeixa: I just followed the existing pattern that patch all resource to avoid deviation, but let me know if Update is the preferred method here @mikkeloscar

pkg/apis/acid.zalan.do/v1/operator_configuration_type.go
tcondeixa: https://github.com/zalando/postgres-operator/pull/3128/commits/2b6b0747d55d890943285491f839ba5019989d2a

tcondeixa:
> move sa sync code to existing sync.go file to be all together

sure, I moved it to the existing sync.go file in https://github.com/zalando/postgres-operator/pull/3128/commits/d150b923731410efe916c219e0d97469aba48b47

pkg/cluster/sync_service_account.go
mikkeloscar: I would recommend we start introducing the Update pattern even if other parts don't do it yet.

If @FxKu disagree we can also leave it as is.

FxKu:
The new option still needs to be documented: https://github.com/zalando/postgres-operator/blob/master/docs/reference/operator_parameters.md#aws-or-gcp-interaction

And added to the [values.yaml](https://github.com/zalando/postgres-operator/blob/master/charts/postgres-operator/values.yaml#L343) file of the helm chart.

And the [example config CRD](https://github.com/zalando/postgres-operator/blob/master/manifests/postgresql-operator-default-configuration.yaml#L166) as well as https://github.com/zalando/postgres-operator/blob/master/manifests/configmap.yaml

tcondeixa:
I addressed all the feedback and I updated all the markdown files (even the administrator.md), so please check again @mikkeloscar @FxKu

mikkeloscar:
@tcondeixa build is failing

tcondeixa:
> @tcondeixa build is failing

sorry, it was a network issues with vendor the module. After a retry in the pipeline the build is working.
I see the github actions failed for the operator-e2e-test, I will check if it's something I introduced or if we have flaky tests.

mikkeloscar:
:+1:

FxKu:
👍

PR #3130, opened 2026-07-14
fix: makefile and scripts to also run on macOS

tcondeixa:
adjust makefile and scripts to also run on macos:

- use POSIX date format command to also work on macos
- using SED that try to use gsed (gnu-seed) for macos


mikkeloscar:
:+1:

FxKu:
👍

PR #3135, opened 2026-07-21
add env var for spilo with irsa

tcondeixa:
### Problem detected on smoke tests

Standby and clone clusters fail to bootstrap when the operator is configured with IRSA (irsa_role_arn). Patroni exits with PatroniFatalException: Failed to bootstrap cluster after wal-g repeatedly fails with NoCredentialProviders.

Root cause: The EKS admission webhook injects AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE into the pod environment (bare, unprefixed). Spilo's configure_spilo.py writes these into the main envdir (/run/etc/wal-e.d/env/) correctly, but for standby and clone it looks for the prefixed versions (STANDBY_AWS_ROLE_ARN, CLONE_AWS_ROLE_ARN) which are never set. wal-g runs under envdir which replaces the entire environment with only the files in the target directory — so without the prefixed vars being written there, wal-g has no credentials.

### Proposed Fix

In generateCloneEnvironment() and generateStandbyEnvironment() in pkg/cluster/k8sres.go, when IRSARoleARN is configured, add the prefixed IRSA env vars to the pod spec:

- CLONE_AWS_ROLE_ARN / CLONE_AWS_WEB_IDENTITY_TOKEN_FILE
- STANDBY_AWS_ROLE_ARN / STANDBY_AWS_WEB_IDENTITY_TOKEN_FILE

Spilo then finds these in the pod environment and writes them into the respective envdirs, giving wal-g valid IRSA credentials for S3 access.

This is the PR created for spilo: https://github.com/zalando/spilo/pull/1206

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -2214,6 +2214,11 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
 		}
 	}
 
+	if c.OpConfig.IRSARoleARN != "" {
+		result = append(result, v1.EnvVar{Name: "CLONE_AWS_ROLE_ARN", Value: c.OpConfig.IRSARoleARN})
+		result = append(result, v1.EnvVar{Name: "CLONE_AWS_WEB_IDENTITY_TOKEN_FILE", Value: "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"})
+	}
+
 	return result
 }
 
@@ -2257,6 +2262,11 @@ func (c *Cluster) generateStandbyEnvironment(description *acidv1.StandbyDescript
 		})
 		result = append(result, v1.EnvVar{Name: "STANDBY_METHOD", Value: "STANDBY_WITH_WALE"})
 		result = append(result, v1.EnvVar{Name: "STANDBY_WAL_BUCKET_SCOPE_PREFIX", Value: ""})
+	}
+
+	if c.OpConfig.IRSARoleARN != "" {
+		result = append(result, v1.EnvVar{Name: "STANDBY_AWS_ROLE_ARN", Value: c.OpConfig.IRSARoleARN})
+		result = append(result, v1.EnvVar{Name: "STANDBY_AWS_WEB_IDENTITY_TOKEN_FILE", Value: "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"})
 	}
 
 	return result


tcondeixa:
I was using AI to investigate the issue with the smoke tests, @FxKu does this change make sense? I see the env vars are copied for spilo.

pkg/cluster/k8sres.go
```diff
@@ -2259,6 +2264,11 @@ func (c *Cluster) generateStandbyEnvironment(description *acidv1.StandbyDescript
 		result = append(result, v1.EnvVar{Name: "STANDBY_WAL_BUCKET_SCOPE_PREFIX", Value: ""})
 	}
 
+	if c.OpConfig.IRSARoleARN != "" {
+		result = append(result, v1.EnvVar{Name: "STANDBY_AWS_ROLE_ARN", Value: c.OpConfig.IRSARoleARN})
+		result = append(result, v1.EnvVar{Name: "STANDBY_AWS_WEB_IDENTITY_TOKEN_FILE", Value: "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"})
```
tcondeixa, line 2269: this might need to be dynamic, not sure if legacy cluster (non eks) use a different path

tcondeixa: @mikkeloscar do you know if this is the same path for both eks and legacy? is there a better way instead of hardcode?

tcondeixa: would the approach with $() work? and is it safe since the env variables might be injected in a later time?
```
result = append(result, v1.EnvVar{Name: "CLONE_AWS_ROLE_ARN", Value: "$(AWS_ROLE_ARN)"})
result = append(result, v1.EnvVar{Name: "CLONE_AWS_WEB_IDENTITY_TOKEN_FILE", Value: "$(AWS_WEB_IDENTITY_TOKEN_FILE)"})
```

mikkeloscar: I think this is something spilo should be responsible for doing. E.g. the standard envars are set by IRSA on the pod and then spilo does the needed envar translation for the subprocesses that it runs.

tcondeixa: The main issue is that in the current approach the operator is the one always responsible to provide the correct env variable, so this will be a change in the behaviour. Another option would be to just do like the Instance Profile and unconditional add them in case they exists.

tcondeixa: @FxKu can you please provide feedback here? Can you also give me permissions to create a branch in spilo repository? I can push the changes and we can even try the smoke tests with that version of spilo. thanks

mikkeloscar: > The main issue is that in the current approach the operator is the one always responsible to provide the correct env variable

But only to some extend IMO. It should set the _standard_ IRSA envars, in this case, and then spilo, which runs custom defined background processes, should make sure to pass this correctly. If spilo changes to do it differently, postgres-operator would not need to be adapted.

tcondeixa: https://github.com/zalando/spilo/pull/1206

tcondeixa:
I'm closing this since this was solved in spilo directly in https://github.com/zalando/spilo/pull/1206

tcondeixa:
This can be temporary merged if that helps, but spilo already provides support

FxKu:
👍

tcondeixa:
👍

Issue #2548, opened 2024-02-21
Pod Priority Class for configConnectionPooler is missing


antonmatsiuk:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?**: registry.opensource.zalan.do/acid/postgres-operator:v1.10.1
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?: Kubernetes on AKS, EKS
- **Are you running Postgres Operator in production?**: yes
- **Type of issue?**: feature request 

It is not possible to specify `priorityClassName` for pooler Deployment under `configConnectionPooler` in [values.yaml](https://github.com/zalando/postgres-operator/blob/e34f19be01c1c8c0eb045c9b74726e70034fb844/charts/postgres-operator/values.yaml#L423) Pooler deployment is created without priority class assigned and gets kicked out on production environments with priority classes.

antonmatsiuk:
any update on this?

Issue #2496, opened 2023-12-15
[Feature Request] Option to choose a PriorityClassName for the pooler pods


capuche2412:
- **Which image of the operator are you using?** registry.opensource.zalan.do/acid/postgres-operator:v1.9.0
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Kubenetes on cloud (GKE) and metal
- **Are you running Postgres Operator in production?** yes
- **Type of issue?** Feature request

Hello,

We utilize multiple postgres-operator instances for various purposes and criticality levels (e.g., production, pre-production, testing) on the same cluster and, in some cases, in the same namespace. We would like to set a different priority class for the poolers across different releases of the postgres-operators and postgres clusters.

Is it possible to add the `priorityClassName` in the pooler configuration ?

I am working on a PR to present a possible solution.

PR #3136, opened 2026-07-22
feat(pooler): support priorityClassName for connection pooler pods

trentontrees:
Add priorityClassName to the connectionPooler cluster spec and a connection_pooler_priority_class_name operator config default. Cluster spec overrides the operator default; unset means no priority class. Applies to both master and replica pooler deployments and is reconciled on config drift.


trentontrees:
Addresses Issues:
* https://github.com/zalando/postgres-operator/issues/2496
* https://github.com/zalando/postgres-operator/issues/2548

PR #3137, opened 2026-07-23
update docker build commands in Makefiles to use buildx

FxKu:
docker build is already deprecated so we switch to docker buildx like in the GitHub actions. This needs to be reflected in the docs, too.

I ran into some memory issues locally and got the recommendation to remove go clean -cache -modcache commands again, as it would try to allocate more memory.

This PR also does a small update to the postgresql CRD in the charts directory to make unit tests work. Was an oversight of a previous merge.


idanovinda:
👍

FxKu:
👍

PR #3139, opened 2026-07-24
change volume specs to int32 and define max iops and throughput values

FxKu:
#3108 was not consistent with using the int type matching AWS' API so we started casting int64 to int32 which could lead to potential overflow exceptions. The volume struct was now changed to use int32 everywhere, which should still provide us with a realistic range for size, iops and throughput. We also define max iops and throughput values in the CRD validation.


sdudoladov:
:+1:

FxKu:
👍

PR #3024, opened 2026-01-02
Create critical-op PDB on-demand to avoid false monitoring alerts

a-thomas-22:
The critical-op PodDisruptionBudget was previously created permanently, but its selector (critical-operation=true) matched no pods during normal operation. This caused false alerts in monitoring systems like [kube-prometheus-stack ](https://github.com/prometheus-community/helm-charts/blob/4f332b4147dcb1890bd2ca0ea9c33ae09dd6bfdc/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/kubernetes-apps.yaml#L568-L601) because the PDB expected healthy pods but none matched.

Changes:
- Modified syncCriticalOpPodDisruptionBudget to check if any pods have the critical-operation label before creating/keeping the PDB
- PDB is now created on-demand when pods are labeled (e.g., during major version upgrades) and deleted when labels are removed
- Updated majorVersionUpgrade to explicitly create/delete the PDB around the critical operation for immediate protection
- Removed automatic critical-op PDB creation from initial cluster setup
- Added test to verify on-demand PDB creation and deletion behavior

The explicit PDB creation in majorVersionUpgrade ensures immediate protection before the critical operation starts. The sync function serves as a safety net for edge cases like bootstrap (where Patroni applies labels) or operator restarts during critical operations.

Fixes #3020

PR #3141, opened 2026-07-24
Use maxUnavailable for the critical-op PDB to stop idle alert noise

### LINKED ISSUE #3020: Newly introduced `critical-op` PDB causes tons of alerts with kube-prometheus-stack
The newly introduced `postgres-<instance>-critical-op-pdb` is created with the maximum of healthy instances and a selector not matching any instances in normal operation. When using the kube-prometheus-stack in its default configuration, this raises an alarm because the PDB does not have any healthy pods:

> PDB does not have enough healthy pods.
> PDB /keycloak/postgres-keycloak-pg-critical-op-pdb expects 3 more healthy pods. The desired number of healthy pods has not been met for at least 15m.

Alert-Rule is defined here: [`templates/prometheus/rules-1.14/kubernetes-apps.yaml#L568-L601`](https://github.com/prometheus-community/helm-charts/blob/4f332b4147dcb1890bd2ca0ea9c33ae09dd6bfdc/charts/kube-prometheus-stack/templates/prometheus/rules-1.14/kubernetes-apps.yaml#L568-L601)

Though after the first "what's going on here?!?" this is not critical, it causes a lot of noise in alerting and mitigation is either

- to disable PDB for postgres-operator
- disable PDB monitoring
- introduce silences, muting the `postgres-<instance>-critical-op-pdb` alerts

which is not ideal, as in all cases parts are missing or alerts are not seen when it's important to see them.

A better solution would be, for example, to create those PDBs only on demand and not to let them stick around all the time. That way the alerts would be meaningful and not noisy while retaining the use case those PDBs fulfill.

unkhz:
## Problem description

Since v1.15.0 the operator creates `postgres-<cluster>-critical-op-pdb` with `minAvailable: <numberOfInstances>` and a selector matching pods labeled `critical-operation=true`. During normal operation no pods carry that label, so the PDB's status is permanently `currentHealthy: 0, desiredHealthy: N` - an unsatisfiable budget. Kubernetes behavior is unaffected (no matched pods, nothing to block), but monitoring stacks alert on exactly this condition: kube-prometheus-stack's `KubePdbNotEnoughHealthyPods` fires forever on every idle cluster.

This PR expresses the same protection as `maxUnavailable: 0` instead:

- During critical operations, behavior is identical - zero voluntary evictions permitted among labeled pods.
- While idle, the budget is satisfied (`desiredHealthy` = matched - 0 = 0) - no alert.
- When PDBs are disabled (`enable_pod_disruption_budget: false`) or instances are scaled to zero, the budget relaxes to `maxUnavailable: "100%"`, replacing the previous `minAvailable: 0`.

Compared to #3024 (on-demand create/delete around operations, later abandoned), this keeps the PDB persistent - no lifecycle choreography and no leaked unsatisfiable PDB if the operator restarts mid-operation - and only changes the budget field.

**Behavioral note:** no user-facing configuration changes; what changes is the field used inside the operator-generated (and operator-owned, reconciled) PDB resource. Blocking semantics are equivalent for every labeling state; any external tooling reading `spec.minAvailable` off this specific PDB would observe the new shape after upgrading.

## Testing

Beyond the updated unit tests, the patched image was deployed to a real 7-node k3s v1.36.2 cluster running an operator-managed postgres: with PDBs enabled, the critical-op PDB is created with `maxUnavailable: 0` and reports a satisfied status while idle (`desiredHealthy: 0` - the #3020 alert can no longer fire); labeling a pod `critical-operation=true` engages the budget (`allowedDisruptions: 0`) and eviction of the labeled pod is denied; the budget remains satisfied during the operation, so no false alert fires mid-operation either.

*Disclosure: this patch was authored with AI assistance (Claude, via Claude Code), reviewed and submitted by a human.*

## Linked issues

Fixes #3020. Alternative to the abandoned #3024.

## Checklist

- [x] Your go code is formatted
- [x] You have updated generated code when introducing new fields to the `acid.zalan.do` api package (n/a - no API changes)
- [x] New configuration options are reflected in CRD validation, helm charts and sample manifests (n/a - no new options)
- [x] New functionality is covered by unit and/or e2e tests (critical-op PDB unit test scenarios updated; `go test ./pkg/cluster/` passes)
- [x] You have checked existing open PRs for possible overlay and referenced them (#3024)

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -653,8 +653,14 @@ The PDB is only relaxed in two scenarios:
 * If a cluster is scaled down to `0` instances (e.g. for draining nodes)
 * If the PDB is disabled in the configuration (`enable_pod_disruption_budget`)
 
-The PDBs are still in place: the primary PDB with `MinAvailable` set to `0` and the
-critical operations PDB with `MaxUnavailable` set to `100%`. Disabling PDBs
+The PDBs are still in place but fully relaxed: the primary PDB with `MinAvailable`
+set to `0` and the critical operations PDB with `MaxUnavailable` set to `100%`.
+The two PDBs intentionally use different budget fields matching their purposes:
+the primary PDB guarantees a minimum count of always-present pods
+(`MinAvailable`), while the critical operations PDB freezes disruptions for
+whatever pods currently carry the `critical-operation=true` label - a
+usually-empty set, which `MaxUnavailable: 0` expresses without producing an
+unsatisfiable budget while idle. Disabling PDBs
 helps avoiding blocking Kubernetes upgrades in managed K8s environments at the
 cost of prolonged DB downtime. See PR [#384](https://github.com/zalando/postgres-operator/pull/384)
 for the use case.
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -639,9 +639,11 @@ masters in single-node clusters and/or the last remaining running instance in a 
 cluster.
 
 ## PDB for critical operations
-The `MinAvailable` parameter of this PDB is equal to the `numberOfInstances` set in the
-cluster manifest, while label selector includes `critical-operation=true` condition. This
-allows to protect all pods of a cluster, given they are labeled accordingly.
+The `MaxUnavailable` parameter of this PDB is set to `0`, while label selector includes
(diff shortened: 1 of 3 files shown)


FxKu (approved):

jopadi (commented):
👍

PR #3142, opened 2026-07-27
remove gp3 migration flow

FxKu:
For the next release lets remove the whole gp3 migration logic, because gp3 is the new AWS standard for a while now. We should also allow modifications when volume type differs (for now only with `mixed` mode). Before we only allowed gp3 and io2 when iops or throughput were configured. Unit tests are changed accordingly.


jopadi (commented):
👍

FxKu:
👍

PR #3144, opened 2026-07-28
Fix invalid JSON (trailing comma) in operator-ui chart OPERATOR_UI_CONFIG

ioanalytica:
### Problem

The `postgres-operator-ui` Helm chart (v2.0.0) renders an invalid `OPERATOR_UI_CONFIG` value. The `postgresql_versions` array in `charts/postgres-operator-ui/templates/deployment.yaml` has a trailing comma after the last element:

```json
"postgresql_versions": [
  "18", "17", "16", "15", "14",
]
```

`OPERATOR_UI_CONFIG` is parsed at startup in `operator_ui/main.py` with Python's `json.loads()`, which does not accept trailing commas. The UI container therefore crash-loops immediately:

```
File "/app/operator_ui/main.py", line 63, in <module>
    OPERATOR_UI_CONFIG = loads(getenv('OPERATOR_UI_CONFIG', '{}'))
json.decoder.JSONDecodeError: Expecting value: line 25 column 3 (char 602)
```

The standalone `ui/manifests/deployment.yaml` is already correct on master — only the chart template carries the trailing comma.

### Fix

Remove the trailing comma so the rendered config is valid JSON.

### Verification

`helm template` on the patched chart now yields an `OPERATOR_UI_CONFIG` that parses cleanly with `json.loads()` (versions `18,17,16,15,14`).

Fixes #3062

PR #3145, opened 2026-07-28
Fix operatorconfigurations CRD: render NamespacedName fields as type string

### LINKED ISSUE #3143: v2.0.0: operatorconfigurations CRD rejected by the apiserver (string default on object-typed oauth_token_secret_name), operator fatals on startup

- **Which image of the operator are you using?** ghcr.io/zalando/postgres-operator:v2.0.0 (upgrading from v1.15.1)
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** Bare Metal K8s, v1.36

## Problem

Upgrading from v1.15.1 to v2.0.0 (helm chart 2.0.0) fails applying the CRDs. The apiserver rejects the `operatorconfigurations.acid.zalan.do` CRD:

```
CustomResourceDefinition.apiextensions.k8s.io "operatorconfigurations.acid.zalan.do" is invalid:
spec.validation.openAPIV3Schema.properties[configuration].properties[kubernetes].properties[oauth_token_secret_name].default:
Invalid value: "string":  in body must be of type object: "string"
```

The generated CRD declares `oauth_token_secret_name` as an object (with `name`/`namespace` properties) but keeps the old string default:

```yaml
oauth_token_secret_name:
  default: postgres-operator
  properties:
    name:
      type: string
    namespace:
      type: string
```

Structural schema validation rejects this on both create and update, so the CRD cannot be applied to any cluster.

## Root cause

`pkg/apis/acid.zalan.do/v1/operator_configuration_type.go`:

```go
// +kubebuilder:default=postgres-operator
// namespaced name of the secret containing the OAuth2 token to pass to the teams API
OAuthTokenSecretName spec.NamespacedName `json:"oauth_token_secret_name,omitempty"`
```

`spec.NamespacedName` serializes to/from a string, but controller-gen does not know that and generates an object schema from the struct fields, while the `+kubebuilder:default` marker keeps the string default. In the v1.x CRDs this field was `type: string`, which matched.

## Impact

This does not only break the helm chart CRD install. The same generated YAML is embedded in the operator binary (`pkg/apis/acid.zalan.do/v1/crds.go`, `//go:embed operatorconfiguration.crd.yaml`). With the default `enable_crd_registration: true` and the OperatorConfiguration config target, `createOperatorCRD` updates the existing CRD at startup and `initController` treats the failure as fatal, so the v2.0.0 operator cannot start at all: it crashloops on the same apiserver error. The v1 -> v2 upgrade path is fully blocked; skipping CRDs at the helm level does not help.

## Suggested fix

Mark the field so codegen emits the string form it actually serializes as, e.g. `+kubebuilder:validation:Type=string` (or a custom schema for `spec.NamespacedName`), then regenerate the CRDs. `master` is still affected. Other `spec.NamespacedName` fields in the CRDs are worth checking for the same object-vs-string mismatch.

ioanalytica:
### Problem

Installing/upgrading to v2.0.0 fails at the CRD apply step. The apiserver rejects the OperatorConfiguration CRD:

```
The CustomResourceDefinition "operatorconfigurations.acid.zalan.do" is invalid:
...properties[oauth_token_secret_name].default: Invalid value: "string":
in body must be of type object
```

This blocks the whole 1.15.x → 2.0.0 upgrade before the operator Deployment is ever touched.

### Root cause

These CRDs are generated by `controller-gen` from the Go types (see the `manifests` target in the `Makefile`). `spec.NamespacedName` is a struct `{Namespace, Name}`, so controller-gen emits an **object** schema for every `NamespacedName` field. But `NamespacedName` has custom `MarshalJSON`/`UnmarshalJSON` that (de)serialize it as a plain JSON **string** `"namespace/name"`:

```go
func (n NamespacedName) MarshalJSON() ([]byte, error) {
    return []byte("\"" + n.String() + "\""), nil
}
func (n *NamespacedName) UnmarshalJSON(data []byte) error {
    var tmp string
    if err := json.Unmarshal(data, &tmp); err != nil { return err }
    // ... Decode(tmp) splits on "/"
}
```

So the generated object schema never matched how the operator actually reads and writes these fields (in 1.15.x they were `type: string`). For `oauth_token_secret_name` it becomes a hard failure, because that field also carries `// +kubebuilder:default=postgres-operator`, and a scalar string default on an object-typed property is rejected by the apiserver.

### Fix

Annotate the `NamespacedName` type itself with `// +kubebuilder:validation:Type=string` and regenerate the CRDs. controller-gen then emits `type: string` for **all** `NamespacedName` fields — `oauth_token_secret_name`, `infrastructure_roles_secret_name`, `pod_environment_configmap`, and the nested infrastructure-role `secretname` — matching their runtime serialization and restoring 1.15.x behaviour.

A field-level `+kubebuilder:validation:Type=string` override does **not** work here: it conflicts with the struct-derived schema (`conflicting types in allOf branches: object vs string`) and fails generation. The override has to be on the type.

Changes:
- `pkg/spec/types.go` — the marker (source of the fix)
- regenerated `manifests/operatorconfiguration.crd.yaml` and its two synced copies (`pkg/apis/acid.zalan.do/v1/operatorconfiguration.crd.yaml`, `charts/postgres-operator/crds/operatorconfigurations.yaml`)

The CRDs were regenerated with the pinned `controller-tools` (`go tool controller-gen ...`), so the only diff is the intended object→string change on the `NamespacedName` fields — no unrelated churn.

### Verification

Server-side dry-run apply against a real cluster:

- unpatched master CRD → rejected with the error above
- regenerated CRD → `customresourcedefinition.apiextensions.k8s.io/operatorconfigurations.acid.zalan.do configured (server dry run)`

Fixes #3143

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -226,24 +226,7 @@ spec:
                       type: string
                     type: array
                   infrastructure_roles_secret_name:
-                    description: |-
-                      NamespacedName comprises a resource name, with a mandatory namespace,
-                      rendered as "<namespace>/<name>".  Being a type captures intent and
-                      helps make sure that UIDs, namespaced names and non-namespaced names
-                      do not get conflated in code.  For most use cases, namespace and name
-                      will already have been format validated at the API entry point, so we
-                      don't do that here.  Where that's not the case (e.g. in testing),
-                      consider using NamespacedNameOrDie() in testing.go in this package.
-
-                      from: https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go
-                    properties:
-                      name:
-                        type: string
-                      namespace:
-                        type: string
-                    required:
-                    - name
-                    type: object
+                    type: string
                   infrastructure_roles_secrets:
                     description: namespaced name of the secret containing infrastructure
                       roles names and passwords
@@ -265,14 +248,7 @@ spec:
                           description: |-
                             Name of a secret which describes the role, and optionally name of a
                             configmap with an extra information
-                          properties:
-                            name:
-                              type: string
-                            namespace:
-                              type: string
-                          required:
-                          - name
-                          type: object
+                          type: string
                         template:
                           type: boolean
                         userkey:
@@ -458,14 +434,7 @@ spec:
                     default: postgres-operator
                     description: namespaced name of the secret containing the OAuth2
                       token to pass to the teams API
-                    properties:
-                      name:
-                        type: string
-                      namespace:
-                        type: string
-                    required:
-                    - name
-                    type: object
+                    type: string
                   pdb_master_label_selector:
                     default: true
                     type: boolean
@@ -485,14 +454,7 @@ spec:
                   pod_environment_configmap:
                     description: namespaced name of the ConfigMap with environment
                       variables to populate on every pod
-                    properties:
-                      name:
-                        type: string
-                      namespace:
-                        type: string
-                    required:
-                    - name
-                    type: object
+                    type: string
                   pod_environment_secret:
                     type: string
                   pod_management_policy:
--- a/manifests/operatorconfiguration.crd.yaml
+++ b/manifests/operatorconfiguration.crd.yaml
@@ -226,24 +226,7 @@ spec:
                       type: string
                     type: array
                   infrastructure_roles_secret_name:
-                    description: |-
-                      NamespacedName comprises a resource name, with a mandatory namespace,
-                      rendered as "<namespace>/<name>".  Being a type captures intent and
-                      helps make sure that UIDs, namespaced names and non-namespaced names
-                      do not get conflated in code.  For most use cases, namespace and name
-                      will already have been format validated at the API entry point, so we
-                      don't do that here.  Where that's not the case (e.g. in testing),
-                      consider using NamespacedNameOrDie() in testing.go in this package.
-
-                      from: https://github.com/kubernetes/apimachinery/blob/master/pkg/types/namespacedname.go
-                    properties:
-                      name:
-                        type: string
-                      namespace:
-                        type: string
-                    required:
-                    - name
-                    type: object
+                    type: string
                   infrastructure_roles_secrets:
                     description: namespaced name of the secret containing infrastructure
                       roles names and passwords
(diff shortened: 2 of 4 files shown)


jpflouret:
Aren't these files generated from the go source? If so, these edits would be reverted.

ioanalytica:
@jpflouret You're right, and thanks — the CRD YAMLs are generated by `controller-gen` from the Go types (the `manifests` target in the `Makefile`, then copied into `pkg/apis/.../` and `charts/.../crds`), so my original YAML-only edit would have been reverted by `make`.

I've reworked the PR to fix it at the source instead. The underlying issue is that `spec.NamespacedName` is a struct `{Namespace, Name}`, so controller-gen infers an **object** schema, whereas the type's custom `MarshalJSON`/`UnmarshalJSON` actually (de)serialize it as a **string** `"namespace/name"`. That mismatch is what produced the object-typed `oauth_token_secret_name` with a scalar default (and it was `type: string` back in 1.15.x).

The fix now annotates the `NamespacedName` type with `// +kubebuilder:validation:Type=string` and regenerates the CRDs, so controller-gen emits `type: string` for all `NamespacedName` fields. A field-level override doesn't work — it fails generation with `conflicting types in allOf branches: object vs string` — so it has to be on the type.

Regenerated with the pinned `controller-tools` (so the diff is only the intended object→string change), and verified with a server-side dry-run apply: unpatched CRD is rejected, regenerated CRD is accepted.

FxKu (approved):

idanovinda (approved):

ioanalytica:
@FxKu thanks again for merging #3144 and #3145 so quickly! One question on release timing: since #3143 makes the apiserver reject the OperatorConfiguration CRD outright, v2.0.0 currently can't be applied through the usual GitOps flows (Helm `install.crds`, Flux, Argo) — the very first CRD apply fails, so it effectively can't be installed that way. Given the fixes are already on master, would you consider cutting a 2.0.1 patch (or adding a short note to the v2.0.0 release/README) rather than waiting for the regular cycle? Happy to help if anything's needed.

FxKu:
Done

jpflouret:
@FxKu thank you again for merging this and making a new release. Would it be possible to also get a new helm chart release with v2.0.1? Thanks again!

FxKu:
Done, too

jpflouret:
@FxKu thank you! Just FYI that the postgres-operator/index.yaml still only has 2.0.0 as the newest version.

FxKu:
interesting how this slipped through. I hope #3148 will fix it then without needing another release. But the v2.0.0 was also fixed and should work.

PR #3160, opened 2026-08-04
fix operatorconfigurations CRD: render sidecars as array

### LINKED ISSUE #3159: `configuration.sidecars` is typed `object` in the OperatorConfiguration CRD but the Go field is a slice
# `configuration.sidecars` is typed `object` in the OperatorConfiguration CRD but the Go field is a slice

## Summary

Since the CRDs are generated from Go structs (#3004, #3102, #3117), `configuration.sidecars` is rendered
as `type: object` while `SidecarContainers` is a `[]v1.Container`. Any OperatorConfiguration that defines
global sidecars is rejected by the API server, so the feature is unusable on v2.0.0, v2.0.1 and `master`.

This worked on v1.14.0, where the same field was correctly rendered as `type: array`.

## Details

Go type ([operator_configuration_type.go#L436](https://github.com/zalando/postgres-operator/blob/v2.0.1/pkg/apis/acid.zalan.do/v1/operator_configuration_type.go#L436)):

```go
SidecarContainers []v1.Container `json:"sidecars,omitempty"`
```

Generated CRD ([operatorconfigurations.yaml#L781-L783](https://github.com/zalando/postgres-operator/blob/v2.0.1/charts/postgres-operator/crds/operatorconfigurations.yaml#L781-L783), identical in the embedded
[operatorconfiguration.crd.yaml](https://github.com/zalando/postgres-operator/blob/v2.0.1/pkg/apis/acid.zalan.do/v1/operatorconfiguration.crd.yaml#L781-L783)):

```yaml
              sidecars:
                type: object
                x-kubernetes-preserve-unknown-fields: true
```

v1.14.0 for comparison ([operatorconfigurations.yaml#L125-L130](https://github.com/zalando/postgres-operator/blob/v1.14.0/charts/postgres-operator/crds/operatorconfigurations.yaml#L125-L130)):

```yaml
              sidecars:
                type: array
                nullable: true
                items:
                  type: object
                  x-kubernetes-preserve-unknown-fields: true
```

Note this is distinct from #3143 / #3145, which fixed `oauth_token_secret_name` in v2.0.1. `sidecars` was
not covered and is still broken on `master`.

## Reproduction

```sh
kind create cluster
kubectl apply -f https://raw.githubusercontent.com/zalando/postgres-operator/v2.0.1/charts/postgres-operator/crds/operatorconfigurations.yaml

cat <<'EOF' | kubectl apply -f -
apiVersion: acid.zalan.do/v1
kind: OperatorConfiguration
metadata:
  name: repro
configuration:
  sidecars:
  - name: exporter
    image: quay.io/prometheuscommunity/postgres-exporter
    ports:
    - name: exporter
      containerPort: 9187
EOF
```

Result:

```
The OperatorConfiguration "repro" is invalid:
* configuration.sidecars: Invalid value: "array": configuration.sidecars in body must be of type object: "array"
```

With `--server-side` (what Argo CD uses):

```
Error from server: failed to create typed patch object: errors:
  .configuration.sidecars: expected map, got &{[map[image:quay.io/... name:exporter ...]]}
```

## Impact

Global sidecars cannot be configured at all in `configTarget: OperatorConfigurationCRD` mode. There is no
equivalent fallback:

- `sidecar_docker_images` only carries a name to image mapping, so no `env`, `args`, `ports`, `resources`
  or `securityContext`.
- Per cluster `spec.sidecars` uses the restricted [`Sidecar`](https://github.com/zalando/postgres-operator/blob/v2.0.1/pkg/apis/acid.zalan.do/v1/postgresql_type.go#L295-L302)
  type, which has no `args` and no `securityContext`.

## Workaround

Patching the CRD back to `type: array` works and survives operator restarts, because in
`OperatorConfigurationCRD` mode `createConfigurationCRD()` is skipped: at
[controller.go#L319-L324](https://github.com/zalando/postgres-operator/blob/v2.0.1/pkg/controller/controller.go#L319-L324)
it is guarded by `c.opConfig.EnableCRDRegistration`, which is still `nil` at that point since the
configuration has not been read yet. Only `postgresqls` gets re-registered.

## Cause

The markers on the field are hand-written, this is not a generator mistake:

```go
// +kubebuilder:validation:XPreserveUnknownFields
// +kubebuilder:validation:Type=object
// +kubebuilder:validation:Schemaless
SidecarContainers []v1.Container `json:"sidecars,omitempty"`
```

`Type=object` is simply wrong for a slice field. Two fields above in the same struct,
`MaintenanceWindows []string` correctly carries `+kubebuilder:validation:Type=array`.

## Suggested fix

Drop the three markers and let controller-gen derive the schema from the Go type, the same way
`spec.initContainers` (also a `[]v1.Container`) is already handled in the Postgresql CRD. It then
renders as `type: array` with a full Container schema for the items.

Note that restoring the exact v1.14.0 shape (`items: {type: object, x-kubernetes-preserve-unknown-fields: true}`)
is *not* achievable by adding `items:` markers: controller-gen rejects them alongside `Schemaless` with
`must apply kubebuilder:validation:items:Type to an array value, found array`, and emits `type: array`
with no `items`, which is not a valid structural schema. Keeping the CRD small would require a named
element type carrying the markers.

Fix proposed in #3160.

## Environment

- Operator: v2.0.1 (also v2.0.0 and `master`)
- Kubernetes: v1.34.0 (kind)

rasoanaivo-r:
Fixes #3159

## Problem

`configuration.sidecars` is annotated with `+kubebuilder:validation:Type=object` while
`SidecarContainers` is a `[]v1.Container`. The generated schema therefore declares an object where the
Go type is a slice, and the API server rejects every list value:

```
* configuration.sidecars: Invalid value: "array": configuration.sidecars in body must be of type object: "array"
```

Global sidecars cannot be configured at all on v2.0.0, v2.0.1 and `master`. This worked on v1.14.0, where
the hand-written CRD declared `type: array`.

The three markers look like an attempt to avoid inlining the full `v1.Container` schema, but
`Type=object` is simply the wrong type for a slice field. Two fields above in the same struct,
`MaintenanceWindows []string` correctly carries `+kubebuilder:validation:Type=array`.

## Fix

Drop the hand-written `Schemaless` / `Type=object` / `XPreserveUnknownFields` markers and let
controller-gen derive the schema from the Go type:

```diff
-	// +kubebuilder:validation:XPreserveUnknownFields
-	// +kubebuilder:validation:Type=object
-	// +kubebuilder:validation:Schemaless
+	// +nullable
 	SidecarContainers []v1.Container `json:"sidecars,omitempty"`
```

This is how `[]v1.Container` is already handled elsewhere in the repo: `spec.initContainers` in the
Postgresql CRD renders as a full typed array. The result is consistent, and it validates sidecars instead
of accepting anything.

`+nullable` is kept so an explicit `sidecars: null` stays valid, matching the v1.14.0 schema.

I first tried keeping `Schemaless` and adding `items:` markers, but controller-gen rejects that
combination:

```
must apply kubebuilder:validation:items:Type to an array value, found array
```

and emits `type: array` with no `items`, which is not a valid structural schema. Inlining the Container
schema is the straightforward option. If you would rather keep the CRD small and restore the v1.14.0
`items: {type: object, x-kubernetes-preserve-unknown-fields: true}` shape, that needs a named element
type carrying the markers, which is a larger change. Happy to switch if you prefer that.

## Result

```yaml
              sidecars:
                items:
                  description: A single application container that you want to run within a pod.
                  properties:
                    args: ...
                    securityContext: ...
                    ...
                  required:
                  - name
                  type: object
                nullable: true
                type: array
```

`charts/postgres-operator/crds/operatorconfigurations.yaml` grows from 39 KB to 132 KB.

## Verification

- CRD generation is reproducible: regenerating before the change reproduces the committed
  `operatorconfigurations.yaml` byte for byte.
- Only `operatorconfigurations` changes. `postgresqls.yaml` and `postgresteams.yaml` are byte identical
  after regeneration.
- `go build ./...` passes.
- On kind v1.34.0 with the regenerated CRD:
  - the minimal reproduction from #3159 is accepted;
  - a full real-world config applies with `--server-side` (the Argo CD path), and the stored object
    keeps `args`, `env` and `securityContext` intact;
  - the v2.0.1 operator image reconciles a cluster and the sidecar lands in the StatefulSet with its
    `args` and `securityContext`:

    ```
    postgres          args= secctx={"allowPrivilegeEscalation":true,...}
    patroni-selfheal  args= secctx=
    exporter          args=["--collector.stat_activity_autovacuum",...]
                      secctx={"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},
                              "readOnlyRootFilesystem":true,"runAsNonRoot":true,"runAsUser":65534,
                              "seccompProfile":{"type":"RuntimeDefault"}}
    ```

## Note on upgrades

Existing `OperatorConfiguration` objects are not revalidated when the CRD is updated, so this only
affects the next write. Sidecar definitions that were valid `v1.Container` values stay valid. `name`
becomes required, which is already an operator level requirement and relates to #2966 / #2997.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -779,8 +779,1527 @@ spec:
                   type: string
                 type: object
               sidecars:
-                type: object
-                x-kubernetes-preserve-unknown-fields: true
+                items:
+                  description: A single application container that you want to run
+                    within a pod.
+                  properties:
+                    args:
+                      description: |-
+                        Arguments to the entrypoint.
+                        The container image's CMD is used if this is not provided.
+                        Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
+                        cannot be resolved, the reference in the input string will be unchanged. Double $$ are reduced
+                        to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will
+                        produce the string literal "$(VAR_NAME)". Escaped references will never be expanded, regardless
+                        of whether the variable exists or not. Cannot be updated.
+                        More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
+                      items:
+                        type: string
+                      type: array
+                      x-kubernetes-list-type: atomic
+                    command:
+                      description: |-
+                        Entrypoint array. Not executed within a shell.
+                        The container image's ENTRYPOINT is used if this is not provided.
+                        Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
+                        cannot be resolved, the reference in the input string will be unchanged. Double $$ are reduced
+                        to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e. "$$(VAR_NAME)" will
+                        produce the string literal "$(VAR_NAME)". Escaped references will never be expanded, regardless
+                        of whether the variable exists or not. Cannot be updated.
+                        More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
+                      items:
+                        type: string
+                      type: array
+                      x-kubernetes-list-type: atomic
+                    env:
+                      description: |-
+                        List of environment variables to set in the container.
+                        Cannot be updated.
+                      items:
+                        description: EnvVar represents an environment variable present
+                          in a Container.
+                        properties:
+                          name:
+                            description: |-
+                              Name of the environment variable.
+                              May consist of any printable ASCII characters except '='.
+                            type: string
+                          value:
+                            description: |-
+                              Variable references $(VAR_NAME) are expanded
+                              using the previously defined environment variables in the container and
+                              any service environment variables. If a variable cannot be resolved,
+                              the reference in the input string will be unchanged. Double $$ are reduced
+                              to a single $, which allows for escaping the $(VAR_NAME) syntax: i.e.
+                              "$$(VAR_NAME)" will produce the string literal "$(VAR_NAME)".
+                              Escaped references will never be expanded, regardless of whether the variable
+                              exists or not.
+                              Defaults to "".
+                            type: string
+                          valueFrom:
+                            description: Source for the environment variable's value.
+                              Cannot be used if value is not empty.
+                            properties:
+                              configMapKeyRef:
+                                description: Selects a key of a ConfigMap.
+                                properties:
+                                  key:
+                                    description: The key to select.
+                                    type: string
+                                  name:
+                                    default: ""
+                                    description: |-
+                                      Name of the referent.
+                                      This field is effectively required, but due to backwards compatibility is
+                                      allowed to be empty. Instances of this type with an empty value here are
+                                      almost certainly wrong.
+                                      More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                                    type: string
+                                  optional:
+                                    description: Specify whether the ConfigMap or
+                                      its key must be defined
+                                    type: boolean
+                                required:
+                                - key
+                                type: object
+                                x-kubernetes-map-type: atomic
+                              fieldRef:
+                                description: |-
+                                  Selects a field of the pod: supports metadata.name, metadata.namespace, `metadata.labels['<KEY>']`, `metadata.annotations['<KEY>']`,
+                                  spec.nodeName, spec.serviceAccountName, status.hostIP, status.podIP, status.podIPs.
+                                properties:
+                                  apiVersion:
+                                    description: Version of the schema the FieldPath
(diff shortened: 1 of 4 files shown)


rasoanaivo-r:
Two things I cannot do myself as an outside contributor:

- `COMPR/pr-labels` is failing because the PR carries no label. Adding `bugfix` requires admin rights on
  the repo (`403 Must have admin rights to Repository`), so a maintainer needs to set it.
- `operator-tests` and `operator-e2e-tests` are sitting in `action_required`, waiting for a maintainer to
  approve workflow runs for a first-time contributor.

Happy to rework the change if you would rather keep the CRD compact. See the trade-off described in the
description: this version inlines the full Container schema (39 KB to 132 KB), the alternative is a named
element type carrying the markers to restore the v1.14.0 `preserve-unknown-fields` shape.

mikkeloscar:
:+1:

FxKu:
👍

PR #3165, opened 2026-08-11
Skip owner references on user secrets when secret deletion is disabled

### LINKED ISSUE #3162: Secrets still deleted when `enable_secrets_deletion: false` + `enable_owner_references: true`
- **Which image of the operator are you using?** ghcr.io/zalando/postgres-operator:v2.0.1 (also reproducible by reading current `master` @ [a7aaad0](https://github.com/zalando/postgres-operator/commit/a7aaad0a0b9517bb46077f2925eb80ae8ebe1a0c))
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** OpenShift Kubernetes Cluster (OKC), but the issue is a generic Kubernetes garbage-collection interaction and is reproducible on plain Kubernetes as well.

## Summary

`enable_secrets_deletion: false` does not actually prevent credential Secrets from being deleted when `enable_owner_references: true` is also set. Kubernetes' own garbage collector removes the Secrets anyway, because the operator unconditionally attaches an `OwnerReference` to every generated Secret whenever `enable_owner_references` is on, regardless of the `enable_secrets_deletion` setting.

## Background

We enabled `enable_owner_references: true` to fix stale ConfigMaps/Endpoints/other leftover resources after deleting a `postgresql` custom resource (our cluster uses `kubernetes_use_configmaps: true` on Openshift, so DCS state lives in ConfigMaps rather than Endpoints). That part works as expected.

To avoid losing credentials on a subsequent reinstall (the Secret's password must survive if the underlying PVC/data is retained), we additionally set `enable_secrets_deletion: false`, expecting this combination to mean: "garbage-collect everything via owner references, except keep the Secrets around."

Instead, deleting the `postgresql` CR still deletes the credential Secrets, because they carry an `OwnerReference` to the CR and Kubernetes garbage-collects them independently of the operator's own delete logic.

## Root cause

1. [`Cluster.Delete()`](https://github.com/zalando/postgres-operator/blob/a7aaad0a0b9517bb46077f2925eb80ae8ebe1a0c/pkg/cluster/cluster.go#L1300-L1308) only guards the operator's own explicit `deleteSecrets()` call with `EnableSecretsDeletion`:

```go
if c.OpConfig.EnableSecretsDeletion != nil && *c.OpConfig.EnableSecretsDeletion {
    if err := c.deleteSecrets(); err != nil {
        ...
    }
} else {
    c.logger.Info("not deleting secrets because disabled in configuration")
}
```

2. But [`generateSingleUserSecret`](https://github.com/zalando/postgres-operator/blob/a7aaad0a0b9517bb46077f2925eb80ae8ebe1a0c/pkg/cluster/k8sres.go#L1910-L1955) sets `OwnerReferences` on every Secret purely based on [`c.ownerReferences()`](https://github.com/zalando/postgres-operator/blob/a7aaad0a0b9517bb46077f2925eb80ae8ebe1a0c/pkg/cluster/k8sres.go#L2659), which only checks `EnableOwnerReferences`:

```go
var ownerReferences []metav1.OwnerReference
if c.Config.OpConfig.EnableCrossNamespaceSecret && c.Postgresql.ObjectMeta.Namespace != pgUser.Namespace {
    ownerReferences = nil
} else {
    ownerReferences = c.ownerReferences()
}
```

There is no check against `EnableSecretsDeletion` here. So once a Secret carries the owner reference, deleting the `postgresql` CR causes the Kubernetes garbage collector to delete the Secret regardless of `enable_secrets_deletion: false` - the flag only ever controlled the operator's own best-effort cleanup path, which predates owner references and was never updated to interact with them.

## Impact

Any user who enables `enable_owner_references: true` (to fix leftover ConfigMaps/Endpoints/PDBs/etc. after cluster deletion) together with `enable_secrets_deletion: false` (to retain credentials for reinstall-with-existing-PVC scenarios) will silently lose their Secrets anyway. On next install, the operator generates a *new* random password for the credential Secret, which then doesn't match the password already baked into the retained PVC/database, breaking authentication.

## Suggested fix

`generateSingleUserSecret` (and any other Secret-generating code path) should skip setting `OwnerReferences` on the Secret when `EnableSecretsDeletion` is explicitly `false`, so the two flags compose as documented/expected. We have a candidate fix (happy to open a PR):

```diff
 	// if secret lives in another namespace we cannot set ownerReferences
 	var ownerReferences []metav1.OwnerReference
-	if c.Config.OpConfig.EnableCrossNamespaceSecret && c.Postgresql.ObjectMeta.Namespace != pgUser.Namespace {
+	secretsDeletionDisabled := c.OpConfig.EnableSecretsDeletion != nil && !*c.OpConfig.EnableSecretsDeletion
+	if secretsDeletionDisabled || (c.Config.OpConfig.EnableCrossNamespaceSecret && c.Postgresql.ObjectMeta.Namespace != pgUser.Namespace) {
 		ownerReferences = nil
 	} else {
 		ownerReferences = c.ownerReferences()
```

## How to reproduce

1. Deploy the operator with:
   ```yaml
   kubernetes:
     enable_owner_references: true
     enable_secrets_deletion: false
   ```
2. Create a `postgresql` cluster, let it reach `Running`, note the generated credential Secret's password.
3. Delete the `postgresql` CR.
4. Observe: the credential Secret is deleted along with the CR (`kubectl get secret <username>.<cluster>.credentials...` returns not found), even though `enable_secrets_deletion` is `false`.

adshin21:
Closes #3162

`enable_secrets_deletion: false` is supposed to keep user-credential Secrets when a `postgresql` CR is deleted. The operator's own `Cluster.Delete()` does guard `deleteSecrets()`, but `generateSingleUserSecret` attaches a controller `OwnerReference` to every generated Secret whenever `enable_owner_references` is `true`, and K8s garbage collection then removes the Secret independently of the operator's deletion logic. This made `enable_secrets_deletion=false` a no-op whenever `enable_owner_references=true`.

This PR makes `generateSingleUserSecret` skip the controller owner reference on user-credential Secrets whenever `enable_secrets_deletion` is `false`, so the two flags compose as documented.

### How to reproduce / verify
With `enable_owner_references: true` and `enable_secrets_deletion: false`, delete a `postgresql` CR — the user-credential Secret now survives. 
With the default (`enable_secrets_deletion: true`) the Secret still cascades-delete as before. See #3162 for the full reproduction steps.

### Notes
- All other operator-managed resources (StatefulSets, Services, PDBs, logical-backup CronJob) still receive the controller owner reference and continue to cascade-delete as before.
- Cross-namespace Secrets remain excluded from owner references, as before.
- A unit test (`TestGenerateSingleUserSecret_OwnerReferences`) covers all four flag combinations plus the cross-namespace cases.

### DIFF
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -285,11 +285,12 @@ will differ and trigger a rolling update of the pods.
 ## Owner References and Finalizers
 
 The Postgres Operator can set [owner references](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) to most of a cluster's child resources to improve
-monitoring with GitOps tools and enable cascading deletes. There are two
+monitoring with GitOps tools and enable cascading deletes. There are three
 exceptions:
 
 * Persistent Volume Claims, because they are handled by the [PV Reclaim Policy]https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/ of the Stateful Set
 * Cross-namespace secrets, because owner references are not allowed across namespaces by design
+* User-credential secrets when [`enable_secrets_deletion`](reference/operator_parameters.md#enable_secrets_deletion) is `false`, so Kubernetes garbage collection does not cascade-delete them after the Postgresql resource is removed (the `enable_secrets_deletion` flag alone only suppresses the operator's own delete path, not K8s GC)
 
 The operator would clean these resources up with its regular delete loop
 unless they got synced correctly. If for some reason the initial cluster sync
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -289,8 +289,12 @@ configuration they are grouped under the `kubernetes` key.
 * **enable_owner_references**
   The operator can set owner references on its child resources (except PVCs,
   Patroni config service/endpoint, cross-namespace secrets) to improve cluster
-  monitoring and enable cascading deletion. The default is `false`. Warning,
-  enabling this option disables configured delete protection checks (see below).
+  monitoring and enable cascading deletion. User-credential secrets are also
+  excluded from controller owner references whenever
+  [enable_secrets_deletion](#enable_secrets_deletion) is `false`, so that
+  Kubernetes garbage collection does not cascade-delete them when the
+  Postgresql resource is removed. The default is `false`. Warning, enabling
+  this option disables configured delete protection checks (see below).
 
 * **delete_annotation_date_key**
   key name for annotation that compares manifest value with current date in the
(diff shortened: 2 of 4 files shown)


FxKu:
👍

idanovinda (approved):

PR #3169, opened 2026-08-13
Protect operator CRDs from accidental deletion via metadata finalizer

### LINKED ISSUE #3168: Protect operator CRDs from accidental deletion via metadata finalizer
## Problem

A `kubectl delete crd <name>` against the operator's `postgresqls`, `operatorconfigurations`, `postgresteams` or `fabriceventstreams` CRD is destructive: the Kubernetes apiserver cascades the delete to every custom resource of that kind in the cluster, so a single mistyped command can wipe out every `postgresql`/`operatorconfiguration`/`PostgresTeam`/`FabricEventStream` object at once. For an operator whose value proposition is the data behind those CRs, this is a one-command data-loss vector.

This is the same class of accidental-cluster-wipe that has been reported across the Kubernetes ecosystem; the standard mitigation is a metadata finalizer on the CRD that blocks deletion until removed explicitly.

## Proposed change

Add an `acid.zalan.do/crd-protection` finalizer to each operator CRD's `metadata`. With the finalizer set, the apiserver keeps the CRD in `Terminating` instead of dropping it, so the cascade to existing CRs never happens. Removing the finalizer is the documented manual step required to delete the CRD.

Concretely:
- The Go loader (`PostgresCRD` / `OperatorConfigurationCRD` in `pkg/apis/acid.zalan.do/v1/crds.go`) sets the finalizer when the CRD object is constructed. This survives the `controller-gen` regeneration that rewrites the CRD YAML on every `make`, so the finalizer cannot be silently dropped by a regeneration.
- The generated CRD YAML manifests (`manifests/`, `charts/postgres-operator/crds/`, the embedded `pkg/apis/.../*.crd.yaml`) also carry the finalizer so that `kubectl apply` / Helm installs put it in place directly.
- On startup, the operator re-applies the finalizer to the `postgresql` and `operatorconfiguration` CRDs if a pre-existing CRD is missing it, so older deployments pick up the safety net on upgrade.
- The `postgresteam` and `fabriceventstream` CRDs are not auto-registered by the operator, so their protection relies on the finalizer shipped in the YAML manifest at install time.

## Manual removal procedure

To intentionally delete a CRD, clear the finalizer first:

```bash
kubectl patch crd postgresqls.acid.zalan.do -p '{"metadata":{"finalizers":[]}}' --type=merge
kubectl delete crd postgresqls.acid.zalan.do
```

Per the Kubernetes docs, delete the CRs first when possible, wait for their finalizers to clear, and only then drop the CRDs.

## Test

A new e2e test (`test_zy_crd_protection_finalizer` in `e2e/tests/test_e2e.py`) targets the most critical CRD - `postgresqls.acid.zalan.do` - and exercises the full lifecycle:
- Verifies the finalizer is present on every operator CRD.
- Snapshots the live `acid-minimal-cluster` postgresql CR, attempts to delete the CRD, and asserts both the CRD and the CR survive.
- Verifies that clearing the finalizer is what actually unblocks the cascade.
- Re-creates the CRD and the cluster in a finally block so a failed assertion cannot leave the suite in a broken state.

The test runs after every other test that depends on the cluster but before the last `test_zz_cluster_deletion` check.

## Out of scope

A unit test for the Go-side `ensureCRDProtectionFinalizer` helper would be a small addition; the e2e already covers the user-visible behavior end-to-end. Happy to add it if maintainers prefer.

A branch with this change is already pushed to `moduon/postgres-operator:crd-protection-finalizer`; a PR from there to upstream will follow once this issue gets traction.

Assisted-by: OpenCode + MiniMax M3

yajo:
## Problem description

A `kubectl delete crd` against the operator's `postgresqls`, `operatorconfigurations`, `postgresteams` or `fabriceventstreams` CRD is destructive: the Kubernetes apiserver cascades the delete to every custom resource of that kind in the cluster, so a single mistyped command can wipe out every `postgresql`/`operatorconfiguration`/`PostgresTeam`/`FabricEventStream` object at once. For an operator whose value proposition is the data behind those CRs, this is a one-command data-loss vector.

Background on the failure mode:
- [What Happens When You Delete a Kubernetes CustomResourceDefinition (CRD)?](https://marcincuber.medium.com/what-happens-when-you-delete-a-kubernetes-customresourcedefinition-crd-d5e741fb6441) walks through the cascade, the role of finalizers, and the recovery steps.
- The Kubernetes docs lay out the [recommended CRD deletion checklist](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#delete-a-customresourcedefinition) (delete CRs first, wait for finalizers to clear, then drop the CRD), which is exactly what a metadata finalizer on the CRD enforces.

The fix is a metadata finalizer (`acid.zalan.do/crd-protection`) on each operator CRD. While the finalizer is set, the apiserver keeps the CRD in `Terminating` instead of dropping it, so the cascade to existing CRs never happens. Removing the finalizer is the documented manual step required to delete the CRD.

## Linked issues

Fixes #3168

## Checklist

- [x] Go code is formatted (`gofmt`).
- [x] Generated code updated: the finalizer is assigned in the Go loader (`PostgresCRD` / `OperatorConfigurationCRD` in `pkg/apis/acid.zalan.do/v1/crds.go`) so it survives `controller-gen` regeneration; the generated YAML manifests also carry the finalizer for `kubectl apply` / Helm installs.
- [x] Helm chart and sample manifests updated: `charts/postgres-operator/crds/*.yaml` and `manifests/*.crd.yaml` carry the finalizer.
- [x] New functionality is covered by e2e tests: `test_zy_crd_protection_finalizer` in `e2e/tests/test_e2e.py` exercises the full lifecycle (finalizer present, deletion blocked, cascade on unblock) against the most critical CRD and re-creates state in a `finally` block.
- [x] Documentation updated: `docs/administrator.md` documents the protection and the manual removal procedure.
- [x] No existing open PRs overlay this change.

Assisted-by: OpenCode + MiniMax M3

### DIFF
--- a/charts/postgres-operator/crds/operatorconfigurations.yaml
+++ b/charts/postgres-operator/crds/operatorconfigurations.yaml
@@ -7,6 +7,10 @@ metadata:
   labels:
     app.kubernetes.io/name: postgres-operator
   name: operatorconfigurations.acid.zalan.do
+  # finalizer to prevent accidental deletion of the CRD which would erase
+  # all operatorconfiguration custom resources. Remove explicitly to delete the CRD.
+  finalizers:
+  - acid.zalan.do/crd-protection
 spec:
   group: acid.zalan.do
   names:
--- a/charts/postgres-operator/crds/postgresqls.yaml
+++ b/charts/postgres-operator/crds/postgresqls.yaml
@@ -7,6 +7,10 @@ metadata:
   labels:
     app.kubernetes.io/name: postgres-operator
   name: postgresqls.acid.zalan.do
+  # finalizer to prevent accidental deletion of the CRD which would erase
+  # all postgresql custom resources. Remove explicitly to delete the CRD.
+  finalizers:
+  - acid.zalan.do/crd-protection
 spec:
   group: acid.zalan.do
   names:
--- a/charts/postgres-operator/crds/postgresteams.yaml
+++ b/charts/postgres-operator/crds/postgresteams.yaml
@@ -7,6 +7,10 @@ metadata:
   labels:
     app.kubernetes.io/name: postgres-operator
   name: postgresteams.acid.zalan.do
+  # finalizer to prevent accidental deletion of the CRD which would erase
+  # all postgresteam custom resources. Remove explicitly to delete the CRD.
+  finalizers:
+  - acid.zalan.do/crd-protection
 spec:
   group: acid.zalan.do
   names:
--- a/docs/administrator.md
+++ b/docs/administrator.md
@@ -22,6 +22,34 @@ By default, the operator will register the CRDs in the `all` category so
 that resources are listed on `kubectl get all` commands. The `crd_categories`
 config option allows for customization of categories.
 
+### Protecting CRDs from accidental deletion
+
+Deleting a CRD in Kubernetes is destructive: the API server also deletes
+every custom resource of that kind across the cluster. A stray
+`kubectl delete crd` therefore wipes out all `postgresql`,
+`operatorconfiguration`, `postgresteam` and `fabriceventstream` objects in
+one go.
+
+To prevent this, the operator ships its CRDs with a `acid.zalan.do/crd-protection`
+finalizer on the `metadata` of each CRD. While the finalizer is present, the
+CRD is stuck in `Terminating` and the custom resources are not removed. On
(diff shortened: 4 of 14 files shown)


yajo:
Looks like the CI failure is unrelated to this change. I can see the same failure in other PRs.

PR #3171, opened 2026-08-13
Skip ALTER ROLE when the stored SCRAM verifier already matches the password

### LINKED ISSUE #3170: Role sync issues ALTER ROLE ... PASSWORD on every sync cycle when password_encryption is scram-sha-256
## Bug

With `password_encryption: scram-sha-256` in the operator configuration, the role sync re-issues `ALTER ROLE ... PASSWORD` for **every managed role on every sync cycle** (every `resync_period`, 30 min by default), even though nothing changed.

### Root cause

In `syncUsers` the stored `rolpassword` is compared to a freshly generated hash by plain string equality:

https://github.com/zalando/postgres-operator/blob/6143460a476c1a06031ae4d1d17777fbee2fa9e6/pkg/util/users/users.go#L63-L69

For md5 this works because the hash is deterministic. A SCRAM-SHA-256 verifier, however, embeds a random salt (`PGUserPasswordScramSHA256` calls `e.random(saltlength)`), so the generated string never equals the stored one, the role is marked for `PGsyncUserAlter`, and the ALTER re-hashes the verifier with yet another random salt — making the next cycle mismatch again, forever.

### Impact

- `pg_authid.rolpassword` of every managed role is rewritten (new salt) every sync cycle; WAL/audit noise.
- Connection poolers doing SCRAM pass-through against these roles (e.g. pgbouncer with `auth_query`, including the operator-deployed `enableConnectionPooler` setup) hold client keys bound to the previous salt. After every sync their next server login fails with `server login failed: FATAL password authentication failed`, followed by `server_login_retry` throttling that hits unrelated clients for a short window — twice per hour, around minute :04 and :34 with the default resync period.

### Reproduction

1. Operator (tested v1.15.1 and current master) with `configuration.postgresql_users_configuration.password_encryption: scram-sha-256`.
2. Any cluster with a user in `spec.users`.
3. `SELECT rolname, rolpassword FROM pg_authid WHERE rolname = '<user>';` before and after a sync cycle — the SCRAM salt changes every cycle while the secret is untouched.

### Expected

No `ALTER ROLE ... PASSWORD` when the stored verifier already corresponds to the password in the secret. The plaintext can be verified against the stored verifier by reusing its salt and iteration count.

I have a fix ready and will open a PR shortly.

g2px1:
Fixes #3170.

With `password_encryption: scram-sha-256`, role sync compared the stored `rolpassword` with a freshly generated hash by string equality. SCRAM-SHA-256 verifiers embed a random salt, so the comparison never matched and every sync cycle re-issued `ALTER ROLE ... PASSWORD` for every managed role, re-salting the verifier each time. Besides WAL/audit noise, each re-salt invalidates SCRAM pass-through credentials cached by connection poolers (pgbouncer behind `auth_query`, including the operator-deployed pooler), producing a short window of `password authentication failed` server logins after every sync.

### Changes

- `util.PGUserPasswordUpToDate(user, storedPassword, encryption)`: verifies the stored hash against the desired plaintext instead of comparing strings. For SCRAM verifiers the salt and iteration count are parsed from the stored value (`SCRAM-SHA-256$<iter>:<salt>$<storedKey>:<serverKey>`) and the derived server/stored keys are compared with `hmac.Equal`.
- Behavior is preserved everywhere else:
  - a stored hash whose type does not match the configured `password_encryption` is reported as outdated, so switching between `md5` and `scram-sha-256` still re-hashes roles;
  - empty and pre-hashed desired passwords are compared verbatim, mirroring the early return in `PGUserPassword`;
  - malformed stored hashes are treated as outdated.
- `users.go` uses the new check in `produceSyncRequests`; the ALTER statement itself is generated exactly as before.
- Extracted `isMD5Hash`/`isScramHash` helpers from `PGUserPassword` (no behavior change) and added a table test covering match/mismatch, both encryptions, migration between them, pre-hashed and malformed inputs.

After the fix, a role whose verifier already corresponds to the secret is left untouched; `pg_authid.rolpassword` stays stable across sync cycles and pooler SCRAM pass-through keeps working.


jopadi (approved):

pkg/util/util.go
```diff
@@ -94,14 +95,21 @@ func NewEncryptor(encryption string) *Encryptor {
 }
 
 func (e *Encryptor) PGUserPassword(user spec.PgUser) string {
-	if (len(user.Password) == md5.Size*2+len(md5prefix) && user.Password[:3] == md5prefix) ||
-		(len(user.Password) > len(scramsha256prefix) && user.Password[:len(scramsha256prefix)] == scramsha256prefix) || user.Password == "" {
+	if isMD5Hash(user.Password) || isScramHash(user.Password) || user.Password == "" {
```
FxKu, line 98: could be a function because you also repeat it in PGUserPasswordUpToDate

pkg/util/util.go
```diff
@@ -127,6 +135,91 @@ func (e *Encryptor) PGUserPasswordScramSHA256(user spec.PgUser) string {
...
+		if encryption != "md5" {
+			return false
+		}
+		return NewEncryptor(encryption).PGUserPassword(user) == storedPassword
+	case isScramHash(storedPassword):
+		if encryption == "md5" {
+			return false
+		}
+		return scramVerifierMatches(user.Password, storedPassword)
+	}
```
FxKu, line 167: maybe it's me but I find this switch case kinda hard to read

pkg/util/util.go
```diff
@@ -127,6 +135,91 @@ func (e *Encryptor) PGUserPasswordScramSHA256(user spec.PgUser) string {
...
+		return false
+	}
+
+	key := pbkdf2.Key([]byte(password), salt, iterationCount, 32, sha256.New)
+	mac := hmac.New(sha256.New, key)
+	mac.Write([]byte("Server Key"))
+	derivedServerKey := mac.Sum(nil)
+	mac = hmac.New(sha256.New, key)
+	mac.Write([]byte("Client Key"))
+	derivedStoredKey := sha256.Sum256(mac.Sum(nil))
```
FxKu:
```suggestion

	serverMAC := hmac.New(sha256.New, key)
	serverMAC.Write([]byte("Server Key"))
	derivedServerKey := serverMAC.Sum(nil)

	clientMAC := hmac.New(sha256.New, key)
	clientMAC.Write([]byte("Client Key"))
	derivedStoredKey := sha256.Sum256(clientMAC.Sum(nil))
```

pkg/util/util.go
g2px1: Tbh, I'm C++ programmer and I suddenly find that bug in production. So it might be not the best code you've seen

FxKu (approved):

jopadi (approved):

PR #3173, opened 2026-08-18
clean the kube2iam annotations when the role is empty

tcondeixa:
clean the kube2iam annotations when the role is empty from the cronjob and job template.

This is a follow up on https://github.com/zalando/postgres-operator/pull/3128 to clean up the annotations for kube2iam when not role is passed (or empty).
During the migration we noticed the `spilo-logical-backup` cronjob and the job template were still with the annotation, this creates confusion.

This PR only deletes the `KubeIAmAnnotation` to reduce the risk of deleting any other annotation created in another place/time. If preferred I can delete all annotations not present when generating the cronjob

### DIFF
--- a/pkg/cluster/sync.go
+++ b/pkg/cluster/sync.go
@@ -1783,31 +1783,22 @@ func (c *Cluster) syncLogicalBackupJob() error {
 				return fmt.Errorf("could not patch labels of the logical backup job %q: %v", jobName, err)
 			}
 		}
-		_, kubeIamInCurrent := job.Annotations[constants.KubeIAmAnnotation]
-		_, kubeIamInDesired := desiredJob.Annotations[constants.KubeIAmAnnotation]
-		kubeIamNeedsRemoval := kubeIamInCurrent && !kubeIamInDesired
-		if changed, _ := c.compareAnnotations(job.Annotations, desiredJob.Annotations, nil); changed || kubeIamNeedsRemoval {
-			patchAnnotations := make(map[string]interface{})
-			for k, v := range desiredJob.Annotations {
-				patchAnnotations[k] = v
-			}
-			if kubeIamNeedsRemoval {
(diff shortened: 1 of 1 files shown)


tcondeixa:
@mikkeloscar @FxKu can you help to review this one. I did not test it yet in a cluster but want to be sure we are aligned in the approach.

pkg/cluster/sync.go
```diff
@@ -1793,6 +1793,31 @@ func (c *Cluster) syncLogicalBackupJob() error {
 				return fmt.Errorf("could not patch annotations of the logical backup job %q: %v", jobName, err)
 			}
 		}
```
FxKu, line 1795: I think the code lines before are there to remove annotations. For the pod template [here](https://github.com/zalando/postgres-operator/blob/952dce08518fc50b30f9ba9f965e520d33181262/pkg/cluster/sync.go#L1754) and for the job itself [here](https://github.com/zalando/postgres-operator/blob/952dce08518fc50b30f9ba9f965e520d33181262/pkg/cluster/sync.go#L1786). Do we really need an extra patch for the KubeIAmAnnotation?

FxKu:
Should we set KubeIAmAnnotation [here](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L2482) similar to inside pod template generation?

tcondeixa:
> Should we set KubeIAmAnnotation [here](https://github.com/zalando/postgres-operator/blob/master/pkg/cluster/k8sres.go#L2482) similar to inside pod template generation?

@mikkeloscar was suggesting to just remove this specific annotation from the cronjob and job template, because it's not needed at all and it can just be simplified. This annotation is only needed inside the pod template when the `kube_iam_role` is not empty, but this was already addressed in the previous PR.

FxKu:
The logical backup generation reuses the generatePodTempate function from the stateful set. So it would always re-add the annotation for it to be removed by your code.

PR #3177, opened 2026-09-01
pgbouncer: change auth_type to use scram sha 256

idanovinda:
as we already move to scram-sha-256, we also need to update pgbouncer auth_type


FxKu:
👍

idanovinda:
👍

PR #3179, opened 2026-09-01
Retry moveMasterPodsOffNode on failure instead of aborting

### LINKED ISSUE #3176: moveMasterPodsOffNode gives up after one failed retry instead of retrying
**Summary**

When a node hosting a Postgres master pod is cordoned, the operator's moveMasterPodsOffNode is supposed to retry finding a Patroni switchover candidate for up to master_pod_move_timeout (default 20m, checked every 1m). In practice it gets exactly one attempt (~4-5s) and then permanently gives up for that node, because retryutil.RetryWorker treats any non-nil error from the retried function as terminal rather than retryable.

If no synced standby exists in that first ~4-second window (e.g. because a replica pod was recreated moments earlier by an unrelated rolling update), the master is never migrated off the node. The node drain then falls back to the plain Kubernetes eviction API, which the Postgres master PodDisruptionBudget (minAvailable: 1, selector spilo-role=master) correctly and indefinitely rejects — since nothing ever moves the master label to another pod. The drain then blocks until an external timeout (e.g. the node-drain timeout used by whatever machine-management layer is orchestrating the drain) forces the node replacement through.

pdixit26:
## What
`moveMasterPodsOffNode` gave up after a single failed attempt instead of
retrying over the configured `master_pod_move_timeout`. Any error from
`attemptToMoveMasterPodsOffNode` (e.g. no synced standby available yet)
was treated as terminal by `retryutil.Retry`, so the operator got one
~4-5s attempt and then permanently gave up, leaving the master stuck on
a draining node.

## Fix
Log the failure and return `(false, nil)` on error so the retry loop
keeps polling every minute until `master_pod_move_timeout` expires,
matching the documented behavior.

## Testing
Added `TestMoveMasterPodsOffNodeRetriesOnError`, which forces
`attemptToMoveMasterPodsOffNode` to fail and asserts the retry loop
runs to timeout rather than aborting on the first error.

Fixes #3176

### DIFF
--- a/docs/reference/operator_parameters.md
+++ b/docs/reference/operator_parameters.md
@@ -556,9 +556,10 @@ configuration they are grouped under the `kubernetes` key.
 * **master_pod_move_timeout**
   The period of time to wait for the success of migration of master pods from
   an unschedulable node. The migration includes Patroni switchovers to
-  respective replicas on healthy nodes. The situation where master pods still
-  exist on the old node after this timeout expires has to be fixed manually.
-  The default is 20 minutes.
+  respective replicas on healthy nodes. A failed migration attempt is retried
+  every minute until this timeout expires. The situation where master pods
+  still exist on the old node after this timeout expires has to be fixed
+  manually. The default is 20 minutes.
 
 * **enable_pod_antiaffinity**
   toggles [pod anti affinity](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/)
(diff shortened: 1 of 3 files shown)


pdixit26:
This is a bugfix for #3176. I can't add the label myself.

pdixit26:
@FxKu Could I get review on this PR?

jopadi (approved):

FxKu:
👍

PR #3180, opened 2026-09-03
Pass CLONE_AWS_REGION and STANDBY_AWS_REGION when IRSA is configured

tcondeixa:
Description:
When irsa_role_arn is set, the operator already injects CLONE_AWS_ROLE_ARN and CLONE_AWS_WEB_IDENTITY_TOKEN_FILE for clone clusters, and the equivalent vars for standby. However, CLONE_AWS_REGION and STANDBY_AWS_REGION were missing.

Without the region, spilo's configure_spilo.py falls back to reading it from IMDS (169.254.169.254). On clusters where IRSA is used and the IMDS hop limit is set to 1 (pods cannot reach IMDS), this fallback returns an empty or malformed value, causing wal-g to construct a broken STS endpoint (e.g. sts.loca.amazonaws.com) and fail to authenticate.

This change adds CLONE_AWS_REGION and STANDBY_AWS_REGION alongside the other IRSA env vars, using c.OpConfig.AWSRegion which defaults to eu-central-1 and can be overridden via the operator configmap.

NOTE: This PR was done with the help of AI agents


tcondeixa:
👍

tcondeixa:
Considering we decided to keep for now the operator injecting the CLONE variables, I'm also adding the CLONE_AWS_REGION and keep the current pattern.

jopadi (approved):

tcondeixa:
@FxKu can you please also review this one since you helped when we introduced IRSA

FxKu (approved):

PR #3181, opened 2026-09-03
add CLONE/STANDBY_AWS_STS_REGIONAL_ENDPOINTS when IRSA is configured

jopadi:
## Description

Follows up on #3180.

When `irsa_role_arn` is set, the operator injects IRSA-related env vars for clone and standby clusters. However, `AWS_STS_REGIONAL_ENDPOINTS` was missing. Without it, WAL-G may fall back to a global STS endpoint or construct a broken one, causing credential retrieval to fail.

This change adds `CLONE_AWS_STS_REGIONAL_ENDPOINTS` and `STANDBY_AWS_STS_REGIONAL_ENDPOINTS` with value `regional`, consistent with how the other IRSA vars are hardcoded.

NOTE: PR created with the help of AI agents

### DIFF
--- a/pkg/cluster/k8sres.go
+++ b/pkg/cluster/k8sres.go
@@ -2224,9 +2224,11 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
 	}
 
 	if c.OpConfig.IRSARoleARN != "" {
+		result = append(result, v1.EnvVar{Name: "CLONE_AWS_STS_REGIONAL_ENDPOINTS", Value: "regional"})
 		result = append(result, v1.EnvVar{Name: "CLONE_AWS_ROLE_ARN", Value: c.OpConfig.IRSARoleARN})
 		result = append(result, v1.EnvVar{Name: "CLONE_AWS_WEB_IDENTITY_TOKEN_FILE", Value: "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"})
 		result = append(result, v1.EnvVar{Name: "CLONE_AWS_REGION", Value: c.OpConfig.AWSRegion})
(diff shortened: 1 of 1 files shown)


tcondeixa:
AWS_STS_REGIONAL_ENDPOINTS is not set anywhere today, so both the main pod and clone/standby currently use the global STS endpoint. Hardcoding "regional" only for clone/standby would make them inconsistent with the main pod. Additionally, changing the default STS endpoint behavior would be a breaking change for existing OSS users who may rely on the global endpoint. IMO the right approach would be to expose this as an operator config field (e.g. aws_sts_regional_endpoints) with no default, so users can opt in explicitly.

I don't have much experience in this project, so I will let other provide also feedback.

pkg/cluster/k8sres.go
```diff
@@ -2224,9 +2224,11 @@ func (c *Cluster) generateCloneEnvironment(description *acidv1.CloneDescription)
 	}
 
 	if c.OpConfig.IRSARoleARN != "" {
+		result = append(result, v1.EnvVar{Name: "CLONE_AWS_STS_REGIONAL_ENDPOINTS", Value: "regional"})
 		result = append(result, v1.EnvVar{Name: "CLONE_AWS_ROLE_ARN", Value: c.OpConfig.IRSARoleARN})
 		result = append(result, v1.EnvVar{Name: "CLONE_AWS_WEB_IDENTITY_TOKEN_FILE", Value: "/var/run/secrets/eks.amazonaws.com/serviceaccount/token"})
 		result = append(result, v1.EnvVar{Name: "CLONE_AWS_REGION", Value: c.OpConfig.AWSRegion})
+		result = append(result, v1.EnvVar{Name: "STANDBY_AWS_STS_REGIONAL_ENDPOINTS", Value: "regional"})
```
tcondeixa, line 2231: this is in the wrong place

PR #3185, opened 2026-09-08
Set SPILO_PROVIDER=aws when using IRSA

tcondeixa:
When IRSA is configured, spilo attempts to detect the provider by probing the EC2 metadata API (169.254.169.254). When access to that endpoint is restricted, this results in a misleading "assuming local Docker setup" log line, even though the cluster is operating correctly.

Since IRSA is AWS-only, explicitly setting SPILO_PROVIDER=aws skips the probe and makes the provider explicit.


tcondeixa:
👍

FxKu:
👍

Issue #2913, opened 2025-05-19
controllerrevision not cleaned up after Postgresql deletion


fabianem:
Please, answer some short questions which should help us to understand your problem / question better?

- **Which image of the operator are you using?** `ghcr.io/zalando/postgres-operator:v1.14.0`
- **Where do you run it - cloud or metal? Kubernetes or OpenShift?** [AWS K8s | GCP ... | Bare Metal K8s] `K8s`
- **Are you running Postgres Operator in production?** [yes | no] `yes`
- **Type of issue?** [Bug report, question, feature request, etc.] `Bug report`

Hi,

I’m running the Zalando Postgres Operator in cluster-wide mode, and overall the resource cleanup during Postgresql instance deletion works well. However, I’ve observed that controllerrevision resources are left behind after deleting a Postgresql CR - even though other resources like StatefulSets, Pods, Services, and Secrets are properly removed.

My operator configuration includes enable_owner_references: true:
```yaml
configKubernetes:
    pod_management_policy: "parallel"
    enable_persistent_volume_claim_deletion: true
    enable_finalizers: false
    enable_owner_references: true
    enable_pod_antiaffinity: false
    cluster_domain: "cluster.local"
...
```

Upon inspecting the controllerrevision resource before deletion, it correctly has an ownerReference pointing to the corresponding StatefulSet. However, after the Postgresql CR is deleted and the StatefulSet is removed, the controllerrevision remains - and its ownerReference is gone.

This suggests that the controllerrevision becomes orphaned during the deletion process. I suspect this might be due to the use of [metav1.DeletePropagationOrphan](https://github.com/zalando/postgres-operator/blob/7c7aa969350bef67d0283d93589499ae3d114edb/pkg/cluster/cluster.go#L112) when [deleting the StatefulSet](https://github.com/zalando/postgres-operator/blob/7c7aa969350bef67d0283d93589499ae3d114edb/pkg/cluster/resources.go#L250).

Would switching to `DeletePropagationForeground` be more appropriate here to allow Kubernetes to garbage collect the controllerrevision via its ownership chain?

Thanks in advance for looking into this!

FxKu:
I never heard of a controllerreversion resource. Is there one created for every custom resource? Does it produce any issues when it's left behind? You can open a PR where you change the setting for the Statefulset and we can check if e2e tests still work.

PR #3189, opened 2026-09-22
Delete services with background propagation so their EndpointSlices are removed

### LINKED ISSUE #2973: EndpointSlices not cleaned up after replica service deletion (orphaned resources)
**Which image of the operator are you using?** 1.14.0
**Where do you run it - cloud or metal?** Vanilla K8S/Bare metall
**Are you running Postgres Operator in production?** Yes
**Type of issue?** Bug report

**Problem Description**
We're experiencing a resource leak where EndpointSlices for **-repl** services are not being cleaned up after the parent Service is deleted. This has triggered quota alerts in our namespace.

**Observed Behavior**
When a PostgreSQL cluster is deleted, the operator properly deletes the replica service (*-repl)
However, the associated **EndpointSlices** remain in the namespace as orphaned resources
This only affects **-repl** services - other services (-config, master endpoint) clean up properly
The **EndpointSlices** have proper **ownerReferences** pointing to the Service:

```yaml
  - apiVersion: v1
    blockOwnerDeletion: true
    controller: true
    kind: Service
    name: endpoints-keys-test-repl
    uid: 4d904c2d-cfff-4ee1-bd50-3215dc16de82
```
**Reproduction Steps**

Create a Spilo PostgreSQL cluster
Delete the cluster
Check for remaining EndpointSlices:

`kubectl get endpointslices -n <namespace> | grep db-repl`

Orphaned EndpointSlices will remain despite parent Services being deleted

**Expected Behavior**
EndpointSlices should be automatically cleaned up by Kubernetes garbage collector when the parent Service is deleted (via ownerReferences).

**Investigation Findings**
The operator does not create an endpoint for -repl during cluster creation, only the service.
However, at the moment of cluster deletion attempts to clean this object. 
From operator logs during deletion:
```
time="2025-10-30T11:26:09Z" level=debug msg="deleting replica endpoint" cluster-name=service-keys/endpoints-keys-test pkg=cluster worker=3
time="2025-10-30T11:26:09Z" level=info msg="replica endpoint \"service-keys/endpoints-keys-test-repl\" has been deleted" cluster-name=service-keys/endpoints-keys-test pkg=cluster worker=3
time="2025-10-30T11:26:09Z" level=debug msg="deleting replica service" cluster-name=service-keys/endpoints-keys-test pkg=cluster worker=3
time="2025-10-30T11:26:09Z" level=info msg="replica service \"service-keys/endpoints-keys-test-repl\" has been deleted" cluster-name=service-keys/endpoints-keys-test pkg=cluster worker=3

```

From Kubernetes API logs, the garbage collector attempts to patch the EndpointSlice but doesn't delete it:
```
Oct 30, 2025 @ 11:26:09.840  /apis/discovery.k8s.io/v1/namespaces/service-keys/endpointslices/endpoints-keys-test-repl-9p7pf
ResponseComplete  patch  
system:serviceaccount:kube-system:generic-garbage-collector
```

**Hypothesis**
It appears the operator may be deleting the Service with orphan propagation policy instead of background/foreground, which would explain why the garbage collector only patches but never deletes the EndpointSlices.

**Impact**
Resource quota exhaustion in namespaces with many demo/test databases
Manual cleanup required periodically
Affects all -repl services across multiple clusters

somaz94:
## Problem description

When a cluster is deleted, deleteService passes the cluster wide deleteOptions, which use DeletePropagationOrphan. The replica service has a selector, so its EndpointSlices are created by the EndpointSlice controller with the Service as owner. With orphan propagation the garbage collector only removes that owner reference, and the slices stay in the namespace after the Service is gone. The master service is not affected because its slices are mirrored from the Endpoints object the operator deletes itself, and the pooler services already use their own propagation policy in deleteConnectionPooler.

This change uses background propagation in deleteService only. The shared deleteOptions and the statefulset path stay as they are, since deleteStatefulSet orphans the pods on purpose and removes them itself.

A new unit test checks the propagation policy sent for the master and replica service deletes. It fails on master with Orphan and passes with this change. make mocks, make linux and go test -race ./... pass locally.

## Linked issues

Fixes #2973

#2913 reports the same orphan behaviour for ControllerRevisions through the statefulset delete. I left it out of this PR so that change can be checked against the e2e tests on its own.

## Checklist

- [x] Your go code is [formatted](https://blog.golang.org/gofmt). Your IDE should do it automatically for you.
- [ ] You have updated [generated code](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#code-generation) when introducing new fields to the `acid.zalan.do` api package.
- [ ] New [configuration options](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#introduce-additional-configuration-parameters) are reflected in CRD validation, helm charts and sample manifests.
- [x] New functionality is covered by [unit](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#unit-tests) and/or [e2e](https://github.com/zalando/postgres-operator/blob/master/docs/developer.md#end-to-end-tests) tests.
- [x] You have checked existing open PRs for possible overlay and referenced them.

I used Claude Code to help investigate the issue and write the test.


somaz94:
Thanks for the label and for merging master in. I've marked this ready for review. The unit tests pass, and the e2e failure is test_major_version_upgrade, which failed the same way on master in run [33736493007](https://github.com/zalando/postgres-operator/actions/runs/33736493007) on September 3. Since the suite stops at the first failure, the deletion tests after it did not run on this commit, so a re-run would cover those too.

